PR #6, opened 2019-10-09
Platform init

rshipp:
Task: https://app.asana.com/0/1133543009734854/1142999165564240/f

Summary:

* Port data-central to Rails 6 and Ruby 2.6.5
* Rename to Platform
* Remove unused gems


app/javascript/packs/application.js
```diff
@@ -0,0 +1,18 @@
...
+
+
+// Uncomment to copy all static images under ../images to the output folder and reference
+// them with the image_pack_tag helper in views (e.g <%= image_pack_tag 'rails.png' %>)
+// or the `imagePath` JavaScript helper below.
+//
+// const images = require.context('../images', true)
+// const imagePath = (name) => images(name, true)
+
+console.log('Hello World from Webpacker')
```
StetsenTech: 👋

.gitignore
```diff
@@ -38,3 +38,10 @@ log/
 /.bundle/
 /vendor/bundle
 /lib/bundler/man/
+
+/public/packs
+/public/packs-test
+/node_modules
+/yarn-error.log
```
sadleb: The two are .gitignored above. Delete the above?

Dockerfile
```diff
@@ -1,7 +1,14 @@
...
-RUN apt-get update -yqq
-RUN apt-get install -yqq --no-install-recommends nodejs
+RUN apk add --update --no-cache \
+    build-base \
+    postgresql-dev \
+    git \
+    imagemagick \
+    nodejs \
+    yarn \
+    tzdata
```
sadleb: Can we add vim? I want to be able to connect to the container and load/edit files while developing if necessary

Dockerfile
```diff
@@ -1,7 +1,14 @@
...
 
 COPY Gemfile* /usr/src/app/
```
sadleb: All our other Rails apps use `/app` as the root of the app in the Docker image. If there is no underlying reason to use `/usr/src/app` I feel like we should be consistent. Note that our wordpress apps use `/var/www/html`. I would just love to get to the point of everything running the same way, same root dir, same permissions, users.

sadleb: > All our other Rails apps use `/app` as the root of the app in the Docker image. If there is no underlying reason to use `/usr/src/app` I feel like we should be consistent. Note that our wordpress apps use `/var/www/html`. I would just love to get to the point of everything running the same way, same root dir, same permissions, users.

Though, Ryan, I think you should punt on changing anything here for now. Getting a consistent Docker setup for all apps where dev, staging, test, and prod all work the same way is part of my overall Q2 roadmap.

Dockerfile
```diff
@@ -10,5 +17,6 @@ ENV BUNDLE_PATH /gems
 
 RUN bundle install
 COPY . /usr/src/app/
+RUN yarn install --check-files
```
sadleb: No npm!! Thank gawd!!

.gitignore
```diff
@@ -38,3 +38,10 @@ log/
 /.bundle/
 /vendor/bundle
 /lib/bundler/man/
+
```
sadleb: Note that ```/config/deploy/production.rb
/config/deploy/staging.rb``` above are for Capistrano deploys. I think we can remove.

sadleb:
I think you can delete the whole `config/deploy` dir. That is for Capistrano.

Dockerfile
rshipp: This is an artifact of the datacentral dockerfile, so I have no argument against changing it.

Dockerfile
rshipp: @sadleb we should talk about this too, I'd argue that anything you want to be able to edit inside the container should be mounted, not copied. I want to be able to edit any file outside the container and have it live-reload.

PR #7, opened 2019-10-11
CI

rshipp:
Task: https://app.asana.com/0/1128645836596228/1143501603378293/f

Summary: Set up CI so we can get test and coverage reports on PRs.

* Codacy: https://app.codacy.com/manual/bebraven/platform/dashboard
* Coverage will start showing up once this is merged.
* CI: (Private, in Heroku -> platform pipeline)
* Docker: Fixed a bunch of issues so it builds and runs correctly and live-updates as you edit.
* Unit tests: fixed a bunch of failing tests, so we can start from green.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+# Git
+.git
+.gitignore
+
+# Logs
+log/*
+
+# Temp files
+tmp/*
+
+# Webpack-related
+node_modules
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,12 +11,15 @@ RUN apk add --update --no-cache \
     tzdata
 
 COPY Gemfile* /usr/src/app/
+COPY package.json /usr/src/app/
+COPY yarn.lock /usr/src/app/
 WORKDIR /usr/src/app
 
 ENV BUNDLE_PATH /gems
 
 RUN bundle install
-COPY . /usr/src/app/
 RUN yarn install --check-files
+
+COPY . /usr/src/app/
 
 CMD ["bundle", "exec", "rails", "s", "-p", "3020", "-b", "0.0.0.0"]
--- a/app.json
+++ b/app.json
@@ -1,0 +1,9 @@
+{
+  "environments": {
+    "test": {
+      "scripts": {
+        "test": "bundle exec rspec"
+      }
+    }
+  }
+}
--- a/app.json
+++ b/app.json
@@ -1,6 +1,7 @@
 {
   "environments": {
     "test": {
+      "addons": ["heroku-postgresql:in-dyno"],
       "scripts": {
         "test": "bundle exec rspec"
       }
--- a/app.json
+++ b/app.json
@@ -1,6 +1,10 @@
 {
   "environments": {
     "test": {
+      "buildpacks": [
+        { "url": "https://github.com/heroku/heroku-buildpack-google-chrome" },
+        { "url": "heroku/ruby" }
+      ],
       "addons": ["heroku-postgresql:in-dyno"],
       "scripts": {
         "test": "bundle exec rspec"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,6 @@
 version: '3'
 
 services:
-      
   platformweb:
     build: 
       context: .
@@ -10,6 +9,9 @@ services:
       - "3020:3020"
     volumes:
       - .:/usr/src/app
+      - node-modules:/usr/src/app/node_modules
+      - /usr/src/app/tmp
+      - /usr/src/app/log
     env_file:
       - .env
 
@@ -22,3 +24,4 @@ services:
 
 volumes:
   db-platform:
+  node-modules:
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ log/
 /yarn-error.log
 yarn-debug.log*
 .yarn-integrity
+
+# Ignore coverage files.
+/coverage
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,7 @@ gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
 gem 'paranoia'
 gem 'will_paginate'
+gem 'bulk_insert'
 
 gem 'devise'
 gem 'devise_cas_authenticatable'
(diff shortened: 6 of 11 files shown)


rshipp:
Note: I've also locked down the `production` and `staging` branches to only allow merges from PRs with 1+ approving review and passing CI (but not codacy, because it has too many false positives for now). (But admins can bypass these limits anyway.)

docker-compose.yml
```diff
@@ -10,6 +9,9 @@ services:
       - "3020:3020"
     volumes:
       - .:/usr/src/app
+      - node-modules:/usr/src/app/node_modules
```
StetsenTech: I'm not sure how others feel about this, but I personally do not include `node_modules` as a volume. Docker image can be using a completely different environment that is not compatibility with the local machine. I know this increase build time, but in the long run I believe it will save heartache of dealing with compatibility issues.

rshipp: this volume lets us run yarn both in the docker container and in the local filesystem, without them breaking each other. node-modules is now a persistent volume within the container.

docker-compose.yml
```diff
@@ -10,6 +9,9 @@ services:
...
+      - /usr/src/app/tmp
+      - /usr/src/app/log
```
rshipp: these two are just mounting *nothing* over these directories, so that we 1) don't have Rails server PID files showing up from the docker container in our local directory (you can run `rails s` both inside and outside the container at the same time) and 2) aren't writing logs out to the local filesystem and wasting space.

app.json
```diff
@@ -0,0 +1,14 @@
+{
+  "environments": {
+    "test": {
+      "buildpacks": [
+        { "url": "https://github.com/heroku/heroku-buildpack-google-chrome" },
```
rshipp: this isn't used yet but it'll let us start using selenium when we get there

rshipp:
@StetsenTech ha- check the comment I just left on that same line 😆 I think it is doing what you want, it's specifically so that the node_modules don't conflict.

.dockerignore
```diff
@@ -0,0 +1,12 @@
+# Git
+.git
```
sadleb: If we're ignoring `.git` in the image, we should probably not install git in the Dockerfile?

Dockerfile
```diff
@@ -11,12 +11,15 @@ RUN apk add --update --no-cache \
...
 
 ENV BUNDLE_PATH /gems
 
 RUN bundle install
-COPY . /usr/src/app/
 RUN yarn install --check-files
 
+COPY . /usr/src/app/
+
 CMD ["bundle", "exec", "rails", "s", "-p", "3020", "-b", "0.0.0.0"]
```
sadleb: We'll want to change this to just run `/bin/bash` or something and choose the command to run in Dev vs Staging/Prod using `docker-compose.yml` and `heroku.yml` respectively.

ACTUALLY, bigger question. Should we just being using Heroku buildpacks (not Docker) for staging and prod and only use the Docker image for development?

docker-compose.yml
sadleb: @StetsenTech this actually does the opposite of what I think you're saying. By creating a named volume at the path inside the container, aka `/usr/src/app/node_modules` it's masking the node_modules directory on your local machine. So like Ryan said, stuff inside and outside are kept separate, but also, the modules persist which speeds up build times. Though, you have to be aware that if you ever really need to blow away this stuff, you'll have to delete the named volume and rebuild.

docker-compose.yml
sadleb: What's the plan for how to read logs when developing locally? Always run `docker-compose logs` ? Maybe that's a good flow? I'm trying to think. I can imagine myself wanting to scan back through the log file, searching around using vim, in which case it may be nice to be able to just open the log file locally outside the container? At the end of the day, this is all on your local system, so whether or not the log dir inside the container is mounted so that you can access it from outside the container, there isn't a space savings -- unless I'm missing something?

Dockerfile
rshipp: yeah we don't use docker in staging/prod at all, heroku uses the app.yml here.

docker-compose.yml
rshipp: oh you're right, it's not saving space. I still would rather everything that conflicts between running locally and running in docker is sectioned off like this, but we should talk about it more if you disagree.

sadleb: I guess I'm confused what you mean about "running locally."  The Docker container runs the web-server, has all dependencies installed inside like the correct version of bundler, ruby, rails, etc. Are you saying there may be a need to setup all the dependencies outside the container to be able to run `rails s` or other commands that execute code from outside the container? If so, what purpose does the container serve? From my perspective, we should be editing code locally (which since it's mounted gets served by docker) and ideally troubleshooting locally (like running a debugger and whatnot, including looking at logs), but for anything that needs to run code and depends on all the dependencies, that happens inside the container. 

If we only had one web app, I could see there being a case for setting up a local dev env that you can use but also using Docker as a way to see how the local stuff would behave in a deployed container -- but since we have multiple apps with different versions of ruby, I feel like the container should be the place where everything that needs to run code happens.

rshipp: ok these are my thoughts, feel free to disagree or point out flaws: 

1. the purpose of the container is
  1.1. avoid the "it works on my machine" problem
  1.2. ease dependency management (don't have to install postgres, redis, ... on your dev box just to start the app)
  1.3. lower barrier to entry for contributing
2. all Platform code should work outside the container
  1.1. there's no reason code we write (including gems we use) should only work on one OS or env.
  1.2. if we need to switch to another platform in the future, we should be able to do that with no code changes
  1.3. this discourages weird hacks that only work in docker for some reason, which hopefully means more simplicity overall
  1.4. this discourages us from pulling in external binary dependencies that can't be managed by bundle/yarn (which is a good thing)
3. we should be able to attach a debugger and view logs from whatever we're running, regardless of where we're running it.
  3.1. I haven't used byebug but i assume this isn't a problem inside docker or out
  3.2. react debugging goes through the browser anyway so its a non-issue here
  3.3. logs for container-internal things should be in `docker-compose log`, or in your dev log receiver of choice. Piping through `less` should let you search just like in vim, but lmk if there's something i'm missing here where you specifically need files.
4. for the purposes of this repo, I'm assuming we only have one app
  4.1. normalizing things between this and our old apps, like the docker-compose service names from earlier, is great
  4.2. we don't want to do things in platform *only* because we're doing them somewhere else; some things should be different here if we can align on them being good practice
  4.3. we should be able to rely on rvm, bundle, and yarn to manage versions for us outside the container

PR #10, opened 2019-10-22
React improvements

rshipp:
* Renamed `entrypoint.js` to `platform.js` because `entrypoint` is both technically wrong and potentially confusing.
* Removed autogenerated example hello-world file.
* Moved axe import inside the env check and made it non-global.


app/javascript/packs/platform.js
```diff
@@ -1,13 +1,14 @@
-// Run this example by adding <%= javascript_pack_tag 'development' %> to the head of your layout file,
+// This is the main entrypoint for Braven Portal.
```
sadleb: Super nit picky, but can we make sure and always say Platform, not Portal. I just don’t want to get off on the wrong foot of confusion when talking about this app by always calling it the Braven Platform (bc the Portal is Canvas-lms).

rshipp: not nitpicky, i definitely just confused the two p-words. will fix!

PR #11, opened 2019-10-24
Adds Platform to Braven Dev Network

StetsenTech:
**Ticket(s)**:
https://app.asana.com/0/1133543009734854/1143501603378304/f

**Summary**
- Adds Platform to the Braven Dev Network so that it can resolve names from other services


config/secrets.yml
```diff
@@ -24,12 +26,14 @@ staging:
   mailer_from_email: <%= ENV['MAILER_FROM_EMAIL'] %>
   devise_secret_key: <%= ENV['DEVISE_SECRET_KEY'] %>
   devise_pepper: <%= ENV['DEVISE_PEPPER'] %>
-  
+  sso_url: <%= ENV['SSO_URL'] %>
+
```
rshipp: sso_url was already in there for all of these

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3'
+version: "3.5"
```
rshipp: is this the version of docker-compose, or something else? just curious

StetsenTech: This is the version of docker-compose we use in other containers. It's the reason we are able to name networks.
https://docs.docker.com/compose/compose-file/compose-versioning/#version-35

config/secrets.yml
StetsenTech: Ah for some reason it didn't show in mine....removing.

StetsenTech:
Removed extra `sso_url`s. Not sure why it was missing when I was developing...

PR #12, opened 2019-10-25
livereload

rshipp:
Task: https://app.asana.com/0/1128645836596228/1145928107151942/f

Summary:

* Live-reload *everything* (see git commits for more detail)
* Works in docker and outside it
* Example hello.js is included to test React live reloading
* Note that the dockerfile now runs guard, instead of `rails s`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,4 +22,4 @@ RUN yarn install --check-files
 
 COPY . /usr/src/app/
 
-CMD ["bundle", "exec", "rails", "s", "-p", "3020", "-b", "0.0.0.0"]
+CMD ["bundle", "exec", "guard", "-i"]
--- a/Guardfile
+++ b/Guardfile
@@ -119,3 +119,24 @@ guard :webpacker do
   watch('config/webpacker.yml')
   watch(%r{^config/webpacker/(.+)$})
 end
+
+# Guard-Rails supports a lot options with default values:
+# daemon: false                        # runs the server as a daemon.
+# debugger: false                      # enable ruby-debug gem.
+# environment: 'development'           # changes server environment.
+# force_run: false                     # kills any process that's holding the listen port before attempting to (re)start Rails.
+# pid_file: 'tmp/pids/[RAILS_ENV].pid' # specify your pid_file.
+# host: 'localhost'                    # server hostname.
+# port: 3000                           # server port number.
+# root: '/spec/dummy'                  # Rails' root path.
+# server: thin                         # webserver engine.
+# start_on_start: true                 # will start the server when starting Guard.
+# timeout: 30                          # waits untill restarting the Rails server, in seconds.
+# zeus_plan: server                    # custom plan in zeus, only works with `zeus: true`.
+# zeus: false                          # enables zeus gem.
+# CLI: 'rails server'                  # customizes runner command. Omits all options except `pid_file`!
+
+guard 'rails', port: 3020, host: '0.0.0.0' do
+  watch('Gemfile.lock')
+  watch(%r{^(config|lib)/.*})
+end
--- a/Guardfile
+++ b/Guardfile
@@ -68,3 +68,43 @@ guard :rspec, cmd: "bundle exec rspec" do
     Dir[File.join("**/#{m[1]}.feature")][0] || "spec/acceptance"
   end
 end
+
+guard 'livereload' do
+  extensions = {
+    css: :css,
+    scss: :css,
+    sass: :css,
+    js: :js,
+    coffee: :js,
+    jsx: :js,
+    html: :html,
+    png: :png,
+    gif: :gif,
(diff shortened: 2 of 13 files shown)


Dockerfile
```diff
@@ -21,4 +21,4 @@ RUN yarn install --check-files
 
 COPY . /usr/src/app/
 
-CMD ["bundle", "exec", "rails", "s", "-p", "3020", "-b", "0.0.0.0"]
+CMD ["bundle", "exec", "guard", "-i"]
```
sadleb: If this entrypoint command is dev only, it should probably be in docker-compose.yml which is what we use to start a dev container. The image itself should prob just have `bash` as the default entrypoint so that whatever env is running the container has to explicitly choose the CMD to run when it's started. For other Dockerfiles in other apps, docker-compose will run the dev server command while heroku.yml will run the production server command.

Guardfile
```diff
@@ -68,3 +68,75 @@ guard :rspec, cmd: "bundle exec rspec" do
...
+# port: 3000                           # server port number.
+# root: '/spec/dummy'                  # Rails' root path.
+# server: thin                         # webserver engine.
+# start_on_start: true                 # will start the server when starting Guard.
+# timeout: 30                          # waits untill restarting the Rails server, in seconds.
+# zeus_plan: server                    # custom plan in zeus, only works with `zeus: true`.
+# zeus: false                          # enables zeus gem.
+# CLI: 'rails server'                  # customizes runner command. Omits all options except `pid_file`!
+
+guard 'rails', port: 3020, host: '0.0.0.0' do
```
sadleb: Does hard-coding this port here have dependencies elsewhere? If so, is there anyway to make it easier to change the port and know all the places that need to be updated?

config/webpacker.yml
```diff
@@ -59,10 +59,10 @@ development:
   # Reference: https://webpack.js.org/configuration/dev-server/
   dev_server:
     https: false
-    host: localhost
+    host: 0.0.0.0
     port: 3035
-    public: localhost:3035
-    hmr: false
+    public: 0.0.0.0:3035
```
sadleb: Continuing to show my ignorance about what all this stuff actually is, but how does port 3035 relate to port 3020 in the Guardfile? What calls into webpacker and guard on these ports?

docker-compose.yml
```diff
@@ -7,6 +7,7 @@ services:
       dockerfile: Dockerfile
     ports:
       - "3020:3020"
+      - "3035:3035"
```
sadleb: I just answered my own question above. Maybe just put a comment here about what each port is used for and which file to look in if it needs to be changed?

Guardfile
StetsenTech: Assuming this is should be dev only...does it matter in this case? I mean...we could make it so `0.0.0.0` can pull from an env for `platformweb`.

config/webpacker.yml
StetsenTech: Correct me if I am wrong, but isn't 3020 so that `rails -s` can run and port 3035 is so that webpack can properly get signals to rebuild based on changes in the working in environment?

Guardfile
rshipp: yeah this is dev-only, rails uses 3020 and webpack-dev-server uses 3035. I think both of those are hardcoded in a couple places in order to get our `http://platformweb` docker network stuff working. it's all ignored on heroku though. it's possible we could put them in variables somehow instead, do you think that's worth the effort?

config/webpacker.yml
rshipp: this config is used for webpack-dev-server, which allows React [HMR](https://webpack.js.org/guides/hot-module-replacement/) - when you visit a rails view in the browser, it attaches a websocket to :3035 so it can get notifications when React components change and live-reload them. this is only true in dev, HMR isn't used in prod.

docker-compose.yml
rshipp: good idea, can do

Dockerfile
rshipp: both of these are dev-only - we only use Dockerfile in dev. You mentioned this before but I think I just didn't understand how our heroku stuff is working for other apps. Do you want to migrate Platform to match what we're doing in other apps? Right now, this is how the platform stuff works:

* dev
  * docker-compose.yml
    * Dockerfile
      * Guardfile

* production
  * *heroku autodetects rails and does magic things*

* test
  * app.json
    * spec/*

sadleb: Generally, I think we can migrate the docker stuff to work like our other apps in a separate PR. In the other apps, Dockerfile just has the CMD be `bash` and docker-compose.yml specifies the command to run in dev.

Guardfile
sadleb: Prob not worth trying to share variables. Helpful context about this being dev only (esp in the comments below)

PR #13, opened 2019-10-29
Add sentry-raven gem

rshipp:
Task: https://app.asana.com/0/1136372351092870/1142999165564244

It only needs the `SENTRY_DSN` env variable set in Heroku. No other config/code changes.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,3 +63,5 @@ gem 'devise'
 gem 'devise_cas_authenticatable'
 
 gem 'react-rails'
+
+gem 'sentry-raven'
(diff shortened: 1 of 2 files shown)


sadleb:
Do you have a plan for how to manage the various ENV vars for dev, staging and prod? I've been working on some of that for the old code and wasn't sure if you are waiting on me to propose / rollout a plan or if you already had something in the works. Ideally, we would have them all in one spot per env and ideally we'd have some sort of versioning/access control.

sadleb:
> @sadleb I was just going to set staging/prod up in the Heroku pipeline. I don't personally need Sentry in dev, but I assume anyone who does can just add it to their local `.env`, which is picked up by the docker-compose config.

Cool, that works for now. At some point, we may want to create a way to back them up and version them in case someone accidentally overwrites them so we can get back up and running quickly instead of hunting down all the values... but I'll consider that piece on my plate with the overall dev infrastructure project. A versioned S3 bucket and a cron that writes the current values there at come regular cadence may be good enough.

PR #14, opened 2019-10-29
Adds RubyCAS Core Gem and Setup for External Communication

StetsenTech:
**Ticket(s)**:
https://app.asana.com/0/1133543009734854/1146936250326599

**Description**:
Breaking out development pieces so it can be merged in. This PR sets up platform to have the RubyCAS Core Server gem and enable communicating with the other boxes in the Braven Dev Network via the `platformweb`.

I also have a PR open on the nginx repo to add the route to the current set up.
https://github.com/beyond-z/nginx-dev/pull/1

**Summary**
- Adds Git to Platform dockerfile
  - This is necessary for bundle to install the RubyCAS Core gem
from GitHub
- Adds RubyCAS Core Gem
- Whitelists `platformweb` so other containers can communicate
  - This is necessary so that other containers to communicate
with the platform box using platformweb:3020

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,12 +7,12 @@ RUN apk add --update --no-cache \
     imagemagick \
     nodejs \
     yarn \
-    tzdata \
-    git
+    tzdata 
 
 COPY Gemfile* /usr/src/app/
 COPY package.json /usr/src/app/
 COPY yarn.lock /usr/src/app/
+COPY vendor/ /usr/src/app/vendor/
 WORKDIR /usr/src/app
 
 ENV BUNDLE_PATH /gems
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,12 +7,12 @@ RUN apk add --update --no-cache \
     imagemagick \
     nodejs \
     yarn \
-    tzdata 
+    tzdata \
+    git 
 
 COPY Gemfile* /usr/src/app/
 COPY package.json /usr/src/app/
 COPY yarn.lock /usr/src/app/
-COPY vendor/ /usr/src/app/vendor/
 WORKDIR /usr/src/app
 
 ENV BUNDLE_PATH /gems
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,8 @@ RUN apk add --update --no-cache \
     imagemagick \
     nodejs \
     yarn \
-    tzdata
+    tzdata \
+    git
 
 COPY Gemfile* /usr/src/app/
 COPY package.json /usr/src/app/
--- a/Gemfile
+++ b/Gemfile
@@ -66,5 +66,6 @@ gem 'react-rails'
 
 gem 'sentry-raven'
 
-gem 'rubycas-server-core', path: '/usr/src/app/vendor/rubycas-server-core'
+# Using a branch for a time being to remove R18n until we decide what we want to do
+gem 'rubycas-server-core', github: 'bebraven/rubycas-server-core', branch: 'platform-compat'
 gem 'rubycas-server-activerecord'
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -7,7 +7,8 @@ RUN apk add --update --no-cache \
     imagemagick \
     nodejs \
     yarn \
-    tzdata
+    tzdata \
+    git
```
sadleb: I don't think we want git inside the container. Is there a reason you added this?

Gemfile
```diff
@@ -65,3 +65,6 @@ gem 'devise_cas_authenticatable'
 gem 'react-rails'
 
 gem 'sentry-raven'
+
+gem 'rubycas-server-core', git: 'https://github.com/vasilakisfil/rubycas-server-core'
```
sadleb: Can we fork this repo to https://github.com/bebraven/rubycas-server-core so that we are pointing to our own version of the "gem" and can tweak that gem if necessary. I don't know what happens if someone deletes a repo or makes it private, but I assume if we fork it, we'll always have the code and not depend on the whims of vasilakisfil's account.

Gemfile.lock
```diff
@@ -1,3 +1,11 @@
+GIT
+  remote: https://github.com/vasilakisfil/rubycas-server-core
+  revision: dfc94f4f55e055d969a6fce599bf48099a4a2589
+  specs:
+    rubycas-server-core (0.2.1)
+      activesupport (>= 3.0)
+      r18n-core (~> 2.0.3)
```
sadleb: I assume that rubycas-server-core depends on this and has actual translations applied? If not, in our fork I think we should remove the dependency and figure out a holistic approach to i18n translations. I wouldn't want some of our codebase using translations tools from this gem and other parts of our codebase using another i18n translation gem / approach. If it's hard to get rid of, let's sync with @rshipp about his thoughts.

config/environments/development.rb
```diff
@@ -59,4 +59,6 @@
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  config.hosts << "platformweb"
```
sadleb: Can you add a comment about why this is required? Like, what are the implications if we don't do this? I can see a future dev coming upon this and spending too much time figuring out what it's for when a quick comment would save them a ton of time understanding why this is necessary.

Gemfile
StetsenTech: Can totally do that.

config/environments/development.rb
StetsenTech: Agreed.

Gemfile.lock
StetsenTech: I will see what I can do here.

Dockerfile
StetsenTech: I added git so I could use `gem 'rubycas-server-core', git: 'https://github.com/bebraven/rubycas-server-core'`

Gemfile.lock
StetsenTech: I was able to remove the code that utilized this gem. I will be making a PR on `bebraven/rubycas-server-core`.

config/environments/development.rb
StetsenTech: Added a comment above this.

Dockerfile
StetsenTech: Removed. See comment here: https://github.com/bebraven/platform/pull/14/commits/e671c98e6ea54f7db0e69101c3cd03981addebcd

Gemfile
StetsenTech: See comment here: https://github.com/bebraven/platform/pull/14/commits/e671c98e6ea54f7db0e69101c3cd03981addebcd

PR #15, opened 2019-10-31
Adds Docker Development Scripts

StetsenTech:
**Ticket(s)**:
https://app.asana.com/0/0/1147512938357231/f

**Summary**:
- Adds Docker Scripts to help with development
  - Pretty much the same as in other environments

### DIFF
--- a/docker-compose/scripts/dbconnect.sh
+++ b/docker-compose/scripts/dbconnect.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+# This connects to the development database. The user and database name are in docker-compose.yml
+docker-compose exec platformdb psql -U user password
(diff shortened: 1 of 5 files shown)


rshipp:
do we need these for setup.sh? i thought we had talked about not using them anymore

docker-compose/scripts/dbconnect.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+# This connects to the development database. The user and database name are in docker-compose.yml
+docker-compose exec platformdb psql -U user password
```
sadleb: Does this actually work? I thought the command to send a password was -p<insert password>

StetsenTech:
You don't need them. Just nice to haves if you don't want to manually run commands. Honestly, I don't remember a conversation about not using them anymore. I made the PR so that this is inline with our current development environments.

docker-compose/scripts/dbconnect.sh
StetsenTech: You are right. That shouldn't be there. Works if it isn't there.

PR #22, opened 2019-11-04
Update README

rshipp:
Task: https://app.asana.com/0/1143269577537245/1147920068856107/f

Added more context to the README, updated seeds to be more useful for our dev work, added example `.env` files.


db/seeds.rb
```diff
@@ -10,57 +10,6 @@ def yaml label
   YAML.load(File.read("#{Rails.root}/db/seeds/#{label}.yml"))
 end
 
-if Industry.count == 0
-  Industry.create yaml(:industries).map{|name| {name: name}}
+if User.count == 0
+  User.create [{email: 'admin@beyondz.org', admin: true}]
```
rshipp: I'm not sure if there's a better way to do this without hardcoding the admin email - I tried looking up best practices but couldn't find anything relevant. @StetsenTech @sadleb thoughts?

StetsenTech: @rshipp I actually brought his up before in one of @sadleb PR's. Ideally, we should have emails etc. in a `secrets.yml` file or similar. From there, It is either read from the secret or exported as an env for the container.

README.md
```diff
@@ -1,53 +1,127 @@
...
+This will take a while the first time, because Docker has to download a Ruby image and a PostgreSQL image and set them
+up. It will also install all the necessary Ruby gems and JavaScript libraries the app requires.
 
-Now create the needed databases.
+Now create the needed databases:
 
     docker-compose exec platformweb bundle exec rake db:create db:schema:load db:migrate db:seed
 
-We've configured Docker to run our Rails app on port 3020, so go to http://localhost:3020 in your favorite browser. If everything's working correctly, you should be brought to a page which says "Welcome!" and has links to the database data. 
+We've configured Docker to run our Rails app on port 3020, so go to http://localhost:3020 in your favorite browser. If
```
StetsenTech: If this is in the development environment, we should we say `http://platformweb:3020`? I am fine with leaving it as `http://localhost:3020`. Maybe say `http://platformweb:3020 (http://localhost:3020)`?

db/seeds.rb
rshipp: hmm, ok. I don't want to block this on figuring out how to distribute secrets to everyone, but maybe we need to. was someone already working on some s3 solution for that?

README.md
rshipp: `localhost:3020` is the lowest barrier to entry; that, `platformweb`, and `platformweb:3020` would all work in theory, but you'd have to edit the hosts file. I can add a note about that too.

README.md
```diff
@@ -1,53 +1,127 @@
...
+
+## Initial setup
 
 First, we need to copy a couple of environment files in the app directory:
 
     cp .env.example .env
     cp .env.database.example .env.database
 
-Again, you don't really need to change the database passwords in these files, so we're doing this mainly to conform to "best practices" for Rails apps in general. But if you do want to pick a different database password, make sure it matches in both files.
+Be sure to set the `SSU_URL` variable in `.env` to point to whatever CAS SSO server you're using.
```
rshipp: note to self: fix `SSU` typo

db/seeds.rb
StetsenTech: I think @sadleb has something cooking for that. Maybe you can just make a `secrets.yml.example` and in the instructions say what email should be used. My 2 cents but this isn't a blocker. More interested in making progress.

README.md
StetsenTech: Holy cow I saw that as a O as many times as I read this line.

db/seeds.rb
sadleb: I have a task way lower on my priority list to come up with a way to manage and distribute secrets. To date, I've only had to deal a couple values that were needed in the dev env where I couldn't just check-in a reasonable fake value to use. For those, I've just had us add the values in corresponding environment variables in our shell and pass it through to the container. I'm not sure this particular scenario warrants taking the time to do that though.

.env.example
```diff
@@ -0,0 +1,5 @@
+DATABASE_HOST=platformdb
+DATABASE_USERNAME=user
+DATABASE_PASSWORD=password
+SSO_URL=https://ssoweb
```
sadleb: In the dev env, we use `http://ssoweb` -- we haven't setup SSL in the dev env yet.

PR #23, opened 2019-11-04
Add content editor Rails base

rshipp:
Task: https://app.asana.com/0/1133543009734854/1143567983456233/f

Note that the tests here do *not* pass - I didn't want to waste time writing tests for Rails views that I'm just going to blow away and replace with React.

This is mostly just the code from a `rails g scaffold`.


rshipp:
@sadleb definitely. This `content` model is just meant to be a temporary [dummy model](https://app.asana.com/0/1136372351092870/1147692631250865) so I can easily create/load test data with ckeditor/react. Next week I'll rip this out and start using Canvas APIs instead, at which point we plan to be reading/writing both individual fragments of pages, and full modules into the existing "Content Library".

PR #28, opened 2019-11-11
Improve livereload

rshipp:
Task: https://app.asana.com/0/1133543009734854/1148142801429929/f

Summary: This fixes some annoyances I ran into last week while working in platform. More things livereload now.

* Added rack-livereload (livereload.js) port in docker-compose.yml
* Added guard plugins for bundle, yarn, and db:migrate
  * Added `GEM_HOME` to `.env` so bundle knows where to look for gems
* Removed rspec from guardfile (idk why it was there, but I can put it back if someone wants it)
* Swapped `guard` command to `docker-compose.yml` instead of `Dockerfile` (per @sadleb and @StetsenTech request)
* Force guard to use file-polling (`-p`) because notify was only working sometimes
* Turn on guard debugging (`-d`) so we can watch what it's doing


Gemfile
```diff
@@ -44,6 +44,9 @@ group :development do
   gem 'rack-livereload'
   gem 'guard-webpacker'
   gem 'webpacker-react', "~> 1.0.0.beta.1"
+  gem 'guard-bundler', require: false
```
sadleb: `require: false` -- thank you! Been dealing with this in the Portal code. Let's make sure and always have `require: false` set for anything that is not necessary during app startup. Otherwise, overtime the app boots with a ton of memory loaded and it's a pain to go back and figure out whether/where all the stuff is needed. I'd much prefer to have ALL the gems we can have require: false and explicitly require them when needed. It's going to help us maintain this long term once we get a ton of stuff in here.

sadleb: require: false -- thank you! Been dealing with this in the Portal code. Let's make sure and always have require: false set for anything that is not necessary during app startup. Otherwise, overtime the app boots with a ton of memory loaded and it's a pain to go back and figure out whether/where all the stuff is needed. I'd much prefer to have ALL the gems we can have require: false and explicitly require them when needed. It's going to help us maintain this long term once we get a ton of stuff in here.

docker-compose.yml
```diff
@@ -19,6 +21,7 @@ services:
       - .env
     networks:
       - bravendev
+    command: bundle exec guard -dip
```
StetsenTech: Thank you!

Gemfile
rshipp: oh interesting, I wasn't sure what that did. some gems say to use it in their readmes but some don't. I'll default to using false from now on.

PR #30, opened 2019-11-14
Add basic Canvas API lib

rshipp:
Task: https://app.asana.com/0/1143269577537247/1143501603378308/f

A first pass at a CanvasAPI lib for Platform.

Note: tests not written yet, API subject to change.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -67,6 +67,8 @@ gem 'paranoia'
 gem 'will_paginate'
 gem 'bulk_insert'
 
+gem 'rest-client'
+
 gem 'devise'
 gem 'devise_cas_authenticatable'
 
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,8 @@ group :test do
   gem 'selenium-webdriver'
   # Easy installation and use of web drivers to run system tests with browsers
   gem 'webdrivers'
+  # HTTP request mocking
+  gem 'webmock', require: false
 
   # Report test coverage
   gem 'codacy-coverage'
--- a/lib/canvas_api.rb
+++ b/lib/canvas_api.rb
@@ -1,0 +1,38 @@
+require 'rest-client'
+
+class CanvasAPI
+
+  DEFAULT_CANVAS_URL = 'https://portal.bebraven.org/api/v1'
+
+  def initialize(canvas_url, auth_token)
+    @canvas_url = canvas_url || DEFAULT_CANVAS_URL
+    @global_headers = {
+      'Authorization' => "Bearer #{auth_token}",
+    }
+  end
+
+  public
+
+  def update_course_page(course_id, wiki_page_id, wiki_page_body)
+    body = {
+      'wiki_page[body]' => wiki_page_body,
+    }
+
+    put("/courses/#{course_id}/pages/#{wiki_page_id}", body)
+  end
+
+  private
+
+  def get(path, params, headers={})
+    RestClient.get("#{@canvas_url}#{path}", params, @global_headers.merge(headers))
+  end
+
+def post(path, body, headers={})
+    RestClient.post("#{@canvas_url}#{path}", body, @global_headers.merge(headers))
+  end
+
+  def put(path, body, headers={})
+    puts @global_headers
+    RestClient.put("#{@canvas_url}#{path}", body, @global_headers.merge(headers))
+  end
+end
--- a/lib/canvas_api.rb
+++ b/lib/canvas_api.rb
@@ -29,7 +29,6 @@ class CanvasAPI
   end
 
   def put(path, body, headers={})
-    puts @global_headers
     RestClient.put("#{@api_url}#{path}", body, @global_headers.merge(headers))
   end
 end
--- a/lib/canvas_api.rb
+++ b/lib/canvas_api.rb
@@ -27,7 +27,7 @@ class CanvasAPI
     RestClient.get("#{@canvas_url}#{path}", params, @global_headers.merge(headers))
   end
 
-def post(path, body, headers={})
+  def post(path, body, headers={})
     RestClient.post("#{@canvas_url}#{path}", body, @global_headers.merge(headers))
   end
 
--- a/lib/canvas_api.rb
+++ b/lib/canvas_api.rb
@@ -18,10 +18,8 @@ class CanvasAPI
     put("/courses/#{course_id}/pages/#{wiki_page_id}", body)
   end
 
-  private
-
-  def get(path, params, headers={})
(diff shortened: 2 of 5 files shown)


rshipp:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           

Coverage increased per file
===========================
+ lib/canvas_api.rb  56
+ spec/lib/canvas_api_spec.rb  100
         

Complexity increasing per file
==============================
- lib/canvas_api.rb  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/platform/platform/pullRequest?prid=4502907&bid=15178983)

lib/canvas_api.rb
```diff
@@ -0,0 +1,38 @@
+require 'rest-client'
+
+class CanvasAPI
+
+  DEFAULT_CANVAS_URL = 'https://portal.bebraven.org/api/v1'
```
sadleb: I'm assuming the plan is to initialize this class with the actual canvas URL pulled from config, so having a fallback default seems good b/c it makes it easier to see an example vs just failing without much insight into what is wrong by requiring the value to be passed in. That said, I lean more towards it taking just the domain in case there are other places across the codebase where we'll need to access the domain but not the API. E.g. if we need to add links or something or hit a health check. So I would change the initializer to construct this base URL for the API  by adding `api/v1` to whatever is passed in.

StetsenTech: Can you make this a config in a yml or defined in the config/environments? I can foresee this being a problem when we want to do local development.

lib/canvas_api.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  public
+
+  def update_course_page(course_id, wiki_page_id, wiki_page_body)
+    body = {
+      'wiki_page[body]' => wiki_page_body,
+    }
+
+    put("/courses/#{course_id}/pages/#{wiki_page_id}", body)
```
sadleb: It's unfortunate that HTTP put made me see this as Ruby `puts` when I first saw it. Just noting what my brain did when I saw this. I was wondering if we should call it `upload` but prob leaving it as `put` is best... I'm torn.

lib/canvas_api.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def get(path, params, headers={})
+    RestClient.get("#{@canvas_url}#{path}", params, @global_headers.merge(headers))
```
sadleb: Do we want to standard error handling in any way? Prob best just to let everything bubble up for now, but when we get to writing tests we should def come up with a robust and consistent way for error handling when using this API.

Gemfile
```diff
@@ -67,6 +67,8 @@ gem 'paranoia'
 gem 'will_paginate'
 gem 'bulk_insert'
 
+gem 'rest-client'
```
StetsenTech: I'm a bit hesitant in adding more gems to this, but this is probably far better than using `Net::HTTP::Post` etc...

lib/canvas_api.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+def post(path, body, headers={})
```
sadleb: Spacing....

lib/canvas_api.rb
rshipp: yeah @sadleb @StetsenTech I just didn't know how to import this from secrets.yml or wherever it should be.

As for the domain part, I can change it to construct from a base URL without the `/api/v1` part, that makes sense.

lib/canvas_api.rb
rshipp: yikes! idk how i missed that.

lib/canvas_api.rb
rshipp: I typed it wrong a couple times too 😆 but `put` is standard and matches the naming convention of `rest-client`, so I'm hesitant to change it.

sadleb:
Just summarizing the offline chat. We're going to wait to approve this until it's fleshed out a bit more with tests, reading from config, etc before merging to staging.

lib/canvas_api.rb
```diff
@@ -0,0 +1,35 @@
...
+  def get(path, params, headers={})
+    RestClient.get("#{@api_url}#{path}", params, @global_headers.merge(headers))
+  end
+
+  def post(path, body, headers={})
+    RestClient.post("#{@api_url}#{path}", body, @global_headers.merge(headers))
+  end
+
+  def put(path, body, headers={})
+    puts @global_headers
```
StetsenTech: Is this for logging purposes?

rshipp: 🙃

spec/lib/canvas_api_spec.rb
```diff
@@ -3,7 +3,27 @@
...
+    expect(WebMock).to have_requested(:get, "#{CANVAS_URL}/api/v1/test").once
+  end
+
+  it 'updates course wiki page' do
+    stub_request(:put, "#{CANVAS_URL}/api/v1/courses/1/pages/test")
+
+    canvas = CanvasAPI.new(CANVAS_URL, 'test-token')
+    canvas.update_course_page(1, 'test', 'test-body')
+
+    expect(WebMock).to have_requested(:put, "#{CANVAS_URL}/api/v1/courses/1/pages/test").
```
sadleb: This WebMock solution seems to be pretty easy. Good stuff!

spec/lib/canvas_api_spec.rb
```diff
@@ -3,7 +3,27 @@
...
 
-  describe 'updates course wiki page' do
-    pending "test the canvas api client"
+  CANVAS_URL = "http://canvas.example.com".freeze
+
+  WebMock.disable_net_connect!
+
+  it 'correctly sets authorization header' do
+    stub_request(:any, /#{CANVAS_URL}.*/).
+      with(headers: {'Authorization'=>'Bearer test-token'})
```
sadleb: Just trying to understand -- since you're only defining a stub_request that has this header, if the CanvasAPI didn't correctly set it to this then the stub wouldn't be hit and the `expect(WebMock).to have_requested` below would be false? What happens between tests? E.g. what if a different test defines a different stub_request that isn't so strict. I assume the tests are isolated so only the stubs defined in that test are registered with WebMock and everything is cleared between tests?

rshipp: all those assumptions are correct! the `with(headers:` line will match requests that include that header and (optionally) any other headers. the `expect` will fail if the method wasn't hit, or was hit without the auth header. and stubs are scoped inside individual `it` blocks.

rshipp:
(merged with #32)

PR #31, opened 2019-11-14
Add Sarah's UI draft

rshipp:
Task: https://app.asana.com/0/1143269577537247/1148998666960360/f

Summary: Copied in the HTML Sarah gave me (https://app.asana.com/0/0/1147724327259437/f), and added raw HTML-edit functionality and working tabs.

(Note: this PR is on top of #29.)

PR #29, opened 2019-11-14
Refactor to use new CourseContent model

rshipp:
Task: https://app.asana.com/0/1143269577537246/1148998666960365/f

Rename `Content` to `CourseContent`, and add some new columns to track metadata.

PR #32, opened 2019-11-15
Implement publish content button

rshipp:
Task: https://app.asana.com/0/1143269577537246/1148998666960354/f

Add a dynamic publish button that pushes to canvas through the API.

Canvas API now pulls from config instead of hardcoded.

(Note: this is based on #31 with cherry-picked commit from #30.)


app/controllers/course_contents_controller.rb
```diff
@@ -61,6 +61,20 @@ def destroy
     end
   end
 
+  # POST /course_contents/1/publish
+  # POST /course_contents/1/publish.json
+  def publish
```
rshipp: new route!

app/controllers/course_contents_controller.rb
```diff
@@ -69,6 +83,6 @@ def set_course_content
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def course_content_params
-      params.require(:course_content).permit(:title, :body, :published_at, :content_type)
+      params.require(:course_content).permit(:title, :body, :published_at, :content_type, :course_id, :course_name, :secondary_id)
```
rshipp: forgot to permit these attributes with #29!

app/javascript/components/ContentEditor.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
 import CKEditor from '@ckeditor/ckeditor5-react';
+import Rails from '@rails/ujs';
```
rshipp: 🛤 this lets us call Rails HTTP APIs more easily

app/javascript/components/ContentEditor.js
```diff
@@ -144,6 +145,7 @@ class ContentEditor extends Component {
             // The initial editor data. It is bound to the editor instance and will change as
             // the user types and modifies the content of the editor.
             editorData: props.course_content['body'] || "",
+            isPublished: false
```
rshipp: React manages state for us! Eventually I'll update this to pull real `isPublished` state from the database instead of just setting it to `false` by default.

app/javascript/components/ContentEditor.js
```diff
@@ -170,15 +172,17 @@ class ContentEditor extends Component {
     // It updates the state of the application.
     handleEditorDataChange( evt, editor ) {
         this.setState( {
-            editorData: editor.getData()
+            editorData: editor.getData(),
+            isPublished: false
```
rshipp: when we make changes to the content, update the `isPublished` state to `false`.

app/javascript/components/ContentEditor.js
```diff
@@ -196,6 +200,31 @@ class ContentEditor extends Component {
         CKEditorInspector.attach( editor );
     }
 
+    handlePublish( evt ) {
+        fetch("/course_contents/1/publish.json", {
```
rshipp: es6 `fetch`/`Promises`! i'm not sure how to use es2017 async/await; I can look into it if yall think we should be doing that instead.

app/javascript/components/ContentEditor.js
```diff
@@ -196,6 +200,31 @@ class ContentEditor extends Component {
...
+                method: 'POST',
+                body: JSON.stringify(this.props.course_content), // data can be `string` or {object}!
+                headers: {
+                    'Content-Type': 'application/json',
+                    'X-CSRF-Token': Rails.csrfToken()
```
rshipp: this is where `@rails/ujs` saves us by providing the csrfToken, so we don't get "invalid authenticity token" errors when POSTing to the publish endpoint

app/javascript/components/ContentEditor.js
```diff
@@ -196,6 +200,31 @@ class ContentEditor extends Component {
...
+                }
+            })
+            .then(res => res.json())
+            .then(
+                (result) => {
+                    this.setState( {'isPublished': true} );
+                    // TODO: provide a way for the user to easily view the result, which will be in:
+                    // `${CANVAS_URL}/courses/${result['course_id']}/pages/${result['secondary_id']}`
```
rshipp: i'll figure out some way to pass this in once we decide how we're going to handle this flow ux-wise

app/javascript/components/ContentEditor.js
```diff
@@ -210,7 +239,9 @@ class ContentEditor extends Component {
                     <span id="autosave-indicator" className="saving">Saving...</span>
 
                     <ul>
-                        <li>Publish</li>
+                        <li onClick={(evt) => this.handlePublish(evt)} className={this.state['isPublished'] ? "success" : ""}>
+                          Publish{this.state['isPublished'] ? "ed" : ""} {this.state['isPublished']}
```
rshipp: please ignore this mess for now, I'll refactor once I have a chance to talk with Sarah

app/javascript/packs/application.js
```diff
@@ -7,6 +7,8 @@
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
 
+import Rails from '@rails/ujs';
+Rails.start();
```
rshipp: have to initialize rails_ujs here in the entrypoint.

app/models/course_content.rb
```diff
@@ -1,2 +1,16 @@
...
+    canvas = CanvasAPI.new(Rails.application.secrets.canvas_url,
+        Rails.application.secrets.canvas_token)
+
+    response = canvas.update_course_page(params[:course_id], params[:secondary_id], params[:body])
+
+    if response.code == 200 then
+      true
+    else
+      false
+    end
```
rshipp: this is ugly and could be a 1-liner but I think we'll have more complex logic here later so I didn't bother to condense it.

app/views/course_contents/_course_content.json.jbuilder
```diff
@@ -1,2 +1,3 @@
-json.extract! course_content, :id, :title, :body, :published_at, :content_type, :created_at, :updated_at
+json.extract! course_content, :id, :title, :body, :published_at, :content_type, :created_at, :updated_at, :course_id,
+        :course_name, :secondary_id
```
rshipp: another missed one from #29, need to make sure these columns get into the json builder

config/routes.rb
```diff
@@ -1,5 +1,8 @@
 Rails.application.routes.draw do
-  resources :course_contents
+  resources :course_contents do
+    post :publish
```
rshipp: hook up the new route!

app/javascript/components/ContentEditor.js
StetsenTech: Nice! If you ever need to handle CSRF Tokens on a form in an ERB fail, you can add this line to the form:
`<%= hidden_field_tag :authenticity_token, form_authenticity_token -%>`

StetsenTech:
👍 Fine to merge into the draft branch.

app/javascript/components/ContentEditor.js
```diff
@@ -196,6 +200,31 @@ class ContentEditor extends Component {
...
+                (result) => {
+                    this.setState( {'isPublished': true} );
+                    // TODO: provide a way for the user to easily view the result, which will be in:
+                    // `${CANVAS_URL}/courses/${result['course_id']}/pages/${result['secondary_id']}`
+                },
+                // Note: it's important to handle errors here
+                // instead of a catch() block so that we don't swallow
+                // exceptions from actual bugs in components.
+                (error) => {
+                    console.log(error);
```
sadleb: This is a bigger topic and outside the scope of this pull request, but I think you should queue up another task to decide on and implement a consistent way to do error handling in javascript. For now, having it in the console log seems fine but as we get closer to something production ready, we should probably be having JS errors flow into our various error monitoring tools like Sentry and to show up in the application logs. I'd be curious to hear other's thoughts on this though. What's best practices for monitoring app health as it relates to JS errors happening?

app/javascript/packs/application.js
sadleb: I'm confused about what this is actually doing. Is it loading the entire Rails app once per browser session? Or is it someone connecting to the existing running Rails app?

app/models/course_content.rb
```diff
@@ -1,2 +1,12 @@
+require 'canvas_api'
+
 class CourseContent < ApplicationRecord
+  def publish(params)
+    canvas = CanvasAPI.new(Rails.application.secrets.canvas_url,
```
sadleb: Do you see any reason to initialize a CanvasAPI object with the target URL/token each time we call it, or should the config info be stored statically on the class (not the instance)?

app/models/course_content.rb
```diff
@@ -1,2 +1,12 @@
...
+        Rails.application.secrets.canvas_token)
+
+    response = canvas.update_course_page(params[:course_id], params[:secondary_id], params[:body])
```
sadleb: I really think we should rename `secondary_id` to be more specific. Is this is page ID? The module ID? The assignment ID? All of the above?

app/javascript/components/ContentEditor.js
rshipp: I'd need to look more into what kind of error can show up here, but yeah I think sending to sentry is probably what we'll want to do once we tie that in. I don't have any tasks for that kind of work on the boards, so I was assuming we'd tackle it after the MVP is done, in december-january. For the time being, since this is in development, console.log is the easiest place for me to see things.

app/javascript/packs/application.js
rshipp: No this is just loading the [ujs module](https://github.com/rails/rails/tree/master/actionview/app/assets/javascripts), the `Rails.start()` method just initializes some JS stuff https://github.com/rails/rails/blob/master/actionview/app/assets/javascripts/rails-ujs/start.coffee so we can call use its `csrfToken()` and whatever else in our React components.

app/models/course_content.rb
rshipp: we could save `canvas` in the class, it just doesn't have much benefit since it's stateless and we're not taking advantage of HTTP sessions or anything.

app/models/course_content.rb
rshipp: left a comment on this in https://github.com/bebraven/platform/pull/29/commits/cd3cd1e336d905695f7c818a6c3bf2f4216dbddd - the name is intentionally vague, because what it points to changes depending on the `content_type`. That said, I think this is a good candidate for refactoring once we get to table design stuff, because it's basically relying on an `enum`, which isn't great practice.

app/javascript/components/ContentEditor.js
rshipp: i made a subtask for adding Sentry in https://app.asana.com/0/1133543009734854/1143521519868577/f

app/models/course_content.rb
sadleb: I'm not thinking in terms of performance regarding CPU/memory. I'm thinking in terms of the number of places in the code we'll end up writing `CanvasAPI.new(Rails.application.secrets.canvas_url, Rails.application.secrets.canvas_token` if we have Canvas around for awhile and slowly cutover functionality that has to interact with Canvas. Seems like any place in the code that wants to call into the Canvas app using the CanvasAPI, should just be able to say `CanvasAPI.new` and have it just work based on the overall config.  Or even just call static methods like `CanvasAPI::update_course_page(...)` without having to always pass in the config values which, when changed, should happen app-wide and require a restart of the app.

rshipp: ahh interesting, ok. I don't know how to have the Rails config inside the CanvasAPI class, I was assuming anything in `lib/` shouldn't know about Rails at all. Is that wrong and I should just be importing Rails there and setting up the config that way? Or is this something that would go in an initializer somehow?

sadleb: Basically, the config of where CanvasAPI calls out to should be controlled inside that CanvasAPI class and not require clients of that class to know the details because we're not going to have different code paths call out to different canvas servers.... lol, after I wrote this, I realized why this may not be true. We may have most every code path call to the production canvas instance, but for Content Editor, it may want to call out to staging or some other server. Given that, if you don't specify config values, I think there should be a default way to use that CanvasAPI without specifying the target server where it uses production. For ContentEditoer, which may need to call into the staging server, either allow the caller to specify the `canvas_url` and `canvas_token` or just pass an enum/something for the server/environment it's targeting and let the `CanvasAPI` read from config to get the connection info for that server.

rshipp: yeah we will need prod and staging access for contenteditor. alright I think I can get that set up to pull from config inside of canvas_api.rb instead of the model.

sadleb: > I was assuming anything in `lib/` shouldn't know about Rails at all. Is that wrong and I should just be importing Rails there and setting up the config that way? Or is this something that would go in an initializer somehow?

Ideally, code in lib wouldn't reference Rails and we would either have it load directly from config or have Rails initializer initialize it. I'm not sure how much of a stickler we should be regarding that though. I'd love to hear from @hcli-work who has had more experience with real-world Rails apps and their structure.

hcli-work: The general belief is that stuff in `lib` shouldn't have knowledge of your specific application. If it's tweaking your models, presenters, and such, then it probably belongs somewhere in your actual `app` directory. If, however, it just has generic knowledge of Rails, that's fine. The exception to this is in the `lib/tasks` directory where app specific rake tasks live.

That said, my previous team definitely violated this purity occasionally and it wasn't a terribly big deal. It made things like Guard and Spring harder to use as those things didn't reload `lib` files proactively the way they reloaded files under `app`

rshipp: Oh, like just load the YAML config directly inside lib instead of loading it through Rails? That still feels a little non-SRP to me, but we'll also be getting rid of the canvas stuff soon enough so I'm not too worried about it.

rshipp: @sadleb check out https://github.com/bebraven/platform/pull/32/commits/95e783ebfac0fd5d059fa39afbe4f41e94df0563 and let me know what you think

app/models/course_content.rb
```diff
@@ -1,11 +1,6 @@
-require 'canvas_api'
-
 class CourseContent < ApplicationRecord
   def publish(params)
-    canvas = CanvasAPI.new(Rails.application.secrets.canvas_url,
-        Rails.application.secrets.canvas_token)
-
-    response = canvas.update_course_page(params[:course_id], params[:secondary_id], params[:body])
+    response = CanvasProdClient.update_course_page(params[:course_id], params[:secondary_id], params[:body])
```
rshipp: calling this `CanvasProdClient` in anticipation of adding a `CanvasStagingClient` in a couple weeks

config/initializers/canvas_api.rb
```diff
@@ -0,0 +1,4 @@
+require 'canvas_api'
+
+canvas_config = YAML.safe_load(ERB.new(File.read("#{Rails.root.to_s}/config/canvas.yml")).result, [], [], true)[Rails.env].with_indifferent_access
+::CanvasProdClient = CanvasAPI.new(canvas_config[:canvas_url], canvas_config[:canvas_token])
```
sadleb: Being a ruby / rails noob myself, I'm not familiar with this syntax. Is this a global variable? Or is this just sort of an alias where when you call `CanvasProdClient` it calls `CanvasAPI.new` under the covers? Either way, this seems like a more maintainable way to use the API instead of reading the config and creating the instance yourself everywhere you want to use it.

rshipp: It's definitely some kind of variable, but I'm not sure how the syntax works exactly. This is where I got it from: https://stackoverflow.com/questions/23588674/rails-how-to-set-global-instance-variables-in-initializers.

sadleb: Results from a little bit of digging on what this syntax means. Anything that starts with a capital letter is a constant in Ruby. So this is defining a `CanvasProdClient` constant (which funny enough can change, but will warn instead of fail unless you add a `.freeze`). The `::` just means top level scope, aka global scope. I'm not sure it's strictly necessary here b/c this is outside of a class or module definition, but if not it at least makes it clear that it's meant to be global scope @rshipp @StetsenTech

PR #33, opened 2019-11-18
First Round Tests and Refactors for CAS Login

StetsenTech:
**Ticket(s)**:

**Description**:
This is the first round of tests and refactors to the login route. Now that I have been exposed to Rails and done some reading, I understand what needs to be done. This adds environment application configs, adds missing db columns for RubyCAS Core ActiveRecord gem, cleans up Double Render errors, removes invalid code, and some styling changes. 

**Summary**:
- Adds environment based configs for RubyCAS YML
- Updates migration schema
- Refactors some logic from the original brought over code
- Brings over BeyondZ authenticator since we will be using the system for a bit
- Removes code that is not longer valid in the transfer Sinatra to Rails
- Adds tests for `/login`

### DIFF
--- a/app/controllers/cas_controller.rb
+++ b/app/controllers/cas_controller.rb
@@ -163,7 +163,6 @@ class CasController < ApplicationController
         logger.info("Credentials for username '#{@username}' successfully validated using #{successful_authenticator.class.name}.")
         logger.debug("Authenticator provided additional user attributes: #{extra_attributes.inspect}") unless extra_attributes.blank?
 
-        # 3.6 (ticket-granting cookie)
         tgt = generate_ticket_granting_ticket(@username, extra_attributes)
         response.set_cookie('tgt', tgt.to_s)
 
@@ -190,15 +189,17 @@ class CasController < ApplicationController
         end
       else
         logger.warn("Invalid credentials given for user '#{@username}'")
-        @message = {:type => 'mistake', :message => "Incorrect username or password."}
-        render :json => @message, status: :unauthorized
+        @message = {
+          :type => 'mistake', :message => "Incorrect username or password."
+        }
+        return render :login, status: :unauthorized
       end
     rescue RubyCAS::Server::Core::AuthenticatorError => e
       logger.error(e)
       # generate another login ticket to allow for re-submitting the form
       @lt = LT.generate_login_ticket(c).ticket
       @message = {:type => 'mistake', :message => e.to_s}
-      render :json => @message, status: :unauthorized
+      return render :login , status: :unauthorized
     end
 
     render :login
--- a/app/controllers/cas_controller.rb
+++ b/app/controllers/cas_controller.rb
@@ -96,9 +96,9 @@ class CasController < ApplicationController
       @form_action = params['submitToURI'] || guessed_login_uri
 
       if @form_action
-        render :login_form
+        return render :_login_form
       else
-        render :json => {:response => "Could not guess the CAS login URI. Please supply a submitToURI parameter with your request."}, status: :internal_server_error
+        return render :json => {:response => "Could not guess the CAS login URI. Please supply a submitToURI parameter with your request."}, status: :internal_server_error
       end
     else
       render :login
@@ -212,7 +212,8 @@ class CasController < ApplicationController
 
     # BZ modification: always use default service so logout/login goes back to our main
     # site (which can redirect) regardless of where they came from
-    @service = Utils.clean_service_url(params['service'] || 'http://canvasweb/login/cas') # params['service'] || params['destination'])
+    settings = RubyCAS::Server::Core::Settings._settings
+    @service = Utils.clean_service_url(params['service'] || settings[:default_service]) # params['service'] || params['destination'])
     @continue_url = params['url']
 
     @gateway = params['gateway'] == 'true' || params['gateway'] == '1'
@@ -260,21 +261,11 @@ class CasController < ApplicationController
     @lt = generate_login_ticket(c)
 
     if @gateway && @service
-      # I'm changing this to render the logout instead of redirecting
-      # because the redirection wouldn't give a chance for the logout
-      # iframe trick to log out of the main site. Since we also have
-      # a default service now in the configuration, it isn't as important
-      # for that to be preserved anyway - it will send them back to the
-      # default service which is good for us too as we can then handle
-      # the redirect based on the user account type automatically.
-
-      # redirect_to @service, 303
-      render :login
+      return render :login
     elsif @continue_url
-      render :logout
-    else
-      render :login
+      return render :logout
     end
+    render :login
   end
 
   def loginTicket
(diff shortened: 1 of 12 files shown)


app/controllers/cas_controller.rb
```diff
@@ -46,55 +50,57 @@ def login
...
     if params.has_key? 'onlyLoginForm'
       if request.env['HTTP_HOST']
-        guessed_login_uri = "#{request.env['HTTPS']}://#{request.env['REQUEST_URI']}#{self / '/login'}"
+        @protocol = request.env['HTTPS'] == 'on' ? 'https' : 'http'
+        guessed_login_uri = "#{@protocol}://#{request.env['REQUEST_URI']}/login"
       else
         guessed_login_uri = nil
       end
 
       @form_action = params['submitToURI'] || guessed_login_uri
```
StetsenTech: I'm thinking of removing this logic for guessing a form action if they `onlyLoginForm`. If someone specifies onlyLoginForm I would prefer if they also specify `submitToURI`. I never want to be in a situation of trying to guess what URL someone else wants (especially for external systems).

app/controllers/cas_controller.rb
```diff
@@ -103,42 +109,33 @@ def login
...
+    c = request.env['HTTP_X_FORWARDED_FOR'] || request.env['REMOTE_HOST'] || request.env['REMOTE_ADDR']
 
-    if error = validate_login_ticket(@lt)
+    if !result = validate_login_ticket(@lt)
       @message = {:type => 'mistake', :message => error}
       # generate another login ticket to allow for re-submitting the form
-      c = request.env['HTTP_X_FORWARDED_FOR'] || request.env['REMOTE_HOST'] || request.env['REMOTE_ADDR']
       @lt = LT.generate_login_ticket(c).ticket
-      return render :login, status: :internal_server_error
+      return render :login, status: :unauthorized
```
StetsenTech: I hate returning 500s in general. If you know there is a problem and purposely doing a return, then it needs to be an error in the 400s.

app/controllers/cas_controller.rb
```diff
@@ -258,25 +252,14 @@ def logout
...
     elsif @continue_url
-      render :logout
-    else
-      render :login
+      return render :logout
     end
+    render :login
   end
 
   def loginTicket
```
StetsenTech: Anything below this point in this file is going to be in another PR. I made these changes so I don't forget when I start working on tests for these.

config/rubycas.yml
```diff
@@ -1,26 +1,76 @@
...
+  cookie_domain: .localhost 
+  organization: "RSPEC-TEST"
+  infoline: "This is an rspec test."  
+  theme: simple
+  database:
+    <<: *default_db
+    database: platform_test
+  authenticator:
+    <<: *default_bz_auth
+    server: joinweb
```
StetsenTech: I may point this to `platform_test` instead...or I may just make another authenticator. 🤔

lib/beyondz.rb
```diff
@@ -0,0 +1,30 @@
...
+      request = Net::HTTP::Post.new('/users/check_credentials')
+      request.set_form_data(
+        'username' => credentials[:username],
+        'password' => credentials[:password]
+      )
+      response = http.request(request)
+      return (response.body == "true")
+    end
+  end
+end
```
StetsenTech: Brought directly from SSO and the class it inherits from.

app/controllers/cas_controller.rb
rshipp: I think that'd be a good change, if we're sure it won't break anything.

app/controllers/cas_controller.rb
```diff
@@ -103,42 +109,33 @@ def login
...
     @password = params['password']
     @lt = params['lt']
 
-    # Remove leading and trailing widespace from username.
-    @username.strip! if @username
-
-    if @username && settings.config[:downcase_username]
-      logger.debug("Converting username #{@username.inspect} to lowercase because 'downcase_username' option is enabled.")
-      @username.downcase!
-    end
```
rshipp: thanks for cleaning up this logic

app/controllers/cas_controller.rb
rshipp: will this only happen when the credentials were actually wrong? if it could be an upstream communication error with join or something, I think it makes sense to have a 500

app/controllers/cas_controller.rb
```diff
@@ -191,16 +183,17 @@ def loginpost
...
       end
-    rescue Core::Authenticator::AuthenticatorError => e
+    rescue RubyCAS::Server::Core::AuthenticatorError => e
       logger.error(e)
       # generate another login ticket to allow for re-submitting the form
-      c = request.env['HTTP_X_FORWARDED_FOR'] || request.env['REMOTE_HOST'] || request.env['REMOTE_ADDR']
       @lt = LT.generate_login_ticket(c).ticket
       @message = {:type => 'mistake', :message => e.to_s}
-      render :json => @message, status: :unauthorized
+      return render :login , status: :unauthorized
```
rshipp: are we no longer using `@message` here?

config/rubycas.yml
```diff
@@ -1,26 +1,76 @@
...
 default: &default
-  log:
-    output: log/casserver.log
-    level: INFO
   maximum_unused_login_ticket_lifetime: 300
   maximum_unused_service_ticket_lifetime: 300
   maximum_session_lifetime: 172800
   maximum_remember_me_lifetime: 604800
-  
+  infoline: Powered by <a href="http://rubycas.github.com">RubyCAS-Server-Core</a>
```
rshipp: do we need this? it's just extra fingerprinting info

rshipp:
Looks like the tests are complaining because they don't know what `SSO_URL` to use

app/controllers/cas_controller.rb
StetsenTech: `@message` and most variable available in scope of the route are passed to render when using an ERB view.

config/rubycas.yml
StetsenTech: Don't really need it. It's pretty common practice say what powers your system. I left it in because it is used in the login route. 

I think I am going to strip out some things in the view pages so we can have a fresh start.

app/controllers/cas_controller.rb
rshipp: ahhh ok, didn't know that

config/rubycas.yml
rshipp: oh does this show up on the login page or something? I was thinking it was just a version banner.

app/controllers/cas_controller.rb
StetsenTech: To clarify and be more specific, the variables with `@` will get passed in.

app/controllers/cas_controller.rb
StetsenTech: This particular path only happens if you fail authentication.

StetsenTech:
I'm going to go ahead and merge this into the feature branch. Feel free to still leave comments and I will make the updates in a future PR.

PR #36, opened 2019-11-20
Adds Tests for CAS Service Validate

StetsenTech:
**Ticket(s)**:

**Summary**:
- Adds test for /serviceValidate
- Cleans up and converts logic into Rails format


app/controllers/cas_controller.rb
```diff
@@ -301,30 +301,32 @@ def validate
...
     @renew = params['renew']
 
+    c = request.env['HTTP_X_FORWARDED_FOR'] || request.env['REMOTE_HOST'] || request.env['REMOTE_ADDR']
+
     st, @error = validate_service_ticket(@service, @ticket)
-    @success = st && !@error
+
+    if @error
+      return render json: {error: @error}, status: :unprocessable_entity
+    end
```
hcli-work: `return render(json: {error: @error}, status: :unprocessable_entity) if @error`

app/controllers/cas_controller.rb
```diff
@@ -301,30 +301,32 @@ def validate
...
+    c = request.env['HTTP_X_FORWARDED_FOR'] || request.env['REMOTE_HOST'] || request.env['REMOTE_ADDR']
+
     st, @error = validate_service_ticket(@service, @ticket)
-    @success = st && !@error
+
+    if @error
+      return render json: {error: @error}, status: :unprocessable_entity
+    end 
+
+    @success = !st.nil?
```
hcli-work: Is there a chance that `st` can be the `false` object? If not, then a more idiomatic way to write this is
```
@success = !!st
```

app/controllers/cas_controller.rb
StetsenTech: Oh snap! I know this syntax, but didn't think of using it for this!

app/controllers/cas_controller.rb
StetsenTech: Found what I have seen, it's either an object or `nil`.

app/controllers/cas_controller.rb
```diff
@@ -301,30 +301,32 @@ def validate
...
 
     if @success
       @username = st.username
       if @pgt_url
         pgt = generate_proxy_granting_ticket(@pgt_url, st, c)
         @pgtiou = pgt.iou if pgt
       end
-      @extra_attributes = st.granted_by_tgt.extra_attributes || {}
+      tgt = TGT.find_by({id: st.ticket_granting_ticket_id})
```
hcli-work: you can leave the `{}` off

PR #38, opened 2019-11-21
Refactors Some Logic on CAS Controller

StetsenTech:
**Ticket(s)**:

**Summary**:
- Simplifies some logic on CAS Controller
- Refactors calls to common parameters
- Fixes a test
- Updates migration


app/controllers/cas_controller.rb
```diff
@@ -4,6 +4,10 @@
 require 'rubycas-server-core/tickets/generations'
 
 class CasController < ApplicationController
+  before_action :set_settings
```
sadleb: Something about this feels wrong. Off the top of my head, I don't know the most Railsy way to do this, but the settings are initialized once when the app is started and the controller should just access that instance methinks. Can you just include the `RubyCAS::Server::Core` module and then access `Settings._settings` directly? Also, if we need to expose something to the view, you can use a `helper_method`. E.g. 

```
def login_domain
   RubyCAS::Server::Core::Settings._settings[:public_site_domain]
end
helper_method :login_domain
```
and then in the view do:
```<a href="//<%= login_domain %>/signup/new">Sign up</a><br />```

File: app/views/cas/_login_form.erb [hunk 2]
```diff
@@ -7,7 +7,7 @@
      Since it is a cross-domain, client side cookie, we had to implement this
      using a helper method on the other side and an iframe on this side.
 -->
-  <iframe style="visibility: hidden; width: 1px; height: 1px;" src="//<%= @domain %>/users/clear_session_cookie"></iframe>
+  <iframe style="visibility: hidden; width: 1px; height: 1px;" src="//<%= @settings[:public_site_domain] %>/users/clear_session_cookie"></iframe>
```
sadleb: See my comment above. There is so much in all of this RubyCAS stuff that isn't really "the rails way", so I'm just picking on this one part to maybe cleanup. The view's should be pretty dumb and just know how to layout the page without needing to know the guts of how the controller (and backing RubyCAS) gem stores stuff. By having it need to know that there is a settings object and the key into the settings object for the domain is called `public_site_domain `, we're breaking down the separation of concerns. Ideally, the view would just link to something like `clear_session_cookie_path`, but maybe that's overkill to add those for all the paths in the view and it would suffice just to expose the domain as a helper method.

app/views/cas/login.erb
```diff
@@ -15,19 +15,19 @@
             response.set_cookie("login_flash", {
               :value => '',
               :expires => Time.now,
-              :domain => @cookie_domain
+              :domain => @settings[:cookie_domain]
```
sadleb: If you end up going with my recommendation, we would end up with two helper methods. `login_domain` and `cookie_domain` and the controller would directly pull those out of the CAS settings.

StetsenTech: I am going with your suggestion for sure.

app/controllers/cas_controller.rb
StetsenTech: I definitely feel you here. It isn't railsy. I couldn't figure out how to function properly. I originally had this broken out into small lib functions, but had a hard time getting everything to communicate properly. I spent a lot of time trying to get the settings to expose as `:cas_settings` throughout the application, but one reason or another, it either was not set at the current time or information was not loaded into the settings. I made the decision to make this draft instead of halting progress on the project any further because I wanted to get things perfect.

I do not consider this the final draft of what I am doing. I mostly did it this so 1) things functional so we can cutover SSO and  2) for testing so that I could confidently refactor without breaking what needs to work.

StetsenTech: Clean up task:
https://app.asana.com/0/1133543009734854/1151447176069705/f

app/views/cas/_login_form.erb
StetsenTech: Clean up task:
https://app.asana.com/0/1133543009734854/1151447176069705/f

app/views/cas/login.erb
StetsenTech: Clean up task:
https://app.asana.com/0/1133543009734854/1151447176069705/f

StetsenTech:
Merging into feature branch. Clean up tasks created for comments.

PR #39, opened 2019-11-21
Adds Tests for Logout

StetsenTech:
Ticket(s):

**Summary**:
- Adds quick and dirty tests for logout


spec/routing/cas/logout_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
+require "capybara_helper"
+
+include ERB::Util
+include Rack::Utils
+
+
+VALID_USERNAME = "platform_user"
```
sadleb: I know these tests were quick and dirty, but something feels off here. Are these users that were setup in the test database? Either way, we should have a more maintainable way of setting up valid users than hard-coding known values across various specs. The same place that sets up valid users for testing should be used to retrieve them.

spec/routing/cas/logout_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+VALID_PASSWORD = "rspec_test"
+
+INVALID_USERNAME = "bad_user"
+INVALID_PASSWORD = "bad_pass"
+
+RETURN_SERVICE = "http://braven/"
```
sadleb: Any reason why this is stored in a variable when it's just used in one spot and the actual value doesn't matter?

sadleb: Any reason why this is stored in a variable when it's just used in one spot and the actual value doesn't matter?

spec/routing/cas/logout_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        fill_in "password", :with => VALID_PASSWORD
+        
+        # Make sure there is a login button that can be clicked
+        # Capybara with Selenium can have problems with javascript
+        expect(page).to have_button("login-submit", disabled: false)
+        click_button "login-submit"
+
+        # Ensure that the login was successful
+        visit "/logout"
+        expect(page).to have_content('You have successfully logged out')
```
sadleb: I think we should look for the cookie being cleared and/or the response code instead of relying on the text of the page. If we ever change the logout page's content, this test will break when it shouldn't.

spec/routing/cas/logout_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        visit "/login?service=#{url_encode(RETURN_SERVICE)}"
+
+        # Input login credentials
+        fill_in "username", :with => VALID_USERNAME
+        fill_in "password", :with => VALID_PASSWORD
+        
+        # Make sure there is a login button that can be clicked
+        # Capybara with Selenium can have problems with javascript
+        expect(page).to have_button("login-submit", disabled: false)
+        click_button "login-submit"
```
sadleb: Is there a way to have both of these tests share the part of the test that does the actual login? A helper or something?

StetsenTech: It is possible. I would just need to make a `before :each` after the describe.

spec/routing/cas/logout_spec.rb
StetsenTech: Valid point.

spec/routing/cas/logout_spec.rb
StetsenTech: Its used multiple times and in future iterations of testing I want this to actually be a microservice that actively communicates with the test instead of leading to a dead end.

spec/routing/cas/logout_spec.rb
StetsenTech: Currently, this test user is setup on the join server (joinweb on tests). I completely agree with. The more appropriate way is to have fixtures set up that load up when you start running tests. The problem currently is that I need those users on the join server, so it would involve extra steps that I did not want to deal with atm.

StetsenTech:
I'm going to close this PR and refactor this on #38

PR #40, opened 2019-12-03
Brings CSS from SSO Login to Platform CAS Login

StetsenTech:
**Ticket(s)**:
https://app.asana.com/0/1133543009734854/1151392840915737/f

**Description**:
Brings over the CSS from the SSO box and breaks it out into header and footer partials


sadleb:
I don't think the TradeGothic font stuff (4 files) belongs in the `cas` subfolder. I assume that's our Braven branded fonts and stuff which we'll end up applying across the entire platform codebase.

sadleb:
This stackoverflow article appears to give decent guidance on where to put fonts (but maybe check the rails 6 docs to make sure nothing has changed): https://stackoverflow.com/a/10907276

StetsenTech:
Valid point. I made a clean up task for this:
https://app.asana.com/0/1133543009734854/1152808652553869/f

PR #41, opened 2019-12-04
Refactors Login Tests

StetsenTech:
**Ticket(s)**:
https://app.asana.com/0/1143269577537245/1152497774391761/f

**Description**:
This is the first step of a larger refactor on how tests should be set up. This leverages the `before(:each)` and `let` functions to setup and dynamically alter how tests are run per tests basis. This also is an example of how you break out tests into sections via `describe`, `context` and `its`.

**Summary**:
- Fixes authentication redirects with CAS routes for tests issue
- Removes JSON from paths that generate XMLs
- Refactors login spec


spec/controllers/cas/login_spec.rb
```diff
@@ -4,113 +4,94 @@
...
+
+          # Revisit login page to validate current ticket
+          visit "/cas/login?service=#{url_encode(return_service)}"
+
+          # New ticket should be generated
+          expect(current_url).not_to include(@params["ticket"])
+        end
+      end
+    end
+    describe "/cas/login sing onlyLoginForm parameter" do
```
rshipp: u 🔍sing

StetsenTech: 👁 c u

app/controllers/application_controller.rb
```diff
@@ -22,4 +26,11 @@ def authorized_by_token?
     
     !!AccessToken.find_by(key: key)
   end
+
+  def cas_ticket?
+    ticket = params[:ticket] || nil
```
hcli-work: why isn't this just `ticket = params[:ticket]`? if nothing is in `params`, it'll give you `nil` anyway, no?

app/controllers/application_controller.rb
```diff
@@ -22,4 +26,11 @@ def authorized_by_token?
...
+    return false if ticket.nil?
+
+    !!ST.find_by(ticket: ticket)
```
hcli-work: instead of incurring the memory overhead of constructing an ActiveRecord model, you could use `#exists?` instead. https://apidock.com/rails/ActiveRecord/FinderMethods/exists%3F

spec/controllers/cas/login_spec.rb
```diff
@@ -4,113 +4,94 @@
...
-        @params = parse_query(current_url, "&?,")
-        expect(@params).to include("ticket")
-
-        # Revisit login page to validate current ticket
-        visit "/cas/login?service=#{url_encode(RETURN_SERVICE)}"
-
-        # New ticket should be generated
-        expect(current_url).not_to include(@params["ticket"])
+    describe "/cas/login without service url" do
+      let(:return_service) {'http://braven/'}
```
hcli-work: let's be consistent with spaces around the curly braces

spec/controllers/cas/login_spec.rb
```diff
@@ -4,113 +4,94 @@
...
+        
+        it "logs in successfully" do
+          # Ensure that the login was successful
+          expect(current_url).to include(return_service)
+          expect(current_url).to include("ticket")
+        end
+
+        it "validates existing tickets and generates new one" do
+          expect(current_url).to include(return_service)
+          expect(current_url).to include("ticket")
```
hcli-work: this overlaps with previous test

spec/controllers/cas/login_spec.rb
```diff
@@ -4,113 +4,94 @@
...
+
+          # Get current ticket to check against next generated ticket
+          @params = parse_query(current_url, "&?,")
+          expect(@params).to include("ticket")
+
+          # Revisit login page to validate current ticket
+          visit "/cas/login?service=#{url_encode(return_service)}"
+
+          # New ticket should be generated
+          expect(current_url).not_to include(@params["ticket"])
```
hcli-work: it feels like you're testing more than one thing in this `it` block
unless it's horrifically slow, i would break it out into separate blocks to make it clear what each's expectation is

spec/controllers/cas/login_spec.rb
StetsenTech: Agreed.

spec/controllers/cas/login_spec.rb
StetsenTech: 🤔 I see your point. No need to test samething twice if the results are the same.

spec/controllers/cas/login_spec.rb
StetsenTech: I see your point.

app/controllers/application_controller.rb
StetsenTech: So the `rubycas-server-core` gem doesn't actually use ActiveRecord directly. `ST` comes from that gem and is suppose to be database agnostic, so ActiveRecord methods don't actually work on it. That's why I did it this way. The result of this can be `nil`, but wasn't usre if it could return another false value.

That said, the ActiveRecord adapter for this gem does create a model for you to use. I will instead use that implementation. I just need to use `ServiceTicket` instead of `ST` then I should be able to use `.exist?`

app/controllers/application_controller.rb
StetsenTech: Totally right here.

StetsenTech:
Merging in this PR. Any additional comments will be added to a cleanup task.

PR #49, opened 2019-12-11
Editor components

rshipp:
I'll squash this into a single commit before merging, so nobody has to see how bad the intermediates were. But i thought it was worth including the full thing here.

What's in this PR:

* Sane sections
* Sane checkbox-questions
* Frontend portions of image insertion
* Major ckeditor5 version upgrade (15.0.0 -> 16.0.0)


rshipp:
@sadleb they have extensive docs https://ckeditor.com/docs/ckeditor5/latest/framework/index.html but frankly they read like a textbook to me and i couldnt follow them at all until i'd spent a few weeks tinkering with things. thats part of why i want to write it up in a blog post (or a few), i think they just make too many assumptions about what the person reading their docs already knows.

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -0,0 +1,438 @@
...
+                    this.editor.execute( 'insertCheckbox' )
+                }
+                data.preventDefault();
+                evt.stop();
+            }
+        });
+
+    }
+
+    _defineSchema() {
```
sadleb: I'm not sure where the best place to put this would be, either as a comment here or if in the process of writing up the sample tests those serve as a good reference, but I think it would be immensely helpful to have some sample HTML to refer to when working with these files. Since you're starting from the HTML that Sarah provided, this stuff makes sense to you. But as a new developer coming in here (or even if you return to it 6 months down the road), what is a way we can make it faster for an engineer to understand visually what the HTML corresponding to this stuff looks like? I'm specifically thinking of the data conversion, not the editing conversion.

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -0,0 +1,438 @@
...
+            view: {
+                name: 'div'
+            }
+        } );
+        conversion.for( 'editingDowncast' ).elementToElement( {
+            model: 'checkboxDiv',
+            view: ( modelElement, viewWriter ) => {
+                const div = viewWriter.createContainerElement( 'div', {} );
+
+                const widgetContents = viewWriter.createUIElement(
```
sadleb: I think you should add more context here about why this is necessary and what it's doing since it's non-standard CKEditor stuff and doesn't match the general pattern for everything else that once a new engineer learns, they can understand the general code. If not inline, b/c we're going to be using this technique a bunch, then maybe make a task to talk about this in the higher level docs you writeup?

app/javascript/ckeditor/insertcheckboxcommand.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default class InsertCheckboxCommand extends Command {
+    execute( id ) {
+        this.editor.model.change( writer => {
+            // NOTE: We're making a huge assumption here that we'll only ever call this command while
+            // the current selection is *inside* a checkbox label. If that ever changes, we'll need to
+            // add some extra logic here.
+            // Before inserting, modify the current selection to after the checkboxDiv (the grandparent
+            // of the current selection, iff the aforementioned assumption holds true).
+            writer.setSelection( this.editor.model.document.selection.focus.parent.parent, 'after' );
```
sadleb: Instead of `this.editor.model.document.selection.focus.parent.parent` I think we should store this in named variables so that it's more clear to the reader just what the heck that thing is expected to be. E.g.
```
const grandparentCheckboxDiv = this.editor.model.document.selection.focus.parent.parent;
writer.setSelection( grandparentCheckboxDiv, 'after' );

File: app/javascript/ckeditor/utils.js [hunk 4]
```diff
@@ -0,0 +1,17 @@
+// See https://stackoverflow.com/questions/59040225/how-do-i-make-an-editable-label-for-an-input-within-a-ckeditor5-widget
+// and https://github.com/ckeditor/ckeditor5-core/compare/proto/input-widget#diff-44ca1561ce575490eac0d660407d5144R239
+export function preventCKEditorHandling( domElement, editor ) {
```
sadleb: I haven't taken the time to really look at where this gets called, but I just want to make sure you've given some thought to "what if it fails, does it set the default behavior back?" I wouldn't want some weird bug where click handling gets completely disabled or something b/c of some error after we disable it.

app/javascript/ckeditor/checklistquestionediting.js
rshipp: yeah that's a good note. I think i'd rather have that in the tests somehow than in a comment, because a comment seems more likely to quickly get outdated and incorrect as we change the schemas.

app/javascript/ckeditor/insertcheckboxcommand.js
rshipp: love this. I'm not used to the idea of adding variables I only use once just to increase readability, but i think its a really good pattern i should make a habit out of using when appropriate.

app/javascript/ckeditor/utils.js
rshipp: This specifically stops CKEditor from interrupting and overriding default browser handling of these events, so anything we call it on will function as normal. I will definitely write up an explanation of how we're using this, since I think it's non-obvious and super important (and never mentioned in the cke docs).

app/javascript/ckeditor/checklistquestionediting.js
rshipp: yeah I'm not overjoyed about the onchange code needing to be embedded here either :/

Once we get some more of these implemented next week, I'll put some thought into where/how to explain what they're doing. This is one of two ways we're doing custom UI work, so I'll have to go over both of them (inline UI vs external UI).

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    init() {
+        this._defineSchema();
+        this._defineConverters();
+
+        this.editor.commands.add( 'insertChecklistQuestion', new InsertChecklistQuestionCommand( this.editor ) );
+        this.editor.commands.add( 'insertCheckbox', new InsertCheckboxCommand( this.editor ) );
+
+        // Override the default 'enter' key behavior for checkbox labels.
```
rshipp: this is another area that needs to get more fully documented/tested. these next 10 lines represent 2 full weeks of work. 😓

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -0,0 +1,438 @@
...
+            }
+        });
+
+    }
+
+    _defineSchema() {
+        const schema = this.editor.model.schema;
+
+        schema.register( 'checklistQuestion', {
+            isObject: true,
```
rshipp: i need to go through all the schema definitions and double-check and document which flags I'm setting. `isInline` is obvious, but `isObject` and `isLimit` are less-so.

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -0,0 +1,438 @@
...
+                        domElement.innerHTML = `
+                            <option value="correct">Correct</option>
+                            <option value="incorrect">Incorrect</option>
+                            <option value="maybe">Maybe</option>`;
+
+                        // Default to the stored value.
+                        domElement.value = modelElement.getAttribute( 'data-correctness' );
+
+                        // Allow toggling this input in the editor UI.
+                        preventCKEditorHandling(domElement, editor);
```
rshipp: @sadleb this is where we're using this function, on the `select` dropdown. if we take this out, CKE interrupts all events on the dropdown. And since this is a UIElement, the behavior it uses is "this element does not exist" - you can't click on it, tab into it, change the options, etc.

spec/feature/course_contents/smoke_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+          find("button.ck-inspector-navbox__navigation__toggle").click
+
+          # Insert a question.
+          find("li", text: "Section").click
+          find("li", text: "Checklist Question").click
+
+          # Make sure the question was inserted.
+          expect(page).to have_selector("h5.ck-editor__editable.ck-editor__nested-editable")
+          expect(page).to have_selector('input[type="checkbox"].ck-widget')
+
```
rshipp: @sadleb @StetsenTech check out these tests! 😎

StetsenTech: 😲

rshipp:
Added 5 followup tasks here: https://app.asana.com/0/1142638035116665/1156610992150988/f to track the feedback in this PR that hasn't already been addressed. 

I'm merging this PR into staging at its previously-reviewed state, as discussed with @sadleb and @StetsenTech in our call today. I will open smaller PRs for the rest of the work that I moved out of this branch.

PR #50, opened 2019-12-16
Adds Honeycomb

StetsenTech:
**Ticket(s)**:
https://app.asana.com/0/1133543009734854/1153847903754889/f

**Summary**:
- Adds Honeycomb

### DIFF
--- a/config/initializers/honeycomb.rb
+++ b/config/initializers/honeycomb.rb
@@ -1,0 +1,4 @@
+Honeycomb.configure do |config|
+  config.write_key = Rails.application.secrets.honeycomb_api_key
+  config.dataset = Rails.application.secrets.honeycomb_dataset
+end
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -6,6 +6,8 @@ development:
   mailer_from_email: <%= ENV['MAILER_FROM_EMAIL'] %>
   devise_secret_key: <%= ENV['DEVISE_SECRET_KEY'] %>
   devise_pepper: <%= ENV['DEVISE_PEPPER'] %>
+  honeycomb_dataset: <%= ENV['HONEYCOMB_DATASET'] %>
+  honeycomb_api_key: <%= ENV['HONEYCOMB_API_KEY'] %>
 
 test:
   database_host: <%= ENV['DATABASE_HOST'] %>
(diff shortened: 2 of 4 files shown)


StetsenTech:
I think I pushed the `Gemfile.lock` at the same time you made the comment. LOL!

config/initializers/honeycomb.rb
```diff
@@ -0,0 +1,4 @@
+Honeycomb.configure do |config|
+  config.write_key = Rails.application.secrets.honeycomb_api_key
```
sadleb: I don't think Honeycomb should be required. E.g. it's fine to have an ENV with no `HONEYCOMB_DATASET` or `HONEYCOMB_API_KEY` -- does this fail gracefully if they aren't set?

Also, given that we'll want to be able to have an env that doesn't use Honeycomb, we should set `require=false` in the Gemfile, check if the ENV vars are set here, and explicitly require the gem if so. That way, if we have a Review App or some other environment where we aren't using Honeycomb, the gem won't load into memory on restart making it faster and using less memory (e.g. if we want to use cheap or free dynos to run the app which come with very little memory).

config/secrets.yml
```diff
@@ -6,6 +6,8 @@ development:
   mailer_from_email: <%= ENV['MAILER_FROM_EMAIL'] %>
   devise_secret_key: <%= ENV['DEVISE_SECRET_KEY'] %>
   devise_pepper: <%= ENV['DEVISE_PEPPER'] %>
+  honeycomb_dataset: <%= ENV['HONEYCOMB_DATASET'] %>
+  honeycomb_api_key: <%= ENV['HONEYCOMB_API_KEY'] %>
```
sadleb: For Portal on Heroku, I used this: https://github.com/honeycombio/heroku-honeycomb-drain -- the ENV key is name `HONEYCOMB_WRITE_KEY`. Can we name it the same here for consistency, esp if we decide to use another instance of this honeycomb drain app for platform?

sadleb:
Here is a task with more info on how I setup the honeycomb drain for Portal: https://app.asana.com/0/search/1154069427979331/1147796411992721

config/initializers/honeycomb.rb
rshipp: good point! i hadnt thought of that

StetsenTech: I tested on my local env with nothing present before pushing and everything still functioned properly.

I am more than happy to set `require=false`

config/secrets.yml
StetsenTech: Done!
https://github.com/bebraven/platform/pull/51/files

config/initializers/honeycomb.rb
StetsenTech: Done!
https://github.com/bebraven/platform/pull/51/files

PR #51, opened 2019-12-16
Updates to Honeycomb Installation 

StetsenTech:
**Ticket(s)**:
https://app.asana.com/0/1133543009734854/1153847903754889/f

**Summary**:
- Adds `require=false` for Gems that are not needed to run the application
- Wraps initialization in check to see if gem is present
- Updates  `HONEYCOMB_API_KEY` to `HONEYCOMB_WRITE_KEY`
  - This is so Honeycomb Drain can be used in the future

### DIFF
--- a/config/initializers/honeycomb.rb
+++ b/config/initializers/honeycomb.rb
@@ -1,11 +1,13 @@
-if Gem.loaded_specs.has_key?('honeycomb-beeline')
-  Rails.logger.info "Honeycomb detected. Initalizing setup."
-  require "honeycomb-beeline"
+if Rails.application.secrets.honeycomb_write_key && Rails.application.secrets.honeycomb_dataset
+  if Gem.loaded_specs.has_key?('honeycomb-beeline')
+    require "honeycomb-beeline"
 
-  Honeycomb.configure do |config|
-    config.write_key = Rails.application.secrets.honeycomb_write_key
-    config.dataset = Rails.application.secrets.honeycomb_dataset
+    Rails.logger.info "Honeycomb Beeline detected. Initalizing setup."
+    Honeycomb.configure do |config|
+      config.write_key = Rails.application.secrets.honeycomb_write_key
+      config.dataset = Rails.application.secrets.honeycomb_dataset
+    end
+  else
+    Rails.logger.warn "Honeycomb Beeline is not install. Skipping initialization."
   end
-else
-  Rails.logger.warn "Honeycomb is not install. Skipping initialization."
 end
--- a/config/initializers/honeycomb.rb
+++ b/config/initializers/honeycomb.rb
@@ -1,6 +1,6 @@
 require "honeycomb-beeline"
 
 Honeycomb.configure do |config|
-  config.write_key = Rails.application.secrets.honeycomb_api_key
+  config.write_key = Rails.application.secrets.honeycomb_write_key
   config.dataset = Rails.application.secrets.honeycomb_dataset
 end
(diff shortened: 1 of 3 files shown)


config/initializers/honeycomb.rb
```diff
@@ -1,4 +1,11 @@
-Honeycomb.configure do |config|
-  config.write_key = Rails.application.secrets.honeycomb_api_key
-  config.dataset = Rails.application.secrets.honeycomb_dataset
-end
\ No newline at end of file
+if Gem.loaded_specs.has_key?('honeycomb-beeline')
```
sadleb: I was expecting something like ```if ENV['HONEYCOMB_DATASET'] && ENV['HONEYCOMB_WRITE_KEY']``` then run the code in this initializer. How does checking if the honeycomb gem is loaded make it load if the ENV vars are set?

StetsenTech: So the reason this is okay is because Honeycomb will fallback to a `nullClient` defined in LibHoney.
https://github.com/honeycombio/beeline-ruby/blob/a0998861164dc7951860f0d35921563b49127cb3/lib/honeycomb/configuration.rb#L34-L36

https://github.com/honeycombio/libhoney-rb/blob/c477f0a1236023ceb36291dcedae4fb75a783e77/lib/libhoney/client.rb#L104-L112

https://github.com/honeycombio/libhoney-rb/blob/1d1c2d5ad6e587486d350b8e481f9354ee02f821/lib/libhoney/null_client.rb

StetsenTech: Also this line of code is to address the concern if you decided not to install `honeycomb-beeline` on an environment. Not having it would cause an error because `Honeycomb.configure` wouldn't exist.

sadleb: The NullTransmissionClient stuff answers my question about what happens if the ENV vars aren't there. Thanks!

As for the "if you decided not to install honeycomb-beeline" point, that gem is not in a group so it's very unlikely for someone to run bundle install in a way that doesn't install the gem. E.g. `bundle install --without <some group>`. I'm pretty much always expecting the gem to be install on the machine, so the require false stuff is really for the situation where you don't set the ENV vars. Setting them implicitly means "initialize honeycomb". I think it's fine to check if the gem is installed (though prob unnecessary in practice), I really think that what should control whether we load the gem and configure honeycomb is if the ENV vars are set. That's what will happen in practice where in dev or in some review app, we'll do everything the same (aka run the same build and deploy scripts), but the only difference will be that we don't configure honeycomb ENV vars.

StetsenTech: Ah! I see what you mean about the groups. It would be difficult not to install Honeycomb with the current set up. I will revise the grouping so that it is possible spin up environments without Honeycomb. Good catch!

I disagree with the ENV check being the deciding factor of whether or not the initializer could be used in this case. For example, let's say we decide to use Honeycomb Drain that uses the exact same envs, but don't want to use Beeline...the check would pass because they envs are present, but an error would occur because Beeline isn't installed.

sadleb: So a honeycomb drain is a separate app. If we use it, we would set those ENV vars on that app and point the app at the platform log output. The ENV vars wouldn't be used in the platform if we switch to the drain instead of using beeline. The point about renaming the var to match was just for consistency sake.

sadleb: In practice, making honeycomb optional just means "configure the dataset / api key if you want to use it, otherwise don't". I don't see the need for the added complexity of "if you don't want to use honeycomb, you need to run bundle install differently. Having the gems installed on the machine isn't a concern since disk space is cheap. The main concern is loading it into memory if you're not going to use it.

StetsenTech:
Tested with and without. Haven't noticed any problems!

PR #54, opened 2020-01-10
Editor components chunk 2

rshipp:
This is a bunch of stuff I did before break, that's hard to split apart now. Basic summary of what's included:

* Some refactoring/bugfixes
* New components (radio, iframe, blockquote, matching game, table)
* Heroku CI config fixes
* Basic passing tests for CKEditor stuff
* VCR mods to work with chromedriver/heroku
* Improved placeholders in editable elements
* A bunch of messing around with retained data stuff that i'll throw away later
* Copy of bz_newui.css with Sarah's mods for our new HTML

### DIFF
--- a/app.json
+++ b/app.json
@@ -7,7 +7,7 @@
       ],
       "addons": ["heroku-postgresql:in-dyno"],
       "scripts": {
-        "setup": "bundle exec rake db:migrate db:seed",
+        "postdeploy": "bundle exec rake db:migrate db:seed",
         "test-setup": "yarn install",
         "test": "bundle exec rspec && yarn test"
       }
--- a/app.json
+++ b/app.json
@@ -2,7 +2,6 @@
   "environments": {
     "test": {
       "buildpacks": [
-        { "url": "https://github.com/heroku/heroku-buildpack-chromedriver" },
         { "url": "https://github.com/heroku/heroku-buildpack-google-chrome" },
         { "url": "heroku/ruby" }
       ],
--- a/app.json
+++ b/app.json
@@ -2,6 +2,7 @@
   "environments": {
     "test": {
       "buildpacks": [
+        { "url": "https://github.com/heroku/heroku-buildpack-chromedriver" },
         { "url": "https://github.com/heroku/heroku-buildpack-google-chrome" },
         { "url": "heroku/ruby" }
       ],
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -11,7 +11,13 @@
  * It is generally better to create a new file per style scope.
  *
  */
-//*= require_tree .
+//*= require course_contents
+//*= require home
+//*= require industries
+//*= require reboot
+//*= require scaffolds
+//*= require style
+//*= require bz_newui
 
 /* custom temporary hacks for content editor demo */
 form {
--- a/app/javascript/ckeditor/checklistquestionediting.js
+++ b/app/javascript/ckeditor/checklistquestionediting.js
@@ -1,4 +1,5 @@
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import { enablePlaceholder } from '@ckeditor/ckeditor5-engine/src/view/placeholder';
 import { toWidget, toWidgetEditable } from '@ckeditor/ckeditor5-widget/src/utils';
 import Widget from '@ckeditor/ckeditor5-widget/src/widget';
 import InsertChecklistQuestionCommand from './insertchecklistquestioncommand';
@@ -49,7 +50,7 @@ export default class ChecklistQuestionEditing extends Plugin {
         schema.register( 'checkboxInput', {
             isInline: true,
             allowIn: 'checkboxDiv',
-            allowAttributes: [ 'id', 'name', 'value' ]
+            allowAttributes: [ 'id', 'name', 'value', 'data-bz-retained' ]
         } );
 
         schema.register( 'checkboxLabel', {
@@ -81,7 +82,7 @@ export default class ChecklistQuestionEditing extends Plugin {
             model: ( viewElement, modelWriter ) => {
                 // Read the "data-id" attribute from the view and set it as the "id" in the model.
                 return modelWriter.createElement( 'checklistQuestion', {
-                    id: parseInt( viewElement.getAttribute( 'data-id' ) )
+                    id: viewElement.getAttribute( 'data-id' )
                 } );
             }
         } );
@@ -131,7 +132,7 @@ export default class ChecklistQuestionEditing extends Plugin {
                     'select',
                     {
                         'name': 'test',
-                        'onchange': 'console.log("TODO: SAVE CORRECTNESS")'
+                        'onchange': 'addRetainedDataID(this)'
                     },
                     function( domDocument ) {
                         const domElement = this.toDomElement( domDocument );
@@ -166,7 +167,8 @@ export default class ChecklistQuestionEditing extends Plugin {
             },
             model: ( viewElement, modelWriter ) => {
                 return modelWriter.createElement( 'checkboxInput', {
-                    'id': viewElement.getAttribute( 'id' )
+                    'id': viewElement.getAttribute( 'id' ),
+                    'data-bz-retained': viewElement.getAttribute('data-bz-retained') || addRetainedDataID(viewElement)
                 } );
             }
 
@@ -176,7 +178,8 @@ export default class ChecklistQuestionEditing extends Plugin {
             view: ( modelElement, viewWriter ) => {
                 const input = viewWriter.createEmptyElement( 'input', {
                     'type': 'checkbox',
-                    'id': modelElement.getAttribute( 'id' )
+                    'id': modelElement.getAttribute( 'id' ),
+                    'data-bz-retained': modelElement.getAttribute('data-bz-retained') || addRetainedDataID(modelElement)
                 } );
                 return input;
             }
@@ -186,7 +189,8 @@ export default class ChecklistQuestionEditing extends Plugin {
             view: ( modelElement, viewWriter ) => {
                 const input = viewWriter.createEmptyElement( 'input', {
                     'type': 'checkbox',
-                    'id': modelElement.getAttribute( 'id' )
+                    'id': modelElement.getAttribute( 'id' ),
+                    'data-bz-retained': modelElement.getAttribute('data-bz-retained') || addRetainedDataID(modelElement)
                 } );
                 return toWidget( input, viewWriter );
             }
@@ -222,6 +226,12 @@ export default class ChecklistQuestionEditing extends Plugin {
                 const label = viewWriter.createEditableElement( 'label', {
                     // NOTE: We don't set the 'for' attribute in the editing view, so that clicking in the label
                     // editable to type doesn't also toggle the checkbox.
+                } );
+
+                enablePlaceholder( {
+                    view: editing.view,
+                    element: label,
+                    text: 'Answer text'
                 } );
 
                 return toWidgetEditable( label, viewWriter );
--- a/app/javascript/ckeditor/contentcommonediting.js
+++ b/app/javascript/ckeditor/contentcommonediting.js
@@ -18,6 +18,7 @@ export default class ContentCommonEditing extends Plugin {
         schema.register( 'content', {
             isObject: true,
             allowIn: ['blockquoteContent', 'tableContent', 'iframeContent'],
+            allowContentOf: '$root'
         });
 
         schema.register( 'question', {
--- a/app/javascript/ckeditor/contentcommonediting.js
+++ b/app/javascript/ckeditor/contentcommonediting.js
@@ -15,9 +15,14 @@ export default class ContentCommonEditing extends Plugin {
     _defineSchema() {
         const schema = this.editor.model.schema;
 
+        schema.register( 'content', {
+            isObject: true,
+            allowIn: ['blockquoteContent', 'tableContent', 'iframeContent'],
+        });
+
         schema.register( 'question', {
             isObject: true,
-            allowIn: ['checklistQuestion', 'radioQuestion'],
+            allowIn: ['checklistQuestion', 'radioQuestion', 'matchingQuestion'],
         } );
 
         schema.register( 'questionTitle', {
@@ -67,6 +72,30 @@ export default class ContentCommonEditing extends Plugin {
         const editor = this.editor;
         const conversion = editor.conversion;
         const { editing, data, model } = editor;
+
+        // <content> converters
+        conversion.for( 'upcast' ).elementToElement( {
+            model: 'content',
+            view: {
+                name: 'div',
+                classes: 'content'
+            }
+        } );
+        conversion.for( 'dataDowncast' ).elementToElement( {
+            model: 'content',
+            view: {
+                name: 'div',
+                classes: 'content'
+            }
+        } );
+        conversion.for( 'editingDowncast' ).elementToElement( {
+            model: 'content',
+            view: ( modelElement, viewWriter ) => {
+                const section = viewWriter.createContainerElement( 'div', { class: 'content' } );
+
+                return toWidget( section, viewWriter );
+            }
+        } );
 
         // <question> converters
         conversion.for( 'upcast' ).elementToElement( {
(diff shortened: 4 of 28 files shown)


app.json
```diff
@@ -7,7 +7,7 @@
       ],
       "addons": ["heroku-postgresql:in-dyno"],
       "scripts": {
-        "setup": "bundle exec rake db:migrate db:seed",
+        "postdeploy": "bundle exec rake db:migrate db:seed",
```
rshipp: fixes autodeploy for review apps

app/assets/stylesheets/application.scss
```diff
@@ -11,7 +11,13 @@
...
  *
  */
-//*= require_tree .
+//*= require course_contents
+//*= require home
+//*= require industries
+//*= require reboot
+//*= require scaffolds
+//*= require style
+//*= require bz_newui
```
rshipp: fixes the issue where login css got included on every page, overwriting a bunch of styles

spec/support/vcr.rb
```diff
@@ -3,5 +3,6 @@
 VCR.configure do |config|
   config.cassette_library_dir = "spec/fixtures/vcr_cassettes"
   config.hook_into :webmock
-  config.ignore_hosts 'platformweb', 'stagingplatform.bebraven.org', 'platform.bebraven.org'
+  config.ignore_hosts 'platformweb', 'stagingplatform.bebraven.org', 'platform.bebraven.org', 'chromedriver.storage.googleapis.com'
+  config.ignore_localhost = true
```
rshipp: lets VCR work with headless chromedriver on heroku

spec/capybara_helper.rb
```diff
@@ -1,3 +1,25 @@
 RSpec.configure do |config|
   config.include Capybara::DSL
-end
\ No newline at end of file
+end
+
+chrome_shim = ENV.fetch('GOOGLE_CHROME_SHIM', nil)
+if chrome_shim
```
rshipp: this is basically a heroku CI detector - we use their chrome binary if it's available

spec/capybara_helper.rb
```diff
@@ -1,3 +1,25 @@
...
+    Capybara::Selenium::Driver.new(
+      app,
+      browser: :chrome,
+      desired_capabilities: Selenium::WebDriver::Remote::Capabilities.chrome(chrome_opts)
+    )
+  end
+
+  Capybara.javascript_driver = :chrome
+else
+  Capybara.javascript_driver = :selenium_chrome_headless
```
rshipp: if we're not on heroku CI (i.e. running locally), use the headless chrome, because its less annoying

app/javascript/components/ContentPartPreview.js
```diff
@@ -6,6 +6,7 @@ export default class ContentPartPreview extends React.Component {
             <li
                 onClick={() => this.props.onClick( this.props.id )}
                 title="Add to the editor"
+                className={this.props.enabled ? '' : 'disabled'}
```
rshipp: add the `disabled` class to buttons if the `enabled` property isn't true

app/javascript/components/ContentEditor.js
```diff
@@ -326,10 +380,41 @@ class ContentEditor extends Component {
...
+                                            onClick={( id ) => {
+                                                this.editor.execute( 'insertBlockquoteContent', id );
+                                                this.editor.editing.view.focus();
+                                            }}
+                                            {...{name: 'Quote', id: uuidv4()}}
+                                        />
+                                        <ContentPartPreview
+                                            id="33"
+                                            key="33"
+                                            enabled={this.state.enabledCommands.includes('insertIFrameContent')}
```
rshipp: all of these `enabled=` things pass in the property to ContentPartPreview. we check if a relevant command is enabled in order to determine whether the corresponding button should be.

app/javascript/components/ContentEditor.js
```diff
@@ -326,10 +380,41 @@ class ContentEditor extends Component {
...
                                         />
                                         <ContentPartPreview
-                                            id="3"
-                                            key="3"
+                                            id="30"
+                                            key="30"
+                                            enabled={this.state.enabledCommands.includes('imageUpload')}
                                             onClick={this.showFileUpload}
-                                            {...{name: 'Image', id: Math.floor(Math.random() * 1e16)}}
+                                            {...{name: 'Image', id: uuidv4()}}
```
rshipp: this is some stuff i was testing for autogenerating retained-data IDs; it'll get replaced eventually with a real API, so don't pay too much attention to it.

app/javascript/components/ContentEditor.js
```diff
@@ -198,6 +220,12 @@ class ContentEditor extends Component {
         } );
     }
 
+    handleEditorFocusChange( ) {
+        this.setState( {
+            enabledCommands: [...this.editor.commands.names()].filter(x => this.editor.commands.get(x).isEnabled)
```
rshipp: update the list of currently enabled commands in the React state, where the buttons can see it.

app/javascript/components/ContentEditor.js
```diff
@@ -219,6 +247,10 @@ class ContentEditor extends Component {
             editorData: editor.getData()
         } );
 
+        // Attach the focus handler, and call it once.
+        editor.editing.view.document.selection.on('change', this.handleEditorFocusChange);
+        this.handleEditorFocusChange();
```
rshipp: every time the focus changes, and once at document load, run the Focus Change handler (above).

app/javascript/components/ContentEditor.js
```diff
@@ -158,7 +178,8 @@ class ContentEditor extends Component {
             // The initial editor data. It is bound to the editor instance and will change as
             // the user types and modifies the content of the editor.
             editorData: props.course_content['body'] || "",
-            isPublished: false
+            isPublished: false,
+            enabledCommands: [],
```
rshipp: initialize an empty list to track CKE's enabled commands inside the React component's state

app/javascript/components/ContentEditor.js
```diff
@@ -146,6 +159,13 @@ BalloonEditor.defaultConfig = {
     language: 'en'
 };
 
+function addRetainedDataID(element) {
+   console.log('todo');
+   return uuidv4();
+}
+
+window.addRetainedDataID = addRetainedDataID;
```
rshipp: Test code please ignore 🙃 - this'll get replaced shortly by the actual retained data API.

app/javascript/ckeditor/insertchecklistquestioncommand.js
```diff
@@ -42,15 +42,12 @@ function createChecklistQuestion( writer, id ) {
...
     // There must be at least one paragraph for the description to be editable.
     // See https://github.com/ckeditor/ckeditor5/issues/1464.
-    writer.insertText( 'Question?', questionTitle );
-    writer.insertText( 'Option 1', checkboxLabel );
-    writer.insertText( 'Answer Title', answerTitle );
-    writer.insertText( 'Answer body', answerParagraph );
+    writer.append( answerParagraph, answerText );
+
+    // Add text to empty editables where placeholders don't work.
+    writer.insertText( 'Answer 1', checkboxLabel );
```
rshipp: we're using real placeholders where possible, and fill text where it doesn't work for whatever reason

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -0,0 +1,287 @@
...
+            model: 'answerText',
+            view: ( modelElement, viewWriter ) => {
+                const div = viewWriter.createEditableElement( 'div', {} );
+
+                enablePlaceholder( {
+                    view: editing.view,
+                    element: div,
+                    text: 'Answer body',
+                    isDirectHost: false
+                } );
```
rshipp: example placeholder. `isDirectHost` must be `true` for elements that can contain paragraphs, and unset/false for elements that can only contain text.

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -324,7 +189,8 @@ export default class ChecklistQuestionEditing extends Plugin {
             view: ( modelElement, viewWriter ) => {
                 const input = viewWriter.createEmptyElement( 'input', {
                     'type': 'checkbox',
-                    'id': modelElement.getAttribute( 'id' )
+                    'id': modelElement.getAttribute( 'id' ),
+                    'data-bz-retained': modelElement.getAttribute('data-bz-retained') || addRetainedDataID(modelElement)
```
rshipp: test code for retained data stuff, pls ignore

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -269,7 +132,7 @@ export default class ChecklistQuestionEditing extends Plugin {
                     'select',
                     {
                         'name': 'test',
-                        'onchange': 'console.log("TODO: SAVE CORRECTNESS")'
+                        'onchange': 'addRetainedDataID(this)'
```
rshipp: test code for retained data stuff, pls ignore

app/javascript/components/ContentEditor.js
sadleb: I could go look at the `handleEditorFocusChange ` myself, but in general, I think comments are very helpful to make it faster to understand the code if they tell why we do something and not just what the code is doing. Small nit and not worth holding back the merge, but "why?"

app/javascript/components/ContentEditor.js
sadleb: `console.log('todo');` -- lol

app/javascript/ckeditor/contentcommonediting.js
sadleb: Well, that's non-intuitive. Hoping that we can make a way for that to be more obvious from just looking at the code and not researching how CKEditor works at some point as we iterate on this stuff.

app/javascript/ckeditor/insertblockquotecontentcommand.js
```diff
@@ -0,0 +1,39 @@
...
+
+    writer.append( content, blockquoteContent );
+    writer.append( quote, content );
+    writer.append( paragraph, quote );
+    writer.append( citation, quote );
+
+    // Add text to empty editables, to get around the lack of placeholder support.
+    // There must be at least one paragraph for the description to be editable.
+    // See https://github.com/ckeditor/ckeditor5/issues/1464.
+    writer.insertText( 'Quote', paragraph );
```
sadleb: How does this compare to the placeholder example in this PR where you said that `isDirectHost` must be true for paragraphs?

app/javascript/ckeditor/inserttablecontentcommand.js
```diff
@@ -0,0 +1,40 @@
...
+
+    const tableContent = writer.createElement( 'tableContent', {id} );
+    const content = writer.createElement( 'content' );
+
+    writer.append(content, tableContent);
+
+            model.insertContent( tableContent );
+            console.log(content);
+        writer.setSelection( writer.createPositionAt( content, 0 ) );
+        this.editor.execute( 'insertTable', { rows: rows, columns: columns } );
```
sadleb: Whitespace formatting is all off here.

app/javascript/ckeditor/matchingquestionediting.js
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    init() {
+        this._defineSchema();
+        this._defineConverters();
+
+        this.editor.commands.add( 'insertMatchingQuestion', new InsertMatchingQuestionCommand( this.editor ) );
+        this.editor.commands.add( 'insertMatchingTableRow', new InsertMatchingTableRow( this.editor ) );
+
+        // Override the default 'enter' key behavior for checkbox labels.
```
sadleb: Why?

app/javascript/components/ContentEditor.js
```diff
@@ -299,22 +331,44 @@ class ContentEditor extends Component {
...
                                                 this.editor.execute( 'insertSection', id );
                                                 this.editor.editing.view.focus();
                                             }}
-                                            {...{name: 'Section', id: Math.floor(Math.random() * 1e16)}}
+                                            {...{name: 'Section', id: uuidv4()}}
                                         />
                                         <ContentPartPreview
-                                            id="2"
-                                            key="2"
+                                            id="20"
```
sadleb: I think we need a better approach to producing unique `id` and `key` values if these could span multiple files. If it's just in this file, fine. But maybe a naming heuristic like `id="content-part-checklist-q"` or something?

spec/capybara_helper.rb
```diff
@@ -1,3 +1,25 @@
...
   config.include Capybara::DSL
-end
\ No newline at end of file
+end
+
+chrome_shim = ENV.fetch('GOOGLE_CHROME_SHIM', nil)
+if chrome_shim
+  Selenium::WebDriver::Chrome.path = chrome_shim
+
+  chrome_opts = chrome_shim ? { "chromeOptions" => { "binary" => chrome_shim, "remote-debugging-port" => "9222", "headless" => true } } : {}
```
sadleb: We should make this port settable from the standard config ENV vars, wherever they are defined, so that devs don't have to go hunting through code to find the port number they should attach to (and can change it if necessary). I assume you're exposing/mapping this port through docker as well?

spec/feature/course_contents/smoke_spec.rb
```diff
@@ -0,0 +1,66 @@
+require "rails_helper"
+require "capybara_helper"
+
+include ERB::Util
+include Rack::Utils
+
+RSpec.describe CourseContentsController, type: :routing do
```
sadleb: Tests!!!!

spec/capybara_helper.rb
rshipp: This one is only for heroku, I'm not sure if its configurable at all. If we're not running on heroku, it uses a different driver and ignores these options.

app/javascript/components/ContentEditor.js
rshipp: true! this one could be something like "...call it once, to set the initial state of the right sidebar buttons" ?

app/javascript/ckeditor/contentcommonediting.js
rshipp: yeah, this is a side effect of a known "bug" in ckeditor, so it doesn't make sense without a whole lot of context. I'll add it to the list of things to explicitly document.

app/javascript/ckeditor/insertblockquotecontentcommand.js
rshipp: that linked issue is exactly the reason why `isDirectHost` exists 😁

For editables that can contain paragraphs (`$root` editables, in cke-terms), they must contain at least one paragraph to be editable. Because they contain a paragraph, they aren't "empty", and `enablePlaceholder` won't add a placeholder to them. As a workaround, you can pass in `isDirectHost = false`, which tells `enablePlaceholder` the element is not the "direct host" of the placeholder, and it should instead look at the children of that element. If there is only one child, and that child is an empty paragraph, `enablePlaceholder` puts the placeholder on that `p` element instead.

app/javascript/ckeditor/inserttablecontentcommand.js
rshipp: yeah :( copy-paste issue. fixed in #56.

app/javascript/components/ContentEditor.js
rshipp: fix: 3877f7d509c7ad0417c47f71646ef7d4db4a0076

app/javascript/ckeditor/contentcommonediting.js
rshipp: task: https://app.asana.com/0/0/1157299394301467/f

app/javascript/ckeditor/insertblockquotecontentcommand.js
rshipp: task: https://github.com/bebraven/platform/pull/54#discussion_r367017287

app/javascript/ckeditor/matchingquestionediting.js
rshipp: this is... not even the right comment, there's no checkboxes here 😣 

fix: 867e7a4dcd5112afe21d64a7b4708f17d13cc5dc

app/javascript/components/ContentEditor.js
rshipp: I looked up how keys work in react, and they're just unique strings: https://reactjs.org/docs/lists-and-keys.html#keys - so I normalized on just using the associated command for each button as the key.

fix: 04d2cc8b376a0cf8f221f4de9c606b37674ef2f0

StetsenTech: Nice catch.

PR #59, opened 2020-01-10
Add question body

rshipp:
Task: https://app.asana.com/0/1142638035116665/1155859225081587/f

### DIFF
--- a/app/javascript/ckeditor/contentcommonediting.js
+++ b/app/javascript/ckeditor/contentcommonediting.js
@@ -36,7 +36,7 @@ export default class ContentCommonEditing extends Plugin {
         schema.register( 'questionBody', {
             isLimit: true,
             allowIn: 'question',
-            allowContentOf: '$block'
+            allowContentOf: '$root'
         } );
 
         schema.register( 'questionForm', {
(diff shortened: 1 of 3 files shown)


app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -36,7 +36,7 @@ export default class ContentCommonEditing extends Plugin {
         schema.register( 'questionBody', {
             isLimit: true,
             allowIn: 'question',
-            allowContentOf: '$block'
+            allowContentOf: '$root'
```
sadleb: What does `root` vs `block` do here and why is root necessary on the questionBody?

rshipp: Short answer: `$root` allows things like multiple paragraphs and bulleted lists that we use in question bodies; `$block` only allows text, with/without formatting (like bold, links, tooltips, etc).

Long answer: https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-SchemaItemDefinition.html

rshipp:
Merged by #69

PR #63, opened 2020-01-10
Add video component

rshipp:
Task: https://app.asana.com/0/1142638035116665/1155859225081597/f

### DIFF
--- a/app/javascript/ckeditor/videocontentediting.js
+++ b/app/javascript/ckeditor/videocontentediting.js
@@ -1,0 +1,306 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import { enablePlaceholder } from '@ckeditor/ckeditor5-engine/src/view/placeholder';
+import { toWidget, toWidgetEditable } from '@ckeditor/ckeditor5-widget/src/utils';
+import Widget from '@ckeditor/ckeditor5-widget/src/widget';
+import InsertVideoContentCommand from './insertvideocontentcommand';
+
+export default class VideoContentEditing extends Plugin {
+    static get requires() {
+        return [ Widget ];
+    }
+
+    init() {
+        this._defineSchema();
+        this._defineConverters();
+
+        this.editor.commands.add( 'insertVideoContent', new InsertVideoContentCommand( this.editor ) );
+    }
+
+    _defineSchema() {
+        const schema = this.editor.model.schema;
+
+        schema.register( 'videoContent', {
+            isObject: true,
+            allowIn: 'section',
+            allowAttributes: [ 'id', 'class' ]
+        } );
+
+        schema.register( 'videoFigure', {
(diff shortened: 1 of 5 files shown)


app/javascript/ckeditor/videocontentediting.js
```diff
@@ -0,0 +1,306 @@
...
+        } );
+        conversion.for( 'dataDowncast' ).elementToElement( {
+            model: 'videoIFrame',
+            view: ( modelElement, viewWriter ) => {
+                return viewWriter.createEmptyElement( 'iframe', {
+                    'src': modelElement.getAttribute( 'src' ),
+                    'allow': 'encrypted-media',
+                    'allowfullscreen': 'allowfullscreen',
+                    'frameborder': '0',
+                    'height': '315',
```
sadleb: I assume this is just the default and the person inserting a video element can resize after? Also, how does this work on different screens? Does React take care of making it responsive?

app/javascript/components/ContentEditor.js
```diff
@@ -438,7 +440,16 @@ class ContentEditor extends Component {
...
-
+                                        <ContentPartPreview
+                                            id="34"
+                                            key="34"
+                                            enabled={this.state.enabledCommands.includes('insertVideoContent')}
+                                            onClick={( id ) => {
+                                                this.editor.execute( 'insertVideoContent', id, 'https://www.youtube.com/embed/yyRrKMb8oIg?rel=0' );
+                                                this.editor.editing.view.focus();
+                                            }}
+                                            {...{name: 'Video', id: uuidv4()}}
```
sadleb: Is this id for retained data? If so, should you use the `getRetainedDataId()` (or whatever you called that method)?

app/javascript/ckeditor/videocontentediting.js
rshipp: I just copied this from our existing html, evidently all our videos are set up like this. I don't think there's any resizing either. If that's something we want to add, it'll be non-MVP since it's not in the current content editor.

app/javascript/components/ContentEditor.js
rshipp: yep! I was lazy replacing all these since I haven't started retained data tasks yet and don't know where I'm going to tie it in exactly - that'll (hopefully) be next week.

app/javascript/ckeditor/videocontentediting.js
rshipp: task: https://app.asana.com/0/0/1157299394301469/f

app/javascript/components/ContentEditor.js
rshipp: task: https://app.asana.com/0/1142638035116665/1154794977027321/f

rshipp:
Merged by #69

PR #65, opened 2020-01-10
Add ratethismodule component

rshipp:
Task: https://app.asana.com/0/1142638035116665/1155859225081589/f

ignore the textarea bit, I'm not sure why that's in here twice.

### DIFF
--- a/app/javascript/ckeditor/ratethismodulequestionediting.js
+++ b/app/javascript/ckeditor/ratethismodulequestionediting.js
@@ -1,0 +1,157 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import { toWidget } from '@ckeditor/ckeditor5-widget/src/utils';
+import Widget from '@ckeditor/ckeditor5-widget/src/widget';
+import InsertRateThisModuleQuestionCommand from './insertratethismodulequestioncommand';
+
+export default class RateThisModuleQuestionEditing extends Plugin {
+    static get requires() {
+        return [ Widget ];
+    }
+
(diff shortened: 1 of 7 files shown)


app/javascript/ckeditor/ratethismodulequestionediting.js
```diff
@@ -0,0 +1,157 @@
...
+            },
+            model: ( viewElement, modelWriter ) => {
+                return modelWriter.createElement( 'rateThisModuleQuestion', {} );
+            }
+        } );
+        conversion.for( 'dataDowncast' ).elementToElement( {
+            model: 'rateThisModuleQuestion',
+            view: ( modelElement, viewWriter ) => {
+                return viewWriter.createContainerElement( 'div', {
+                    'class': 'module-block module-block-rate-this-module',
```
sadleb: Is `module-block-rate-this-module` the class that we scrape for to know that this is a rate this module magic field? I just want to make sure we still have a way to find all the rate this module magic fields based on the HTML content or retainedData names.

rshipp: it's not, the old name was bz-something. I was planning to add in a Rails API that just returns all the rate-this-module IDs, since scraping seems fragile. But I don't think I have an explicit task for this yet, so I'll add that.

rshipp: task: https://app.asana.com/0/0/1157299394301471/f

rshipp:
Merged by #69

PR #70, opened 2020-01-16
Implement full Assignment support

rshipp:
This PR contains everything necessary to support assignments in CKE5.

Task: https://app.asana.com/0/1142638035116665/1156287933839540/f

My messy notes from combing through assignments:

<details>

new components: 

* dropdowns (select) with a ton of answers

modify existing: 

* sign in with LinkedIn button/input that autofills LI url - (assignment id 1914)
* Collapsible rubric link thing
*  things that are layed out like questions but without the surrounding div / done buttons
* bz-helpful-tip-box ($root)
* bz-example ($root)
* bz-assignment-intro -> table class=steps
* bz-watch-out-box ($root)
* Inputs with placeholder in table
* sliders that autoupdate a number above them with the current selection 
* Just textareas/inputs in $root??
* bullets with input+textarea in each (might already be done)
</details>

What's actually included here: 

* Preserve classes for tables (see https://github.com/ckeditor/ckeditor5/issues/6076#issuecomment-575064953 and https://github.com/ckeditor/ckeditor5/issues/5569)
* Allow custom attributes on `div`, `p`, and `td` (with modified solutions from https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/conversion/conversion-preserving-custom-content.html)
* Allow placeholders in input/textarea
* Allow text `input`s in table cells
* Allow text `input`s/`textarea`s in text blocks (including paragraphs, bullets, etc)
* Add dropdown component (`select` and `option`s)
* Improve radio group logic (now works outside of radio groups, and keeps existing group names when imported from HTML)
* Allow linking images (modified solution from https://github.com/ckeditor/ckeditor5/issues/702)
* Preserve `class` attributes on `a` elements
* Add `span`s with custom preserved classes


app/javascript/ckeditor/radioquestionediting.js
```diff
@@ -178,12 +178,26 @@ export default class RadioQuestionEditing extends Plugin {
...
+                    radioGroupName = viewElement.getAttribute('name') || moduleBlockRadioDiv.getAttribute('id');
+                }
+                catch (e) {
+                    if (e instanceof TypeError) {
+                        // We're not in a radio question; use something else for the group name.
+                        radioGroupName = viewElement.getAttribute('name');
+                    } else {
+                        throw e;
+                    }
+                }
```
rshipp: i'm not sure this is the correct pattern to use in JS, it seems like exceptions are treated differently here than in Python. But i'm also not sure how to do this with an `if`.

PR #72, opened 2020-01-16
Add left nav sidebar

rshipp:
Adds a super basic left sidebar, with a wonky button and a list of created pages.

Task: https://app.asana.com/0/1142638035116665/1155859225081656/f

Screenshots: 
Screen Shot 2020-01-16 at 5 31 18 PM
Screen Shot 2020-01-16 at 5 31 23 PM

### DIFF
--- a/app/controllers/course_contents_controller.rb
+++ b/app/controllers/course_contents_controller.rb
@@ -1,4 +1,6 @@
 class CourseContentsController < ApplicationController
+  before_action :set_course_content, only: [:show, :edit, :update, :destroy, :publish]
+  before_action :set_course_contents, only: [:index, :edit, :new]
 
   # GET /course_contents
   # GET /course_contents.json
@@ -12,13 +14,11 @@ class CourseContentsController < ApplicationController
 
   # GET /course_contents/new
   def new
-    @course_contents = CourseContent.all
     @course_content = CourseContent.new
   end
 
   # GET /course_contents/1/edit
   def edit
-    @course_contents = CourseContent.all
   end
 
   # POST /course_contents
@@ -76,6 +76,10 @@ class CourseContentsController < ApplicationController
   end
 
   private
+
+    def set_course_contents
+      @course_contents = CourseContent.all
+    end
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def course_content_params
--- a/app/controllers/course_contents_controller.rb
+++ b/app/controllers/course_contents_controller.rb
@@ -12,10 +12,13 @@ class CourseContentsController < ApplicationController
 
   # GET /course_contents/new
   def new
+    @course_contents = CourseContent.all
+    @course_content = CourseContent.new
   end
 
   # GET /course_contents/1/edit
   def edit
+    @course_contents = CourseContent.all
   end
 
   # POST /course_contents
--- a/app/controllers/course_contents_controller.rb
+++ b/app/controllers/course_contents_controller.rb
@@ -76,6 +76,9 @@ class CourseContentsController < ApplicationController
   end
 
   private
+    def set_course_content
+      @course_content = CourseContent.find(params[:id])
+    end
 
     def set_course_contents
       @course_contents = CourseContent.all
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "prop-types": "^15.7.2",
     "raw-loader": "4.0.0",
     "react": "^16.11.0",
+    "react-burger-menu": "^2.6.13",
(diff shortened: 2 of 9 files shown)


app/controllers/course_contents_controller.rb
```diff
@@ -14,11 +14,13 @@ def show
 
   # GET /course_contents/new
   def new
+    @course_contents = CourseContent.all
     @course_content = CourseContent.new
   end
 
   # GET /course_contents/1/edit
   def edit
+    @course_contents = CourseContent.all
```
rshipp: is there a better way to do this?

StetsenTech: What is the goal you are trying to achieve here? If you are just trying to get the result of all CourseContent, then this is fine.

rshipp: yeah that's all. It just seemed weird to me, idk if this is "rails-y".

StetsenTech: `Modal.all` is pretty common when you are trying to get all the information. Just be careful of the amount of data is can pull in.

sadleb: I think there is something off here. It makes sense to call `all` in the index, but not in the `new` and `edit`? Is the sidebar showing in the layout to create a new course_contents or the layout to edit one? I'm still too much of a rails noob to be able to say exactly how this should be refactored, but I think it's something like splitting out the sidebar layout into its own view or partial and bind it to course_contents.index. The view to edit or create a new course_contents object would just need to access the individual course_content instance that it's working on.

rshipp: Yeah that's what I was worried about. 

My intent is to give the `new` and `edit` views access to the full list of `course_contents`, because we want to show navigation in the left sidebar. I thought it'd be faster on the frontend if those were loaded directly into the view rather than having to make an AJAX call, but maybe that's the wrong way to do it. Maybe I should just have the left sidebar load from the index API instead?

> something like splitting out the sidebar layout into its own view or partial and bind it to course_contents.index

I totally don't know how to do this. From what I understood looking through the code, it seemed like partials don't inherently have access to any data, only the actual views do.

Maybe @hcli-work would have some input here?

StetsenTech: @sadleb That is accurate.

StetsenTech: @rshipp You can still pass data into partials even if they don't inherently have it. When you include the partial into your view, you can give it info through `locals` parameters.

Example from Canvas code:
`<%= render :partial => "shared/form_errors", locals: { object: task } %>`
`<%= render :partial => 'student', :locals => { :f => f} %>`

rshipp: Oh that part I understand, but I'd still have to have access to the variables in the view in order to pass them in to the partial.

sadleb: After more thought and discussion, if the performance starts to get bad b/c we're dumping the entire course_contents table into a variable every time we hit new/edit actions then let's solve it either by some paging solution or by loading only the columns we need and storing them in a collection that drives the sidebar view. E.g. id and name. Then when selecting an item in the sidebar, we look up the course_contents record using the id.

app/controllers/course_contents_controller.rb
```diff
@@ -14,11 +14,13 @@ def show
 
   # GET /course_contents/new
   def new
+    @course_contents = CourseContent.all
```
sadleb: We could make this a bit more railsy by using `before_action :set_course_contents, only: [:index, :show, :edit]` and define a `set_course_contents` method that runs this line of code. I think it's worth doing before merging this PR.

package.json
```diff
@@ -42,6 +42,7 @@
     "prop-types": "^15.7.2",
     "raw-loader": "4.0.0",
     "react": "^16.11.0",
+    "react-burger-menu": "^2.6.13",
```
sadleb: Can we rename this to something a bit more animal conscious and environmentally friendly? 😅

app/controllers/course_contents_controller.rb
StetsenTech: Definitely agree here. Didn't think about that.

package.json
StetsenTech: That is why Black Bean Burgers exist.

rshipp: 😂 I can't rename external packages, unfortunately.

app/controllers/course_contents_controller.rb
rshipp: Fixed in fbdb8e3

PR #73, opened 2020-01-17
Use window.prompt for 'modal' input

rshipp:
An ugly, but quick workaround for "modal" inputs in the editor UI.

When you click on the buttons in the right sidebar in the editor, some of them need a way to input different options before inserting a component. E.g. we use this to set the height/width of tables, set the URL to use for iframes/videos/images, etc.

Example workflow:

1. Click the "IFrame" button to insert an iframe
2. You get a popup asking for a URL
3. The popup returns the URL you enter to the JS and stores it in a variable
4. We pass the URL variable into the 'insertIFrame' command, so it knows what URL to use
5. A frame with your custom URL is inserted into the editor

I tried using some actual modal prompt stuff, but I kept hitting dead ends and decided it's more complicated than we have time for. We should get a UX / frontend person to give this a makeover later.

Task: https://app.asana.com/0/1142638035116665/1155859225081638/f

### DIFF
--- a/app/controllers/course_contents_controller.rb
+++ b/app/controllers/course_contents_controller.rb
@@ -14,11 +14,13 @@ class CourseContentsController < ApplicationController
 
   # GET /course_contents/new
   def new
+    @course_contents = CourseContent.all
     @course_content = CourseContent.new
   end
 
   # GET /course_contents/1/edit
   def edit
+    @course_contents = CourseContent.all
   end
 
   # POST /course_contents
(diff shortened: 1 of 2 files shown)


rshipp:
Not sure if this is what you mean, but `window.prompt` is just a browser builtin, it looks like this:

Screen Shot 2020-01-17 at 9 13 26 AM

I'd love to use modals instead, but I didn't find an easy way to do that.

StetsenTech:
Is this just for the editor and not the generated page?

rshipp:
Oh yeah this is just in the editor itself - definitely useful context to have, let me edit the issue body!

rshipp:
followup: https://app.asana.com/0/0/1163803258210895/f

PR #74, opened 2020-01-17
Some dev env cleanup.

sadleb:
- Nginx isn't running inside the docker network so http://ssoweb doesn't resolve.
  Need to use the explicit port that our Docker SSO container listens on.
- ruby:2.6.5-alpine Docker image doesn't seem to have /bin/bash. Using /bin/sh.

Note: I tried switching the SSO_URL over to platformweb instead of ssoweb (the old SSO box), but was getting authenticity token verification issues so I'm punting on that for now.


rshipp:
Should we change the example `SSO_URL` to `localhost:3020` so it uses the new platform SSO code?

StetsenTech:
@sadleb Try `SSO_URL=http://platformweb:3020/cas`

sadleb:
> Try `SSO_URL=http://platformweb:3020/cas`

Bah, that was it. Needed the `cas` path. I updated the README and `.env.example`.

sadleb:
> Should we change the example `SSO_URL` to `localhost:3020` so it uses the new platform SSO code?

I updated it to `http://platformweb:3020/cas` (assuming we're all adding platformweb to our /etc/hosts).

README.md
```diff
@@ -5,14 +5,18 @@
...
 
 First, we need to copy a couple of environment files in the app directory:
 
     cp .env.example .env
     cp .env.database.example .env.database
 
-Be sure to set the `SSO_URL` variable in `.env` to point to whatever CAS SSO server you're using.
+*Note:* At the time of writing, the SSO login requires the Join server to be running locally at [http://joinweb](http://joinweb).
+Look at the `server` config value in `config/rubycas.yml` for where that comes from. You can login with any user that
+exists in the Join server's database and admins are anyone with an `@bebraven.org` email.
```
rshipp: oh good note, i forgot about this.

PR #75, opened 2020-01-17
Add context-sensitive right toolbar

rshipp:
Summary: Add a toolbar to the right sidebar that only shows up if you have certain elements selected. E.g. when you click on a textarea, it shows a toolbar that lets you change the textarea's placeholder attribute.

Task: https://app.asana.com/0/1142638035116665/1155859225081719/f

We didn't design anything for this, so it's not pretty. Here's what this looks like:

Screen Shot 2020-01-17 at 3 37 55 PM

Screen Shot 2020-01-17 at 3 39 24 PM

This PR also includes some other, smaller right-sidebar tasks: 

✓ Editor JS: Implement UI for sliders (right sidebar)
✓ Editor JS/CK: Allow setting placeholder text for textareas/inputs (right sidebar)
✓ Editor JS: Remove tabs from right sidebar
✓ Design/Editor JS: organize buttons on right sidebar

### DIFF
--- a/app/javascript/ckeditor/contentcommonediting.js
+++ b/app/javascript/ckeditor/contentcommonediting.js
@@ -2,6 +2,7 @@ import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import { enablePlaceholder } from '@ckeditor/ckeditor5-engine/src/view/placeholder';
 import { toWidget, toWidgetEditable } from '@ckeditor/ckeditor5-widget/src/utils';
 import Widget from '@ckeditor/ckeditor5-widget/src/widget';
+import InsertTextInputCommand from './inserttextinputcommand';
 
 export default class ContentCommonEditing extends Plugin {
     static get requires() {
@@ -11,6 +12,8 @@ export default class ContentCommonEditing extends Plugin {
     init() {
         this._defineSchema();
         this._defineConverters();
+
+        this.editor.commands.add( 'insertTextInput', new InsertTextInputCommand( this.editor ) );
     }
 
     _defineSchema() {
@@ -460,7 +463,7 @@ export default class ContentCommonEditing extends Plugin {
             model: ( viewElement, modelWriter ) => {
                 return modelWriter.createElement( 'textInput', {
                     'data-bz-retained': viewElement.getAttribute('data-bz-retained'),
-                    'placeholder': viewElement.getAttribute('placeholder')
+                    'placeholder': viewElement.getAttribute('placeholder') || ''
                 } );
             }
         } );
@@ -470,7 +473,7 @@ export default class ContentCommonEditing extends Plugin {
                 const input = viewWriter.createEmptyElement( 'input', {
                     'type': 'text',
                     'data-bz-retained': modelElement.getAttribute('data-bz-retained'),
-                    'placeholder': modelElement.getAttribute('placeholder')
+                    'placeholder': modelElement.getAttribute('placeholder') || ''
                 } );
                 return input;
             }
@@ -481,7 +484,7 @@ export default class ContentCommonEditing extends Plugin {
                 const input = viewWriter.createEmptyElement( 'input', {
                     'type': 'text',
                     'data-bz-retained': modelElement.getAttribute('data-bz-retained'),
-                    'placeholder': modelElement.getAttribute('placeholder')
+                    'placeholder': modelElement.getAttribute('placeholder') || ''
                 } );
                 return toWidget( input, viewWriter );
             }
--- a/app/javascript/ckeditor/contentcommonediting.js
+++ b/app/javascript/ckeditor/contentcommonediting.js
@@ -495,7 +495,7 @@ export default class ContentCommonEditing extends Plugin {
             model: ( viewElement, modelWriter ) => {
                 return modelWriter.createElement( 'textArea', {
                     'data-bz-retained': viewElement.getAttribute('data-bz-retained'),
-                    'placeholder': viewElement.getAttribute('placeholder')
+                    'placeholder': viewElement.getAttribute('placeholder') || ''
                 } );
             }
         } );
@@ -504,7 +504,7 @@ export default class ContentCommonEditing extends Plugin {
             view: ( modelElement, viewWriter ) => {
                 const textarea = viewWriter.createEmptyElement( 'textarea', {
                     'data-bz-retained': modelElement.getAttribute('data-bz-retained'),
-                    'placeholder': modelElement.getAttribute('placeholder')
+                    'placeholder': modelElement.getAttribute('placeholder') || ''
                 } );
                 return textarea;
             }
@@ -514,7 +514,7 @@ export default class ContentCommonEditing extends Plugin {
             view: ( modelElement, viewWriter ) => {
                 const textarea = viewWriter.createEmptyElement( 'textarea', {
                     'data-bz-retained': modelElement.getAttribute('data-bz-retained'),
-                    'placeholder': modelElement.getAttribute('placeholder')
+                    'placeholder': modelElement.getAttribute('placeholder') || ''
                 } );
                 return toWidget( textarea, viewWriter );
             }
@@ -573,9 +573,9 @@ export default class ContentCommonEditing extends Plugin {
                 return modelWriter.createElement( 'slider', {
                     'class': viewElement.getAttribute('class') || '',
                     'data-bz-retained': viewElement.getAttribute('data-bz-retained'),
-                    'min': viewElement.getAttribute('min'),
-                    'max': viewElement.getAttribute('max'),
-                    'step': viewElement.getAttribute('step') || '',
+                    'min': viewElement.getAttribute('min') || 0,
+                    'max': viewElement.getAttribute('max') || 10,
+                    'step': viewElement.getAttribute('step') || 1,
                 } );
             }
         } );
@@ -586,9 +586,9 @@ export default class ContentCommonEditing extends Plugin {
                     'type': 'range',
                     'class': modelElement.getAttribute('class') || '',
                     'data-bz-retained': modelElement.getAttribute('data-bz-retained'),
-                    'min': modelElement.getAttribute('min'),
-                    'max': modelElement.getAttribute('max'),
-                    'step': modelElement.getAttribute('step') || '',
+                    'min': modelElement.getAttribute('min') || 0,
+                    'max': modelElement.getAttribute('max') || 10, 
+                    'step': modelElement.getAttribute('step') || 1,
                 } );
                 return input;
             }
@@ -600,9 +600,9 @@ export default class ContentCommonEditing extends Plugin {
                     'type': 'range',
                     'class': modelElement.getAttribute('class') || '',
                     'data-bz-retained': modelElement.getAttribute('data-bz-retained'),
-                    'min': modelElement.getAttribute('min'),
-                    'max': modelElement.getAttribute('max'),
-                    'step': modelElement.getAttribute('step') || '',
+                    'min': modelElement.getAttribute('min') || 0,
+                    'max': modelElement.getAttribute('max') || 10,
+                    'step': modelElement.getAttribute('step') || 1,
                 } );
                 return toWidget( input, viewWriter );
             }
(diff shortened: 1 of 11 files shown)


app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -573,9 +573,9 @@ export default class ContentCommonEditing extends Plugin {
                 return modelWriter.createElement( 'slider', {
                     'class': viewElement.getAttribute('class') || '',
                     'data-bz-retained': viewElement.getAttribute('data-bz-retained'),
-                    'min': viewElement.getAttribute('min'),
-                    'max': viewElement.getAttribute('max'),
-                    'step': viewElement.getAttribute('step') || '',
+                    'min': viewElement.getAttribute('min') || 0,
+                    'max': viewElement.getAttribute('max') || 10,
```
rshipp: I used these default values because they make more sense for our modules than HTML's default (max: 100) for `<input type=range>` elements. Embedding magic numbers is a little icky, but I don't see a cleaner way to factor these out.

app/javascript/components/ContentEditor.js
```diff
@@ -236,10 +239,46 @@ class ContentEditor extends Component {
...
+
+        // If one of the above cases was true, we'll have a selectedViewElement, and we can
+        // pull its ancestor chain.
+        if ( selectedViewElement ) {
+            const viewAncestorNames = selectedViewElement.getAncestors().map( x => x.name );
+
+            this.setState( {
+                viewPath: viewAncestorNames.concat(selectedViewElement.name)
+            } );
+        }
```
rshipp: This is the meat of this PR. Let me know if you see a way to improve the logic and/or comments here.

app/javascript/components/ContentEditor.js
```diff
@@ -269,7 +308,10 @@ class ContentEditor extends Component {
 
         // CKEditor 5 inspector allows you to take a peek into the editor's model and view
         // data layers. Use it to debug the application and learn more about the editor.
-        CKEditorInspector.attach( editor );
+        // Disable unless debug mode is enabled.
+        if ( window.location.search === '?debug' ) {
+            CKEditorInspector.attach( editor );
+        }
```
rshipp: threw in a way to turn off the cke5 inspector.

app/javascript/components/ContentEditor.js
```diff
@@ -333,162 +375,232 @@ class ContentEditor extends Component {
...
+                                                        step: evt.target.value,
+                                                    } );
+                                                }}
+                                            />
+                                            <label htmlFor='input-step'>Step</label>
+                                        </>
+                                    );
+                                }
+                            } ) }
+                        </div>
```
rshipp: This is the other important chunk. We're getting to a point where we should probably start refactoring out pieces of the `ContentEditor` react component into smaller components. :/

app/javascript/components/ContentEditor.js
```diff
@@ -333,162 +375,232 @@ class ContentEditor extends Component {
                                        placeholder="Page Title"
                                 />
                             </h2>
+                        </div>
+                        <div id="toolbar-contextual">
+                            {this.state.modelPath.map( modelElement => {
+                                if ( ['textArea', 'textInput'].includes( modelElement ) ) {
```
rshipp: this only shows up if we're in a textarea or text input

app/javascript/components/ContentEditor.js
```diff
@@ -333,162 +375,232 @@ class ContentEditor extends Component {
...
+                                    // Text inputs and textareas have placeholder settings.
+                                    return (
+                                        <>
+                                            <h4>Text Input</h4>
+                                            <input
+                                                type='text'
+                                                id='input-placeholder'
+                                                defaultValue={this.state['selectedElement'].getAttribute('placeholder')}
```
rshipp: this reads the existing value (if any) of the placeholder attribute into the text input

app/javascript/components/ContentEditor.js
```diff
@@ -333,162 +375,232 @@ class ContentEditor extends Component {
...
+                                                onChange={( evt ) => {
+                                                    this.editor.execute( 'insertTextArea', 'retained-data-todo', evt.target.value );
+                                                }}
```
rshipp: `onChange` lets us update things on the fly as you type, instead of needing a submit button or something!

note that the actual logic for "changing the placeholder" is actually replacing the entire element in the editor - this is unfortunately the best way I could figure out how to do this. :/

app/javascript/ckeditor/contentcommonediting.js
StetsenTech: I agree with defaulting values to min 0 and max 10.  I can't recall any sliders that go up to 100. I know some sliders do start at 1 instead of 0, but that is minor since you can change the min yourself.

I am a bit curious about what you mean by "Embedding magic numbers is a little icky".

app/javascript/components/ContentEditor.js
StetsenTech: This would be ideal. I would make tests for the current state of ContentEditor.js before making refactors.

app/javascript/components/ContentEditor.js
StetsenTech: How often is this `onChange` triggered? Triggering `onChange` frequently can lock up the browser. Ideally, a timeout would be set to prevent a flood of DOM changes from happening at once.

app/javascript/ckeditor/contentcommonediting.js
rshipp: Just the [hardcoded 0, 1, 10](https://en.wikipedia.org/wiki/Magic_number_%28programming%29) part. Generally I'd opt to use constants like `SLIDER_MAX=10` kind of thing, but I don't think that increases readability here.

app/javascript/components/ContentEditor.js
rshipp: every time you type a letter. we can use `onBlur` instead, if we want to sacrifice a little UX for a little performance. My opinion here is that we shouldn't prematurely optimize, and should wait until we actually see perf issues here to change this, but feel free to push back on that if you have a different perspective.

app/javascript/ckeditor/contentcommonediting.js
StetsenTech: I think that would be a good use of a constant since it is used more than once. Easy enough to read. That way you don't forget to update it in multiple places if we make some discovery that we want to use different values.

rshipp: I could put constants at the top of the file, that just seems kind of weird since sliders are only a super small part of what's in this file. I'm not sure if this is an issue with me not being super familiar with JS patterns (should constants go at the top of the file? top of the function? right above the function?) or a code smell (contentcommonediting.js too broad and needs to be broken up?)

app/javascript/components/ContentEditor.js
StetsenTech: I'm willing to wait and see. Let's just make a note of this.

rshipp: task: https://app.asana.com/0/0/1158367893927205/f

app/javascript/components/ContentEditor.js
rshipp: task: https://app.asana.com/0/0/1158367893927206/f

app/javascript/components/ContentEditor.js
rshipp: note to self: this breaks the tests, fix them before merging! ❗️❗️❗️❗️❗️❗️

app/javascript/ckeditor/contentcommonediting.js
sadleb: This feels like a case of balancing pragmatism with ideals. Ideally, since the values are used in 3 places, they would be in constants. But as you said, there is no good place to define the constants. I don't like the idea of having a big list of constants at the top of the file. I think it makes it less readable. I don't like the idea of having a big list of constants at the top of the function for all converters for the same reason. If we were going to do it, I think that a set of converters for a single thingie would be in a function and you would define the constants at the top of that function. But, since each set of converters fits on a page and you can see them all at once, I don't think it increases readability to factor them into a dedicated function. Also, since you see all three converters on a single screen you see the hardcoded values repeated and it's highly unlikely that you would change it in one spot and not the other three that you're looking at. So I'm fine with the pragmatism of leaving them hard-coded. 

If you do decide to define constants, I would do it right before the first of the three converters that use it.

app/javascript/ckeditor/customelementattributepreservation.js
```diff
@@ -175,7 +175,6 @@ function setupAllowedAttributePreservation( editor ) {
     editor.conversion.for( 'upcast' ).elementToElement( {
         view: 'p',
         model: ( viewElement, modelWriter ) => {
-        console.log(viewElement.getAttributes());
```
sadleb: good catch

app/javascript/components/ContentEditor.js
sadleb: I don't have enough JS experience to know how to do this better. I appreciate the comments though b/c it's slightly complicated and they help understand what is going on.

app/javascript/components/ContentEditor.js
sadleb: Can we add this `?debug` option to the README? You're in the code now so you know that it's an option, but as we build up more and more features, it would be good to have the various ways to configure, dev, and troubleshoot those features in the README (or README for that feature).

app/javascript/components/ContentEditor.js
sadleb: nice

app/javascript/components/ContentEditor.js
```diff
@@ -333,162 +375,236 @@ class ContentEditor extends Component {
...
+                                    }
+                                    return (
+                                        <>
+                                            <h4>Text Input</h4>
+                                            <input
+                                                type='text'
+                                                id='input-placeholder'
+                                                defaultValue={this.state['selectedElement'].getAttribute('placeholder')}
+                                                onChange={( evt ) => {
+                                                    this.editor.execute( commandMap[modelElement], 'retained-data-todo', evt.target.value );
```
sadleb: what's a `retained-data-todo` ?

app/javascript/components/ContentEditor.js
```diff
@@ -333,162 +375,236 @@ class ContentEditor extends Component {
...
 
+                            <ul key="content-part-list-section" className="widget-list">
+                                <ContentPartPreview
+                                    key="insertSection"
+                                    enabled={this.state.enabledCommands.includes('insertSection')}
+                                    onClick={( id ) => {
+                                        this.editor.execute( 'insertSection', id );
+                                        this.editor.editing.view.focus();
+                                    }}
+                                    {...{name: 'Section', id: uuidv4()}}
```
sadleb: Can we wrap `uuidv4` with a nicer function name and use that everywhere? Something folks can remember and type without having to search for another place that inserts a unique id? Even just `uuid()` would be better. Not a big deal though. Feel free to punt and just leave it if it's too much of a hassle and doesn't really make the code easier to read/maintain.

app/javascript/components/ContentEditor.js
```diff
@@ -333,162 +375,236 @@ class ContentEditor extends Component {
...
+                                    enabled={this.state.enabledCommands.includes('insertSlider')}
+                                    onClick={( id ) => {
+                                        this.editor.execute( 'insertSlider', id );
+                                        this.editor.editing.view.focus();
+                                    }}
+                                    {...{name: 'Slider', id: uuidv4()}}
+                                />
+                                <input
+                                    type="file"
+                                    style={{ display: "none" }}
```
sadleb: Why `display: "none"` ?

app/javascript/components/ContentEditor.js
rshipp: @sadleb It was a placeholder so I didn't have to refactor the InsertTextareaCommand execute() function. This is gone in the editor-retained-data branch.

app/javascript/components/ContentEditor.js
rshipp: Agreed, uuidv4 is gone and removed as a dependency entirely in the editor-retained-data branch.

app/javascript/components/ContentEditor.js
rshipp: This is evidently the only way to create buttons that act as file uploads 😞 - you create a hidden `<input type="file">` and then activate its onClick function from your button:

* https://github.com/bebraven/platform/pull/75/files/0a11bf128bb8ad961c427d9fd06a6736cd0d6cfa#diff-92838aff2423c239e8050b1d702f1a39R582
* https://github.com/bebraven/platform/pull/75/files/0a11bf128bb8ad961c427d9fd06a6736cd0d6cfa#diff-92838aff2423c239e8050b1d702f1a39R227

PR #76, opened 2020-01-22
DO NOT MERGE: Allow running selenium / chrome specs inside the Docker container.

sadleb:
NOTE: this doesn't work yet. I got to the point where I can run the specs
and it finds the driver and connects, but it can't find the "username" element.
Putting this up to show other devs my code.


rshipp:
I don't fully understand the VCR/Selenium interaction, but if we can't have them on separate hosts, can we just install chrome inside the platformweb container? That should allow the regular headless driver to work.

PR #77, opened 2020-01-22
Editor versioning

rshipp:
Summary: 

Cherry-picked over the versioning models from the autosave branch, and tied in some new functionality:

* Clicking Publish publishes to Canvas, updates `published_at`, saves a "version", and opens the newly published page on Canvas.
* Visiting `/course_contents/{id}/versions/` shows a list of all versions
* Visiting `/course_contents/{id}/versions/{version_id}` shows the selected version

Tasks:

[✓ Rails: Implement version detail view that just shows HTML](https://app.asana.com/0/0/1155859225081652/f)
[✓ Rails: Implement version list views per-page](https://app.asana.com/0/0/1155859225081650/f)
[✓ Rails: Publish button saves a new HTML version](https://app.asana.com/0/0/1155859225081658/f)
[✓ Rails: Implement "Create new version" API endpoint](https://app.asana.com/0/0/1154794977027318/f)

Additional notes:

`CourseContentHistory` is the version model. `CourseContentUndo` was used by autosave, but isn't used in this PR.


app/controllers/course_contents_controller.rb
```diff
@@ -68,6 +68,18 @@ def destroy
...
+
+        # update publish time, save a version
+        @course_content.published_at = DateTime.now
+        new_version = CourseContentHistory.new({
+          course_content_id: @course_content.id,
+          title: @course_content.title,
+          body: @course_content.body
+        })
+        new_version.save
+        @course_content.save
```
rshipp: I'd like some extra eyes here especially. I'm fairly sure there should be more error checking going on, but I don't know how much. Should I really be checking every `.save` call's response code?

Keep in mind I'm used to a more Python-ish flow where you rarely check return codes, and rely heavily on exceptions instead.

app/models/course_content.rb
```diff
@@ -1,7 +1,10 @@
 class CourseContent < ApplicationRecord
+  has_many :course_content_history
+  has_many :course_content_undo
+
   def publish(params)
     response = CanvasProdClient.update_course_page(params[:course_id], params[:secondary_id], params[:body])
 
-    response.code == 200
+    response.code >= 200 && response.code < 300
```
rshipp: sometimes we get 201s or something back instead, so this just accepts all of them.

config/routes.rb
```diff
@@ -1,6 +1,7 @@
 Rails.application.routes.draw do
   resources :course_contents do
     post :publish
+    resources :course_content_histories
```
sadleb: You can use `resources :course_content_histories, , :path => 'histories'` if you want a nicer route like `/course_contents/1/histories`

sadleb: Is it worth limiting the routes to just what we support. E.g. `only: [:index, etc]`

db/migrate/20191212193241_create_course_content_histories.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCourseContentHistories < ActiveRecord::Migration[6.0]
+  def change
+    create_table :course_content_histories do |t|
+      t.references :course_content, null: false, foreign_key: true
+      t.string :title
```
sadleb: Can this be null? Same with body?

db/migrate/20191212193546_create_course_content_undos.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCourseContentUndos < ActiveRecord::Migration[6.0]
+  def change
+    create_table :course_content_undos do |t|
+      t.references :course_content, null: false, foreign_key: true
+      t.text :operation
+      t.integer :version
```
sadleb: Can these be null?

db/schema.rb
```diff
@@ -34,6 +34,35 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "contact_owners", force: :cascade do |t|
```
sadleb: contact_owners?

spec/factories/course_content_histories.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :course_content_history do
+    course_content { nil }
```
sadleb: I think instead of setting it to nil, i think you should just give the factory. E.g.

```
title { "MyString" }
body { "MyText" }

course_content
```

spec/factories/course_content_undos.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :course_content_undo do
+    course_content { nil }
```
sadleb: Ditto

spec/factories/course_content_undos.rb
```diff
@@ -0,0 +1,7 @@
...
+    operation { "MyText" }
+    version { "" }
```
sadleb: Why is this empty? Is an empty version valid?

spec/models/course_content_history_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe CourseContentHistory, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
sadleb: I would take this advice. Either add some specs of delete it for now.

spec/models/course_content_undo_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe CourseContentUndo, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
sadleb: Ditto

app/controllers/course_contents_controller.rb
rshipp: aha, ok. i'll fix that.

app/javascript/components/ContentEditor.js
```diff
@@ -329,6 +329,7 @@ class ContentEditor extends Component {
                     this.setState( {'isPublished': true} );
                     // TODO: provide a way for the user to easily view the result, which will be in:
```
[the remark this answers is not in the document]
rshipp: good catch, the example url is wrong too actually

app/controllers/course_content_histories_controller.rb
```diff
@@ -0,0 +1,46 @@
+class CourseContentHistoriesController < ApplicationController
+  before_action :set_course_content_history, only: [:show, :edit, :update, :destroy]
+  before_action :set_course_content, only: [:index]
```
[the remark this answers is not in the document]
rshipp: @sadleb what would it take for me to have these use drycrud? if it's something I can do relatively quickly, I'd rather do it before merging

db/migrate/20191212193241_create_course_content_histories.rb
rshipp: They can be empty, do you think I should set them non-null and `default: ''`?

rshipp:
Finally, passing tests....

app/controllers/course_contents_controller.rb
```diff
@@ -14,11 +14,13 @@ def show
 
   # GET /course_contents/new
   def new
+    @course_contents = CourseContent.all
     @course_content = CourseContent.new
```
sadleb: I think you can just delete both `before_action`s and the associated `set_course_content(s)` methods and remove all the guts from new and edit and it will just work.

app/controllers/course_contents_controller.rb
```diff
@@ -14,11 +14,13 @@ def show
...
   end
 
   # GET /course_contents/1/edit
   def edit
+    @course_contents = CourseContent.all
```
sadleb: This seems ... wrong? Shouldn't edit be to edit a single `@course_content` gotten by looking it up using the `params[:id]` (which DryCrud takes care of)?

db/migrate/20191212193241_create_course_content_histories.rb
sadleb: Maybe doesn't matter then unless it would simplify some ActiveRecord query or prevent some faulty logic during development. But I think we would just use `blank?` to check for both null and empty.

config/routes.rb
```diff
@@ -1,6 +1,7 @@
 Rails.application.routes.draw do
   resources :course_contents do
     post :publish
+    resources :course_content_histories, path: 'versions', only: [:index, :show]
```
sadleb: `versions` is so much nicer as a URL :)

app/controllers/course_contents_controller.rb
rshipp: Here's the previous discussion on this https://github.com/bebraven/platform/pull/72#discussion_r367703831, but I'll add a comment to the code itself to make the purpose more clear.

app/controllers/course_contents_controller.rb
rshipp: I definitely fixed this already https://github.com/bebraven/platform/pull/72#discussion_r370471104, I must have chosen the wrong diff in a rebase or something. I'll remove them again.

PR #80, opened 2020-01-23
Cleanup old data-central Person model.

sadleb:
Consolidate Person and User together into one and remove all the stuff
around Contact Owner and List. I think "Contact" was meant to represent
a way for a person to be contacted, e.g. Phone, Email, Address but the naming
made it very confusing. I'm leaving Phone, Email, and Address in place for now
in case we want to use them, but we may end up deleting them too)

Since we're going to be designing the rostering data model from the ground up I
just want to start from something basic. A User, a Role, and the Program they're in
and we'll work up from there.

Consider renaming Program to Course as we start to flesh out the roster data model.

TESTING:
- all specs pass except the Selenium smoke tests for Content Editor (which didn't pass in my dev env before these changes).
- able to login and see Content Editor still
- able to load /users route and see a list of users

NOTE:
- the UI around all the models created for the data-central
  project is broken with this commit:
    https://github.com/bebraven/platform/commit/eeb022de85da74928d577a9fa2b8f199a99fadd0
  However, this stage of the project is focused on the models so
  I'm not putting any effort into fixing the UI.


spec/factories/users.rb
```diff
@@ -1,6 +1,21 @@
...
+        [
+          ['Aaron', 'Anderson'], ['Anthony', 'Anderson'], ['Barry', 'Boswick'], ['Bill', 'Boswick'], 
+          ['Bob', 'Boswick'], ['Cara', 'Clark'], ['Candice', 'Clark'], ['Carmen', 'Clark'], 
+          ['Diana', 'Davis'], ['Daisy', 'Davis'], ['Deirdre', 'Davis'], ['Fred', 'Fox'], 
+          ['Fran', 'Fox'], ['Gertie', 'Gray'], ['Greg', 'Gray'], ['Henry', 'Hill'], 
+          ['Harrieta', 'Hill'], ['Harvey', 'Hill'], ['Jack', 'Jones'], ['Jenny', 'Jones'], 
+          ['John', 'Jones'], ['Jeff', 'Jones']
+        ]
+      end
+    end
```
StetsenTech: I'm fine with this as an updated, but we really should leverage Faker to handle random name generation.

sadleb:
> Can we make it so these migrations can't be reversed?
> 
> ```
> def down
>   raise IrreversibleMigration 
> end
> ```
> 
> https://api.rubyonrails.org/classes/ActiveRecord/IrreversibleMigration.html

They actually can be reversed b/c of how I defined `drop_table` by specifying the column names/types -- it's just that we would lose the data and have to restore from a backup. http://joshfrankel.me/blog/ensure-dropping-a-database-table-is-reversible/

StetsenTech:
The data loss and data integrity is my biggest concern once we start having data in our production database especially on user tables. If we were to rollback migrations to a point before this, how do we associate the data from other tables like program members, roles, etc. between Users and Person? We could restore the database with a backup, but imo that defeats the purpose of migrations.

Also, since you have started reworking the foundation for our tables, it would nice to have a point where we know that this is the beginning of our current infrastructure.

sadleb:
> Also, since you have started reworking the foundation for our tables, it would nice to have a point where we know that this is the beginning of our current infrastructure.

This is a good point. I'll make it irreversible. See https://github.com/bebraven/platform/pull/80/commits/bfae7b187dab120db1d0e9e0ea295521d7b81db7

PR #81, opened 2020-01-24
Add a convenience script to connect to the Docker rails console.

sadleb:
I have aliases for these so I can just type `devc` in any of the
Braven repos and connect to the rails console for that project.


sadleb:
Not worth a review. This just mimic what all the other Braven repos have setup currently.

PR #82, opened 2020-01-24
Fix high CPU usage on a Mac when running Docker container.

sadleb:
See: https://app.asana.com/0/1156506513832825/1158558207611705

We were running guard with force polling on because using the filesystem
event notificaiton stuff wasn't working. It was pinning the CPU at 100%+
This commit adds libnotify to the Docker container, adds the gem to use it,
and then runs gaurd without forced polling.

**IMPORTANT NOTE:** there are issues with certain editors, like vim, when you
edit from the host machine. For vim, you have to add this setting to your
.vimrc file:
`set backupcopy=yes`

See the following for more details
- https://github.com/guard/listen/issues/434
- https://github.com/guard/listen/issues/420

TESTING:
- Rebuilt the container and edited an .erb file from vim on the host Mac machine
  with `backupcopy=yes` turned on. The change was live reloaded.
- Tried with VSCODE as well. That worked.
- Oh, and my machine is quite as a mouse now. Going into the container and running `top` shows no pinned CPU for the guard process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,9 @@ RUN apk add --update --no-cache \
     nodejs \
     yarn \
     tzdata \
-    git 
+    libnotify \
+    git \
+    vim
 
 COPY Gemfile* /usr/src/app/
 COPY package.json /usr/src/app/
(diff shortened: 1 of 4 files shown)


rshipp:
Wild. Thanks for doing the research.

sadleb:
I couldn’t take it anymore. It wasn’t a priority on my list but it was like a mosquito in the room that I had to squash

PR #84, opened 2020-01-28
Add retained data scaffolding, new right-sidebar options, and done buttons

rshipp:
Sorry these are all in the same PR, I got halfway through retained data and realized I needed to change things around in my planned implementation.

Changes: 

* ~~Added Rails APIs (`rails g scaffolding <...> --api`) for `CourseContentAnswers`~~
* Remove the correctness dropdowns from the editor inline view
* Add correctness dropdowns (when a radiobutton/checkbox is selected) to the right sidebar
  * (And implement functionality.)
* Add instant-feedback toggles to the right sidebar (when a question is selected)
* Add done button elements, and place them in questions by default
  * No UI for adding them in custom locations yet.
* Added a `RetainedData` plugin that manages IDs
* Implemented retained data throughout
* Some refactoring

Related tasks: 

[✓ Rails: Implement magic-field/correctness/(mastery?) API with view/create/update](https://app.asana.com/0/0/1154794977027322/f)
[✓ Editor JS: Add a checkbox for toggling instant feedback on each checkbox/radio](https://app.asana.com/0/0/1155859225081672/f)
[* Editor JS: Tie in correctness dropdown to Rails API](https://app.asana.com/0/1142638035116665/1154794977027323/f)
[* CK/Editor JS: Implement UI for done buttons (in both questions and sections)](https://app.asana.com/0/1142638035116665/1155859225081635/f)
[✓ Editor JS: Generate and manage magic field IDs](https://app.asana.com/0/1142638035116665/1154794977027321/f)


app/javascript/components/ContentEditor.js
```diff
@@ -452,8 +456,47 @@ class ContentEditor extends Component {
...
+                                                <option value="">CHOOSE ONE</option>
+                                                <option value="correct">Correct</option>
+                                                <option value="incorrect">Incorrect</option>
+                                                <option value="maybe">Maybe</option>
+                                            </select>
+                                            <label htmlFor='input-correctness'>Correctness</label>
+                                        </>
+                                    );
+                                } else if ( ['question'].includes( modelElement ) ) {
+                                    const questionElem = getNamedAncestor( 'question', this.state['selectedElement'] );
```
rshipp: we get the named ancestor here because we want to set the 'instant-feedback' attribute on the `question` element, but allow the current selection to be *any element inside the `question` element*

db/migrate/20200124194648_create_course_content_answers.rb
```diff
@@ -0,0 +1,15 @@
...
+    create_table :course_content_answers do |t|
+      t.string :uuid
+      t.references :course_content, null: false, foreign_key: true
+      t.boolean :correctness
+      t.boolean :mastery
+      t.boolean :instant_feedback
+
+      t.timestamps
+
+      t.index [:course_content_id, :uuid], unique: true
```
rshipp: this might be overkill, but I can only enforce same-page uuid uniqueness from the frontend. chances of a collision are extremely low, but i don't want to deal with some frustrating-to-debug birthday problem nonsense years from now when I've forgotten how all this works.

rshipp:
note: the tests will fail until I update the URLs (because course_content_answers is a nested resource), so I won't merge until then.

app/javascript/ckeditor/retaineddata.js
```diff
@@ -0,0 +1,48 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+
+const RETAINED_DATA_ATTRIBUTE = 'data-bz-retained';
+
+export default class RetainedData extends Plugin {
```
rshipp: this is where the retained data magic happens! ✨

app/helpers/course_content_answers_helper.rb
```diff
@@ -0,0 +1,2 @@
+module CourseContentAnswersHelper
+end
```
StetsenTech: What's going to go in here?

app/javascript/ckeditor/iframecontentediting.js
```diff
@@ -95,7 +88,6 @@ export default class IFrameContentEditing extends Plugin {
         conversion.for( 'editingDowncast' ).elementToElement( {
             model: 'iframe',
             view: ( modelElement, viewWriter ) => {
-                alert('here');
```
StetsenTech: 👀 Good thing to remove.

db/migrate/20200124194648_create_course_content_answers.rb
StetsenTech: Fair.

app/helpers/course_content_answers_helper.rb
rshipp: idk, this is just part of the rails scaffold

db/migrate/20200124194648_create_course_content_answers.rb
rshipp: i ended up using ints instead of uuids anyway, I'm currently not using course_content_answers at all. We should maybe remove it tbh.

app/controllers/course_content_answers_controller.rb
```diff
@@ -0,0 +1,58 @@
+class CourseContentAnswersController < ApplicationController
+  before_action :set_course_content_answer, only: [:show, :update, :destroy]
```
sadleb: You should be able to just delete `set_course_content_answer ` and DryCrud will take care of it (but double check that it runs on destroy too, b/c i may have missed that one). But see the other comment about "we should prob just get rid of it tbh"...

app/controllers/course_content_answers_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  before_action :set_course_content, only: [:index]
```
sadleb: Ditto on DryCrud if you make this nestable.

app/javascript/ckeditor/retaineddata.js
rshipp: will do

app/javascript/ckeditor/retaineddata.js
```diff
@@ -0,0 +1,48 @@
...
+        return 'RetainedData';
+    }
+
+    init() {
+        const editor = this.editor;
+        const editingDoc = editor.editing.view.document;
+
+        const pageId = editor.config.get( 'retainedData.pageId' );
+
+        this._idCounter = 1;
```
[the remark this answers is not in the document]
rshipp: A couple corrections:

* When the content editor is loaded, the RetainedData plugin will be loaded and the init() method is called setting `_idCounter` to 1.
* When content is loaded into the editor, each retained data ID will fire a `change:attributes` event, which will be handled by the `_consumeAttributeEvent` method, setting the `_idCounter` to the higher of itself or the newly processed ID. Once the content is fully loaded, the ID will be the highest existing ID (that matches the prefix) in the document.
* The life of that plugin object is until the CKEditor component is destroyed (page reload, or clicking the HTML tab).
* _idCounter will be reset to 1 on plugin initialization (see above).

app/javascript/ckeditor/retaineddata.js
rshipp: added some comments to the plugin, let me know if that clears it up at all.

app/assets/stylesheets/bz_newui.css
```diff
@@ -1273,7 +1273,7 @@ table {
 	height: 378px;
 }
 
-.bz-box:not(.video):not(.flush) figure iframe {
+.bz-box:not(.module-block-video > div.content):not(.module-block-flush > div.content) figure iframe {
```
StetsenTech: This isn't valid CSS if I recall. `:not()` can only take a single condition as an argument.

rshipp: hmm i think you're right. I'll figure out what I was doing here and fix it.

app/javascript/ckeditor/retaineddata.js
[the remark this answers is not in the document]
sadleb: Super helpful!

PR #85, opened 2020-01-29
First pass at some grading related data models

sadleb:
I'm starting to flesh out the grading data model. I'm going to use the [grading feature branch](https://github.com/bebraven/platform/tree/grading) for this and make smaller PRs against that for incremental review before we merge back to staging since this touches the User model which is in live use on production.

To this end, I'm mostly focused on the models and their associations to one another. There are no specs, validations, or even most columns on the models yet. I have added a basic UI to be able to navigate between the models to get a sense of what it would be like to start writing views and logic using these models. Also, to be able to visualize the models and how they relate. I added some dummy records for this purpose which you can load by doing the following

`docker-compose exec platformweb bundle exec rake db:migrate db:dummies`

Here are some screenshots of what this looks like:

image

image

image

image


rshipp:
Can you add `db:dummies` to the `postdeploy` [script](https://github.com/bebraven/platform/blob/grading_poc/app.json#L10)? It's [supposed](https://devcenter.heroku.com/articles/github-integration-review-apps#the-postdeploy-script) to make review apps work, though it doesn't seem to be running correctly.

README.md
```diff
@@ -118,6 +126,25 @@ Tests:
...
+for the particular type of thing you are adding. Rails has a bunch of default generators you can use.
+For example, to create a new model you can do:
+
+    rails generate scaffold User name:string
+
+which will create the model, migration, routes, views, controllers, tests, etc.
+
+However, the default ones create a lot of cruft and things we don't use. Also, we can enforce our
+project structure and make it easier to get any boilerplate in that matches how we've set things up
+by creating our own generators. See [here](https://guides.rubyonrails.org/generators.html) for more info.
```
rshipp: +1 I've been thinking about doing this but I wasn't sure how

app/controllers/course_modules_controller.rb
```diff
@@ -0,0 +1,21 @@
+class CourseModulesController < ApplicationController
+  before_action :set_course_module, only: [:show]
+
+  # GET /course_modules
+  # GET /course_modules.json
+  def index
+    @course_modules = params[:program_id] ? CourseModule.where(program_id: params[:program_id]) : CourseModule.all
```
rshipp: i've been using a nested resource pattern instead of query params for my dummy editor models. we should come up with some Rails URI/API design guidelines or something before we get too deep into fleshing all these out.

app/models/lesson_submission.rb
```diff
@@ -0,0 +1,9 @@
+class LessonSubmission < ApplicationRecord
+  belongs_to :user
+  belongs_to :lesson
+
+  # Example Usage: 
+  # submissions = LessonSubmission.for_lessons_and_user(course_module.lessons, user)
+  scope :for_lessons_and_user, ->(ls, u) { where(lesson: ls, user: u) }
```
rshipp: cool! I had to look up what [scope](https://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-scope) does, I haven't seen this before

app/views/users/show.html.erb
```diff
@@ -22,10 +22,12 @@
           <% @user.program_memberships.each do |membership| %>
             <li class="list-group-item">
               <b><%= membership.program.name %> <%= membership.role.name %></b>
-              <br>
+              <br/>
```
rshipp: relatable

config/routes.rb
```diff
@@ -11,7 +11,25 @@
   resources :interests, except: [:show]
   resources :locations, only: [:index, :show]
   resources :majors, except: [:show]
-  resources :programs, except: [:show]
+
+  # See this for why we nest things only 1 deep:
+  # http://weblog.jamisbuck.org/2007/2/5/nesting-resources
```
rshipp: ooooh this is exactly what I was looking for

app/models/lesson_submission.rb
sadleb: I’m just learning about scopes. Scopes are dope! They’re lamdas/closures with a pretty name. In other words, they’re a block of code with a nice name that you can stick in a chain of calls and they abstract out some logic and deliver it as a convenient name to use, almost as if it’s an attribute of an object.

app/controllers/course_modules_controller.rb
sadleb: Yeah, I want to talk about this pattern. It felt wrong / smelled bad — but for the sake of getting some thing that showed the relationships b/n objects in a UI I didn’t take more time to figure out a better way. I want a better way tho!

config/routes.rb
sadleb: You should have seen what I had going before I found this technique. It was fugly. I like this approach a lot

StetsenTech: Elegant. 😲

hcli-work:
I still don't quite understand the distinction between Program and CourseModule. I remember we talked about how that would make historical data hard.

app/controllers/course_modules_controller.rb
sadleb: Single id. This shows the list of course modules for a single program if it's called as `program/:id/course_modules`

app/controllers/course_modules_controller.rb
```diff
@@ -0,0 +1,21 @@
+class CourseModulesController < ApplicationController
```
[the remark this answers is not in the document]
sadleb: The Braven App is supposed to become "the app" for anything we build at Braven. CourseContents and CourseModules were named this way as a first pass b/c Content and Module seemed too generic and as we built out more functionality over the years there may be other things other than a course that have content and modules. Maybe we decide to namespace everything from the get-go? E.g. `CAS::Blah`, `Course::Modules`, `Course::Contents`, `Course::Lessons`, etc? Or I guess just use the generic name since the course is the only thing in the app and then split things out into namespaces as we add more functionality?

app/models/program.rb
```diff
@@ -2,6 +2,7 @@ class Program < ApplicationRecord
   has_many :program_memberships
```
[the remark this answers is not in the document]
sadleb: That was a hold-over from the data central project. Since there are more types of "programs" at Braven than just those with an associated online course, I think that's why it was named that way. E.g. the Pre-Accelerator program didn't have an online course, but it had start dates and end dates and people in the program.

app/controllers/lesson_submissions_controller.rb
```diff
@@ -0,0 +1,21 @@
+class LessonSubmissionsController < ApplicationController
```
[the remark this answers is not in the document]
sadleb: See app/models/lesson.rb for a description. Ditto on app/models/course_module. Lmk if the difference doesn't make sense once you read the descriptions.

app/models/course_module.rb
```diff
@@ -0,0 +1,16 @@
...
+# to the topic of the first week of class.
+class CourseModule < ApplicationRecord
+  
+  belongs_to :program
+  has_many :projects
+  has_many :lessons
+
+  validates :name, presence: true
+
+  def grade_for(user)
```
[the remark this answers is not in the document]
sadleb: Feels more intuitive that if you're looking for the grade for a course module, it would have a method there instead of having to hunt for the helper that actually has the logic. I'm a rails noob, so maybe in the rails world people know to look for helpers first when trying to find logic applied to a model?

app/helpers/grades_helper.rb
```diff
@@ -0,0 +1,8 @@
+module GradesHelper
+
+  # E.g. the float 0.5 would return 50% as a string
+  def as_percent decimal_value
```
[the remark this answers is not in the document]
sadleb: Done. See https://app.asana.com/0/1156506513832825/1161295460290153 too - we should create a linter to enforce this style guideline.

app/models/course_module.rb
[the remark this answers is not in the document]
sadleb: fyi: Harry and I discussed this offline and after much consideration agreed that it's more "railsy" to have these convenience methods on the model. In rails, things should just work without having to hunt down things. So it's natural to expect a `grade_for(user)` to exist on the model instead of having to hunt down the helper/service/presenter class that calculates the grade.

app/models/user.rb
```diff
@@ -44,6 +50,10 @@ def current_membership program_id
     program_memberships.current.find_by program_id: program_id
   end
 
+  def total_grade program
```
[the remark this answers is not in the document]
sadleb: Decision is not to. Make it convenient for someone using a `user`

sadleb:
> I still don't quite understand the distinction between Program and CourseModule. I remember we talked about how that would make historical data hard.

Think of Program as a Course (though there can be programs that are not courses, like the pre-accelerator program). A CourseModule is a grouping of things in a course. A course has a bunch of modules, which are just a group of things to do such as a lesson and a survey or a lesson and project and a survey or just a lesson.

app/controllers/lesson_submissions_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  # GET /lesson_submissions/1
+  # GET /lesson_submissions/1.json
+  def show
+  end
+
+  private
+
+  # Use callbacks to share common setup or constraints between actions.
```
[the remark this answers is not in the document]
sadleb: See: https://github.com/bebraven/platform/pull/96 -- I'm going to pull this in and update the dry_crud controller to handle my weird `params` index logic.

app/controllers/course_modules_controller.rb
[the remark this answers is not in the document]
hcli-work: I definitely lean towards simplest first and then renaming if needed.

sadleb:
> Can you add `db:dummies` to the `postdeploy` [script](https://github.com/bebraven/platform/blob/grading_poc/app.json#L10)? It's [supposed](https://devcenter.heroku.com/articles/github-integration-review-apps#the-postdeploy-script) to make review apps work, though it doesn't seem to be running correctly.

Update @rshipp - I tried this but `FactoryBot` class's gem isn't installed in a prod environment. db:dummies relies that. Longer term, I'm not sure if we'll maintain a db:dummies file as the way we populate dev and review app DBs so I punted on doing this and just used a plain ol' db:seed command which currently just creates that one user to login with. 

Long term my gut is that we should have a way to populate dev, staging, and review app DBs from sanitized (potentially stripped down) versions of the prod DB.

PR #86, opened 2020-01-29
Allow running Selenium specs inside the Docker container.

sadleb:
This uses a separate Docker container that has chrome setup in it.
It sets up a remote driver to use that container to execute Selenium
tests.

Note: I had a heck of a time getting this working b/c Capybara
sets up a server listening on localhost and some seemingly random port.
The way this works is that you run rspec which is configured with
Capybara. It uses a test database, mocks out and stubs out various things,
and runs as a web server. When Selenium tests are used (aka those that
need a javascript driver) the chrome binary on the other Docker container
needs to be able to call back into the web-server that rspec/capybara
have set up.

TESTING:
- Ran the following and those tests pass now:
    `docker-compose exec platformweb bundle exec rspec spec/feature`


spec/rails_helper.rb
```diff
@@ -2,6 +2,10 @@
 require 'codacy-coverage'
 Codacy::Reporter.start
 
+# Make sure the local ENV variables are loaded
+require 'dotenv'
+Dotenv.load('.env')
```
rshipp: can we get rid of these lines? or do we need them somewhere? My local setup (outside of docker) errors out if it loads the .env like this, because it tries to use the wrong database. But inside of docker seems to work even without this here, since the env vars are already loaded by docker-compose.

sadleb: Done.

PR #88, opened 2020-02-09
Add rake task to restart dynos.

sadleb:
This is intended to be used from the Scheduler add-on
(or some other add-on) to restart the dyno in the middle of
the night daily so that we're not at the whims of the dyno
manager cycling it.

TESTING:
- After creating an oauth token using
   heroku authorizations:create -d "Platform Admin API token" -s write
 I ran:
   HEROKU_OAUTH_TOKEN=my-token HEROKU_APP_NAME=my-app bundle exec rake heroku:app:restart_dyno[web]
 and verified that the dyno restarted

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,3 +94,9 @@ gem 'rubycas-server-activerecord'
 
 # Honeycomb
 gem 'honeycomb-beeline', require: false
+
+# Allows us to write rake tasks that can programatticaly run Heroku commands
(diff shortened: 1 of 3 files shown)


sadleb:
There are no active users right now and I want to deploy so I'm going to override and just merge before getting a review. This rake task is copied from the Portal code which is already tested, so the review is more of an FYI.

PR #90, opened 2020-02-09
Attempt to fix review apps database population.

sadleb:
The app.json had everything in the "test" environment, so the
postdeploy script wasn't running. Move the review apps related
postdeploy to the "review" environment and cleanup the buildpack
that applies to all environments.

TESTING:
- F.It, we'll do it live!
  (it's the only option)

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,15 +1,21 @@
 {
+  "buildpacks": [
+    { "url": "heroku/ruby" }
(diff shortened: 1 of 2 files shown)


sadleb:
Not waiting for a code review b/c we won't know until we release. The review request is more of an FYI.

PR #94, opened 2020-02-09
Think I finally fixed Review Apps database population

sadleb:
Also, make db:migrate automatically run during releases.


sadleb:
Yup, it works! Now review apps can be logged into without having to setup the DB.

PR #96, opened 2020-02-11
DRY-up our controllers a bit with some meta-programming

sadleb:
This allows us to create new controllers that inherit from ApplicationController and they auto-magically have the instance variables set that `index`, `show`, and `edit` CRUD methods rely on. I'll be adding more functionality to this for nested resources [over here](https://github.com/bebraven/platform/pull/85#discussion_r376598338) which sparked this PR.

Note: this is by no means high on the priority list of things to do, but it was fun to tinker with and getting some of this stuff setup early one just means less and less code to maintain in multiple spots, so I think it's worth it on both fronts.

TESTING
- all specs pass
- clicked around the various routes to make sure that show, edit, and index routes don't throw exceptions

NOTE:
- I borrowed heavily from for implementation inspiration: https://github.com/codez/dry_crud

### DIFF
--- a/lib/dry_crud.rb
+++ b/lib/dry_crud.rb
@@ -1,0 +1,22 @@
+# Helps DRY up your controllers. 
+module DryCrud
+
+  module Controllers
+    extend ActiveSupport::Concern
+  
+    # DRY's up the show method by setting the instance variable
+    # to the current instance of the module
+    included do
+      before_action :set_model_instance, only: [:show, :edit, :update, :destroy, :publish]
+    end
+      
+    # E.g. if your controller is called LessonsController this will set the instance variable like so:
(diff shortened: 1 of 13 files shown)


lib/dry_crud.rb
```diff
@@ -0,0 +1,58 @@
...
+    def dry_crud_enabled?
+      true
+    end
+
+  protected
+
+    # E.g. if your controller is called LessonsController this will set the instance variable like so:
+    # @lessons = Lessons.all
+    def set_models_instance
+      instance_variable_set(:"@#{instance_variable_name.pluralize}", model_class.all) if dry_crud_enabled? && model_class
```
hcli-work: it's so DRY it hurts

hcli-work:
love this.

PR #98, opened 2020-02-11
Adds Course model along with specs

hcli-work:
This pull request adds a couple of columns to Program so that it can belong to an organization. Made a Course a subclass of Program. Popped in some validations along with requisite specs.

Test Plan: `bundle exec rspec spec/models`


db/schema.rb
```diff
@@ -146,6 +146,10 @@
     t.string "name", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "term", null: false
```
sadleb: The concept of a `term` you're introducing here is related to `program_memberships.start_date` and `end_date`. I suspect we'll eventually want `term` to be its own table with an id, when it starts and ends, it's name, etc. b/c we'll want to map different types of things other than just a course to that term. E.g. volunteers, employer partnership programs.

hcli-work:
Yeah, I somehow completely broke my branches and didn't pull in the commit that fixed the issue with other specs breaking. Rebasing is hard.

db/schema.rb
hcli-work: it looks like ProgramMembership has dates on it, which can be different per user in the Program. That feels weird. If/when we have start/end dates on the course, wouldn't we just want one for the entire course?

hcli-work:
Alright. This is ready for re-review. All checks pass and dealt with that migration Codacy thing with NOT NULL and default value.

db/schema.rb
sadleb: I agree that the way program membership does it is probably not the best way. I'm just musing over whether it's worth refactoring program membership to use the same data model when defining a `term` for a program or course with start and end dates. It feels messy to introduce what is logically supposed to represent the same thing, when a program/course is running, but not refactor. Totally fine if that's outside the scope of what you're trying to accomplish in which case I'd queue up an Asana task to do that refactor in a future iteration @hcli-work

sadleb: I agree that the way program membership does it is probably not the best way. I'm just musing over whether it's worth refactoring program membership to use the same data model when defining a term for a program or course with start and end dates. It feels messy to introduce what is logically supposed to represent the same thing, when a program/course is running, but not refactor. Totally fine if that's outside the scope of what you're trying to accomplish in which case I'd queue up an Asana task to do that refactor in a future iteration @hcli-work

db/migrate/20200211154757_add_organization_to_program.rb
```diff
@@ -0,0 +1,15 @@
+class AddOrganizationToProgram < ActiveRecord::Migration[6.0]
+  def change
+
+  	# in prod, there are currently no rows in the programs table
+    add_column :programs, :term, :string, null: false, default: ''
```
sadleb: It's a bit bonkers that this is the cleanest way to accomplish this, but hey, if it works! I'm going to try and bookmark this in head for when I have to do this next.

spec/controllers/programs_controller_spec.rb
```diff
@@ -71,60 +68,77 @@
...
+          }
+        )
+      end
+
       context "with valid params" do
         it "creates a new Program" do
-          expect {
-            post :create, params: {program: valid_attributes}, session: valid_session
-          }.to change(Program, :count).by(1)
+          expect { post_create }.to change(Program, :count).by(1)
```
sadleb: clean

db/migrate/20200211154757_add_organization_to_program.rb
rshipp: we can also dismiss specific instances of complaints in Codacy, without disabling a rule entirely. I think that's appropriate in a case like this.

sadleb: good to know!

PR #99, opened 2020-02-12
DRY-up nested controllers

sadleb:
I already created a module, dry_crud, that uses meta-programming to do the repeated code that gets and sets the instance variables to the correct models for CRUD operations. This PR extends that module to handle nested resources. Here's how it works:

Say we have a `ParentModel` that `has_many` `ChildModels`. If the controller for showing `ChildModels` is configured with nested resource routes such that [https://<root>/parent_model/:id/child_models]() should show a list of all `ChildModels` for that particular `ParentModel`, then in your `ChildModelsController` simply do this:
```
ChildModelsController < ApplicationController
  include DryCrud::Controllers::Nestable
  nested_resource_of ParentModel
  ... 
end
```
Then if you hit the above route it will only show ChildModels for that particular parent. But if you hit [https://<root>/child_models]() it will show them all since the route isn't nested.

TESTING:
- Set this up on a couple nested resource, like LessonsController
  and CourseModulesController and clicked around as well as manually putting
  in both nested routes and root routes to make sure it all worked.


lib/dry_crud.rb
```diff
@@ -54,6 +54,95 @@ def instance_variable_name
...
+    #
+    # class ProjectSubmissionsController < ApplicationController 
+    #   include DryCrud::Controllers::Nestable
+    #   nested_resource_of Project
+    #   ...
+    # end
+    #
+    # Multiple potential parents can be specified like:
+    #  nested_resource_of Project, Course
+    module Nestable
```
sadleb: This is the meat of the PR.

lib/dry_crud.rb
```diff
@@ -54,6 +54,95 @@ def instance_variable_name
...
+      extend ActiveSupport::Concern
+      include Controllers
+
+      included do
+        delegate :parent_resource_classes, to: 'self.class'
+      end
+
+      class_methods do
```
sadleb: This tripped me up a bit when trying to implement, so here's more color to what is going on here. When the module is included that means that the class that included it will have instance methods injected. However, `nested_resource_of` needs to be a class method. Something like `self.nested_resource_of` if you were defining it in the class directly. This `class_methods` block is provided by `ActiveSupport::Concern` as a convenient way to essentially call the following which "extends" the class including this module adding class level methods:

```
    included do |base|
      base.extend ClassMethods
    end

    module ClassMethods
      def nested_resource_of(classes)
      ...
      end
    end
```

sadleb: And also, I needed to delegate `parent_resource_classes` to the class including this module so that the other methods defined in this module could see it.

lib/dry_crud.rb
```diff
@@ -54,6 +54,95 @@ def instance_variable_name
...
+          end
+        else # Single nestable class
+          ret = set_parent_instance_var(parent_resource_classes)
+        end
+        ret
+      end
+
+      def set_parent_instance_var(klass)
+        id = parent_id(klass)
+        instance_variable_set(parent_instance_var_name(klass), klass.find(id)) if id
```
sadleb: This is designed to return `nil` if there is no parent so that the `models_list` knows to just use the default scope which is `all`

lib/dry_crud.rb
rshipp: So this specific method has to be a class method, but the rest have to be instance methods, am I understanding that correctly?

rshipp:
This is beyond my Ruby knowledge so I can't critique the implementation. The `nested_resource_of` interface seems clean.

One question, will this handle https:///parent_model/:id/child_model/:child_id too? or does it stop at the listview?

lib/dry_crud.rb
sadleb: Yeah, the rest are instance methods added to the class that includes this module. I was having trouble figuring out how to access the class method `parent_resource_classes` from the instance method `parent` and this seemed to work. The more I think about it, if I had done `self.class. parent_resource_classes` from within the `parent` method it prob would have worked.

sadleb:
> This is beyond my Ruby knowledge so I can't critique the implementation. The `nested_resource_of` interface seems clean.
> 
> One question, will this handle https:///parent_model/:id/child_model/:child_id too? or does it stop at the listview?

That route is for a specific ChildModel, so it would call the `show` method of the `ChildModelsController` passing in the `params[:id]` which isn't impacted by this change. This change only deals with how to set the `@child_models` instance variable used by `index` when a list of models needs to be shown.

sadleb:
> > This is beyond my Ruby knowledge so I can't critique the implementation. The `nested_resource_of` interface seems clean.
> > One question, will this handle https:///parent_model/:id/child_model/:child_id too? or does it stop at the listview?
> 
> That route is for a specific ChildModel, so it would call the `show` method of the `ChildModelsController` passing in the `params[:id]` which isn't impacted by this change. This change only deals with how to set the `@child_models` instance variable used by `index` when a list of models needs to be shown.

Also, to be fair, this is beyond my Ruby knowledge too. It's stuff I'm just learning and tinkering with as I go!

sadleb:
> zowie. you've pushed into some advanced Ruby and Rails here. this is awesome.
> I think after committing this you should look into writing controller specs in RSpec. I think you would have really enjoyed DRYing this up with controller specs that you could do the whole red, green, refactor thing with.

100% agree. I created a task to put in some specs at some point (need to balance with overall goal of data modeling) -- but I def need to get into the habit of writing tests first and refactoring until they're green. https://app.asana.com/0/1156247804053609/list

PR #100, opened 2020-02-12
Add honeycomb release markers

rshipp:
This adds Honeycomb markers on Heroku releases. What it looks like:

image

Tested on the release app.


release-tasks.sh
```diff
@@ -0,0 +1,9 @@
+# Migrate
+bundle exec rake db:migrate
+
+# Send a Honeycomb marker
+RELEASE_CREATED_AT=$(date +%s -ud ${HEROKU_RELEASE_CREATED_AT})
+curl https://api.honeycomb.io/1/markers/platform.bebraven.org \
```
sadleb: Can you use `HONEYCOMB_DATASET` instead of `platform.bebraven.org`

sadleb: Also, this is a sweet PR. I'm pumped to have these!

PR #102, opened 2020-02-14
First pass at rubrics related data models

sadleb:
[Here is an overview](https://docs.google.com/document/d/160uyajp_be6V8v2aaoIa6NxrlT4DmXXqa5SJRLaJ3Rc/edit#heading=h.d5ihwqwhakzx) of these data models and what they represent:


app/models/rubric_grade.rb
```diff
@@ -0,0 +1,8 @@
+# This represents the grade given for a project submission using a rubric.
+# It is the collection of grades chosen for each row using guidance from
+# the ratings defined for the row.
+class RubricGrade < ApplicationRecord
```
hcli-work: I don't know if this is a good idea...
The naming here is slightly confusing because it's not clear which things are part of the rubric and which things are tied to the submission. Like the difference between a `RubricGrade` and a `RubricRowGrade` is going to cause my brain to explode one day.

Could we call this something like `SubmissionRubricGrade`. I know it's longer...

db/migrate/20200214051057_create_rubric_grades.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRubricGrades < ActiveRecord::Migration[6.0]
+  def change
+    create_table :rubric_grades do |t|
+      t.belongs_to :project_submission, foreign_key: true, null: false, index: { unique: true }
```
hcli-work: oooh. I didn't know you could do that with `foreign_key`

app/models/rubric_grade.rb
sadleb: I'm envisioning eventually having a `RubricGrade` an `EngagmentGrade` (or something like that), a `QuizGrade`, etc. 

A `RubricGrade` isn't too complicated, b/c its just comprised of a bunch of grades given for each rubric row, so `SubmissionRubricGrade` and `SubmissionRubricRowGrade` isn't the worst. But once we get to grading lessons with all the complicated stuff that goes into that automatic grade we'll likely end up with things like `SubmissionChecklistGrade`, `SubmissionRadioListGrade`, `SubmissionDragAndDropGrade`.  I'm leaning towards just sticking with `RubricGrade` for now because I can't really imagine the concept of giving a grade on something that isn't submitted. I suspect that once we use the code a little, it'll be more natural that an XXXGrade is the grade given for something using XXX as the grading strategy and implicitly everything graded is a submission of some sort @hcli-work

sadleb:
> Just one comment/suggestion.
> Could you also add the appropriate entries into the factories in a separate PR? That'll save the next person time.

Here is the followup PR with specs: https://github.com/bebraven/platform/pull/105

PR #103, opened 2020-02-20
Adding Section and Logistic models

hcli-work:
Logistic model captures the day of week and time of day that a section might meet. It belongs to the Course/Program.

A section belongs to a course/program. It also belongs to a logistic model, thereby determining when that section meets.

Test Plan: `bundle exec rspec spec`


db/schema.rb
```diff
@@ -34,6 +34,15 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "cohorts", force: :cascade do |t|
```
sadleb: Did you mean to add the cohorts model too or is this for a future PR?

hcli-work: Oh, weird. I decided to call them sections instead of cohorts to be consistent with Canvas. This must be because my database had a cohorts table leftover from a different branch. Let me figure out how to drop that.

spec/factories/logistics.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :logistic do
+    day_of_week { 'Tuesday' }
+  	time_of_day { '8:00pm' }
```
sadleb: Small nit, is there a tab in there? Time of day looks nested under day of week.

rshipp:
What's the intent behind using strings for date/time fields in the model, vs converting from datetime to human-readable on the frontend?

spec/factories/logistics.rb
hcli-work: good catch

hcli-work:
My ideal would have been to have the `time_of_day` field as a time, but there's apparently not a way to just have a time column type. I've only got datetimes, so it'll put a misleading date on there as well.

rshipp:
Oh, wild. I wonder why they did that. Is there a gem we could use that would allow stricter validation of text-based time fields? Only because I assume doing more validation here would save us effort having to check for bad data when we use this later on.

hcli-work:
> Oh, wild. I wonder why they did that. Is there a gem we could use that would allow stricter validation of text-based time fields? Only because I assume doing more validation here would save us effort having to check for bad data when we use this later on.

I like that idea. Different pull request though?

PR #104, opened 2020-02-20
add UserSection

hcli-work:
Adds a join table between users and sections. The join record has a type column associated to determine the reason why that user is associated with the section: enrolled as fellow? facilitates as leadership coach? assists as a TA?

Also adds some finder methods to make things convenient so you can do things like
```
u = User.last
u.sections.as_fellow

s = Section.last
s.users.only_tas
```

Test Plan: nothing really to test here. it's just defining relationships, db constraints, etc.


rshipp:
What do the `as_X` methods do? I haven't seen that before. Would it just return `nil` if they're not `X`?

hcli-work:
> What do the `as_X` methods do? I haven't seen that before. Would it just return `nil` if they're not `X`?

They wouldn't return `nil`.  They would return an empty Arel, which is like an empty list but fancier.

PR #105, opened 2020-02-24
Some basic specs for rubrics

sadleb:
Just tests that the validations work.

Test Plan:
`bundle exec rspec spec/models`


spec/factories/rubric_row.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :rubric_row do
+    criterion { 'Criterion to get credit for this row.' }
+    points_possible { 10 }
+    sequence(:position)
+
+    rubric_row_category
```
hcli-work: whoa. does this do the association thing I was doing manually?

sadleb: yup, newer version of FactoryBot allow [implicit associations](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#associations).

StetsenTech: TIL

PR #106, opened 2020-02-25
Cleanup grading data models based on feedback

sadleb:
See:
- https://app.asana.com/0/1156247804053609/1161955104333047
- https://app.asana.com/0/1156247804053609/1161955104333045
- https://app.asana.com/0/1156247804053609/1161615040658281

**Note: ignore all the controller, view, and route stuff when reviewing this. Just focus on the models, migrations, and specs.**

### Changes:
1) **Remove CourseModules and replace with GradeCategory.**
Eventually, we may bring the concept of CourseModules back,
but for the purposes of grading it didn't make sense to define
the percent of your total grade that some set of lessons or projects
made up on a CourseModule. The concept of a course module is simply to
group a related set of learning content into a topic or theme, mostly impacting
course navigation. For grading, we want to categorize projects
and lessons and set the percentage of the total grade based on that. E.g.
all lessons related to course engagement would be in an Engagement category,
which may span multiple weeks or modules but the entire category is worth 15%
of your total grade.

Note: this breaks the GradeCalculator so I just made it throw NotImplementedError until we actually need it.

2) **Switch grades within a category to be weighted based on percent instead of relative points.**

Points for one project/lesson shouldn't have an impact on it's weight towards your final grade. They are
only used for making an intuitive way to grade that particular project/lesson.

3) **Add some database constraints, model validations, and related specs.**

### Test Plan:
  `bundle exec rspec spec/models`


app/models/course.rb
```diff
@@ -1,4 +1,8 @@
 class Course < Program
+  has_many :grade_categories
+  has_many :projects
+  has_many :lessons
```
rshipp: are projects and lessons going to be linked both in courses and in grade_categories inside a course? and if so, which one is the source of truth / what do we do if they differ?

app/models/grade_category.rb
```diff
@@ -0,0 +1,16 @@
+# A GradeCategory groups various things that are graded,
+# like lessons and projects, into a category where collectively
+# everything in that category counts as X percent of your final grade.
+class GradeCategory < ApplicationRecord
+  
+  belongs_to :program
+  has_many :projects
+  has_many :lessons
+
+  validates :name, :percent_of_grade, presence: true
```
rshipp: should we validate that percent_of_grade is a float between 0 and 1? or do we want to allow things outside that range? (and same q for Lesson/Project's .percent_of_grade_category)

app/models/course.rb
sadleb: You just made me realize that this should be:
 ```
has_many :projects, :through => : grade_categories
has_many :lessons, :through => : grade_categories
```

I'll make the change.

app/models/grade_category.rb
sadleb: Hmm. Good call pointing this out. I think for now I should validate b/n 0..1. If we implement some extra credit thing in the future, we can revisit if it makes sense to do so by allowing more than 1.

I'll make the change

sadleb:
> I'm okay with this clean up. Kinda feel like we could somehow simply the tests in some way...but nothing comes to mind.

I have the same gut feeling. There must be a simpler, less redundant way to test the validations. For another day though...

PR #108, opened 2020-02-26
Editor assignment components

rshipp:
Reopening this (was PR #70) because these commits seem to have been lost from `staging` somehow.


rshipp:
...huh? i'm so confused

PR #113, opened 2020-03-02
Set new HTML flag before publishing

rshipp:
Task: https://app.asana.com/0/1142638035116665/1155859225081693/f

See also https://github.com/beyond-z/canvas-lms/pull/569

### DIFF
--- a/lib/canvas_api.rb
+++ b/lib/canvas_api.rb
@@ -2,6 +2,11 @@ require 'rest-client'
 
 class CanvasAPI
 
+  # Custom HTML to prepend to each body.
+  PrependHTML = %q(
+    <!-- BRAVEN_NEW_HTML -->
+  )
+
   def initialize(canvas_url, canvas_token)
     @canvas_url = canvas_url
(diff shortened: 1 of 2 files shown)


sadleb:
I was getting all sorts of confused about what this was. Just looking at the code without thinking back to what we've discussed about needing to look for a switch to tell if we need to use new JS/CSS, this looked like the API was giving a way to insert HTML before the body before sending it over. It wasn't intuitive that this was just a marker/flag for the Canvas side to be able to see and make decisions around.

I think just renaming `PrependHTML` to something like `PrependFlagMarkingThisAsNewHTML` (a mouthful, i know) would make future readers immediately understand that this is just a flag/marker. @rshipp

rshipp:
Valid complaint. My thought process was that we might want to prepend other things later on, but that's probably YAGNI. Followup task: https://app.asana.com/0/0/1164949031126081/f

rshipp:
Update: in #116 I'm prepending and appending other things, does that make that variable name make more sense?

PR #118, opened 2020-03-12
parameterized task to delete old rows from tickets tables

hcli-work:
A rake task to let us delete old rows from tickets tables. I envision using this with the Heroku Scheduler add on to run these tasks each night
```
rake maintenance:delete_old_tickets[service_tickets,7]
rake maintenace:delete_old_tickets[login_tickets,7]
```

Test plan: ran it with `bundle exec rake ...` and it removed rows


hcli-work:
I am aware of the SQL injection vulnerability here. I'm okay with that given the isolated use case. Also, I tried to get `ActiveRecord::Base.sanitize_sql` to work for me, but it kept on inserting single quotes around stuff which was messing up the table name after the FROM part of the SQL statement.

lib/tasks/maintenance/tickets.rake
```diff
@@ -0,0 +1,9 @@
+namespace :maintenance do
+  desc "delete old records in tickets tables"
+  task :delete_old_tickets, [:tablename, :num_days] => :environment do |_, args|
+    tbl = args[:tablename]
```
sadleb: It's unlikely, but hoo boy would it suck if we accidentally ran something like `rake maintenance:delete_old_tickets[users,7]` or less catastrophic `rake maintenance:delete_old_tickets[service_tickets,-7]` or more possible and probably not what we want `rake maintenance:delete_old_tickets[service_tickets,0]`.

Might be worth some safeguards. Only for certain tables and only for more than, say, 1 day ago (even though we should probably set this up for things in excess of 14 days or so when we actually schedule it.

hcli-work: I'm actually not that worried because the DELETE statement is based on a `created_on` column. By default, ActiveRecord tables have `created_at` and `updated_at` columns. It's only these tickets tables that have the `created_on` columns.

PR #120, opened 2020-03-12
Revert "marks some failing tests as pending"

sadleb:
Reverts bebraven/platform#119

The `Test run config vars` under Heroku CI settings needed to be updated to change `SSO_URL` from `http://stagingplatform.bebraven.org/cas/` to `https://stagingplatform.bebraven.org/cas/`.

For good measure, I added `FORCE_SSL=true` since that's the mode we run staging and prod in.


sadleb:
Added a review request as an FYI. I'm going to just merge this revert though since I didn't change any code, just ENV vars for the test environment.

PR #121, opened 2020-03-13
Extend checkbox/radio attribute support; refactor module blocks

rshipp:
This PR introduces some necessary refactoring to make module-block handling more simple and flexible. It also adds support for all currently known data-bz-* attributes to checkboxes and radio buttons (other elements in a later PR).

Tasks: 

* https://app.asana.com/0/1142638035116665/1165345563821547/f
* https://app.asana.com/0/1142638035116665/1165345563821545/f


app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -54,7 +55,7 @@ export default class ChecklistQuestionEditing extends Plugin {
             isInline: true,
             isObject: true,
             allowIn: [ 'checkboxDiv', 'tableCell' ],
-            allowAttributes: [ 'id', 'name', 'value', 'data-bz-retained', 'data-correctness' ]
+            allowAttributes: [ 'id', 'name', 'value', 'data-correctness' ].concat(ALLOWED_ATTRIBUTES),
```
rshipp: see the imported file for what's contained in this constant, but it's mostly data-bz-* things.

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -150,37 +151,40 @@ export default class ChecklistQuestionEditing extends Plugin {
...
-                    'id': id,
-                    'data-bz-retained': id,
-                    'data-correctness': viewElement.getAttribute('data-correctness') || ''
-                } );
+                return modelWriter.createElement( 'checkboxInput', new Map( [
+                    ...filterAllowedAttributes(viewElement.getAttributes()),
+                    ['id', id],
+                    ['data-bz-retained', id],
+                    ['data-correctness', viewElement.getAttribute('data-correctness') || '']
+                ] ) );
```
rshipp: Changing this from a raw JS Object to a Map, so we can use `filterAllowedAttributes` and the iterable "spread" syntax. Putting `...filterAllowedAttributes` first means the hardcoded values after that will override anything in the `filterAllowedAttributes` Map. This is important bc data-bz-retained is explicitly overriden here to match `id`.

app/javascript/ckeditor/moduleblockediting.js
```diff
@@ -0,0 +1,73 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import { toWidget } from '@ckeditor/ckeditor5-widget/src/utils';
+import Widget from '@ckeditor/ckeditor5-widget/src/widget';
+
+export default class ModuleBlockEditing extends Plugin {
```
rshipp: This plugin replaces all other `*Question` / `*Content` (e.g. `checklistQuestion`) elements with a single `moduleBlock` element that supports arbitrary classes, as long as each one has the `module-block` class.

PR #138, opened 2020-04-01
Handle share-release divs/checkboxes

rshipp:
Task: https://app.asana.com/0/1142638035116665/1169382763057601/f

Summary: allows the custom "div" element inside `checkboxDiv` and `questionFieldset`, because those are the two places we currently use `div class="share-release"`.

example: 
Screen Shot 2020-04-02 at 1 42 59 PM

context / things to note: This "custom div" code is super fragile, and our goal should be to eventually phase it out. We can do this by implementing more things (like this share-release div) as fully supported elements, and/or by removing them from the modules entirely (this should be preferred).

### DIFF
--- a/app/javascript/ckeditor/checklistquestionediting.js
+++ b/app/javascript/ckeditor/checklistquestionediting.js
@@ -54,7 +54,7 @@ export default class ChecklistQuestionEditing extends Plugin {
         schema.register( 'checkboxInput', {
             isInline: true,
             isObject: true,
-            allowIn: [ 'checkboxDiv', 'tableCell', 'div' ],
+            allowIn: [ 'checkboxDiv', 'tableCell', '$root' ],
             allowAttributes: [ 'id', 'name', 'value', 'data-correctness' ].concat(ALLOWED_ATTRIBUTES),
         } );
 
--- a/app/javascript/ckeditor/checklistquestionediting.js
+++ b/app/javascript/ckeditor/checklistquestionediting.js
@@ -6,11 +6,12 @@ import RetainedData from './retaineddata';
 import InsertChecklistQuestionCommand from './insertchecklistquestioncommand';
 import InsertCheckboxCommand from './insertcheckboxcommand';
 import SetAttributesCommand from './setattributescommand';
-import { ALLOWED_ATTRIBUTES, filterAllowedAttributes } from './customelementattributepreservation.js';
+import CustomElementAttributePreservation from './customelementattributepreservation';
+import { ALLOWED_ATTRIBUTES, filterAllowedAttributes } from './customelementattributepreservation';
 
 export default class ChecklistQuestionEditing extends Plugin {
     static get requires() {
-        return [ Widget, RetainedData ];
+        return [ Widget, RetainedData, CustomElementAttributePreservation ];
     }
 
     init() {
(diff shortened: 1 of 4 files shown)


app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -6,7 +6,7 @@ import RetainedData from './retaineddata';
 import InsertChecklistQuestionCommand from './insertchecklistquestioncommand';
 import InsertCheckboxCommand from './insertcheckboxcommand';
 import SetAttributesCommand from './setattributescommand';
-import { ALLOWED_ATTRIBUTES, filterAllowedAttributes } from './customelementattributepreservation.js';
+import { ALLOWED_ATTRIBUTES, filterAllowedAttributes } from './customelementattributepreservation';
```
rshipp, line 9: unrelated change I made because I noticed it at the same time 😞

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -54,7 +54,7 @@ export default class ChecklistQuestionEditing extends Plugin {
         schema.register( 'checkboxInput', {
             isInline: true,
             isObject: true,
-            allowIn: [ 'checkboxDiv', 'tableCell' ],
+            allowIn: [ 'checkboxDiv', 'tableCell', 'div' ],
```
rshipp, line 57: circular dependency with https://github.com/bebraven/platform/pull/138/files#diff-562cb730384caf99704ad067f46e2f56R162 😟

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -349,13 +349,14 @@ export default class ContentCommonEditing extends Plugin {
...
-                name: 'div'
+                name: 'div',
             },
         } );
         conversion.for( 'dataDowncast' ).elementToElement( {
             model: 'questionBody',
             view: {
-                name: 'div'
+                name: 'div',
+                classes: [ 'question-body' ],
```
rshipp, line 359: this is a migration change I'm making to prepare for eventually adding `class="question-body"` to all question body `div`s. (same with answer body below).

by adding this ONLY in the dataDowncast converter first, we allow reading the old-style question bodies (`div` without any classes inside a `div class="question"`) through the upcast converter, and writing them back out through dataDowncast to HTML. That means when you open a module, then save it, it'll have the new format. I decided this was an easier route than trying to script out the change or manually updating all the existing new-HTML.

context for this change: when I wrote these converters, I didn't know this would be a problem - now that I understand CKE better, it's glaringly obvious that ambiguous elements like this are a really bad idea. I've already lost time into debugging issues with this, and this is my attempt at getting ahead of the problem before it causes even more pain.

app/javascript/ckeditor/customelementattributepreservation.js
```diff
@@ -159,12 +159,15 @@ function setupAllowedAttributePreservation( editor ) {
     // Allow <div> elements in the model.
     editor.model.schema.register( 'div', {
         allowWhere: '$block',
+        allowIn: [ 'questionFieldset', 'checkboxDiv' ],
```
rshipp, line 162: this line is the meat of the PR. 😑

app/javascript/ckeditor/customelementattributepreservation.js
```diff
@@ -159,12 +159,15 @@ function setupAllowedAttributePreservation( editor ) {
...
         allowContentOf: '$root'
     } );
 
     // Allow <span> elements in the model.
     editor.model.schema.register( 'span', {
+        isInline: true,
```
rshipp, line 168: this fixes an unrelated rendering bug where spans were getting extra newlines added in the editing view. there's no change to the output HTML, it just makes it look nicer in the editor.

app/javascript/ckeditor/customelementattributepreservation.js
```diff
@@ -206,6 +217,8 @@ function setupAllowedAttributePreservation( editor ) {
         model: ( viewElement, modelWriter ) => {
             return modelWriter.createElement( 'span', filterAllowedAttributes( viewElement.getAttributes() ) );
         },
+        // Use low priority to make sure existing converters run first.
+        converterPriority: 'low'
```
rshipp, line 221: changing this to `low` because it's interfering with `span class="has-tooltip"`

app/javascript/ckeditor/checklistquestionediting.js
rshipp: fixed this by allowing checkboxInput in $root, which is much safer.

PR #146, opened 2020-04-09
Render portal images correctly in the editor

rshipp:
Task: https://app.asana.com/0/1142638035116665/1170693002395763/f

Summary: Prepend `https://portal.bebraven.org` to image `src` attributes that begin with `/`, only in the editing view.

Context: We have a lot of images in modules that have `src='/blah'`, and they don't render in the editor because it's hosted on a different domain than canvas. This PR makes it easier to edit existing modules. By using this custom converter on the editingDowncast chain ONLY, we avoid modifying any of the actual content, so it works transparently when publishing to portal. 

Before:
Screen Shot 2020-04-09 at 3 50 01 PM

After:
Screen Shot 2020-04-09 at 3 50 15 PM

### DIFF
--- a/app/javascript/ckeditor/portalimageediting.js
+++ b/app/javascript/ckeditor/portalimageediting.js
@@ -1,0 +1,43 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+
+import { getViewImgFromWidget } from '@ckeditor/ckeditor5-image/src/image/utils';
+
+const PORTAL_URL = 'https://portal.bebraven.org'
+
+export default class PortalImageEditing extends Plugin {
+
+    static get pluginName() {
+        return 'PortalImageEditing';
+    }
+
+    init() {
+        const editor = this.editor;
+        const conversion = editor.conversion;
+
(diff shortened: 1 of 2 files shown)


app/javascript/ckeditor/portalimageediting.js
```diff
@@ -0,0 +1,43 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+
+import { getViewImgFromWidget } from '@ckeditor/ckeditor5-image/src/image/utils';
+
+const PORTAL_URL = 'https://portal.bebraven.org'
```
sadleb, line 5: can you queue up a task to not hardcode this for the day that we actually have our staging env in decent shape and allow people to either use the staging editor or use the prod editor in staging mode where we work there before pushing to prod?

rshipp: ✅ https://app.asana.com/0/0/1170759051787969/f

worth noting that this is only for existing images - once we add a fully-featured "image upload" button that sends to canvas (or platform, in the case of LTI), we would use that instead and presumably re-upload all the existing images there.

PR #147, opened 2020-04-09
Improve checkboxDiv insertion/deletion

rshipp:
Don't know why this works. :/

I wish I could figure out what's going on here, but this still seems to be better than what we have now. At least it doesn't crash the whole editor. At this point I'm resigned to pushing it out and iterating as needed.

Task: https://app.asana.com/0/1142638035116665/1169461014498429/f

### DIFF
--- a/app/javascript/ckeditor/checklistquestionediting.js
+++ b/app/javascript/ckeditor/checklistquestionediting.js
@@ -24,6 +24,13 @@ export default class ChecklistQuestionEditing extends Plugin {
         // Add a shortcut to the retained data ID function.
         this._nextRetainedDataId = this.editor.plugins.get('RetainedData').getNextId;
 
+        // Listen for 'delete' events (includes Backspace).
+        this.listenTo( this.editor.editing.view.document, 'delete', ( evt, data ) => {
+            console.log(evt, data, this.editor.model.document.selection.getLastPosition());
+            data.preventDefault();
+            evt.stop();
+        } );
+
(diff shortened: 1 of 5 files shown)


app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -24,6 +24,13 @@ export default class ChecklistQuestionEditing extends Plugin {
         // Add a shortcut to the retained data ID function.
         this._nextRetainedDataId = this.editor.plugins.get('RetainedData').getNextId;
 
+        // Listen for 'delete' events (includes Backspace).
+        this.listenTo( this.editor.editing.view.document, 'delete', ( evt, data ) => {
+            console.log(evt, data, this.editor.model.document.selection.getLastPosition());
```
sadleb, line 29: I assume this logging is for dev purposes and you meant to delete?

rshipp: yep! deleting now, thanks for the catch

rshipp:
This page describes isObject vs isLimit: https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-SchemaItemDefinition.html. I still don't fully understand what they do though. :/

jessouyangbraven:
Trying to repro, and I must be doing something wrong since I get a crash when trying to insert rather than when trying to delete. You should land this, though, and we can screen share to figure out what's going on during our call on Friday.

PR #151, opened 2020-04-13
Fix iframe handling

rshipp:
Task: https://app.asana.com/0/1170776727341290/1171073105499761/f


jessouyangbraven:
Just to check that I understand the intended plugin behavior. This change lets you edit the title and body when you insert a new iframe in "Design"? 

Screen Shot 2020-04-13 at 7 32 08 PM

rshipp:
Yep. Now that i've seen some iframes being used in the actual content, I realized they have a title/body just like a video block. There's no way for a designer to add those elements to a "module-block", so it's better to have the button insert them.

PR #152, opened 2020-04-13
Fix editor performance

jessouyangbraven:
Task: https://app.asana.com/0/1170776727341290/1159217855939649

Test: 
- Copy-paste code from here: https://stagingplatform.bebraven.org/course_contents/1/edit
- Change text in "Design" tab, no lag
- Switch to "Code" tab, check that change is there
- Edit in "Code" tab, switch back, make sure changes are in "Design" tab

Details:
- When in "Code" tab, track the data in `this.state.editorData` because there's not logically a WYSIWYG `editor` in this view.
- When in "Design" tab, initialize with data from `this.state.editorData`, but then let `editor` track the content and don't touch `this.state`.
- This means we don't have to call `this.setState()` for the content and cause the entire editor to re-render every time we type, which was causing the lag. 
- Only when switching to "Code," update the state so it's reflected in the other view. 
- Also, only flip the `isPublished` bit once, because changing that state will also cause slow re-rendering.

### DIFF
--- a/app/javascript/components/ContentEditor.js
+++ b/app/javascript/components/ContentEditor.js
@@ -232,6 +232,7 @@ class ContentEditor extends Component {
         this.handleEditorDataChange = this.handleEditorDataChange.bind( this );
         this.handleEditorFocusChange = this.handleEditorFocusChange.bind( this );
         this.handleEditorInit = this.handleEditorInit.bind( this );
+        this.handleTabSelect = this.handleTabSelect.bind(this);
 
         // Non-CKE UI functions.
         this.fileUpload = React.createRef();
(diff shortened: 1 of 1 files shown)


jessouyangbraven:
Just noticed I developed this on top of `editor-class-conflict` and not `staging`. Will fix after meeting so the edited files are less confusing.

rshipp:
The lag is gone! 🥳

I am seeing that changes made in the design tab aren't saved unless you toggle the code tab first. Is there a way to have the save button update the editorData in the state or something?

jessouyangbraven:
@rshipp that should fix it for the "Save" button.

I'm having trouble with "Publish" both locally and on `stagingplatform`, any advice on testing would be nice. Happy to Slack/Zoom tomorrow if it's easier to screen share.

rshipp:
The publish button only publishes what is already saved in the database, not the live contents of the editor, so that one's fine.

PR #155, opened 2020-04-15
Fix performance regression

jessouyangbraven:
**Blame**: https://github.com/bebraven/platform/commit/b93fb9a8b36d190b257318daa8176996d7f0880a#diff-92838aff2423c239e8050b1d702f1a39R771

I wrote that line to naively fix the `Save` issue, but didn't test with the massive module data. Now I know `editor.getData()` is an expensive call. 

**Clean up**: 

Assuming this fixes the issue and doesn't regress anything, I'd like to try and clean this up a bit, starting with the `document.forms[0].submit()`. I tried this: https://pastebin.com/JTefjiqC but got a 404 error I couldn't figure out.

**Test**:
- Same as [#152](https://github.com/bebraven/platform/pull/152), with some additions
- Always use this input to test: https://stagingplatform.bebraven.org/course_contents/1/edit
- Check that editing in "Design" and clicking "Save" works
- Test that using `?html=true` in the URL works

**Details**:

Admittedly, the data flow here is really hard to follow. Where the content data lives depends on the whether we're in the Design or Code tab now. 

- If we're in Design, `editor.getData()` keeps track of it. 
- If we're in Code, `this.state.editorData` has it. 
- When switching between tabs, we need to sync `this.state.editorData` and re-render. 
- When we `Save`, we need to know which tab we're in so we read the correct data.

### DIFF
--- a/app/javascript/components/ContentEditor.js
+++ b/app/javascript/components/ContentEditor.js
@@ -66,6 +66,12 @@ import ContentPartPreview from './ContentPartPreview';
 // Other local imports.
 import { getNamedAncestor, getNamedChildOrSibling } from '../ckeditor/utils';
 
+// Tabs for different content views
+const TABS = {
+    DESIGN: 0,
+    CODE: 1,
+};
+
 // Plugins to include in the build.
 BalloonEditor.builtinPlugins = [
     Essentials,
@@ -207,6 +213,7 @@ class ContentEditor extends Component {
             modelPath: [],
             viewPath: [],
             selectedElement: undefined,
+            tabIndex: window.location.search.includes('html=true') ? TABS.CODE : TABS.DESIGN,
         };
 
         // The configuration of the <CKEditor> instance.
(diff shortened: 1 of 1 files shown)


app/javascript/components/ContentEditor.js
```diff
@@ -357,7 +359,28 @@ class ContentEditor extends Component {
             );
     }
 
+    handleSave(event) {
+        if (this.state.tabIndex == 0) { // We're on the "Design" tab and saving
```
sadleb, line 363: It may be more readable / maintainable to define constants for the Design vs Code tabIndex. E.g. 
```
const tabs { 
    DESIGN:  0,
    CODE: 1
}

if (this.state.tabIndex == tabs.DESIGN) {
...
```

jessouyangbraven: Updated! Still investigating the `fetch()` issue.

jessouyangbraven:
I'm really struggling with getting this to work without sneakily updating `secret-html`'s `value` (e.g, https://pastebin.com/JTefjiqC). 

If it's okay, I'd like to review and land this fix, verify it in platformstaging/platform so the editor is usable and then get some help to get the above working and clean up the code.

jessouyangbraven:
Thanks @rshipp! I'll merge and check it in `platformstaging`.

PR #161, opened 2020-04-20
Copy-paste for checklist question

jessouyangbraven:
**Task**: https://app.asana.com/0/1168520191527013/1170510474174757

**Test**
- Create two empty sections
- Add a question to one section, fill in some fields
- Select (blue perimeter), Ctrl+C to copy
- Click into other section, Ctrl+P to paste
- Also used `Module 1` to test copy-paste on large data sets, the performance and behavior are fine. 

**Details**

The key things to note in our implementation:
- All of the data we need is in `data.getData('text/html')`. The default behavior we get from CKEditor is pasting in what is returned by `data.getData('text/plain')`. 
- We need to call `evt.stop()` to prevent other input transformations from running, or else we end up pasting in the same old, messed up plain text.
- We have to let model conversion know we're in the context of `section`, so the right converters run and we get the right model fragment to insert. 

References:
- This is the most relevant code snippet from CKEditor:  https://github.com/ckeditor/ckeditor5-clipboard/blob/4141ef641dd96defbce5c542401a16c2bebecd8f/src/clipboard.js#L88-L97
- This is the CKEditor `Clipboard` documentation, which is useful for an overview of how events are fired, but less so for code: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/clipboard.html

### DIFF
--- a/app/javascript/ckeditor/moduleblockediting.js
+++ b/app/javascript/ckeditor/moduleblockediting.js
@@ -1,15 +1,55 @@
+import Clipboard from '@ckeditor/ckeditor5-clipboard/src/clipboard';
+
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
-import { toWidget } from '@ckeditor/ckeditor5-widget/src/utils';
+
 import Widget from '@ckeditor/ckeditor5-widget/src/widget';
+import { toWidget } from '@ckeditor/ckeditor5-widget/src/utils';
+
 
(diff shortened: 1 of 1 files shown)


rshipp:
this seems relevant https://github.com/ckeditor/ckeditor5-clipboard/blob/4141ef641dd96defbce5c542401a16c2bebecd8f/src/clipboard.js#L88-L97

PR #162, opened 2020-04-21
Allow inserting options with the entire option selected, instead of just inside the label

rshipp:
Task: https://app.asana.com/0/1170776727341290/1170308508489046/f

This is another one where I don't fully understand why it works this way.

As a side-effect of this PR, the `insertCheckbox` / `insertRadio` commands are much more robust and less hardcoded-y.

To test: 

1. Insert a section
2. insert a checkbox (or radio) question
3. select the option, outside the checkbox label or inline feedback, so the blue outline shows up.
4. hit enter; note that a new option is inserted after the current option.

### DIFF
--- a/app/javascript/ckeditor/checklistquestionediting.js
+++ b/app/javascript/ckeditor/checklistquestionediting.js
@@ -30,8 +30,10 @@ export default class ChecklistQuestionEditing extends Plugin {
             evt.stop();
         } );
 
-        // For some reason 'enter' events don't fire when the current selection is a checkboxDiv,
-        // so fix that explicitly.
+        // Because 'enter' events are consumed by Widget._onKeydown when the current selection is a non-inline
+        // block widget, we have to re-fire them explicitly for checkboxDivs.
+        // https://github.com/ckeditor/ckeditor5-widget/blob/bdeec63534d11a4fa682bb34990c698435bc13e3/src/widget.js#L174
+        // https://github.com/ckeditor/ckeditor5-widget/blob/bdeec63534d11a4fa682bb34990c698435bc13e3/src/widget.js#L408
         this.listenTo( this.editor.editing.view.document, 'keydown', ( evt, data ) => {
             const selection = this.editor.model.document.selection;
             const selectedElement = selection.getSelectedElement();
@@ -44,6 +46,8 @@ export default class ChecklistQuestionEditing extends Plugin {
                     evt.stop();
                 }
             }
+        // Use 'highest' priority, because Widget._onKeydown listens at 'high'.
+        // https://github.com/ckeditor/ckeditor5-widget/blob/bdeec63534d11a4fa682bb34990c698435bc13e3/src/widget.js#L92
         }, { priority: 'highest' } );
 
         // Override the default 'enter' key behavior to allow inserting new checklist options.
--- a/app/javascript/ckeditor/checklistquestionediting.js
+++ b/app/javascript/ckeditor/checklistquestionediting.js
@@ -30,19 +30,34 @@ export default class ChecklistQuestionEditing extends Plugin {
             evt.stop();
         } );
 
-        // Override the default 'enter' key behavior for checkbox labels.
-        this.listenTo( this.editor.editing.view.document, 'enter', ( evt, data ) => {
-            const positionParent = this.editor.model.document.selection.getLastPosition().parent;
-            if ( positionParent.name == 'checkboxLabel' ) {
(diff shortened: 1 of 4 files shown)


app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -30,19 +30,34 @@ export default class ChecklistQuestionEditing extends Plugin {
...
+
+            if ( selectedElement && selectedElement.name == 'checkboxDiv' ) {
+                if ( data.domEvent.key === 'Enter' ) {
+                    // This will end up calling our enter listener below.
+                    this.editor.editing.view.document.fire( 'enter', { evt, data } );
+                    data.preventDefault();
+                    evt.stop();
                 }
+            }
+        }, { priority: 'highest' } );
```
rshipp, line 47: why does this have to be 'highest' priority?

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -30,19 +30,34 @@ export default class ChecklistQuestionEditing extends Plugin {
...
 
-        // Override the default 'enter' key behavior for checkbox labels.
-        this.listenTo( this.editor.editing.view.document, 'enter', ( evt, data ) => {
-            const positionParent = this.editor.model.document.selection.getLastPosition().parent;
-            if ( positionParent.name == 'checkboxLabel' ) {
-                // Only insert a new checkbox if the current label is empty, but stop the event from
-                // propogating regardless.
-                if (!positionParent.isEmpty) {
-                    this.editor.execute( 'insertCheckbox' )
+        // For some reason 'enter' events don't fire when the current selection is a checkboxDiv,
```
rshipp, line 33: why do we need this at all? it seems like something inside cke is listening for enter events with an element selected, and stopping them (after inserting a paragraph - could this be in the paragraph plugin?)

jessouyangbraven: Maybe this in the `Widget` plugin? https://github.com/ckeditor/ckeditor5-widget/blob/master/src/widget.js#L174

rshipp: yep that's it exactly 🤦‍♀️thank you!! i'll add a code comment about this before merging

jessouyangbraven: Yeah, I'm pretty sure it's the `Widget` plugin. 

For some reason, CKEditor wants widgets's `enter` events to be converted into new paragraph inserts for particular types of widgets: https://github.com/ckeditor/ckeditor5-widget/blob/master/src/widget.js#L408

To test, I (intentionally mis-)configured `radioDiv` to `isInline: true`, deleted this handler, and saw that you get the `enter` event in the original handler.

Logically, `radioDiv` and `checkBoxDiv` aren't `inline`, so I think your fix is the best way to go about this.

app/javascript/ckeditor/checklistquestionediting.js
jessouyangbraven: Probably this in `Widget`: https://github.com/ckeditor/ckeditor5-widget/blob/master/src/widget.js#L92

rshipp: 🙃

rshipp:
> How do we decide which of our plugins are widget/inline/object/limit/etc.

in the beginning - my best guess, which was not much better than random. now, I have a little better understanding, so I pay more attention, but its still just a guess. as you might tell from this PR, there are a lot of behaviors for isLimit/isObject that aren't clearly documented.

> and do we get reasonable CKEditor behavior configuring them as such?

frequently, if they're defined correctly, yes. more recently I've started having to override their behavior (eg `findAllowedParentIgnoreLimit`) bc they (imo) overload each property too much. the more complicated we make our components, the less the CKE builtin properties make sense.

> I ask because I could totally see hacking this by adding isInline: true in the radio/checkBoxDivs, but there are probably unintended Widget behaviors from that approach.

we could probably use isInline + CSS widths to get the desired behavior, but it seems easy to break. i'm not sure what else might be tied to "inline" behavior either.

> Can we de-dupe some of the shared code between the radio and checkbox questions? I almost always test with checkbox because it's first in the list and I can totally see shooting myself in the foot here.

ideally, yes. i've almost done it in the past, but I thought it'd be more effort to figure out a clean way to do it because of the large checkboxInput/radioInput converter differences.

I think the cke-way to do this (see their todolist plugin for a perfect example) would be to have a single base plugin with attributes like `optionType = checkbox` or `optionType = radio`, and base different behavior off that.

PR #163, opened 2020-04-21
Don't delete options by typing

jessouyangbraven:
**Task**: https://app.asana.com/0/1170776727341290/1171026326095793

**Test**: 
- Add section, add {radio, checklist}question
- Click on option, verify Enter adds a new one, Backspace deletes, any other keydown does nothing

This is really straightforward to do on top of Ryan's `editor-insertcheckbox` changes.

### DIFF
--- a/app/javascript/ckeditor/checklistquestionediting.js
+++ b/app/javascript/ckeditor/checklistquestionediting.js
@@ -37,12 +37,14 @@ export default class ChecklistQuestionEditing extends Plugin {
             const selectedElement = selection.getSelectedElement();
 
             if ( selectedElement && selectedElement.name == 'checkboxDiv' ) {
+                if (data.domEvent.key === 'Backspace') {
+                    return;
+                }
                 if ( data.domEvent.key === 'Enter' ) {
                     // This will end up calling our enter listener below.
                     this.editor.editing.view.document.fire( 'enter', { evt, data } );
-                    data.preventDefault();
-                    evt.stop();
                 }
+                evt.stop();
             }
         }, { priority: 'highest' } );
 
--- a/app/javascript/ckeditor/moduleblockediting.js
+++ b/app/javascript/ckeditor/moduleblockediting.js
@@ -1,15 +1,55 @@
+import Clipboard from '@ckeditor/ckeditor5-clipboard/src/clipboard';
+
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
-import { toWidget } from '@ckeditor/ckeditor5-widget/src/utils';
+
 import Widget from '@ckeditor/ckeditor5-widget/src/widget';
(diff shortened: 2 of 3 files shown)


app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -37,12 +37,14 @@ export default class ChecklistQuestionEditing extends Plugin {
...
+                if (data.domEvent.key === 'Backspace') {
+                    return;
+                }
                 if ( data.domEvent.key === 'Enter' ) {
                     // This will end up calling our enter listener below.
                     this.editor.editing.view.document.fire( 'enter', { evt, data } );
-                    data.preventDefault();
-                    evt.stop();
                 }
+                evt.stop();
```
rshipp, line 47: this also breaks ctrl-z when a checkboxDiv is selected. (e.g. select one, hit enter to create a new one, ctrl-z to remove it, notice it fails, select something not a checkboxDiv, ctrl-z, notice it works)

app/javascript/ckeditor/checklistquestionediting.js
```diff
@@ -37,12 +37,14 @@ export default class ChecklistQuestionEditing extends Plugin {
             const selectedElement = selection.getSelectedElement();
 
             if ( selectedElement && selectedElement.name == 'checkboxDiv' ) {
+                if (data.domEvent.key === 'Backspace') {
```
rshipp, line 40: i can't test this on Mac bc the "Delete" button is actually "Backspace", but will this work with regular Delete too?

rshipp: for some reason this is also not stopping ctrl-c/ctrl-x/ctrl-p, which I would expect it to. is that part of preventDefault?

jessouyangbraven:
I was naive about this. What I want to do is intercept printable characters and not have them do anything with the `checkboxDiv` is selected so it doesn't get deleted, while the rest of the keystroke events (like delete, enter, etc.) do what they've always done.

Poking around, there's no reliable way to test `printable character`: https://stackoverflow.com/questions/12467240/determine-if-javascript-e-keycode-is-a-printable-non-control-character/12467610 and most of the suggestions use deprecated `KeyboardEvent` features: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent

On the CKEditor side, there's a `Typing` plugin, a Delete observer, etc., but none of them seem to capture what we want here. 

I think this might not be worth it given how fragile the code will be so long as we provide reliable "Undo" behavior for the designers. I'll ask in retro.

rshipp:
For posterity - I implemented a naive version of this for sections [here](https://github.com/bebraven/platform/blob/7f6840c1b277f79ed8553d47bc75e95d31c772bd/app/javascript/ckeditor/sectionediting.js#L33-L38).

PR #164, opened 2020-04-22
Prototype of user story to allow users to create an account and login.

sadleb:
# User story:
1. Exposes URLs like: http://platformweb/users/sign_up?u=<salesforce_id>
2. Presents a signup form to enter a password.
3. Emails a confirmation link to the email address in Salesforce for that Salesforce Contact.
4. After confirmation, redirects to sign in through SSO.
5. Redirects you to Portal homepage if you are provisioned in the Portal (and are a non-admin user, aka an @bebraven.org account)

# Context
The main thing this prototype is aimed at is getting Devise
cas_authenticable and database_authenticable to play nice.
They are only meant to be one or the other. This sort of hacks
around that by keeping the User model cas_authenticable, aka
you login through CAS SSO, but it adds the database_authenticable
module in and creates some controllers, routes and views to allow registration
and confirmation to happen through the database_authenticable
framework.

# Test Plan:
- Set the APPLICATION_HOST env var. Then run through the above steps,
  grabbing the confirmation link out of the server logs since the mailer isn't actually setup.

# TODOs
- automatically log them in when they click the confirmation link
- redirect somewhere, like a page telling them who to contact for help, if they are not provisioned in teh portal
- make sure non-admin users can't hit the content editor and other platform pages (any non @bebraven.org email is non-admin by default)
- write moar specs
- grep for TODO in this commit and do those
- registration should provision them in the Portal if their Salesforce Participant Record is Enrolled (means Sync To LMS hasn't run yet, but they should be there)
- implement forgot password and trackable


app/controllers/users/confirmations_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  # def create
+  #   super
+  # end
+
+  # GET /resource/confirmation?confirmation_token=abcdef
+  # def show
+  #   super
+  # end
+
+  # protected
```
rshipp, line 19: should this `protected` be commented?

app/controllers/users/registrations_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  # GET /resource/cancel
+  # Forces the session data which is usually expired after sign
+  # in to be expired now. This is useful if the user wants to
+  # cancel oauth signing in/up in the middle of the process,
+  # removing all OAuth session data.
+  # def cancel
+  #   super
+  # end
+
+  # protected
```
rshipp, line 46: same q here

config/environments/test.rb
```diff
@@ -43,6 +43,7 @@
   end
 
   config.action_mailer.perform_caching = false
+  config.action_mailer.default_url_options= {host: Rails.application.secrets.application_host, port: '3020'}
```
rshipp, line 46: should this port be hardcoded? iirc tests don't run through the default port, they use some custom rspec mock object.

config/initializers/devise.rb
```diff
@@ -170,7 +177,7 @@
 
   # ==> Configuration for :validatable
   # Range for password length.
-  config.password_length = 4..128
+  config.password_length = 8..128
```
rshipp, line 189: 😅

lib/cas_authenticator.rb
```diff
@@ -0,0 +1,28 @@
...
+      # TODO: check whether they are confirmed as well and clean this up.
+      #valid_password?(credentials[:password]) && active_for_authentication?
+      valid_password?(credentials[:password])
+    rescue ActiveRecord::RecordNotFound
+      false
+    end
+
+    def valid_password?(password)
+      return false if @user.encrypted_password.blank?
+      BCrypt::Password.new(@user.encrypted_password) == password
```
rshipp, line 16: what's this do? I'm reading it as "return true if `hash(hash(password))` is equal to `password`", which doesn't seem right.

config/environments/test.rb
sadleb: Good call. Fixed. This only impacts links in outgoing mail which we currently have no tests for, but putting the port in there serves no purpose.

app/controllers/users/registrations_controller.rb
sadleb: Good catch.

app/controllers/users/confirmations_controller.rb
sadleb: Good catch.

lib/cas_authenticator.rb
sadleb: So `password` is the plain-text password they just posted. `@user.encrypted_password` is the hashed password (using `BCrypt::Password.create`). `BCrypt::Password.new(@user.encrypted_password)` reads in the hashed password into an object that implements `==` to do comparisons.

https://www.rubydoc.info/github/codahale/bcrypt-ruby/BCrypt/Password

rshipp: interesting, thanks for the link. weird design decision on their part.

PR #170, opened 2020-04-23
Render HTML on the show page

rshipp:
Task: none

Summary: Renders HTML in the `show` page, to make it easier for me to download the full HTML without having to unescape a bunch of characters or something. I did this now so I can get the HTML from the modules on prod and clean up the duplicate-tooltip-span mess.

Screenshot: 
Screen Shot 2020-04-23 at 10 16 14 AM

### DIFF
--- a/app/assets/stylesheets/course_contents.scss
+++ b/app/assets/stylesheets/course_contents.scss
@@ -3,6 +3,16 @@
 // You can use Sass (SCSS) here: http://sass-lang.com/
 /* assets/styles.css */
 
+/* --- Show module --- */
+.show-module-notice {
+        margin: 2em;
+	font-size: 20px;
+
+	strong {
+                color: red;
+	}
+}
+
 /* --- Editor / Content styles --- */
 .bz-module,
 .bz-view-box {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/course_contents.scss
```diff
@@ -3,6 +3,16 @@
...
 /* assets/styles.css */
 
+/* --- Show module --- */
+.show-module-notice {
+        margin: 2em;
+	font-size: 20px;
+
+	strong {
+                color: red;
+	}
```
sadleb, line 13: Weird whitespace.

spec/views/course_contents/show.html.erb_spec.rb
```diff
@@ -2,17 +2,30 @@
...
-  it "renders attributes in <p>" do
+  it "renders HTML" do
     render
-    expect(rendered).to match(/Title/)
-    expect(rendered).to match(/MyText/)
-    expect(rendered).to match(/MyText/)
+    expect(rendered).to match(/<p>MyText<\/p>/)
+  end
+
+  it "adds appropriate div class for content_type" do
```
sadleb, line 17: It feels like this should be in two separate examples.

PR #171, opened 2020-04-23
Add ability to adjust log level using LOG_LEVEL env var.

sadleb:
Note that this impacts specs too. In your local dev env,
usually you would just leave LOG_LEVEL unset so that specs
only log errors and the server logs at the debug level.


sadleb:
> looks fine to me. did you need logs for something?

I was trying to get some specs I was writing to pass and there were already logs in the areas of the code I needed visibility into so I figured I'd just make it easier to change the log levels across the board.

PR #172, opened 2020-04-23
Allow changing module-block "icon"

rshipp:
Task: https://app.asana.com/0/1170776727341290/1172273996801759/f

This one is bad. I spent way too long trying and failing to get `attributeToAttribute` to work on classes. Finally gave up and tried the low-level event `dispatcher` we use for `customelementattributepreservation`, and it seems to work.

I would recommend ignoring commit d38e0de and looking at the first and third commits instead.

Side note, radio questions might be way more broken than I realized. Opened https://app.asana.com/0/1170776727341290/1172643959550099/f for that.

Screenshot:
Screen Shot 2020-04-23 at 5 07 25 PM

This (or some variation) is what I'd rather use in place of the dispatcher, but it doesn't work. It just keeps adding things on to the `class` attribute in the editingView instead of replacing them.

```js
        } ).attributeToAttribute( {
            model: 'blockClasses',
            view: modelAttributeValue => {
                return {
                    key: 'class',
                    value: modelAttributeValue,
                };
            },
        } ).
```

Resources:

* https://github.com/ckeditor/ckeditor5-engine/pull/1804/files
* https://github.com/ckeditor/ckeditor5/issues/4517
* https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_conversion_conversion-Conversion.html#function-attributeToAttribute


rshipp:
Hmm. I think the part that bothers me is that I can't tell why `attributeToAttribute` doesn't work, our usecase matches their docs exactly. It's possible this is another upstream bug though :/

PR #175, opened 2020-04-28
Fix tooltip converter

rshipp:
This addresses both the double-tooltip upcast issue, and the tooltip-within-list issue.

Task: https://app.asana.com/0/1170776727341290/1161453930156673/f
Task: https://app.asana.com/0/1142638035116665/1169229931357117/f

Test with:

```html
        <ol>
          <li><span class="has-tooltip" title="test-title">Text</span></li>
        </ol>
```

Toggle back and forth between Design/Code a few times, and note it doesn't create duplicate spans, nor does it crash the editor.

Reference: https://github.com/ckeditor/ckeditor5/issues/4529

### DIFF
--- a/app/javascript/ckeditor/tooltipediting.js
+++ b/app/javascript/ckeditor/tooltipediting.js
@@ -35,9 +35,12 @@ export default class TooltipEditing extends Plugin {
 
         // We have to use a lower-level dispatcher because otherwise this conflicts with the span upcast in
         // customelementattributepreservation.js and we end up with double spans.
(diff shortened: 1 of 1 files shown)


rshipp:
Just for posterity - the reason this fixed the issue (and also caused the can't-edit-tooltips issue) is NOT because `lowest` ran after the parent conversion as mentioned in the linked CKE issue, but because `lowest` ran after `customelementattributepreservation.js` span converters... which is not helpful.

PR #178, opened 2020-04-30
Implement image upload

rshipp:
Task: https://app.asana.com/0/1170776727341290/1172626862010553/f

I'm not sure how to rspec the controller for this, @sadleb any suggestions?

### DIFF
--- a/app/controllers/file_upload_controller.rb
+++ b/app/controllers/file_upload_controller.rb
@@ -1,0 +1,25 @@
+class FileUploadController < ApplicationController
+  # POST /course_contents.json
+  def create
+    respond_to do |format|
+      upload = CanvasAPI.client.upload_file_to_course(
+          file_upload_params.tempfile,
+          file_upload_params.original_filename,
+          file_upload_params.content_type
+      )
+
+      if upload[:url]
+        format.json { render json: upload, status: :ok }
+      else
+        info = JSON.parse(upload.body)
+        format.json { render json: info, status: 500 }
+      end
+    end
+  end
+
+  private
+    # Never trust parameters from the scary internet, only allow the white list through.
+    def file_upload_params
+      params.require(:upload)
+    end
+end
--- a/app/controllers/file_upload_controller.rb
+++ b/app/controllers/file_upload_controller.rb
@@ -8,7 +8,7 @@ class FileUploadController < ApplicationController
           file_upload_params.content_type
       )
 
-      if upload[:url]
+      if upload.is_a? Hash and upload[:url]
         format.json { render json: upload, status: :ok }
       else
         info = JSON.parse(upload.body)
--- a/app/controllers/file_upload_controller.rb
+++ b/app/controllers/file_upload_controller.rb
@@ -1,5 +1,5 @@
 class FileUploadController < ApplicationController
-  # POST /course_contents.json
+  # POST /file_upload.json
   def create
     respond_to do |format|
       upload = CanvasAPI.client.upload_file_to_course(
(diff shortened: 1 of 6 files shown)


sadleb:
> I’ve been punting on controller specs too. I need to spend more time researching. So, no ideas, but I would queue up a task to write a controller spec for this. That’s what I’ve been doing.
> 
> That said, this is way cleaner than I expected!

Oh wait, I just noticed the hard coded course id = 1. Let’s introduce a constant for content library.

sadleb:
> > I’ve been punting on controller specs too. I need to spend more time researching. So, no ideas, but I would queue up a task to write a controller spec for this. That’s what I’ve been doing.
> > That said, this is way cleaner than I expected!
> 
> Oh wait, I just noticed the hard coded course id = 1. Let’s introduce a constant for content library.

Eg CONTENT_LIBRARY_COURSE_ID

rshipp:
@sadleb I used `ContentLibraryCourseID` because it matches the other constants in that file and I don't know ruby formatting rules. lmk if it should be all-caps and/or somewhere else.

sadleb:
> @sadleb I used `ContentLibraryCourseID` because it matches the other constants in that file and I don't know ruby formatting rules. lmk if it should be all-caps and/or somewhere else.

That works for me. In ruby, anything that starts with a capital letter is a constant. E.g. class names are actually constants. I'm in the habit of using all caps just b/c it's been consistent across the programming languages i've used, but i don't really care at the moment. At some point, we should have a discussion if we want to enforce a consistent constant naming format.

sadleb:
> > @sadleb I used `ContentLibraryCourseID` because it matches the other constants in that file and I don't know ruby formatting rules. lmk if it should be all-caps and/or somewhere else.
> 
> That works for me. In ruby, anything that starts with a capital letter is a constant. E.g. class names are actually constants. I'm in the habit of using all caps just b/c it's been consistent across the programming languages i've used, but i don't really care at the moment. At some point, we should have a discussion if we want to enforce a consistent constant naming format.

And as for location, there makes sense for now. We can refactor to a more appropriate place if we start needing to use that constant elsewhere in ruby code.

lib/canvas_api.rb
```diff
@@ -157,4 +160,28 @@ def get_all_from_pagination(response)
     info
   end
 
+  # See: https://canvascoach.instructure.com/doc/api/file.file_uploads.html#method.file_uploads.post
+  # If successful, URL will be available as the hash value of the 'url' key.
+  # If unsuccessful, returns a Response object.
```
rshipp, line 165: I'm not happy with this interface, but I'm not sure what the "correct" way to do it is. seems like it should return the same kind of object whether or not it fails.

app/controllers/file_upload_controller.rb
```diff
@@ -0,0 +1,25 @@
+class FileUploadController < ApplicationController
+  # POST /course_contents.json
```
jessouyangbraven, line 2: Is this supposed to be `#POST /file_upload.json`?

lib/canvas_api.rb
```diff
@@ -18,6 +18,9 @@ class CanvasAPI
     </div>
   )
 
+  # Canvas course ID for the Braven Content Library.
+  ContentLibraryCourseID = 1
```
jessouyangbraven, line 22: Will we need different IDs for Booster? Should this eventually be something we configure in `.env`?

rshipp: afaik contentlibrary (which is part of Braven's custom modifications to our Canvas fork) is always course 1, and the editor will only ever need to publish to contentlibrary. @sadleb correct me if I'm wrong?

app/controllers/file_upload_controller.rb
rshipp: yep, thanks for the catch

PR #183, opened 2020-05-01
Add textInput button, fix textArea/Input placeholders

rshipp:
Two changes:

1. Add a textInput button (the command already existed).
2. Set the default placeholder for both textInput/textArea to the empty string `''`, to fix the following 2 issues:
  a. When inserting a textarea/input, the default placeholder shows up as the literal string `undefined`.
  b. When attempting to edit the default placeholder, the editor crashes, and undo breaks.

Button Task: https://app.asana.com/0/1170776727341290/1173719663284395/f
Placeholder Task: https://app.asana.com/0/1170776727341290/1173719663284397/f

### DIFF
--- a/app/javascript/components/ContentEditor.js
+++ b/app/javascript/components/ContentEditor.js
@@ -771,11 +771,21 @@ class ContentEditor extends Component {
                                     key="insertTextArea"
                                     enabled={this.state.enabledCommands.includes('insertTextArea')}
                                     onClick={( id ) => {
-                                        this.editor.execute( 'insertTextArea' );
+                                        this.editor.execute( 'insertTextArea', '' );
                                         this.editor.editing.view.focus();
                                     }}
                                     onClickDisabled={() => this.editor.editing.view.focus()}
                                     {...{name: 'Text Area'}}
+                                />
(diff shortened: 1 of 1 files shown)


jessouyangbraven:
I might be missing something, but it looks like `InsertTextInputCommand` is never enabled, e.g., I never see it in the right column as an option. How do I get the option to show up? Do I need to be inside another component?

rshipp:
Weird, it should be allowed just about anywhere - outside/inside sections, inside questions, tables, bulleted lists... 

Screen Shot 2020-05-04 at 3 41 16 PM

Sometimes things mess up if the container is in a weird state, if you bring it down and back up does it fix anything?

jessouyangbraven:
I think it's because the base branch changed to staging post-full-upgrade. Rebuilding now to try again.

jessouyangbraven:
Thanks for the screenshot! I know it might seem dumb, but it is really helpful for me to know what the code's final result should be. I just had to pull and re-build everything and now it works.

rshipp:
Good note, I'll up my screenshotting. Looking back, the details for this one were especially vague.

PR #185, opened 2020-05-04
Highlight invalid things inside sections, and disallow adding all but a single module-block to sections.

rshipp:
Problem: The module JS expects sections to only have one module-block and nothing else. The booster content has many things inside sections. This causes the done-button JS to break and show everything inside the section.

Proposal: Highlight things that are invalid inside a section, to encourage the designer to fix them. I didn't enforce this from CKE because it doesn't seem possible from [addChildCheck](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_schema-Schema.html#function-addChildCheck), and I don't know another good place to put it.

Proposal part 2 (2020-05-07): Fix remaining pain points, without invalidating/mangling existing incorrect content. Additions as follows - 

* Disallow typing within sections. (Still allows ctrl-z/{ctrl-shift-z,ctrl-y}, backspace/delete, clipboard actions, etc.)
* Hitting enter within a section now creates a new section, instead of inserting a paragraph inside the section.
* Hitting enter inside a section with a module-block selected now creates a new section, instead of inserting a paragraph inside the section.
* The above makes it impossible for a designer to add multiple module-blocks inside a section, or to add paragraphs/headings/etc inside the section.
* If there is existing incorrect content, it will be highlighted by the CSS change, but the designer should never be able to create new incorrect content.

Pros:

* This solution takes minimal effort on our part; no digging around in CKE to figure out how to enforce this.
* Designers can easily tell what is invalid.

Cons:

* ~~The only way to add a new section while you're in a section that already has a module-block is to hit enter (which will show an alert for invalid `p` inside section), then click Section (which will show another alert as there are now two `p`s inside section), then click a Question/Content button (which will still show an alert above the newly inserted content, because of the remaining `p`). This may be confusing.~~ I fixed this.

Screenshot:

Screen Shot 2020-05-04 at 2 12 26 PM

A better way to do this would be to get rid of sections entirely, but that would take more work and testing.

Sections task: https://app.asana.com/0/1170776727341290/1172766526299174/f
Done button task: https://app.asana.com/0/1170776727341290/1173524622792753/f

### DIFF
--- a/app/assets/stylesheets/course_contents.scss
+++ b/app/assets/stylesheets/course_contents.scss
@@ -42,6 +42,25 @@
 	}
 }
 
+/* --- Warnings for invalid content --- */
+.content-section > :not(.module-block):not([data-placeholder="[Empty section]"]):before,
+.content-section > .module-block:nth-of-type(2):before {
+        content: "INVALID CONTENT IN SECTION \A";
+        white-space: pre;
+        color: black;
+        background-color: red;
+        border: 1px black solid;
+        padding: 0.2em;
+        font-weight: bold;
+        font-size: 12pt;
+        font-family: sans-serif;
+}
+
+.content-section > :not(.module-block):not([data-placeholder="[Empty section]"]),
+.content-section > .module-block:nth-of-type(2) {
+        background-color: khaki;
+}
+
 /* --- Editor / Content styles --- */
 .bz-module,
 .bz-view-box {
--- a/app/assets/stylesheets/course_contents.scss
+++ b/app/assets/stylesheets/course_contents.scss
@@ -45,7 +45,7 @@
 /* --- Warnings for invalid content --- */
 .content-section > :not(.module-block):not([data-placeholder="[Empty section]"]):before,
 .content-section > .module-block:nth-of-type(2):before {
-        content: "INVALID CONTENT IN SECTION \A";
+        content: "INVALID CONTENT IN SECTION - Please remove or ask tech team for help\A";
         white-space: pre;
         color: black;
         background-color: red;
--- a/app/javascript/ckeditor/insertsectioncommand.js
+++ b/app/javascript/ckeditor/insertsectioncommand.js
@@ -1,9 +1,22 @@
 import Command from '@ckeditor/ckeditor5-core/src/command';
+import { getNamedAncestor } from './utils';
 
 export default class InsertSectionCommand extends Command {
     execute() {
         this.editor.model.change( writer => {
-            this.editor.model.insertContent( createSection( writer ) );
+            const selection = this.editor.model.document.selection;
+            const position = selection.getFirstPosition();
+
+            const section = getNamedAncestor( 'section', position );
+            if ( section ) {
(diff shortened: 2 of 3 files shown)


rshipp:
the rules are: 

1. one and only one question/content block in each section
2. only question/content blocks inside sections, nothing else

are those clear enough? 

I can add a link to a google doc or something, but it wouldn't be clickable, since you can't have HTML elements inside CSS `content`.

jessouyangbraven:
>This solution takes minimal effort on our part; no digging around in CKE to figure out how to enforce this.

Since we're planning on getting rid of `<section>` anyway, I agree that this isn't worth doing in the CKE code.

> Designers can easily tell what is invalid.

Yeah, I think is a good thing, but it might be more frustrating in the short term to be able to see something's wrong and not be able to directly fix it. This seems most useful for us to go in and quickly be able to figure out what to fix in the HTML. In the ideal case, designers wouldn't see this ever. 

> The only way to add a new section while you're in a section that already has a module-block is to hit enter (which will show an alert for invalid p inside section), then click Section (which will show another alert as there are now two ps inside section), then click a Question/Content button (which will still show an alert above the newly inserted content, because of the remaining p). This may be confusing.

...Yes. That's really confusing.

I'm cool with landing this, since it'll help us fix the content quickly, but I don't think it's a great long-term solution without giving designers straightforward ways to fix. 

For now, can we fix Booster Module 1 & 2 manually and ask Tess to keep to one module-block per section when she's editing? She shouldn't see the error at all and if she does, it'd be rare, and then we can just fix them as one-offs.

rshipp:
> For now, can we fix Booster Module 1 & 2 manually and ask Tess to keep to one module-block per section when she's editing? She shouldn't see the error at all and if she does, it'd be rare, and then we can just fix them as one-offs.

Yeah, I'd definitely manually fix the modules first. Tess would see the alert every time she tried to insert a new section though - that may be more pain than it's worth. What I'm trying to avoid is us having to manually fix modules over and over, because it's easy to lose track and put invalid things in sections.

Maybe I'll try adding a little JS and see if I can get rid of the alert on every new section. That'd remove the worst of the designer pain here.

rshipp:
before i forget again, I had an idea about changing the insert-section behavior away from cke's autoparagraphing, and towards out custom insert-checkbox behavior. i think that'd solve this nicely, if it doesn't take too much to get working.

rshipp:
Alright, I think this is a more full solution (but should still remove it once we get rid of sections).

rshipp:
I just realized this is all wrong, and now you can't type outside of sections. 😞

rshipp:
OK, now that's fixed too. Hopefully that's the last issue...

rshipp:
Fun sidebar: https://github.com/ckeditor/ckeditor5/issues/407#issuecomment-602111695 this is exactly the same issue as https://github.com/bebraven/platform/pull/185#issuecomment-625528263.

rshipp:
Merging this, but will keep a close eye on it come Monday.

PR #186, opened 2020-05-05
Allow adding done buttons to content blocks

rshipp:
Task: https://app.asana.com/0/1170776727341290/1172766526299157/f

Summary: Allow designers to add "Done" buttons to content blocks, but nowhere else.

Testing procedure:

* Load empty editor, note that "Done Button" button is disabled
* Insert section, note that "Done Button" button is disabled
* Insert question (any type), click around inside editor, note that "Done Button" button is always disabled
* Delete question, so you have an empty section again
* Insert "Text" content block (not "Text Area" or "Text Area Question")
* Note that "Done Button" button is enabled.
* Click around inside the text content-block, note that "Done Button" button remains enabled.
* Click on module-block so the border turns blue. 
  * Click "Done Button" button
  * Verify button is inserted at bottom of block.
  * Verify selection is returned to the module-block, so border turns blue.
* Click inside Content Title.
  * Click "Done Button" button
  * Verify button is inserted at bottom of block.
  * Verify selection is returned to the Content Title.
* Click inside Content Body.
  * Click "Done Button" button
  * Verify button is inserted at bottom of block.
  * Verify selection is returned to the Content Body.
* You should now have 3 done buttons.
* Undo, note the done buttons are removed.

Screenshots:

Screen Shot 2020-05-05 at 1 30 13 PM
Screen Shot 2020-05-05 at 1 30 19 PM

### DIFF
--- a/app/javascript/ckeditor/contentcommonediting.js
+++ b/app/javascript/ckeditor/contentcommonediting.js
@@ -448,7 +448,7 @@ export default class ContentCommonEditing extends Plugin {
         conversion.for( 'dataDowncast' ).elementToElement( {
             model: 'doneButton',
             view: ( modelElement, viewWriter ) => {
-                return viewWriter.createContainerElement( 'input', {
+                return viewWriter.createEmptyElement( 'input', {
                     'type': 'button',
                     'value': 'Done',
                     'class': 'done-button',
@@ -460,7 +460,7 @@ export default class ContentCommonEditing extends Plugin {
         conversion.for( 'editingDowncast' ).elementToElement( {
             model: 'doneButton',
             view: ( modelElement, viewWriter ) => {
-                return viewWriter.createContainerElement( 'input', {
+                return viewWriter.createEmptyElement( 'input', {
                     'type': 'button',
                     'value': 'Done',
                     'class': 'done-button',
--- a/app/javascript/ckeditor/insertdonebuttoncommand.js
+++ b/app/javascript/ckeditor/insertdonebuttoncommand.js
@@ -1,24 +1,59 @@
 import Command from '@ckeditor/ckeditor5-core/src/command';
+import { getNamedChildOrSibling, getNamedAncestor } from './utils';
 
 export default class InsertDoneButtonCommand extends Command {
     execute() {
         this.editor.model.change( writer => {
-            const doneButton = createDoneButton( writer );
-            this.editor.model.insertContent( doneButton );
-            writer.setSelection( doneButton, 'on' );
+            // Note: This command only works in content blocks, not question or answer blocks.
+            // The selection is expected to be somewhere in or on the module-block.
+            // Before inserting, we must modify the current selection to the end of the content block.
+            const selection = this.editor.model.document.selection;
+            const selectedElement = selection.getSelectedElement();
+            const position = selection.getFirstPosition();
+
+            const contentBlock = findContentBlock( selectedElement, position );
+            if ( !contentBlock ) {
+                return;
+            }
+
+            writer.setSelection( contentBlock, 'end' );
+            this.editor.model.insertContent( createDoneButton( writer ) );
+
+            // ♪ Put that thing back where it came from, or so help meeee ♫
+            if ( selectedElement ) {
+                writer.setSelection( selectedElement, 'on' );
+            } else {
+                writer.setSelection( position );
+            }
         } );
     }
 
     refresh() {
         const model = this.editor.model;
         const selection = model.document.selection;
-        const allowedIn = model.schema.findAllowedParent( selection.getFirstPosition(), 'doneButton' );
+        const selectedElement = selection.getSelectedElement();
+        const position = selection.getFirstPosition();
+        const allowedIn = findContentBlock( selectedElement, position );
 
-        this.isEnabled = allowedIn !== null;
+        this.isEnabled = allowedIn !== undefined;
     }
 }
 
 function createDoneButton( writer ) {
     const doneButton = writer.createElement( 'doneButton' );
     return doneButton;
+}
+
+function findContentBlock( selectedElement, position ) {
+    // Find the content block.
+    let contentBlock;
+    if ( selectedElement && 'moduleBlock' === selectedElement.name ) {
+        // The current selection is a moduleBlock; check for a content block inside it.
+        contentBlock = getNamedChildOrSibling( 'content', selectedElement );
+
+    } else if ( contentBlock = getNamedAncestor( 'content', position ) ) {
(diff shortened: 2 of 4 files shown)


app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -448,7 +448,7 @@ export default class ContentCommonEditing extends Plugin {
         conversion.for( 'dataDowncast' ).elementToElement( {
             model: 'doneButton',
             view: ( modelElement, viewWriter ) => {
-                return viewWriter.createContainerElement( 'input', {
+                return viewWriter.createEmptyElement( 'input', {
```
rshipp, line 451: noticed this while I was messing with done buttons - `<input>`s should *always* be EmptyElements.

app/javascript/ckeditor/insertdonebuttoncommand.js
```diff
@@ -1,24 +1,59 @@
...
     refresh() {
         const model = this.editor.model;
         const selection = model.document.selection;
-        const allowedIn = model.schema.findAllowedParent( selection.getFirstPosition(), 'doneButton' );
+        const selectedElement = selection.getSelectedElement();
+        const position = selection.getFirstPosition();
+        const allowedIn = findContentBlock( selectedElement, position );
 
-        this.isEnabled = allowedIn !== null;
+        this.isEnabled = allowedIn !== undefined;
```
rshipp, line 38: weird change, but `findContentBlock` never returns null, only undefined. it'd be more work to change that.

app/javascript/ckeditor/insertdonebuttoncommand.js
```diff
@@ -1,24 +1,59 @@
...
 }
+
+function findContentBlock( selectedElement, position ) {
+    // Find the content block.
+    let contentBlock;
+    if ( selectedElement && 'moduleBlock' === selectedElement.name ) {
+        // The current selection is a moduleBlock; check for a content block inside it.
+        contentBlock = getNamedChildOrSibling( 'content', selectedElement );
+
+    } else if ( contentBlock = getNamedAncestor( 'content', position ) ) {
```
rshipp, line 54: note this works for contentBody, contentTitle, videos/video transcripts, and everything else that might possibly be inside a content block. the empty if body is a little weird, but i thought it looked nicer than calling getNamedAncestor twice.

app/javascript/ckeditor/insertdonebuttoncommand.js
```diff
@@ -1,24 +1,59 @@
 import Command from '@ckeditor/ckeditor5-core/src/command';
+import { getNamedChildOrSibling, getNamedAncestor } from './utils';
 
 export default class InsertDoneButtonCommand extends Command {
     execute() {
         this.editor.model.change( writer => {
-            const doneButton = createDoneButton( writer );
-            this.editor.model.insertContent( doneButton );
-            writer.setSelection( doneButton, 'on' );
+            // Note: This command only works in content blocks, not question or answer blocks.
```
jessouyangbraven, line 7: I think I get what's happening from the code, but I want to make sure I understand the difference between this Done button and the existing ones. 

The `InsertDoneButtonCommand` is only for the situation where the designer wants to put in a custom Done button. 

This command is separate from the Done buttons we currently have in the content editor, because those are part of question/answer blocks. Those buttons were inserted by their respective "Insert*QuestionCommand". 

And we're only allow `InsertDoneButtonCommand` on content blocks so we don't end up with malformed HTML with multiple Done buttons in, for example, a question block, which would be really weird for the Portal JS.

app/javascript/ckeditor/insertdonebuttoncommand.js
```diff
@@ -1,24 +1,59 @@
...
+
+            const contentBlock = findContentBlock( selectedElement, position );
+            if ( !contentBlock ) {
+                return;
+            }
+
+            writer.setSelection( contentBlock, 'end' );
+            this.editor.model.insertContent( createDoneButton( writer ) );
+
+            // ♪ Put that thing back where it came from, or so help meeee ♫
```
jessouyangbraven, line 22: Lol.

app/javascript/ckeditor/contentcommonediting.js
jessouyangbraven: ...wait, why? And was this causing other weirdness?

rshipp: `<input>` elements are defined as empty by the HTML spec - this is more apparent in HTML5, where closing tags for inputs are actually an error. (Not that browsers care about HTML specs...)

`createContainerElement` creates code like `<input></input>`, which, because CKE doesn't allow empty containers, gets changed in the editing view into `<input><br data-cke-filler="true"></input>`. As a side effect, it's possible to get the cursor "inside" the done button, leading to undefined behavior.

`createEmptyElement` creates code like `<input>`, so CKE doesn't try to shove fillers inside it. This also ensures our output HTML is valid HTML5.

app/javascript/ckeditor/insertdonebuttoncommand.js
rshipp: > The InsertDoneButtonCommand is only for the situation where the designer wants to put in a custom Done button.

Yes and, only inside content blocks, never question blocks.

> This command is separate from the Done buttons we currently have in the content editor, because those are part of question/answer blocks. Those buttons were inserted by their respective "Insert*QuestionCommand".

Yes. Maybe worth noting here that commands are only required if you need a designer to be able to modify the model. I.E. you don't need an `InsertTextAreaCommand` to add a textarea inside a question, and when you put `<textarea>` inside the Code tab and switch to Design, that doesn't go through the commands. If we ever allow inserting done buttons in other places, we should probably rename this command to make its purpose clearer.

> And we're only allow InsertDoneButtonCommand on content blocks so we don't end up with malformed HTML with multiple Done buttons in, for example, a question block, which would be really weird for the Portal JS.

We only allow it on content blocks because all the `Insert*QuestionCommand`s already have done buttons, while content block commands don't. This is an intentional design decision, but not enforced by anything in the code. 

The idea is that you will always need to click "Done" on a question (with rare exceptions), because it has some sort of input you want to save / show the answer for. For content-blocks, you don't need to save or show an answer. But if there are a bunch of content blocks in a row with no questions, the done-button JS will expand a bunch of the module at once, down to the next done button. If the designer wants to add an artificial pause partway through, they can add a done button to one of the content blocks, forcing the fellow to stop and interact with the page rather than just skipping over everything. 

Note that there's nothing stopping you from adding multiple done buttons to a content block. We could enforce that from the "Done Button" button, but it seems unnecessary. If you add too many, you can undo to remove them.

PR #187, opened 2020-05-05
Add checklist-other button

rshipp:
Task: https://app.asana.com/0/1170776727341290/1172646780976523/f

Summary: Add an `"Other" Checkbox with Text Area` button, to let designers add "Other:" checkboxes to checklist questions.

Testing procedure:

1. Note the "Other" button is disabled
1. Insert section, insert checklist question
2. Click on/in the first checklist option; note the "Other" button changes to enabled
3. Click the "Other" button; verify a new option with a textarea is added after the currently selected option.

Other notes:

* "Other" options have no inline-feedback, by design.
* Both the checkbox `<input>` and the `<textarea>` will have the `data-bz-optional-magic-field="true"` attribute. This is used by the module JS as part of autograding.
* The "Other" checkbox will have `data-correctness` automatically set to `maybe`. ("Other" options are never right or wrong.)
* The "Other" textarea should not have a placeholder set.

Screenshots

Screen Shot 2020-05-05 at 2 20 19 PM
Screen Shot 2020-05-05 at 2 20 36 PM

### DIFF
--- a/app/javascript/ckeditor/checklistquestionediting.js
+++ b/app/javascript/ckeditor/checklistquestionediting.js
@@ -5,6 +5,7 @@ import Widget from '@ckeditor/ckeditor5-widget/src/widget';
 import RetainedData from './retaineddata';
 import InsertChecklistQuestionCommand from './insertchecklistquestioncommand';
 import InsertCheckboxCommand from './insertcheckboxcommand';
+import InsertChecklistOtherCommand from './insertchecklistothercommand';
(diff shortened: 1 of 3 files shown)


rshipp:
> Just out of curiosity, are we going to need one of these for radio questions?

I haven't seen a radio question with an "Other" field in any of the modules, but I suppose it's possible? I'd lean towards not implementing it unless someone asks for it though.

PR #190, opened 2020-05-05
Make docker better, faster, stronger 

rshipp:
A few fixes to the dev env to make it faster / less crashy. Changes described in comments below.


sadleb:
Wow, these are changes I wouldn’t have expected to be the root of the problem.

Guardfile
```diff
@@ -99,11 +67,41 @@ guard :bundler do
   files.each { |file| watch(helper.real_path(file)) }
 end
 
-guard :yarn do
-  watch("package.json")
-end
+group :development do
+  guard :yarn do
+    watch("package.json")
+    watch("yarn.lock")
```
rshipp, line 73: watch yarn.lock too, just in case.

Guardfile
```diff
@@ -99,11 +67,41 @@ guard :bundler do
   files.each { |file| watch(helper.real_path(file)) }
 end
 
-guard :yarn do
-  watch("package.json")
-end
+group :development do
```
rshipp, line 70: moved everything inside a group and into an order that makes more sense. now it checks yarn *before* starting Rails, so _hopefully_ we get less Rails crashes with `yarn install --check-files` errors.

Guardfile
```diff
@@ -99,11 +67,41 @@ guard :bundler do
...
-  watch('db/seeds.rb')
+  ### Guard::Webpacker
+  #  available options:
+  #  - :bin (defaults to "webpack-dev-server") to run
+  #  - :watch (defaults to "default") can be an array
+  #  - :colors (defaults to 1)
+  #  - :progress
+  guard :webpacker do
+    watch('config/webpacker.yml')
+    watch(%r{^config/webpack/(.+)$})
```
rshipp, line 84: there was a typo in the second watch here, so it didn't actually reload on webpack config changes 🤦‍♀️

config/webpack/development.js
```diff
@@ -5,4 +5,5 @@ const webpackerReactconfigureHotModuleReplacement = require('webpacker-react/con
 
 const config = environment.toWebpackConfig()
 
-module.exports = webpackerReactconfigureHotModuleReplacement(config)
+module.exports = webpackerReactconfigureHotModuleReplacement(config);
+module.exports.cache = true;
```
rshipp, line 9: in addition to @sadleb's persistent caching volume from before, enable in-memory caching for livereloading JS.

docker-compose/scripts/docker_compose_run.sh
```diff
@@ -35,6 +35,16 @@ else
     cp_built_lock_file
 fi
 
+
+# Migrate the db, if needed.
+bundle exec rake db:migrate
```
rshipp, line 40: got rid of the migrate guard bc it wasn't very good, so let's migrate at startup instead. this is super fast if it doesn't need to migrate anything.

docker-compose/scripts/docker_compose_run.sh
```diff
@@ -35,6 +35,16 @@ else
...
+
+# Fix "inotify event queue has overflowed."
+# https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers
+sysctl fs.inotify.max_user_watches=524288
+sysctl fs.inotify.max_queued_events=524288
+sysctl fs.inotify.max_user_instances=524288
```
rshipp, line 46: this crashed constantly when you made a bunch of JS changes, but I never knew why until @sadleb's earlier change to have Rails log to stdout. now that the error isn't being swallowed, let's fix it.

sadleb: Whoa, interesting. So much trickiness with Listen. Remember the CPU running like a jet engine issue a few months back 🙃


~~Prob not worth doing since you already got it working this way, but I think ideally we would change Dockerfile to configure these in the image itself using something this:~~
```
RUN echo "fs.inotify.max_user_watches=524288 
fs.inotify.max_queued_events=524288
fs.inotify.max_user_instances=524288" >> /etc/sysctl.conf && \
sysctl -p
```
Welp, that doesn't work. I guess docker low-level kernel stuff is tricky. Lgtm!

rshipp: Yeah I tried this a bunch of different ways and this was the only one that worked 🙃. Evidently you can't set persistent flags unless they're "[namespaced](https://docs.docker.com/engine/reference/commandline/run/#configure-namespaced-kernel-parameters-sysctls-at-runtime)", which fs.inotify isn't.

PR #191, opened 2020-05-06
Refactor Sync To LMS and account creation to work with Salesforce Programs

sadleb:
https://app.asana.com/0/1170770467635599/1173648174910505

We have two separate servers running the Portal, one for normal Braven and one
for Braven Booster. There may be conflicts with the course_id. Also, for Booster
we're going to have a "Holding" Program and then split folks out into Program's
specific to the run of the program they're taking. Also, folks may sign-up last
minute and we need to provision their account in Canvas. Lastly, Booster folks will not
be mapped to a Cohort. They will have a scheduled day/time and we need to add them
into Canvas in corresponding sections.

This commit aims to solve all that.

In addition to making Sync To LMS run off the Salesforce Program, it also changes
the account creation / sign_up to create their Canvas account on the fly at the time they
register their account. This was last minute sign-ups will work without running the sync.

Test Plan:
- `bundle exec rspec spec`
- Fully confirm and enroll a Participant in a Salesforce Program. Try signing up using their
  Contact Id and notice that they get Canvas access.
- Ditto but run Sync To LMS with the Program and see that they get Canvas access and are provisioned in a section for their scheduled day/time. It can be run with something like:
 `bundle exec rake salesforce:sync_to_lms[a2Y1J000000YpQFUA0]`

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,7 @@
 require 'grade_calculator'
 require 'salesforce_api'
 require 'canvas_api'
+require 'sync_to_lms'
 
 class User < ApplicationRecord
   include Devise::Models::DatabaseAuthenticatable
@@ -125,8 +126,8 @@ class User < ApplicationRecord
     if existing_user
       self.canvas_id = existing_user['id']
     else
-      Rails.logger.error("User is trying to create an account but is not in Canvas: #{inspect}")
-      throw :abort # TODO: create them on the fly by running a Sync To LMS for just this user. Maybe it just wasn't run yet.
+      Rails.logger.info("Creating Canvas account and enrollments for new user: #{inspect}")
+      self.canvas_id = SyncToLMS.new.for_contact(salesforce_id)
     end
   end
 
--- a/lib/canvas_api.rb
+++ b/lib/canvas_api.rb
@@ -36,20 +36,28 @@ class CanvasAPI
 
   def get(path, params={}, headers={})
     RestClient.get("#{@api_url}#{path}", {params: params}.merge(@global_headers.merge(headers)))
+  rescue => e
+    handle_rest_client_error(e)
   end
 
   def post(path, body, headers={})
     RestClient.post("#{@api_url}#{path}", body, @global_headers.merge(headers))
+  rescue => e
+    handle_rest_client_error(e)
   end
 
   def put(path, body, headers={})
     RestClient.put("#{@api_url}#{path}", body, @global_headers.merge(headers))
+  rescue => e
+    handle_rest_client_error(e)
   end
 
   def delete(path, body={}, headers={})
     # Delete helper method doesn't accept a payload. Have to drop down lower level.
     RestClient::Request.execute(method: :delete, 
       url: "#{@api_url}#{path}", payload: body, headers: @global_headers.merge(headers))
+  rescue => e
+    handle_rest_client_error(e)
   end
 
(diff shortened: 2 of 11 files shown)


lib/canvas_api.rb
```diff
@@ -183,5 +201,12 @@ def upload_file_to_course(file, original_filename, content_type, course_id=Conte
     # Return preview URL.
     info = JSON.parse(response.body)
     {url: "#{canvas_url}/courses/#{course_id}/files/#{info['id']}/preview"}
+
+  end
+
+  def handle_rest_client_error(e)
```
sadleb, line 207: This is hacky, but i didn't want to invest time at the moment so I opened this task to refactor: https://app.asana.com/0/search/1174509298383602/1174359056398761

lib/canvas_api.rb
```diff
@@ -36,20 +36,28 @@ def initialize(canvas_url, canvas_token)
 
   def get(path, params={}, headers={})
     RestClient.get("#{@api_url}#{path}", {params: params}.merge(@global_headers.merge(headers)))
+  rescue => e
+    handle_rest_client_error(e)
```
sadleb, line 40: When RestClient throws, if we don't read the body out of the exception it's hard to diagnose. We'll just see something like 400 Bad but not get the details of what was bad.

lib/sync_to_lms.rb
```diff
@@ -13,20 +13,33 @@ class SyncToLMS
   def initialize
     @salesforce_api = SalesforceAPI.client
     @canvas_api = CanvasAPI.client
+    @programs = {}
   end
 
-  # Syncs all folks from Salesforce to Canvas for the specified course.
-  def execute(course_id)
```
sadleb, line 19: This class is the meat of the PR. `for_program` can be used to sync all participants in a program and `for_contact` is mean to be used to provision a user on the fly when they create their account.

app/models/user.rb
```diff
@@ -125,8 +126,8 @@ def setup_canvas_access
     if existing_user
       self.canvas_id = existing_user['id']
     else
-      Rails.logger.error("User is trying to create an account but is not in Canvas: #{inspect}")
-      throw :abort # TODO: create them on the fly by running a Sync To LMS for just this user. Maybe it just wasn't run yet.
+      Rails.logger.info("Creating Canvas account and enrollments for new user: #{inspect}")
+      self.canvas_id = SyncToLMS.new.for_contact(salesforce_id)
```
sadleb, line 130: Here is the on-the-fly provisioning / sync.

lib/canvas_api.rb
rshipp: something to keep in mind - it'd be amazing to get Honeycomb instrumentation in all of this. it's probably not a priority, but the earlier we start doing it, the easier it'll be when we need it https://docs.honeycomb.io/getting-data-in/ruby/beeline/#adding-context-to-events

sadleb:
> a lot of this is way over my head, but it's useful to see how we're handling all the salesforce stuff. i don't know how to give feedback tho. were there specific parts you wanted more eyes on?

Not really. We're working independently at the moment so it's not worth the effort to get each of us steeped in the other's code. I think the main point of sticking with requesting reviews is mainly to give each other a vague understanding of what is going on in the other's world -- and of course in case we can spot any big code smells or something that overlaps / could benefit from each other's work.

PR #194, opened 2020-05-07
Attempt to fix Codacy code coverage.

sadleb:
See: "Cannot send coverage when using VCR" here
https://www.rubydoc.info/gems/codacy-coverage/2.1.0


sadleb:
Just an fyi @rshipp

PR #199, opened 2020-05-08
Move yarn install from build to startup

rshipp:
We never use the packages we install at build time, because of the `node-modules` volume that gets mounted over top of them. Move yarn to the start script instead.

Results: 

* Container build is much faster ✅ 
* Container is smaller ✅ 
* Startup is 7s slower (on subsequent runs) ❌ 
* Startup is more reliable (no more yarn check failures! auto-installs new JS packages on container restart!) ✅

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,6 @@ WORKDIR /usr/src/app
 ENV BUNDLE_PATH /gems
 
 RUN bundle install && cp Gemfile.lock /tmp
-RUN yarn install --check-files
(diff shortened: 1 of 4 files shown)


sadleb:
> Wait, can we run yarn check and only install if necessary?

nm, just read that they recommend not to use that. lgtm!

rshipp:
yeah, `yarn check` is way slower (like 1min+) than `yarn install --check-files`

PR #200, opened 2020-05-08
Editor slider

jessouyangbraven:
**Task**: https://app.asana.com/0/1170776727341290/1174986069483581

**Test**
- 2 new sections, 2 slider questions, change min/max/step
- Click between two, see the slider editing inputs update
Screen Shot 2020-05-08 at 2 32 01 PM
Screen Shot 2020-05-08 at 2 31 55 PM

- Publish and make sure things behave sanely with our HTML
Screen Shot 2020-05-08 at 2 29 15 PM

- This is the HTML I found from an existing module `new.html` to test with: https://pastebin.com/8WMy6aGY

- This is the default HTML inserted when you put in a slider question:
Screen Shot 2020-05-08 at 2 26 48 PM

**Details**

- Commit 1: Refactor out some of the constants used for slider questions (min/max/step). 
- Commit 2: Fix the slider editing input so they update when you click between sliders
- Note: we're not giving designers the option to change `data-bz-range-clg` and `data-bz-range-flr` in the UI right now, but we support it for existing HTML.


app/javascript/components/ContentEditor.js
```diff
@@ -565,7 +562,7 @@ class ContentEditor extends Component {
                                             <input
                                                 type='number'
                                                 id='input-answer'
-                                                defaultValue={answer}
+                                                value={answer}
                                                 disabled={!(answer || rangeAnswer)}
```
jessouyangbraven, line 566: I'm not able to get into a state from the UI-side that lets me change the answer. When should this be disabled/enabled?

rshipp: this is https://app.asana.com/0/1170776727341290/1173560888333813/f. it's supposed to let you change the correct answer IFF both the data-answer and data-range-answer attributes are set (not undefined), because we don't want anyone to accidentally add a "correct answer" to the Rate This Module slider. the insertslidercommand is supposed to set both those attributes, so you can edit the correct answer for sliders you create. but that's evidently not working as intended.

app/javascript/components/ContentEditor.js
```diff
@@ -521,10 +521,7 @@ class ContentEditor extends Component {
                                     );
                                 } else if ( 'slider' === modelElement ) {
                                     // Sliders have several different settings to change.
-                                    const min = this.state['selectedElement'].getAttribute('min');
-                                    const max = this.state['selectedElement'].getAttribute('max');
-                                    const step = this.state['selectedElement'].getAttribute('step');
-
+                                    const selectedElement = this.state['selectedElement'];
```
rshipp, line 524: any time you grab a persistent reference to the selectedElement like this, you'll want to check it for undefined. for some reason it can magically disappear and crash the editor. see https://github.com/bebraven/platform/blob/7458f5cbd149868409bc64e96d949f42092dd6dd/app/javascript/components/ContentEditor.js#L472-L477

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -9,6 +9,7 @@ import InsertDoneButtonCommand from './insertdonebuttoncommand';
 import InsertContentBlockCommand from './insertcontentblockcommand';
 import InsertFileUploadQuestionCommand from './insertfileuploadquestioncommand';
 import { ALLOWED_ATTRIBUTES, filterAllowedAttributes } from './customelementattributepreservation';
+import * as Slider from '../constants/sliderquestionconstants';
```
rshipp, line 12: this is a cool syntax

app/javascript/components/ContentEditor.js
jessouyangbraven: Got it! Thanks! I'll leave the logic as-is for this diff. 

Is that task going to be an issue for Tess when she's editing? Does she need to update the correct answer on slider questions or can she leave what's already in the HTML?

rshipp: we decided it's not a blocker, Tess is converting the one slider she wants to edit into a radio question

rshipp:
> Note: we're not giving designers the option to change data-bz-range-clg and data-bz-range-flr in the UI right now, but we support it for existing HTML.

This is one of the things where we technically allow it to support existing content, but don't want to let anyone create new content that uses it.

PR #204, opened 2020-05-11
Add ability to run in production mode in your local dev env.

sadleb:
Test Plan
- switch the RAILS_ENV value in .env from development to production and
  rebuild the container. Ensure there are no dev related paths/links in
  the HTML or server logs. Switch it back to development and do the same.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -18,8 +18,12 @@ MAILER_FROM_EMAIL=someone@example.com
 MAILER_DELIVERY_OVERRIDE_ADDRESS=yourownemail@example.com
 VCR_JOIN_SERVER=joinweb:3001
 VCR_CANVAS_SERVER=canvasweb:3000
+RAILS_ENV=development
 RAILS_LOG_TO_STDOUT=true
+RAILS_SERVE_STATIC_FILES=true
 SALESFORCE_HOST=test.salesforce.com
+# Needed if you want to compile the app in production mode locally.
+SECRET_KEY_BASE=f843142daa4f4c69c7ac67bacd5837fd43f7b5390a922d9ac43692db24eb7a0f490229354db1720cd7e6f2cf1533a6560324a0584ebf87e2ba496c2d3e92fb9c
 SELENIUM_HOST=platform_chrome
 SELENIUM_PORT=4444
 SPEC_HOST=platformweb
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,13 +13,20 @@ RUN apk add --update --no-cache \
     python2 \
     vim
 
-COPY Gemfile* /usr/src/app/
-WORKDIR /usr/src/app
+# Allow local builds to change it. Defaults to development.
+ARG RAILS_ENV=development
+ENV RAILS_ENV ${RAILS_ENV}
+ENV NODE_ENV=${RAILS_ENV}
+
+WORKDIR /app
+
+RUN mkdir -p /app/tmp/pids/
 
 ENV BUNDLE_PATH /gems
 
+COPY Gemfile Gemfile.lock /app/
 RUN bundle install && cp Gemfile.lock /tmp
 
-COPY . /usr/src/app/
+COPY . /app/
 
 CMD ["/bin/sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,6 @@ ENV NODE_ENV=${RAILS_ENV}
 
 WORKDIR /app
 
-RUN mkdir -p /app/tmp/pids/
-
 ENV BUNDLE_PATH /gems
 
 COPY Gemfile Gemfile.lock /app/
--- a/config/application.rb
+++ b/config/application.rb
@@ -15,5 +15,10 @@ module Platform
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
+    # Allow `bundle exec rake assets:precompile` without loading the whole rails app. 
+    # E.g. if we want to run in production mode locally.
+    #config.assets.initialize_on_precompile=false
+
   end
 end
--- a/config/webpack/environment.js
+++ b/config/webpack/environment.js
@@ -1,23 +1,7 @@
 const { environment } = require('@rails/webpacker')
-const dotenv = require('dotenv')
 const ckeditorSVG = require('./loaders/ckeditor-svg')
 const ckeditorCSS = require('./loaders/ckeditor-css')
 const CKEditorWebpackPlugin = require( '@ckeditor/ckeditor5-dev-webpack-plugin' )
-
-const dotenvFiles = [
-  `.env.${process.env.NODE_ENV}.local`,
-  '.env.local',
-  `.env.${process.env.NODE_ENV}`,
-  '.env',
-]
-dotenvFiles.forEach((dotenvFile) => {
-  dotenv.config({ path: dotenvFile, silent: true })
-})
-
-environment.plugins.prepend('CKEditor', new CKEditorWebpackPlugin({
-        language: 'en'
-    })
-)
 
 environment.loaders.append('ckeditorCSS', ckeditorCSS)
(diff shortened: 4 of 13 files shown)


.env.example
```diff
@@ -18,8 +18,12 @@ MAILER_FROM_EMAIL=someone@example.com
 MAILER_DELIVERY_OVERRIDE_ADDRESS=yourownemail@example.com
 VCR_JOIN_SERVER=joinweb:3001
 VCR_CANVAS_SERVER=canvasweb:3000
+RAILS_ENV=development
 RAILS_LOG_TO_STDOUT=true
+RAILS_SERVE_STATIC_FILES=true
 SALESFORCE_HOST=test.salesforce.com
+# Needed if you want to compile the app in production mode locally.
+SECRET_KEY_BASE=f843142daa4f4c69c7ac67bacd5837fd43f7b5390a922d9ac43692db24eb7a0f490229354db1720cd7e6f2cf1533a6560324a0584ebf87e2ba496c2d3e92fb9c
```
sadleb, line 26: I generated this for dev. It just needs to be there for the precompile assets rake task in production mode.

.env.example
```diff
@@ -18,8 +18,12 @@ MAILER_FROM_EMAIL=someone@example.com
 MAILER_DELIVERY_OVERRIDE_ADDRESS=yourownemail@example.com
 VCR_JOIN_SERVER=joinweb:3001
 VCR_CANVAS_SERVER=canvasweb:3000
+RAILS_ENV=development
 RAILS_LOG_TO_STDOUT=true
+RAILS_SERVE_STATIC_FILES=true
```
sadleb, line 23: If this isn't set, production mode won't server static images.

Dockerfile
```diff
@@ -13,13 +13,20 @@ RUN apk add --update --no-cache \
     python2 \
     vim
 
-COPY Gemfile* /usr/src/app/
-WORKDIR /usr/src/app
+# Allow local builds to change it. Defaults to development.
+ARG RAILS_ENV=development
+ENV RAILS_ENV ${RAILS_ENV}
+ENV NODE_ENV=${RAILS_ENV}
```
sadleb, line 19: Needed for webpacker to build in prod mode.

config/application.rb
```diff
@@ -15,5 +15,10 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
+    # Allow `bundle exec rake assets:precompile` without loading the whole rails app. 
+    # E.g. if we want to run in production mode locally.
+    #config.assets.initialize_on_precompile=false
```
sadleb, line 21: Didn't end up needing to use this, but leaving it here in case anyone wants to try and get precompile assets to be built into the docker image in which case this would be helpful b/c you don't have to copy in a bunch of unnecessary files just to be able to run the rake task.

config/webpack/environment.js
```diff
@@ -1,8 +1,19 @@
 const { environment } = require('@rails/webpacker')
+const dotenv = require('dotenv')
 const ckeditorSVG = require('./loaders/ckeditor-svg')
 const ckeditorCSS = require('./loaders/ckeditor-css')
 const CKEditorWebpackPlugin = require( '@ckeditor/ckeditor5-dev-webpack-plugin' )
 
+const dotenvFiles = [
```
sadleb, line 7: I also didn't end up needing this, but it could be useful to have all the .env vars available for use in JS. @rshipp this wouldn't expose any secrets to the browser, right?

docker-compose/scripts/docker_compose_run.sh
```diff
@@ -48,13 +54,21 @@ sysctl fs.inotify.max_user_watches=524288
...
+  # Also, if you force polling it will absolutely destroy your CPU.
+  echo "Starting the rails app using guard"
+  bundle exec guard -di
+else
 
+  echo "Precompiling assets in production mode"
+  bundle exec rake assets:precompile
+
+  echo "Starting the rails app using puma"
+  bundle exec puma -C config/puma.rb
```
sadleb, line 73: I was doing so many iterations of trying to get things working and I think I need to directly run puma instead of through guard. @rshipp do you think guard would work with a prod build?

docker-compose/scripts/docker_compose_run.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/sh
 set -e
 
+# When you stop the container, it doesn't clean itself up properly so it fails to start next time. Cleanup!
```
rshipp, line 4: doesn't /app/tmp get destroyed when you stop the container?

Dockerfile
```diff
@@ -13,13 +13,20 @@ RUN apk add --update --no-cache \
...
+
+WORKDIR /app
+
+RUN mkdir -p /app/tmp/pids/
```
rshipp, line 23: what is this needed for? won't docker-compose mount /app/tmp overtop?

config/webpack/environment.js
rshipp: from what I understand, it'll cause all the secrets to get compiled in and make them available in `process.env` 😳 - especially if we're not using this, I'd remove it. there's nothing in `.env` i'll ever need access to from the editor at least.

docker-compose/scripts/docker_compose_run.sh
rshipp: Guard is only used to reload things when you make changes - we'd never want that in prod, so I think doing it this way is ideal.

docker-compose/scripts/docker_compose_run.sh
sadleb: Yup, this was left over from when I was trying to precompile assets into the container image and was copying stuff from the canvas-lms repo (which is setup differently) to try and fix. I punted on that. I'll yank this back out.

Dockerfile
sadleb: Ditto on: the comment above "this was left over from when I was trying to precompile assets into the container image ... I'll yank this back out."

docker-compose/scripts/docker_compose_run.sh
sadleb: Actually, i take it back. I think this does happen sometimes. In docker-compose.yml we make /app/tmp of a volume so it persists b/n restarts.

rshipp: I thought unnamed volumes were ephemeral and only named volumes were persistent? Not sure where I got that impression though. I don't have the correct vocab here bc searching for how unnamed volumes work isn't giving me any results.

sadleb: I guess I was confused about this as well. Just did some research. Named and unnamed volumes are the exact same in the sense that they persist between container restarts. The difference is where the data stored in that volume lives on the host machine. An unnamed volume will be stored at that path in the current directory. A named volume will be stored somewhere else, depending on the Docker host config. So the difference is just whether you want the stuff in a volume available outside the docker container in the current directory structure (e.g. to manage changes using src ctrl or to be able to run things against those files from outside the container, like perhaps grep through the node_modules) @rshipp

sadleb: Hmm, i just tried verifying this and it does seem to get blown away. I'm confused at this point.

sadleb: Ok, so what I said above is accurate if you're doing stuff directly with docker commands. But for docker-compose, the way it handle anonymous volumes is to create a new one everytime it starts the service. So yeah, I think you're right that I can remove this. I'm not sure how I was running into it at this point...

rshipp: interesting, thanks for the clarification!

Dockerfile
sadleb: Turns out i did need to create it. Puma doesn't create it on the fly like gaurd and all that dev stuff do. I moved it from the image to the startup script though and only run for prod builds.

PR #206, opened 2020-05-11
Add formatting to the accounts views (signup, forgot password, etc)

sadleb:
This is what the previous PRs have been building up to. It nicely formats the account related views. See: https://app.asana.com/0/1170770467635599/1173699558502375

Check them out:

image

image

image

image

image


app/views/cas/_login_form.erb
```diff
@@ -10,25 +10,34 @@
   <iframe style="visibility: hidden; width: 1px; height: 1px;" src="//<%= @settings[:public_site_domain] %>/users/clear_session_cookie"></iframe>
 <% end %>
 <form method="post" action="<%= @form_action || "login" %>" id="login-form"
-      onsubmit="submitbutton = document.getElementById('login-submit'); submitbutton.value='Please wait'; submitbutton.disabled=true; return true;">
-        <div><label id="username-label" for="username">
-          Email
-        </label><br />
-        <input class="form-control" autofocus="autofocus" type="text" id="username" name="username"
-               size="32" tabindex="1" accesskey="u" /></div>
```
rshipp, line 18: any reason these were removed? I don't know best practice around them, I assume tabbing between inputs is fine for a11y.

sadleb: Yeah, they don't recommend we use them anymore: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey

image

PR #208, opened 2020-05-12
Format the confirmation email and reset password email.

sadleb:
Nicely formattted HTML emails are notoriously hard to get right in a way that
they look good in all email cients. For this basic pass at making those emails
look nice, I grabbed this template: https://github.com/leemunroe/responsive-html-email-template
updated the colors, fonts, and some small other formatting. Then ran them through the CSS
inliner tool here: https://htmlemail.io/inline/ (according to the instructions in the .template file
for each).

### Test Plan:
- Try sign up a new account and use forgot password. View the emails you get in
various email clients, both web based and phone. Or use a tool like https://litmus.com/
(or maybe Campaign Monitor which we already pay for) to detect compatibility issues.


sadleb:
Foregoing a review. This is eye-bleeding stuff that no one reviewing would catch anything. The topic of avoiding being marked as spam is the main thing I would be concerned about, but I've already addressed most of that in previous commits and this one is simply the HTML.

PR #209, opened 2020-05-12
Apply new account layout to the edit password form.

sadleb:
Test Plan:
1. On the sign-in page click "Forgot password"
2. Click the link in the email
3. The "create a new password" page should look like the other signin/signup/password pages.


sadleb:
Foregoing a code review. This is the same as the other ones in the prior commits. I just missed a file.

PR #211, opened 2020-05-12
Delete <checklistQuestion>

jessouyangbraven:
**Task**: none

I noticed this when working on table of contents for the `<questionTitle>` and couldn't stop myself from poking around. 

**Test**

Use `?debug` and check the model when I insert a checklist question. Inserting a new question gives me `<question data-grade-as="checklist">`:
Screen Shot 2020-05-12 at 4 09 08 PM

Manually changing the HTML to `<div class="module-block module-block-checkbox">`. I never get a `<checklistQuestion>`, only `<question>` with extra classes on it: 
Screen Shot 2020-05-12 at 4 09 28 PM
Screen Shot 2020-05-12 at 4 09 56 PM

I think `<checklistQuestion>` is dead code at this point?

Also published to see if it messes things up in `canvasweb`, but that also seems fine. 

**Details**

- I think this is dead in favor of using `<question data-grade-as="checklist">` and we're using `data-grade-as` to distinguish between question types (e.g., slider/range, matrix, etc.) going forward. 
- I don't think this messed up existing modules of `modules/*/new.html`. 
- I don't know if it's an issue that some of the existing Booster modules have `data-grade-as="undefined"` right now.


rshipp:
Good catch, this should have been removed with my mass-module-block cleanup.

rshipp:
> I don't know if it's an issue that some of the existing Booster modules have data-grade-as="undefined" right now.

It is, but we have tasks for that. (We'll have to correct all those before launch)

PR #213, opened 2020-05-13
Make SSO logout work properly

sadleb:
Enable single-signout setting in SSO server.

With enable_single_sign_out set to true, it calls into each app that has authenticated
with this ticket and tells them to logout.

The `enable_single_sign_out` setting was only enabled in the test
configuration. It should be enabled for all environments. I think this was an
oversight. Refactor the whole rubycas.yml settings file so that shared settings
are defined in one place and each env only changes settings specific to it. 

### Test Plan:

platform logout
1. Login to platform.
2. Login to portal in a separate tab.
3. Logout of platform.
4. Try to hit a new page in portal, it should redirect to the sign-in page.

portal logout
1. Repeat step 1 & 2 above.
2. Logout of Portal
3. Try to hit a new page in platform, redirects to sign-in page.

### DIFF
--- a/config/rubycas.yml
+++ b/config/rubycas.yml
@@ -1,4 +1,5 @@
-production: &default
+default: &default
+  enable_single_sign_out: true
   maximum_unused_login_ticket_lifetime: 300
   maximum_unused_service_ticket_lifetime: 300
   maximum_session_lifetime: 172800
@@ -32,7 +33,7 @@ production: &default
     source: cas_authenticator.rb
 <% end %>
 
-development:
+development: &development_config
   <<: *default
   log:
     output: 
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -19,7 +19,7 @@
 
 Dir["./spec/support/**/*.rb"].sort.each{|f| require f}
 
-ENV['RAILS_ENV'] ||= 'test'
+ENV['RAILS_ENV'] = 'test'
```
sadleb, line 22: This is the fix for not having to run specs with `RAILS_ENV=test bundle exec rspec` -- in this PR https://github.com/bebraven/platform/pull/204 we build the RAILS_ENV into the image so specs were running with `development` instead of `test` b/c before the ENV var wasn't set at all in the shell.

config/rubycas.yml
```diff
@@ -1,4 +1,5 @@
-production: &default
+default: &default
+  enable_single_sign_out: true
```
sadleb, line 2: This is the main change. Everything else is just a refactor to not repeat stuff in the various environments.

sadleb:
> i never understood why people don't like yaml, but pointers in yaml makes my head spin

Word.

PR #214, opened 2020-05-13
Fix radio questions

rshipp:
Task: https://app.asana.com/0/1170776727341290/1172643959550099/f

Basically, I have no idea how radio buttons work, so I implemented this totally backwards. This PR fixes that.

I'd appreciate if someone can double-check my logic, bc this is confusing.

Things to know:

* Radio group **MUST** be unique for each question. Every button in the question shares the group.
* Group is tracked for convenience in `module-block[data-radio-group]`, but is actually determined by the radio `input`'s `name` attribute.
* Each radio has an `id` attribute that **MUST** be unique in the document.
* Each radio has a `data-bz-retained` ID that **MAY** match the `name` attribute, and **MUST** be the same for all radio `input`s in a group.
* Each radio has a `value` attribute that **MUST** be unique within each group (uniqueness within the document does not matter).
* Each radio has a `data-correctness` attribute that is used to show color / tally results for autograding.
* Each radio has a corresponding `label`, with a `for` attrubute that matches the `input`'s `id` attribute.

Issues I noticed while doing this PR, but consider out of scope for Booster:

* When you add a radio / checklist option in the editor, the attributes are not set. The attributes are only set once you reload the editor (e.g. toggle code->design, or save the document and let it reload.) This means there's an edge-case bug where if you add a radio button or checklist option, save, then publish, that radio button will not work. There are two workarounds: 1) after adding an option, save, save again, publish or 2) after adding an option, toggle code tab, toggle design tab, save, publish. Task: https://app.asana.com/0/1170776727341290/1175871850555492/f
* A similar issue when you insert a checklist question - the first option will be non-functional until you follow the steps above.

### DIFF
--- a/app/javascript/ckeditor/insertradioquestioncommand.js
+++ b/app/javascript/ckeditor/insertradioquestioncommand.js
@@ -19,6 +19,8 @@ export default class InsertRadioQuestionCommand extends Command {
 
 function createRadioQuestion( writer ) {
     const radioGroup = uid();
+    // Value must be unique within each group, but otherwise doesn't matter.
+    // No reason to tie in retained data or anything, we can just use "1".
     const radioFirstValue = '1';
     const radioFirstID = [radioGroup, radioFirstValue].join('_');
 
--- a/app/javascript/ckeditor/insertradioquestioncommand.js
+++ b/app/javascript/ckeditor/insertradioquestioncommand.js
@@ -18,7 +18,11 @@ export default class InsertRadioQuestionCommand extends Command {
 }
 
 function createRadioQuestion( writer ) {
-    const radioQuestion = writer.createElement( 'moduleBlock', {'data-radio-group': uid()} );
+    const radioGroup = uid();
+    const radioFirstValue = '1';
+    const radioFirstID = [radioGroup, radioFirstValue].join('_');
+
+    const radioQuestion = writer.createElement( 'moduleBlock', {'data-radio-group': radioGroup} );
     const question = writer.createElement( 'question', { 'data-grade-as': 'radio' } );
     const questionTitle = writer.createElement( 'questionTitle' );
     const questionBody = writer.createElement( 'questionBody' );
@@ -26,8 +30,12 @@ function createRadioQuestion( writer ) {
     const questionFieldset = writer.createElement( 'questionFieldset' );
     const doneButton = writer.createElement( 'doneButton' );
     const radioDiv = writer.createElement( 'radioDiv' );
-    const radioInput = writer.createElement( 'radioInput' );
-    const radioLabel = writer.createElement( 'radioLabel' );
+    const radioInput = writer.createElement( 'radioInput', {
+        name: radioGroup,
+        id: radioFirstID,
+        value: radioFirstValue,
+    } );
+    const radioLabel = writer.createElement( 'radioLabel', { 'for': radioFirstID } );
     const radioInlineFeedback = writer.createElement( 'radioInlineFeedback' );
     const answer = writer.createElement( 'answer' );
     const answerTitle = writer.createElement( 'answerTitle' );
--- a/app/javascript/ckeditor/moduleblockediting.js
+++ b/app/javascript/ckeditor/moduleblockediting.js
@@ -61,6 +61,7 @@ export default class ModuleBlockEditing extends Plugin {
                 // FIXME: Camelcase is broken with CKE built-in conversions, esp. on upcast
                 'blockClasses',
                 'data-icon',
+                'data-radio-group',
             ],
         } );
     }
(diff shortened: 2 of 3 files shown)


app/javascript/ckeditor/insertradioquestioncommand.js
```diff
@@ -18,16 +18,24 @@ export default class InsertRadioQuestionCommand extends Command {
 }
 
 function createRadioQuestion( writer ) {
-    const radioQuestion = writer.createElement( 'moduleBlock', {'data-radio-group': uid()} );
+    const radioGroup = uid();
+    const radioFirstValue = '1';
```
rshipp, line 22: value must be unique within each group, but otherwise doesn't matter. no reason to tie in retained data or anything, we can just use "1".

app/javascript/ckeditor/insertradioquestioncommand.js
```diff
@@ -18,16 +18,24 @@ export default class InsertRadioQuestionCommand extends Command {
 }
 
 function createRadioQuestion( writer ) {
-    const radioQuestion = writer.createElement( 'moduleBlock', {'data-radio-group': uid()} );
+    const radioGroup = uid();
```
rshipp, line 21: this could technically collide, but ....

app/javascript/ckeditor/moduleblockediting.js
```diff
@@ -61,6 +61,7 @@ export default class ModuleBlockEditing extends Plugin {
                 // FIXME: Camelcase is broken with CKE built-in conversions, esp. on upcast
                 'blockClasses',
                 'data-icon',
+                'data-radio-group',
```
rshipp, line 64: need to allow this in moduleBlock, since we no longer have a radioQuestion

app/javascript/ckeditor/moduleblockediting.js
```diff
@@ -94,26 +95,55 @@ export default class ModuleBlockEditing extends Plugin {
                 // Remove icon from class
                 const blockClasses = srcClasses.replace(icon, "").trim();
 
-                return modelWriter.createElement( 'moduleBlock', {
+                let attributes = {
```
rshipp, line 98: factoring this out here and using the `if` below so that we don't get `data-radio-group="undefined"` on every moduleBlock that isn't a radio question.

app/javascript/ckeditor/radioquestionediting.js
```diff
@@ -195,10 +195,10 @@ export default class RadioQuestionEditing extends Plugin {
 
                 return modelWriter.createElement( 'radioInput', new Map( [
                     ...filterAllowedAttributes(viewElement.getAttributes()),
-                    [ 'id', id ],
+                    [ 'id', [ radioGroupName, id ].join( '_' ) ],
                     [ 'name', radioGroupName ],
-                    [ 'value', viewElement.getAttribute('value') ],
-                    [ 'data-bz-retained', id ],
+                    [ 'value', viewElement.getAttribute('value') || id ],
+                    [ 'data-bz-retained', radioGroupName ],
```
rshipp, line 155: this is the most important bit

app/javascript/ckeditor/moduleblockediting.js
```diff
@@ -94,26 +95,55 @@ export default class ModuleBlockEditing extends Plugin {
...
                     'blockClasses': blockClasses,
                     'data-icon': icon,
-                });
+                }
+
+                // Special radio-question support.
+                // This attribute is set in insertradioquestioncommand.js.
+                const radioGroup = viewElement.getAttribute( 'data-radio-group' );
```
jessouyangbraven, line 105: Is there a reason why we can't use `data-grade-as="radio"` to detect radio questions? Do we not have the element subtree at that point?

app/javascript/ckeditor/insertradioquestioncommand.js
jessouyangbraven: Leave as comment in code?

app/javascript/ckeditor/radioquestionediting.js
jessouyangbraven: I'm seeing the same `id` repeated for every radio/`<input>` in a group/`<fieldset>`:

Screen Shot 2020-05-14 at 11 42 36 AM

Am I doing something weird in the editor to cause this? I toggled between Code/Design a couple of times to make sure the downcast ran. 

I can' find the `id="{uid}_1"` that came from the `createRadioQuestion()`. Maybe the `id` isn't being set/propagated at some point? Do we need an `attributeToAttribute()` for id?

app/javascript/ckeditor/moduleblockediting.js
rshipp: We don't really care if it's a radio question or not, we just want the group ID without having to traverse up to the `<question>` and then back down hoping to run into a radio `input` and hoping again it has a `name` attribute.

jessouyangbraven:
>Things to know:

This entire section. So helpful. Thank you!!

Also, I look at this https://www.w3.org/WAI/tutorials/forms/grouping/ a lot for simple examples of `name/value/id`. Because I can never remember what they mean.

app/javascript/ckeditor/radioquestionediting.js
rshipp: hm yeah somethings wrong, it should use uid_value not uid_name

rshipp: > Do we need an attributeToAttribute() for id?

We only ever need attributeToAttribute if we want to see model changes live in the view, which we only care about for things like the icons, correctness, etc that have CSS attached.

rshipp: ok I think i have this fixed in the latest commit, can you check again?

app/javascript/ckeditor/radioquestionediting.js
```diff
@@ -186,29 +138,36 @@ export default class RadioQuestionEditing extends Plugin {
...
                     ...filterAllowedAttributes(viewElement.getAttributes()),
-                    [ 'id', id ],
+                    [ 'id', [ radioGroupName, radioValue ].join( '_' ) ],
                     [ 'name', radioGroupName ],
-                    [ 'value', viewElement.getAttribute('value') ],
-                    [ 'data-bz-retained', id ],
+                    [ 'value', radioValue ],
+                    [ 'data-bz-retained', retainedID ],
                     [ 'data-correctness', viewElement.getAttribute('data-correctness') || '' ]
                 ] ) );
```
rshipp, line 164: changed all of this.

app/javascript/ckeditor/radioquestionediting.js
```diff
@@ -186,29 +138,36 @@ export default class RadioQuestionEditing extends Plugin {
...
             }
 
         } );
-        conversion.for( 'dataDowncast' ).elementToElement( {
+        conversion.for( 'downcast' ).elementToElement( {
```
rshipp, line 168: i merged both downcasts into one, since they've been the same for a while now.

PR #215, opened 2020-05-14
Auto login the user when they confirm their new account.

sadleb:
### Test Plan:
- Register a new user who has Portal access and when you click the email
  confirmation link it logs you into the Portal.
- Note: we currently don't have a way to register an admin user without
  Portal access, so we can't test that. When the day comes it'll break
  if I got it wrong

### DIFF
--- a/app/controllers/cas_controller.rb
+++ b/app/controllers/cas_controller.rb
@@ -256,7 +256,9 @@ class CasController < ApplicationController
 
     @lt = LT.create! @request_client
 
-    if @gateway && @service
+    if current_user
+      sign_out_and_redirect(current_user) and return
+    elsif @gateway && @service
       return render :login
     elsif @continue_url
       return render :logout
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Users::ConfirmationsController < Devise::ConfirmationsController
+  include RubyCAS::Server::Core::Tickets
+
   layout 'accounts'
   before_action :configure_permitted_parameters
 
@@ -18,9 +20,14 @@ class Users::ConfirmationsController < Devise::ConfirmationsController
   end
 
   # GET /resource/confirmation?confirmation_token=abcdef
-  # def show
-  #   super
-  # end
+  def show
+    super do
+      Rails.logger.debug("Account creation confirmed for #{resource}. Signing them in using CAS SSO.")
+      redirect_path_for_user = sign_in_and_get_redirect_path
+      set_flash_message!(:notice, :confirmed)
+      redirect_to redirect_path_for_user and return
+    end
+  end
 
   protected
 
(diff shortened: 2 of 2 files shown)


app/controllers/cas_controller.rb
```diff
@@ -255,7 +255,9 @@ def logout
 
     @lt = generate_login_ticket(@request_client)
 
-    if @gateway && @service
+    if current_user
```
sadleb, line 258: See the commit for why this is here: https://github.com/bebraven/platform/pull/215/commits/ab7c106a37e6796c3c3f5f5b8d7951494406b748

app/controllers/users/confirmations_controller.rb
```diff
@@ -18,9 +20,14 @@ def create
...
 
   # GET /resource/confirmation?confirmation_token=abcdef
-  # def show
-  #   super
-  # end
+  def show
+    super do
+      Rails.logger.debug("Account creation confirmed for #{resource}. Signing them in using CAS SSO.")
+      redirect_path_for_user = sign_in_and_get_redirect_path
+      set_flash_message!(:notice, :confirmed)
```
sadleb, line 27: This is built in devise stuff, I just ripped it from the base class. We're not currently using the flashes in our view, but if we do this would make that continue to work.

app/controllers/users/confirmations_controller.rb
```diff
@@ -29,23 +36,40 @@ def after_resending_confirmation_instructions_path_for(_)
...
-    # After confirming their account, have them login through CAS instead of being immediately signed in.
-    # Note: if you wanted to sign them in, call this:
-    # sign_in(resource)
-    cas_login_url
+#  # The path used after confirmation.
+#  def after_confirmation_path_for(resource_name, resource)
+#    cas_login_url
+#  end
+
+  def cas_login_url
```
sadleb, line 44: Leaving this here in case we need it even though it's not currently being used.

app/controllers/users/confirmations_controller.rb
```diff
@@ -29,23 +36,40 @@ def after_resending_confirmation_instructions_path_for(_)
...
+    ::Devise.cas_client.add_service_to_login_url(::Devise.cas_service_url(request.url, devise_mapping))
   end
+  helper_method :cas_login_url
+
+  # Does a CAS SSO login for the app they are redirected to on login. E.g. if they have Canvas access,
+  # auto-log them in there, otherwise do it here in the platform.
+  def sign_in_and_get_redirect_path
+     username = resource.email
+     if resource.canvas_id
+       login_service_url = URI.join(CanvasAPI.client.canvas_url, "/login/cas").to_s
```
sadleb, line 54: When they got to canvas, canvas calls back into platform to validate the SSO ticket. This associates the tickets we create with the callback we expect from canvas so they match up.

app/controllers/users/confirmations_controller.rb
```diff
@@ -29,23 +36,40 @@ def after_resending_confirmation_instructions_path_for(_)
...
+     end
+
+     # TGT = Ticket Granting Ticket helper, ST = Service Ticket helper. See: lib/rubycas-server-core/tickets.rb 
+     tgt = TGT.create! username, ::Devise.cas_client
+     response.set_cookie('tgt', tgt.to_s) # When the app calls back to validate the ticket, this is what makes that work.
+     st = ST.create! login_service_url, username, tgt, ::Devise.cas_client
 
+     Rails.logger.debug("Done signing in confirmed user #{username} with CAS Ticket granting cookie '#{tgt.inspect}' and Service Ticket #{st.inspect}")
+
+     Utils.build_ticketed_url(login_service_url, st)
```
sadleb, line 66: This causes them to be redirected to portal which immediately calls back into the SSO server to validate the ticket and then create a logged-in session on the portal side.

app/controllers/cas_controller.rb
rshipp: why is current_user not @current_user? I don't see that variable getting set in this function

app/controllers/users/confirmations_controller.rb
rshipp: this URI.join looks like it works whether canvas_url has a trailing slash or not, but heads up that function isn't generally very good https://stackoverflow.com/questions/8900782/how-do-i-safely-join-relative-url-segments. (Something I noticed when messing with the canvas api lib.)

sadleb:
I've got to get this live before I go on vaca, so I'm merging. I'll address any code review feedback in future PRs.

app/controllers/cas_controller.rb
sadleb: This is a helper method provided by devise. Tons of stuff behind the scenes going on here.

app/controllers/users/confirmations_controller.rb
sadleb: Thanks for the heads up!

PR #216, opened 2020-05-14
Cleanup the CAS controller a bit.

sadleb:
This controller was a copy/paste of bits and pieces of the old Join
server code along with the rubycas SSO server. It's a mess. I just
opened up a PR to run a tiny bit of this logic when confirming your
new account here: https://github.com/bebraven/platform/pull/215
and learned something about the rubycas-server-core gem's code.
This takes a small pass at starting to clean this controller up by only
including the top-level Tickets utility class and using that to do all the
ticket generation and validation.

Note: the proxy ticket stuff isn't implemented in that Tickets module
so I hacked that logic to follow the same patterns with the goal of moving
that code to the gem at some point.

Test Plan:
- Login/logout of both platform and portal still work.
- bundle exec rspec spec/controllers/cas

### DIFF
--- a/app/controllers/cas_controller.rb
+++ b/app/controllers/cas_controller.rb
@@ -12,10 +12,13 @@ class CasController < ApplicationController
   before_action :set_request_client
   before_action :set_params
 
-  include RubyCAS::Server::Core::Util
+  # Note that the following helper classes in this controller come from this module
+  # TGT = Ticket Granting Ticket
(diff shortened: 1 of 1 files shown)


sadleb:
Foregoing a code review b/c this doesn't change any logic or really refactor anything significant. It's just a step in the right direction. Specs work and I confirmed login/logout between platform/portal still works.

PR #222, opened 2020-05-15
Store newly created Canvas user's canvas_id in Salesforce.

sadleb:
Test plan:
- Sign-up a new Enrolled Participant from Salesforce. See that their
  Canvas User ID field gets set on the Contact record


sadleb:
Huge no, no to do on a Friday before I head out to vacation, but I'm just going to push this live. It's tested in staging and specs pass.

PR #223, opened 2020-05-15
Prevent display value from appearing on top of slider

jessouyangbraven:
**Task**: https://app.asana.com/0/1170776727341290/1174547828723905

**Test**: Same as https://github.com/beyond-z/canvas-lms-js-css/pull/303, but check that number appears below slider now. 

Screen Shot 2020-05-15 at 12 28 45 PM


app/javascript/ckeditor/insertsliderquestioncommand.js
```diff
@@ -47,7 +47,7 @@ function createSliderQuestion( writer ) {
     writer.append( questionForm, question );
     writer.append( questionFieldset, questionForm );
     writer.append( slider, questionFieldset );
-    writer.append( displayValueDiv, questionFieldset );
+    writer.append( displayValueDiv, questionForm );
```
jessouyangbraven, line 50: I got the slider and dispalyValueDiv to stop flipping by moving the latter out of the fieldset. 

Alternatively, I can apply `dont-mix` to the classes on the fieldset if you think that's cleaner. I don't think that's ever widget-ized, so we won't get into selection handle issues?

jessouyangbraven:
I don't know why my Gemfile.lock tagged along in the commit. Will fix.

app/javascript/ckeditor/insertsliderquestioncommand.js
rshipp: I think fixing this from the [module-js side](https://github.com/beyond-z/canvas-lms-js-css/blob/staging/braven_custom.js#L618) would be ideal. We should only ever be mixing checkboxDivs and radioDivs. 

That said, I don't actually know what fieldsets are for, that's only there because it was in the HTML I got from our designer. If you think only checkboxes and radio buttons should be in a fieldset, we can move everything else up to the form. If we go this route, we'll also need to modify the associated schemas (e.g. to allow `displayValueDiv` in `questionForm`s, so this is valid model content), and migrate all existing booster and non-booster content (likely by hand, because this doesn't seem like something we can easily take care of in a converter).

jessouyangbraven:
Closing out in favor of module-js side fix.

PR #224, opened 2020-05-18
Add anchor link to <{question, answer, content}Title>

jessouyangbraven:
**Task**: https://app.asana.com/0/1170776727341290/1170693002395753

**Test**

**Link on `<contentTitle>`**: check with `?debug` in model/view of CKE inspector
Screen Shot 2020-05-18 at 4 14 50 PM

**Link on `<questionTitle>`**: check with `?debug` in model/view of CKE inspector
Screen Shot 2020-05-18 at 4 15 06 PM

**Link on `<answerTitle>`**: check in `Code` tab:
Screen Shot 2020-05-19 at 10 52 48 AM

**Published to `canvasweb`**: click and make sure anchor links work
*question and content titles are linked*
Screen Shot 2020-05-18 at 4 15 21 PM
*answer title is linked*
Screen Shot 2020-05-19 at 10 52 34 AM

**Details**

- CKE gave us a solution here: https://github.com/ckeditor/ckeditor5/issues/6815
- The individual commits here a mess, but if you look at the `Files changed`, that should look sane. 
- We need to Save twice to get all the conversions to run to assign the `id` and `toc-link-href`, which isn't ideal. 
- The implementation steps are roughly:
-- Expose the counter from `retaineddata.js` so we can use it as a unique `id`
-- Make schema changes for `<contentTitle>` and `<questionTitle>` to allow attributes `id` and `toc-link-href`
-- On upcast/downcast conversions, propagate the `id` already on the element (in case we've already assigned it in a previous conversion or someone assigned on manually in the HTML). If there isn't one, get a unique one from `retaineddata` to use
-- On upcast for `<contentTitle>` and `<questionTitle>`, use that `id` to assign the `toc-link-href`, this is our attribute to let us know to link the text in these titles
-- Then we need to add a low-level `dispatcher` that listens on `element:h5`, to look for `toc-link-href` and attach the `linkHref` supported by CKE. 
-- I think roughly the same steps apply for our other headings, heading{1-5}, which map to `<h{2-6}>`. There might be a lot of boiler plate, but I'll see what I can refactor out.

### DIFF
--- a/app/javascript/ckeditor/contentcommonediting.js
+++ b/app/javascript/ckeditor/contentcommonediting.js
@@ -305,35 +305,24 @@ export default class ContentCommonEditing extends Plugin {
         } );
 
         // <tocLinkHref> converters
-        editor.conversion.for( 'upcast' ).elementToAttribute( {
-            view: {
-                name: 'a', 
-                attributes: {
-                    toc: true, // I think this is a canary value?
-                },
-            },
-            model: {
-                key: 'questionTitle',
-                value: ( viewElement, modelWriter ) => {
-                    debugger;
-                    return viewElement.getAttribute( 'tocLinkHref' );
-                },
-            },
-        });
-
+        // Using the low-level dispatcher because what we want is elementToNothing, which isn't a CKE funciton.
+        editor.conversion.for( 'upcast' ).add( dispatcher => { dispatcher.on( 'element:a', ( evt, data, conversionApi ) => {
+            if ( conversionApi.consumable.consume( data.viewItem, { name: true, attributes: [ 'href', 'toc' ] } ) ) {
+                // <a> element is inline and is NOT represented in the model.
+                // This is why we need to convert only children.
+                conversionApi.convertChildren( data.viewItem, data.modelCursor );
+            }
+        }, { priority: 'high' } ) } );
         editor.conversion.for( 'dataDowncast' ).attributeToElement( {
             model: 'tocLinkHref',
             view: ( href, viewWriter ) => {
-                //debugger;
-
-                if (!href) {
-                    return;
-                }
-                const linkElement = viewWriter.createAttributeElement( 'a', { href } );
+                const linkElement = viewWriter.createAttributeElement( 'a', { href }, { priority: 5 });
                 viewWriter.setCustomProperty( 'toc', true, linkElement );
+                console.log(linkElement);
                 return linkElement;
             },
-        });
+            converterPriority: 'highest',
+        } );
         conversion.for ('editingDowncast' ).attributeToElement( {
             model: 'tocLinkHref',
             view: (href, viewWriter) => {
--- a/app/javascript/ckeditor/contentcommonediting.js
+++ b/app/javascript/ckeditor/contentcommonediting.js
@@ -43,7 +43,7 @@ export default class ContentCommonEditing extends Plugin {
         schema.register( 'contentTitle', {
             isLimit: true,
             allowIn: 'content',
-            allowAttributes: [ 'id' ],
+            allowAttributes: [ 'id', 'toc-link-href' ],
             allowContentOf: '$block'
         } );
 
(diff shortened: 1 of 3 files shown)


app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -303,36 +304,123 @@ export default class ContentCommonEditing extends Plugin {
...
+                    return;
+                }
+                const linkElement = viewWriter.createAttributeElement( 'a', { href });
+                viewWriter.setCustomProperty( 'toc', true, linkElement );
+                return linkElement;
+            },
+        });
+        */
+
+        conversion.for( 'upcast' ).add( dispatcher => {
```
jessouyangbraven, line 369: This is the key: using CKE's `linkHref` and applying it to the text in the headings we care about. https://github.com/ckeditor/ckeditor5/issues/6815

jessouyangbraven: After Savings/toggling between Code/Design, we get this for a `<questionTitle>`:

Screen Shot 2020-05-18 at 1 14 30 PM

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -165,6 +166,28 @@ export default class ContentCommonEditing extends Plugin {
         const editor = this.editor;
         const conversion = editor.conversion;
         const { editing, data, model } = editor;
+        
+        // Table of contents converters for <contentTitle> and <questionTitle>
+        conversion.for( 'upcast' ).add( dispatcher => {
+            dispatcher.on( 'element:h5', ( evt, data, conversionApi ) => {
```
jessouyangbraven, line 172: I really hope there's something like `element:h*` for all headings or there's going to be a lot of duplication. :(

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -194,8 +217,10 @@ export default class ContentCommonEditing extends Plugin {
                 name: 'h5'
             },
             model: ( viewElement, modelWriter ) => {
+                const id = viewElement.getAttribute( 'id' ) || this._nextId();
```
jessouyangbraven, line 220: Should I be initializing this in the `insert*questioncommand` when the element is first created so we don't have to toggle between design/code? Is there a way to access `retaineddata` from a command?

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -42,7 +43,7 @@ export default class ContentCommonEditing extends Plugin {
         schema.register( 'contentTitle', {
             isLimit: true,
             allowIn: 'content',
-            allowAttributes: [ 'id' ],
+            allowAttributes: [ 'id', 'toc-link-href' ],
```
rshipp, line 46: i changed this from tocLinkHref just to see if it was breaking from that camelcase issue you ran into earlier, we can change it back if you want?

app/javascript/ckeditor/contentcommonediting.js
rshipp: we already loop through all the predefined heading elements in customelementattributepreservation.js, just need to replace that upcast with one like this.

app/javascript/ckeditor/contentcommonediting.js
rshipp: we can use the retained data plugin from commands, but i think it's better to keep it here, since that's consistent with how all the other converters work. if we ever decide to change that, we should do all of them at once.

app/javascript/ckeditor/retaineddata.js
```diff
@@ -38,11 +38,16 @@ export default class RetainedData extends Plugin {
 
         this._consumeAttributeEvent = this._consumeAttributeEvent.bind(this);
         this.getNextId = this.getNextId.bind(this);
+        this.getNextCount = this.getNextCount.bind(this);
 
         // Consume every attribute change event.
         editingDoc.listenTo( editingDoc.roots.first, 'change:attributes', this._consumeAttributeEvent );
     }
 
+    getNextCount() {
```
rshipp, line 47: can you add a note about this function also incrementing the ID to the class comment up top? right now the comment says only getNextId does it

app/javascript/ckeditor/contentcommonediting.js
rshipp: for the toggle issue, we're going to run into that in a few places here, since we're adding the <a> on upcast instead of downcast. not sure how to fix that... I'm leaning towards don't worry about it for now, and tackle all the existing issues like this at the same time later on.

app/javascript/ckeditor/contentcommonediting.js
jessouyangbraven: I'm so relieved.

app/javascript/ckeditor/contentcommonediting.js
jessouyangbraven: This all makes sense. Thanks!

app/javascript/ckeditor/contentcommonediting.js
jessouyangbraven: I prefer your naming convention, to be honest. And I'd rather not stub my toe on the camel-case problem again.

PR #225, opened 2020-05-18
Fix slider answer attributes

rshipp:
Task: https://app.asana.com/0/1170776727341290/1176387214965594/f
Task: https://app.asana.com/0/1170776727341290/1173560888333813/f

We don't actually want data-bz-answer on all sliders, that attribute is only used for mastery questions.


rshipp:
No we still need to allow that attribute on things, we just don't use it at all in the booster.

PR #228, opened 2020-05-19
Add id and toc-link-href to h1-5

jessouyangbraven:
**Task**: https://app.asana.com/0/1170776727341290/1176606860502193

**Test**

Manually type some text, highlight and set heading in drop-down:
Screen Shot 2020-05-19 at 3 10 24 PM

Toggle between Design/Code and check with `?debug` in CKE inspector that `id` and `toc-link-href` are being set:
Screen Shot 2020-05-19 at 2 57 15 PM

Publish to `canvasweb` and check that table of contents is generated properly:
Screen Shot 2020-05-19 at 2 59 18 PM

To test the anchor links, I padded out with some questions & answers to make the browser scroll. I also made sure the `<h5>` from the `<{question, answer, content}Title>`s weren't showing up in the table of contents:
Screen Shot 2020-05-19 at 3 01 13 PM

Made sure that merging of attributes works by manually changing code to add `class=jess-test` on a heading: 
Screen Shot 2020-05-19 at 3 01 45 PM

Go back to `Design` and make sure it's still in the model:
Screen Shot 2020-05-19 at 3 01 53 PM

**Details**

- Need to update `retaineddata`'s `_consumeAttributeEvent()` to look for `id`, in addition to our retained data ID to initialize the id counter. 
- Need **both** an upcast and downcast to call `_getNextId()` in order for `retaineddata` to increment properly. See comment in code.

### DIFF
--- a/app/javascript/ckeditor/customelementattributepreservation.js
+++ b/app/javascript/ckeditor/customelementattributepreservation.js
@@ -28,7 +28,6 @@ export const ALLOWED_ATTRIBUTES = [
     'tabindex',
     'style',
     'align',
-    'toc-link-href',
 ];
 
 export default class CustomElementAttributePreservation extends Plugin {
@@ -92,14 +91,25 @@ export default class CustomElementAttributePreservation extends Plugin {
             editor.conversion.for( 'upcast' ).elementToElement( {
                 view: key,
                 model: ( viewElement, modelWriter ) => {
-                    const id = viewElement.getAttribute( 'id' ) || this._nextId();
-                    return modelWriter.createElement( elements[key], {
-                        'id': id,
-                        'toc-link-href': '#' + id,
-                        ...filterAllowedAttributes( viewElement.getAttributes() ),
-                    } );
+                    let attributes = filterAllowedAttributes( viewElement.getAttributes() );
+                    if ( !attributes.get( 'id' ) ) {
+                        attributes.set( 'id', this._nextId() );
+                    }
+                    attributes.set( 'toc-link-href', '#' + attributes.get( 'id' ) );
+                    return modelWriter.createElement( elements[key], attributes );
                 },
             } );
+
+            editor.conversion.for( 'downcast' ).elementToElement( {
+                model: elements[key], 
+                view: ( modelElement, viewWriter ) => {
+                    let attributes = filterAllowedAttributes( modelElement.getAttributes() );
+                    if ( !attributes.get( 'id' ) ) {
+                        attributes.set( 'id', this._nextId() );
+                    }
+                    return viewWriter.createContainerElement( key, attributes );
+                },
+            });
 
             // Table of contents converter for headings
             editor.conversion.for( 'upcast' ).add( dispatcher => {
--- a/app/javascript/ckeditor/customelementattributepreservation.js
+++ b/app/javascript/ckeditor/customelementattributepreservation.js
@@ -91,10 +91,11 @@ export default class CustomElementAttributePreservation extends Plugin {
             editor.conversion.for( 'upcast' ).elementToElement( {
                 view: key,
                 model: ( viewElement, modelWriter ) => {
-                    return modelWriter.createElement(
-                        elements[key],
-                        filterAllowedAttributes( viewElement.getAttributes() ),
-                    );
+                    let attributes = filterAllowedAttributes( viewElement.getAttributes() );
+                    if ( attributes.get( 'id' ) ) {
+                        attributes.set( 'toc-link-href', '#' + attributes.get( 'id' ) );
+                    }
+                    return modelWriter.createElement( elements[key], attributes );
                 },
             } );
 
@@ -105,7 +106,6 @@ export default class CustomElementAttributePreservation extends Plugin {
                     if ( !attributes.get( 'id' ) ) {
                         attributes.set( 'id', this._nextId() );
                     }
-                    attributes.set( 'toc-link-href', '#' + attributes.get( 'id' ) );
                     return viewWriter.createContainerElement( key, attributes );
                 },
             });
--- a/app/javascript/ckeditor/customelementattributepreservation.js
+++ b/app/javascript/ckeditor/customelementattributepreservation.js
@@ -1,6 +1,7 @@
 // See https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/conversion/conversion-preserving-custom-content.html#adding-extra-attributes-to-elements-contained-in-a-figure
 // and https://github.com/ckeditor/ckeditor5/issues/5569
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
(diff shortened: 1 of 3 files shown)


app/javascript/ckeditor/customelementattributepreservation.js
```diff
@@ -48,6 +50,148 @@ export default class CustomElementAttributePreservation extends Plugin {
...
+        // in the CKE heading options are set to `low` priority.
+        const elements = {
+            'h2': 'heading1',
+            'h3': 'heading2',
+            'h4': 'heading3',
+            'h5': 'heading4',
+            'h6': 'heading5',
+        }
+
+        Object.keys( elements ).forEach( ( key ) => {
```
jessouyangbraven, line 90: All the new code for `id` and `toc-link-href` is in here. 

The rest was just tabbed in.

app/javascript/ckeditor/customelementattributepreservation.js
```diff
@@ -27,18 +28,19 @@ export const ALLOWED_ATTRIBUTES = [
...
-        return [ ];
+        return [ RetainedData ];
     }
 
     init() {
         const editor = this.editor;
 
         // Enable custom attributes on several predefined elements (see function).
-        setupAllowedAttributePreservation( editor );
+        this._setupAllowedAttributePreservation();
```
jessouyangbraven, line 43: I moved this so I could call `this._nextId()`.

app/javascript/ckeditor/retaineddata.js
```diff
@@ -55,7 +55,17 @@ export default class RetainedData extends Plugin {
     _consumeAttributeEvent( evt, elem ) {
         for ( let attribute of elem.getAttributeKeys() ) {
 
-            // Ignore all but retained data IDs.
+            // ID
+            if ( attribute === 'id' ) {
```
jessouyangbraven, line 59: Alternatively, I could set retained data ID every time I set id. 

And I think I was getting lucky with the questions/answers/content because those seem to always have retained data ID on them.

app/javascript/ckeditor/customelementattributepreservation.js
```diff
@@ -48,6 +49,159 @@ export default class CustomElementAttributePreservation extends Plugin {
...
+                    return modelWriter.createElement( elements[key], attributes );
+                },
+            } );
+
+            editor.conversion.for( 'downcast' ).elementToElement( {
+                model: elements[key], 
+                view: ( modelElement, viewWriter ) => {
+                    let attributes = filterAllowedAttributes( modelElement.getAttributes() );
+                    if ( !attributes.get( 'id' ) ) {
+                        attributes.set( 'id', this._nextId() );
```
jessouyangbraven, line 108: @rshipp helped me figure this out. To use `retaineddata`, we need both an upcast and downcast defined that calls `this._nextId()`. With only the upcast, I was always getting `id=1` whenever I toggled between Design/Code, rather than a monotonically increasing identifier. 

To quote @rshipp 
`yeah. which makes sense, because the upcast creates the model element, sets its attributes, which fires off an attribute:change event, which is picked up by consumeAttributeEvent`

So the upcast would call `this._nextId()` before `consumeAttributeEvent` ran to properly initialize the value.

app/javascript/ckeditor/retaineddata.js
```diff
@@ -55,7 +55,17 @@ export default class RetainedData extends Plugin {
...
+                const value = elem.getAttribute( 'id' );
+                // Assume if it's an integer value, we could have assigned it
+                if ( !isNaN( value ) ) {
+                    this._idCounter = Math.max( parseInt(value) + 1, this._idCounter );
+                }
+                break;
```
rshipp, line 65: i think retained data needs to be first, then id, otherwise this will `break` on an input that has both a numeric id and a retained data id. If the retained data id was higher than the numeric id, this would miss it.

app/javascript/ckeditor/customelementattributepreservation.js
```diff
@@ -48,6 +49,159 @@ export default class CustomElementAttributePreservation extends Plugin {
...
+            'h2': 'heading1',
+            'h3': 'heading2',
+            'h4': 'heading3',
+            'h5': 'heading4',
+            'h6': 'heading5',
+        }
+
+        Object.keys( elements ).forEach( ( key ) => {
+            // View-to-model converter converting a view element with all its attributes to the model.
+            editor.conversion.for( 'upcast' ).elementToElement( {
```
rshipp, line 91: this being in a different upcast makes me nervous, but it's probably the right call to keep it and refactor later. this whole file is tech debt.

app/javascript/ckeditor/retaineddata.js
rshipp: wait... that's not how this works at all, it doesn't look over attributes. disregard.

jessouyangbraven: Flipping the order, I think the same applies if `id > retained-data-id`. I think what we want is to take the max of both (checking that they're valid) and `this._idCounter` so we ensure there's no collision. 

The way that made most sense for me was to hoist all the attribute stuff out of the loop because I'm pretty sure it's not necessary. 

But now that I think about it, I can put the loop back in and keep tabs on a running max and set it at the end. 

I can do this if you think it's less risky.

I did some manual testing by setting the retained data/IDs to ridiculously high numbers and making sure it worked, but I honestly have no idea to test thoroughly with what's in booster modules now.

rshipp: To test, maybe take one of the booster modules, find-and-replace a bunch (but not all) of its data-bz-retained attributes, load that back into the editor, and check that
1. the resulting IDs it populates are all unique, and
2. the new IDs it adds start at a number higher than the highest existing ID

that's probably enough to make sure this still works? if we can get rid of a loop that runs on every attribute change, that seems like a positive.

jessouyangbraven: Thanks! I'll try that out, alternatively, the loop with running max would be: https://github.com/bebraven/platform/commit/c9484dca356f238d259184ab48324e33a6b2fa2c

app/javascript/ckeditor/customelementattributepreservation.js
jessouyangbraven: Tested with Booster Module #2 and this gives `id`s to existing `<h*>` elements and they're larger than the existing `data-bz-retained` IDs. 

If I delete some of the existing `data-bz-retained`, they get reassigned, also at a larger value. 

I noticed while testing that it's really hard to guarantee that `_consumeAttributeEvent()` has run for all elements in the document before some other element accesses `this._idCounter`. This means that without reasoning a lot about downcast/upcast/event handling, it's hard to ensure the element's getting a properly initialized value. :(

The fix for the headings was to not assign `id` in `upcast` and only do it in `downcast`, which works for these non-designer edited attributes, but won't work for things we need to modify in the `Design` tab.

PR #238, opened 2020-06-02
Update rowan_bot gem revision to get latest changes.

sadleb:


sadleb:
foregoing a review. just a revision update so I can get the docusign stuff live to prod.

PR #240, opened 2020-06-02
Fix Course Copy on the Portal side of things.

sadleb:
This is dumb, but when we do course copy or content import in the Portal,
it strips out the leading comments. Nested comments inside the div work.
The JS doesn't care where this token is.

See: https://app.asana.com/0/1170770467635599/1178021488868222

### Test Plan:
- Push content to Portal. Copy the course with the content and make sure in the
  copied course the BRAVEN_NEW_HTML token is in the HTML and the new JS and CSS
  runs/is applied.

### DIFF
--- a/lib/canvas_api.rb
+++ b/lib/canvas_api.rb
@@ -9,8 +9,8 @@ class CanvasAPI
   # We also wrap the entire contents in a "bz-module" div, so the CSS selectors
   # work as expected.
   PrependHTML = %q(
-    <!-- BRAVEN_NEW_HTML -->
     <div class="bz-module">
+    <!-- BRAVEN_NEW_HTML -->
   )
 
   # Custom HTML to append to each body.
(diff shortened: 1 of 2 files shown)


sadleb:
@rshipp what should we do after this to make sure the BRAVEN_NEW_HTML token is moved in all release content and won't get overwritten by the old location?

rshipp:
Not sure if I'm interpreting the question correctly, but as soon as this change is deployed, clicking Publish from the content editor will move the comment to the new location.

sadleb:
> Not sure if I'm interpreting the question correctly, but as soon as this change is deployed, clicking Publish from the content editor will move the comment to the new location.

@rshipp so if we just go through all live modules and re-publish them, that's all we need to do? There isn't any HTML stored in src ctrl somewhere that has it which we need to update as well?

rshipp:
There are only 2 live modules, the Booster ones, but yes that's correct.

PR #260, opened 2020-06-24
Ensure that all new sign creates go through canvas setup

skaex:
This ensures that all new users go through canvas setup. There was an issue where returning LCs will no be placed in the right sections. This solves it by:
- [x] Ensuring returning new user sign ups go through the canvas setup again.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -128,19 +128,8 @@ class User < ApplicationRecord
   def setup_canvas_access
     return if canvas_id
 
-    Rails.logger.info("Creating Canvas account and enrollments for new user: #{inspect}")
+    Rails.logger.info("Setting up Canvas account and enrollments for user: #{inspect}")
     self.canvas_id = SyncToLMS.new.for_contact(salesforce_id)
-
-    # Stopping this short circuit so returning user can go through the canvas
-    # process again
-    #
-    # existing_user = CanvasAPI.client.find_user_in_canvas(email)
-    # if existing_user
-    #   self.canvas_id = existing_user['id']
-    # else
-    #   Rails.logger.info("Creating Canvas account and enrollments for new user: #{inspect}")
-    #   self.canvas_id = SyncToLMS.new.for_contact(salesforce_id)
-    # end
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -127,13 +127,20 @@ def sync_salesforce_info
...
-      self.canvas_id = existing_user['id']
-    else
-      Rails.logger.info("Creating Canvas account and enrollments for new user: #{inspect}")
-      self.canvas_id = SyncToLMS.new.for_contact(salesforce_id)
-    end
+
+    Rails.logger.info("Creating Canvas account and enrollments for new user: #{inspect}")
+    self.canvas_id = SyncToLMS.new.for_contact(salesforce_id)
+
+    # Stopping this short circuit so returning user can go through the canvas
```
sadleb, line 134: I would just remove this. I don’t see a reason to leave the commenter our code

app/models/user.rb
```diff
@@ -127,13 +127,20 @@ def sync_salesforce_info
...
     return if canvas_id
-    existing_user = CanvasAPI.client.find_user_in_canvas(email)
-    if existing_user
-      self.canvas_id = existing_user['id']
-    else
-      Rails.logger.info("Creating Canvas account and enrollments for new user: #{inspect}")
-      self.canvas_id = SyncToLMS.new.for_contact(salesforce_id)
-    end
+
+    Rails.logger.info("Creating Canvas account and enrollments for new user: #{inspect}")
```
sadleb, line 131: Can we make this log more accurate? Maybe just remove the word “new” since we don’t know.

spec/models/user_spec.rb
```diff
@@ -76,7 +76,8 @@
       end   
     end
 
-    context 'when canvas user exists' do
+    # Skipping this because this is not valid anymore.
+    xcontext 'when canvas user exists' do
```
sadleb, line 80: If we already have a test of whether en existing user is added to the correct new course and section, let’s remove this. If not, we should change it to test that

PR #261, opened 2020-06-24
Implement a Public JWK Url endpoint 

sadleb:
Adds an endpoint to expose a public Json Web Key (JWK).
    
When implementing a JWK authentication flow (e.g. for LTI Extension),
you need to expose a public JWK that the external service will use to encode
payloads. You need to keep the private key and be able to decode those payloads
(e.g. a JWT id_token) using it. This commit adds the endpoint to expose the public JWK
along with the ability to store and decode payloads using the private key.

See:
-  https://www.imsglobal.org/spec/lti/v1p3#json-web-token-0
- Step 3 here: https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#launch-overview

    
### Test Plan:
    - In the Rails console run Keypair.create! and make sure that when you go to
      https://platformweb/public_jwk that you can see the public JWK for it.
    - Encode a payload using that pulic JWK and make sure you can decode it with
      Keypair.jwt_decode(encoded_payload)
    - Run: `bundle exec rspec spec/modles/keypair_spec.rb`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -107,3 +107,8 @@ gem 'platform-api', require: false
 # gem 'rowan_bot', path: '/app/rowan_bot'
 gem 'rowan_bot', github: 'bebraven/rowan_bot', branch: 'master'
 
+# Implementation of JSON Web Token (JWT) standard: https://github.com/jwt/ruby-jwt
+gem 'jwt'
+
+# Generates attr_accessors that encrypt and decrypt attributes transparently in the database
+gem 'attr_encrypted'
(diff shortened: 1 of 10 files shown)


rshipp:
does this ever clear out old keys from the db?

sadleb:
> does this ever clear out old keys from the db?

This PR itself has no logic related to actually creating or deleting Keypairs. When we implement the Deep Linking flow or any other flow that needs to send encoded payloads to Canvas that's when we'll need to address that concern. E.g. in the hacked up prototype, here is where that call is: https://github.com/bebraven/platform/compare/lti_extension#diff-0de407d94dbd99d1a6fef3a27d42f6a6R169

sadleb:
I need to get another PR that depends on this up so imma go ahead and merge.

PR #265, opened 2020-06-26
Implement the LTI Launch flow.

sadleb:
This implements the flow described here:
https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#launch-overview

### Test Plan:
- Configure your Developer Key with the Redirect URI that exactly matches
  `https://ENV['APPLICATION_HOST']/lti/launch`
- Do an LTI Launch, e.g. if your LTI has a Link Selection placement,
  go to a Module, click the + to add something, select External Tool,
  then click on the name of your LTI. It should start the launch flow and
  successfully load the Target URI that you configured in the Developer Key
  for the Link Selection placement. E.g. https://platformweb/whatever/endpoint/i/implement in the image below
image
- In future requests, you can access the metadata about the current LTI Launch (e.g. user_id, etc) using: `LtiLaunch.current(session[:session_id])`
- The following specs should pass:
    `bundle exec rspec spec/models/lti_launch_spec.rb`
    `bundle exec rspec spec/controllers/lti_launch_controller_spec.rb `

### DIFF
--- a/app/controllers/lti_launch_controller.rb
+++ b/app/controllers/lti_launch_controller.rb
@@ -47,7 +47,7 @@ class LtiLaunchController < ApplicationController
     idt = LtiIdToken.parse_and_verify(params[:id_token])
 
     @lti_launch = LtiLaunch.current(params[:state])
-    @lti_launch.id_token_payload = idt.payload
+    @lti_launch.id_token_payload = idt.payload.to_json
     @lti_launch.save!
 
     # Step 4 in the flow, show the target resource now that we've saved the id_token payload that contains
--- a/app/controllers/lti_launch_controller.rb
+++ b/app/controllers/lti_launch_controller.rb
@@ -1,0 +1,54 @@
+require 'lti_id_token'
+
+# Implement the LTI 1.3 Launch Flow. Here is a nice summary of that flow by Canvas, our LMS:
+# https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#launch-overview
+#
+# Summary of flow:
+# Step 1: Login Initiation - Canvas calls into here letting us know a launch is starting
+# Step 2: Authentication Response - We initiate an handshake to authenticate the launch establish a trust in the launch
+# Step 3: LTI Launch - The launch is initated with meta-data about the context and the final target resource to launch
+# Step 4: Resource Display - We display the target resource
+class LtiLaunchController < ApplicationController
+  skip_before_action :authenticate_user!
+  skip_before_action :ensure_admin!
+  skip_before_action :verify_authenticity_token
+
+  # Non-standard controller without normal CRUD methods. Disable the convenience module.
+  def dry_crud_enabled?() false end
+
+  # POST /lti/login
+  #
+  # This is the OIDC url Canvas is calling in Step 1 above.
+  def login
+    raise ActionController::BadRequest.new(), "Unexpected iss parameter: #{params[:iss]}" if params[:iss] != ENV['CANVAS_LTI_URL']
+
+    # Force load the session by writing something to it. Reading from it isn't enough to initialize it and get a session_id
+    session[:init] = true
+
+    @lti_launch = LtiLaunch.create!(params.except(:iss, :canvas_region).permit(:client_id, :login_hint, :target_link_uri, :lti_message_hint)
+                                         .merge(:state => session[:session_id], :nonce => SecureRandom.hex(10)))
+
+    # Step 2 in the flow, do the handshake
+    redirect_to "#{ENV['CANVAS_LTI_URL']}/api/lti/authorize_redirect?#{@lti_launch.auth_params.to_query}"
+  end
+
+
+  # POST /lti/launch
+  #
+  # This is the endpoint configured as the Redirect URI in the Canvas Developer Key. It's Step 3 in the launch flow.
+  def launch
+    params.require([:id_token, :state])
+
+    # This also verifies the request is coming from Canvas using the public JWK 
+    # as described in Step 3 here: https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html
+    idt = LtiIdToken.parse_and_verify(params[:id_token])
+
+    @lti_launch = LtiLaunch.current(params[:state])
+    @lti_launch.id_token_payload = idt.payload
+    @lti_launch.save!
+
+    # Step 4 in the flow, show the target resource now that we've saved the id_token payload that contains
+    # user identifiers, course contextual data, custom data, etc.
+    redirect_to @lti_launch.target_link_uri
+  end
+end
--- a/app/controllers/lti_launch_controller.rb
+++ b/app/controllers/lti_launch_controller.rb
@@ -16,11 +16,14 @@ class LtiLaunchController < ApplicationController
   # Non-standard controller without normal CRUD methods. Disable the convenience module.
   def dry_crud_enabled?() false end
 
+  LTI_ISS=Rails.application.secrets.lti_oidc_base_uri
+  LTI_AUTH_RESPONSE_URL="#{Rails.application.secrets.lti_oidc_base_uri}/api/lti/authorize_redirect".freeze
+
   # POST /lti/login
   #
   # This is the OIDC url Canvas is calling in Step 1 above.
   def login
-    raise ActionController::BadRequest.new(), "Unexpected iss parameter: #{params[:iss]}" if params[:iss] != ENV['CANVAS_LTI_URL']
+    raise ActionController::BadRequest.new(), "Unexpected iss parameter: #{params[:iss]}" if params[:iss] != LTI_ISS
 
     # Force load the session by writing something to it. Reading from it isn't enough to initialize it and get a session_id
     session[:init] = true
@@ -29,7 +32,7 @@ class LtiLaunchController < ApplicationController
                                          .merge(:state => session[:session_id], :nonce => SecureRandom.hex(10)))
 
     # Step 2 in the flow, do the handshake
-    redirect_to "#{ENV['CANVAS_LTI_URL']}/api/lti/authorize_redirect?#{@lti_launch.auth_params.to_query}"
+    redirect_to "#{LTI_AUTH_RESPONSE_URL}?#{@lti_launch.auth_params.to_query}"
   end
 
 
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -10,6 +10,8 @@ development:
   devise_pepper: <%= ENV['DEVISE_PEPPER'] %>
   honeycomb_dataset: <%= ENV['HONEYCOMB_DATASET'] %>
   honeycomb_write_key: <%= ENV['HONEYCOMB_WRITE_KEY'] %>
+  # This value comes from here: https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#launch-overview
+  lti_oidc_base_uri: <%= ENV.fetch('LTI_OIDC_BASE_URI') { 'https://canvas.instructure.com' } %>
   salesforce_consumer_key: <%= ENV['SALESFORCE_PLATFORM_CONSUMER_KEY'] %>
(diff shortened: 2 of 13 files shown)


sadleb:
I'm going to add some tests before I open the PR for real.

app/controllers/lti_launch_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  # This is the endpoint configured as the Redirect URI in the Canvas Developer Key. It's Step 3 in the launch flow.
+  def launch
+    params.require([:id_token, :state])
+
+    # This also verifies the request is coming from Canvas using the public JWK 
+    # as described in Step 3 here: https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html
+    idt = LtiIdToken.parse_and_verify(params[:id_token])
+
+    @lti_launch = LtiLaunch.current(params[:state])
+    @lti_launch.id_token_payload = idt.payload
```
rshipp, line 50: This implicitly coerces `idt.payload` from a `Hash` into a `String` (because `id_token_payload` is a `text` field), which makes reading from it hard. Should we JSONify it first?

app/controllers/lti_launch_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    @lti_launch.save!
+
+    # Step 4 in the flow, show the target resource now that we've saved the id_token payload that contains
+    # user identifiers, course contextual data, custom data, etc.
+    redirect_to @lti_launch.target_link_uri
```
rshipp, line 55: `target_link_uri` can be different from `idt.payload["https://purl.imsglobal.org/spec/lti/claim/target_link_uri"]`, but I haven't found a case where the latter is wrong. The former is wrong when you try to view an assignment added through the assignment placement flow. Do you know what the difference is? [This](https://github.com/bebraven/platform/compare/lti_extension#diff-0de407d94dbd99d1a6fef3a27d42f6a6R101) prefers the latter, and falls back to the former.

app/controllers/lti_launch_controller.rb
sadleb: Yeah, that's what I was talking about yesterday. I made that change when implementing the specs.

app/controllers/lti_launch_controller.rb
sadleb: I don't know what the difference is. Let's just use what is in the payload and if we run into a case where that doesn't exist, then we'll add the fallback.

spec/api_helper.rb
```diff
@@ -6,7 +6,12 @@ def initialize
   delegate :association, to: :@strategy
 
   def result(evaluation)
-    @strategy.result(evaluation).to_json
+    result = @strategy.result(evaluation)
+    evaluation.notify(:before_json, result)
```
sadleb, line 10: This stuff is taken from [here](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-callbacks) so that I can add keys to the hash/json that aren't valid ruby class attributes. See here: https://github.com/bebraven/platform/pull/265/files#diff-489c08e04d2e1afd30720e0a582669cdR22

app/controllers/lti_launch_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+# Implement the LTI 1.3 Launch Flow. Here is a nice summary of that flow by Canvas, our LMS:
+# https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#launch-overview
+#
+# Summary of flow:
+# Step 1: Login Initiation - Canvas calls into here letting us know a launch is starting
+# Step 2: Authentication Response - We initiate an handshake to authenticate the launch establish a trust in the launch
+# Step 3: LTI Launch - The launch is initated with meta-data about the context and the final target resource to launch
+# Step 4: Resource Display - We display the target resource
+class LtiLaunchController < ApplicationController
+  skip_before_action :authenticate_user!
```
sadleb, line 12: We skip this because the authentication is essentially the handshake b/n us and Canvas. This implies that we'll need to come up with a way to verify that other end-points have been authenticated through this flow. E.g. there is a valid `LtiLaunch` with an `id_token_payload` for the session.

app/controllers/lti_launch_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+  LTI_AUTH_RESPONSE_URL="#{Rails.application.secrets.lti_oidc_base_uri}/api/lti/authorize_redirect".freeze
+
+  # POST /lti/login
+  #
+  # This is the OIDC url Canvas is calling in Step 1 above.
+  def login
+    raise ActionController::BadRequest.new(), "Unexpected iss parameter: #{params[:iss]}" if params[:iss] != LTI_ISS
+
+    # Force load the session by writing something to it. Reading from it isn't enough to initialize it and get a session_id
+    session[:init] = true
```
sadleb, line 29: I'm still not 100% sure why this is needed, I think it's because the session initialization stuff happens when you authenticate a user. I got this solution from [here](https://stackoverflow.com/questions/14665275/how-force-that-session-is-loaded/14684171#14684171).

config/secrets.yml
```diff
@@ -10,6 +10,8 @@ development:
   devise_pepper: <%= ENV['DEVISE_PEPPER'] %>
   honeycomb_dataset: <%= ENV['HONEYCOMB_DATASET'] %>
   honeycomb_write_key: <%= ENV['HONEYCOMB_WRITE_KEY'] %>
+  # This value comes from here: https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#launch-overview
+  lti_oidc_base_uri: <%= ENV.fetch('LTI_OIDC_BASE_URI') { 'https://canvas.instructure.com' } %>
```
sadleb, line 14: We don't currently use this ENV var. I put it in place in case we ever start using the Canvas test or beta environments which have different URLs this.

lib/lti_id_token.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  private
+
+  def self.jwks
+    @jwks ||= fetch_jwks # These are rotated monthly for Canvas and the server restarts once a day. Safe to cache.
+  end
+
+  def self.fetch_jwks
+    response = RestClient.get(PUBLIC_JWKS_URL)
+    JSON.parse(response.body, symbolize_names: true)
```
sadleb, line 40: If you ever have to do anything with the JWT gem, pay extra special close attention to the `symbolize_names` arg here. This led to several hours of wasted time b/c things fails deep in the gem's code if the key's are strings and not symbols.

spec/controllers/lti_launch_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      lti_launch = LtiLaunch.current(session_id)
+      expect(response).to redirect_to("#{LtiLaunchController::LTI_AUTH_RESPONSE_URL}?#{lti_launch.auth_params.to_query}")
+    end 
+  end
+
+  describe 'POST #launch' do
+    let!(:target_link_uri) { 'https://target/link' }
+    let!(:lti_launch) { create(:lti_launch_model, target_link_uri: target_link_uri, state: session_id) }
+    let!(:id_token_payload) { FactoryBot.json(:lti_link_launch_request) }
+    let!(:id_token) { Keypair.jwt_encode(JSON.parse(id_token_payload, symbolize_names: true)) }
```
sadleb, line 30: I made the exact same mistake here with `symbolize_names` and wasted another 2 hours. smitf

spec/controllers/lti_launch_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    let!(:lti_launch_params) { { :id_token => id_token, :state => session_id } }
+
+    before(:each) do
+      public_jwks = JSON.parse({ keys: [ Keypair.current.public_jwk_export ] }.to_json, symbolize_names: true)
```
sadleb, line 34: What we're doing here is essentially mimicking what Canvas would do to encode an `id_token`. I think it makes more sense to have this test a real JWT encoded id_token payload here vs creating a separate spec for the `LtiIdToken` class and stubbing that stuff out here. Lmk if you disagree though.

spec/factories/lti_launch.rb
```diff
@@ -0,0 +1,24 @@
+FactoryBot.define do
+  factory :lti_launch do
+    client_id { '160040000000055555' }
+    login_hint { '914c015dc6637dfb1518ef9c9e02a5918940f67d' }
+    target_link_uri { 'https://some/target/link/uri' }
+    lti_message_hint { 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ2ZXJpZmllciI6ImQ0OTZjNGMxMGQzMTc5ZGRhNTI4Yjg2MjM3YTY4ZDAyNjEzMDEzYmFkMDUwMGZhZDM5ZDY1N2E0MzZmZDMyMWJjYTBjMWJmYzVmNTY3OWI5ZTQyMDY0NTRhNzQ4ZjAyODM0NmRmNTVhY2VlM2FiOTFjZDA0ZmQwNzU2NGFlZDJkIiwiY2FudmFzX2RvbWFpbiI6ImJyYXZlbi5pbnN0cnVjdHVyZS5jb20iLCJjb250ZXh0X3R5cGUiOiJDb3Vyc2UiLCJjb250ZXh0X2lkIjoxNjAwNTAwMDAwMDAwMDAwNDAsImV4cCI6MTU5MzA4NzYyNX0.mUvaDSdCo-dh-PDCHhL6iDs5Py2hzPU1_bJcKQyWveM' }
+   
+    factory :lti_launch_login_params, class: Hash do
```
sadleb, line 8: These are the launch params that Canvas sends us.

spec/factories/lti_link_launch_request.rb
```diff
@@ -0,0 +1,119 @@
...
+      request_msg.merge!({
+        'https://purl.imsglobal.org/spec/lti/claim/message_type' => 'LtiResourceLinkRequest',
+        'https://purl.imsglobal.org/spec/lti/claim/version' => '1.3.0',
+        'https://purl.imsglobal.org/spec/lti/claim/deployment_id' => '59:id_of_lti_deployment',
+        'https://purl.imsglobal.org/spec/lti/claim/target_link_uri' => 'https://platformweb/some/target/uri/to/launch',
+        'https://purl.imsglobal.org/spec/lti/claim/resource_link' => {
+          'id' => 'id_of_the_resource_being_launched',
+          'description' => nil,
+           'title' => nil
+        },
```
sadleb, line 34: I tried doing this sort of thing in the `initialize_with` block, but it didn't play nicely with nested factories.

spec/factories/lti_link_launch_request.rb
```diff
@@ -0,0 +1,119 @@
...
+#
+# IMPORTANT: this is meant to be built with FactoryBot.json(:lti_link_launch_request)
+# and if you don't then it will be missing the json keys that are URLs
+
+  factory :lti_link_launch_request, class: Hash do
+    skip_create # This isn't stored in the DB.
+    iss { 'https://some.lti.platform' }
+    aud { '160040000000000055' }
+    azp { '160040000000000055' }
+    exp { Time.now.to_i + 50000 }
```
sadleb, line 16: Note that the controller specs will fail if the expiration isn't in the future. That's why it's nice to have this whole things as a factory instead of a fixture.

lib/lti_id_token.rb
```diff
@@ -0,0 +1,43 @@
...
+  rescue => e
+    Rails.logger.error("{\"Error\":\"#{e.message}\"}")
+    Rails.logger.error(e.response.body)
+    raise
+  end
+
+  private
+
+  def self.jwks
+    @jwks ||= fetch_jwks # These are rotated monthly for Canvas and the server restarts once a day. Safe to cache.
```
sadleb, line 35: TODO: remove this in case Canvas rotates the key and we don't refresh it until tomorrow. 1 day window?

PR #272, opened 2020-07-13
Update sync to lms

skaex:
### Description
This PR implements the following:

- Use a background job to run the sync.
- Send emails on success or failures.
- Reactivate SIS instead of throwing error.

### Motivation and Context

https://app.asana.com/0/1172786304330247/1172130267952718

### Types of changes

### Checklist:

### DIFF
--- a/app/controllers/salesforce_controller.rb
+++ b/app/controllers/salesforce_controller.rb
@@ -13,7 +13,7 @@ class SalesforceController < ApplicationController
     program_id = params[:program_id]
     email = params[:email]
     if program_id && email
-      SalesforceProgramToLmsSyncJob.perform_now(program_id, email)
+      SalesforceProgramToLmsSyncJob.perform_later(program_id, email)
       @user_notification = "Sync To LMS has been submitted for program #{program_id}, #{email} will be notified"
     else
       @user_notification = 'Please enter the Program ID to sync below and hit Enter'
--- a/app/controllers/salesforce_controller.rb
+++ b/app/controllers/salesforce_controller.rb
@@ -1,7 +1,7 @@
-require 'sync_to_lms'
+# frozen_string_literal: true
 
+# Salesforce controller
 class SalesforceController < ApplicationController
-
   def sync_to_lms
     # TODO: this is a complete hack just to get a prototype going. When we implement this for real,
     # it should be an API that Salesforce can call into (and we need some sort of auth / protection).
@@ -11,13 +11,12 @@ class SalesforceController < ApplicationController
     # If not, queue this up on the background and have a page listing recent runs and their status?
     # Send an email?
     program_id = params[:program_id]
-    if program_id
-      SyncToLMS.new.for_program(program_id) 
-      @user_notification = "Sync To LMS has been submitted for program #{program_id}"
+    email = params[:email]
+    if program_id && email
+      SalesforceProgramToLmsSyncJob.perform_now(program_id, email)
+      @user_notification = "Sync To LMS has been submitted for program #{program_id}, #{email} will be notified"
     else
-      @user_notification = "Please enter the Program ID to sync below and hit Enter"
+      @user_notification = 'Please enter the Program ID to sync below and hit Enter'
     end
-     
   end
-
 end
--- a/app/jobs/salesforce_program_to_lms_sync_job.rb
+++ b/app/jobs/salesforce_program_to_lms_sync_job.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# Salesforce program sync to lms job
+class SalesforceProgramToLmsSyncJob < ApplicationJob
+  queue_as :default
+
+  def perform(program_id)
+    SyncToLMS.new.for_program(program_id)
+  end
+end
--- a/app/jobs/salesforce_program_to_lms_sync_job.rb
+++ b/app/jobs/salesforce_program_to_lms_sync_job.rb
@@ -4,7 +4,12 @@
 class SalesforceProgramToLmsSyncJob < ApplicationJob
   queue_as :default
 
-  def perform(program_id)
+  def perform(program_id, email)
     SyncToLMS.new.for_program(program_id)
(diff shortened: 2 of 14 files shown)


app/controllers/salesforce_controller.rb
```diff
@@ -11,13 +11,12 @@ def sync_to_lms
     # If not, queue this up on the background and have a page listing recent runs and their status?
     # Send an email?
     program_id = params[:program_id]
-    if program_id
-      SyncToLMS.new.for_program(program_id) 
-      @user_notification = "Sync To LMS has been submitted for program #{program_id}"
+    email = params[:email]
+    if program_id && email
+      SalesforceProgramToLmsSyncJob.perform_later(program_id, email)
```
skaex, line 16: Moved sync process to a background job.

app/views/salesforce/_form.html.erb
```diff
@@ -4,6 +4,12 @@
     <%= form.text_field :program_id %>
   </div>
 
+  <div class="field">
```
skaex, line 7: Collected email to notify success or failure of sync job

lib/canvas_api.rb
```diff
@@ -84,7 +84,8 @@ def create_user(first_name, last_name, username, email, salesforce_id, student_i
         'communication_channel[confirmation_url]' => true,
          # Note: the old code used the Join user.id and not the SF id. But now the user account may not
          # be created yet when we're running Sync To LMS.
-        'pseudonym[sis_user_id]' => "BVSFID#{salesforce_id}-SISID#{student_id}" 
+        'pseudonym[sis_user_id]' => "BVSFID#{salesforce_id}-SISID#{student_id}",
+        'enable_sis_reactivation' => true
```
skaex, line 88: Canvas doesn't delete SIS so if any user gets deleted on portal and then later gets sync it'll error out saying the SIS already exist. This updates the creation process to reactivate the deleted user instead.

lib/sync_to_lms.rb
```diff
@@ -141,6 +142,7 @@ def complete_user_enrollment(canvas_user_id)
   #
   # Note: canvas_role = [:StudentEnrollment, :TaEnrollment, :DesignerEnrollment, :TeacherEnrollment]
   def sync_course_enrollment(canvas_user_id, course_id, canvas_role, section_name)
+    section_name ||= 'Default Section'
```
skaex, line 145: This handles for when the program has no cohort or cohort schedules. The section name will be nil and there will be duplicate sections on the portal.

lib/canvas_api.rb
sadleb: Oh nice, this is a good enhancement. I've run into this in the past.

app/controllers/salesforce_controller.rb
```diff
@@ -1,7 +1,7 @@
-require 'sync_to_lms'
+# frozen_string_literal: true
 
+# Salesforce controller
```
sadleb, line 3: This comment doesn't really add any value. I like the idea of commenting classes, but they should add information.

app/controllers/salesforce_controller.rb
```diff
@@ -11,13 +11,12 @@ def sync_to_lms
     # If not, queue this up on the background and have a page listing recent runs and their status?
     # Send an email?
```
sadleb, line 12: It looks like this PR makes this comment obsolete. Errors are emailed, correct?

app/controllers/salesforce_controller.rb
```diff
@@ -11,13 +11,12 @@ def sync_to_lms
...
-    if program_id
-      SyncToLMS.new.for_program(program_id) 
-      @user_notification = "Sync To LMS has been submitted for program #{program_id}"
+    email = params[:email]
+    if program_id && email
+      SalesforceProgramToLmsSyncJob.perform_later(program_id, email)
+      @user_notification = "Sync To LMS has been submitted for program #{program_id}, #{email} will be notified"
     else
-      @user_notification = "Please enter the Program ID to sync below and hit Enter"
+      @user_notification = 'Please enter the Program ID to sync below and hit Enter'
```
sadleb, line 19: This comment should say `Program ID and email` now.

app/jobs/salesforce_program_to_lms_sync_job.rb
```diff
@@ -0,0 +1,15 @@
...
+class SalesforceProgramToLmsSyncJob < ApplicationJob
+  queue_as :default
+
+  def perform(program_id, email)
+    SyncToLMS.new.for_program(program_id)
+    SalesforceToLmsSyncMailer.with(email: email).success_email.deliver_now
+  end
+
+  rescue_from(StandardError) do |_exception|
+    SalesforceToLmsSyncMailer.with(email: arguments.second).failure_email.deliver_now
```
sadleb, line 13: Since only staff run this, wouldn't it be useful to put the exception in the email for faster troubleshooting?

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,7 @@
+# frozen_string_literal: true
+
+# Base mailer class
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
+  default from: 'noreply@braven.org'
```
sadleb, line 5: There is a `Rails.application.secrets.mailer_from_email` value we should probably use. I don't even thing `noreply@braven.org` is valid yet. We bought that domain but haven't set up emails yet I don't think.

lib/sync_to_lms.rb
sadleb: This seems a bit risky to me. Is there a valid reason for having a user with no no cohort or cohort schedule? I think it's going to happen at times those execution folks mess up with setting up a new Program and end up accidentally sync'ing everyone into the `Default Section`. Should we just fail if no section is set? Maybe this is for staff or TAs? If so, I think we probably want to handle that on the Salesforce side and send the section name to use. Salesforce should be the source of truth.

PR #275, opened 2020-07-14
Add ENV vars and helper for the LRS launch query string.

sadleb:
The nested name and email param was causing me issues using the various
ways to convert a hash to a query string, so this is a bit hacked up
but I tested it using a real xApi package and LRS and it works.

Test Plan:
- Upload an Articule Rise360 xApi package to S3 or extract it into your public
  directory. Then hit it's index.html with the query string produced by this helper.
  You should see xApi statements sent to it when doing the lesson.


config/secrets.yml
```diff
@@ -36,7 +38,9 @@ test:
   devise_pepper: <%= ENV['DEVISE_PEPPER'] %>
   honeycomb_dataset: <%= ENV['HONEYCOMB_DATASET'] %>
   honeycomb_write_key: <%= ENV['HONEYCOMB_WRITE_KEY'] %>
-  lti_oidc_base_uri: <%= ENV.fetch('LTI_OIDC_BASE_URI') { 'https://example.lms.com' } %>
+  lrs_uri: 'https://example.lrs.com/path/to/xapi'
+  lrs_auth_token: 'some_basic_auth_token_for_lrs_xapi_statements'
+  lti_oidc_base_uri: 'https://example.lms.com'
```
rshipp, line 43: is this intentional?

sadleb: yeah, we don't want the test suite to rely on actual ENV vars and potentially break if they are changed. right?

PR #277, opened 2020-07-15
Add assignment xAPI code

rshipp:
Task: https://app.asana.com/0/1174274412967132/1183339952232786/f

This is mostly the 'xapi' JS pack, and some basic selenium tests. Everything else is support to get the tests to work correctly.

Things to look for:

* What other tests should we add?
* What should we do with JS `TODO`s? Ignore for now?
* Should we use retained-data attributes, or something else?
* Do we want different data in the xAPI payload?

Follow-up tasks:
* Fix and unmark as pending the CAS specs that don't work without a real staging server configured as the SSO URL: https://app.asana.com/0/1174274412967132/1166260321465780


config/secrets.yml
```diff
@@ -38,8 +38,8 @@ test:
   devise_pepper: <%= ENV['DEVISE_PEPPER'] %>
   honeycomb_dataset: <%= ENV['HONEYCOMB_DATASET'] %>
   honeycomb_write_key: <%= ENV['HONEYCOMB_WRITE_KEY'] %>
-  lrs_uri: 'https://example.lrs.com/path/to/xapi'
-  lrs_auth_token: 'some_basic_auth_token_for_lrs_xapi_statements'
+  lrs_uri: <%= ENV['LRS_URI'] %>
+  lrs_auth_token: <%= ENV['LRS_AUTH_TOKEN'] %>
```
rshipp, line 42: We need these to be the real values if we want to test against the real LRS.

spec/external_integration/xapi/assignment_xapi_spec.rb
```diff
@@ -0,0 +1,68 @@
+require "rails_helper"
```
rshipp, line 1: note the folder name - might want to discuss where e2e tests like this should go. they're flaky, which is bad for a CI, and they're slow, which is bad for running locally.

config/secrets.yml
sadleb: After sleeping on it, i realized you're right. No tests should rely on a hard-coded value anyway. They should read the current ENV and use that to test against.

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
...
+} catch (e) {
+    console.log("Failed to setup LRS object: ", e);
+    // TODO: do something with error, can't communicate with LRS
+}
+
+function sendStatement(e) {
+    const input = e.target;
+    const text = input.value;
+    const id = input.attributes['data-bz-retained'].value;
+    const full_id = `${window.location.origin}${window.location.pathname}#/${id}`;
```
rshipp, line 23: i don't like this variable name

config/secrets.yml
sadleb: I think you should add a fallback though instead of requiring it to be set. E.g. `ENV.fetch('LRS_URI') { 'https://example.lrs.com/path/to/xapi'' }`

spec/external_integration/xapi/assignment_xapi_spec.rb
sadleb: I think these should go in the `spec/feature` dir?

config/secrets.yml
rshipp: added. this e2e test will still error out if that variable isn't a real LRS though.

spec/external_integration/xapi/assignment_xapi_spec.rb
rshipp: I don't think feature specs are intended to access external resources like this. Our other ones just use VCR/webmock. Relying on a real network like this is way more flaky, I'm not sure we even want the CI to run these. We might need something else that runs periodically and reports.

config/initializers/content_security_policy.rb
```diff
@@ -16,9 +16,13 @@
 
   # Customizable through .env
   # Allow plain http, webpack-dev-server, and livereload connections in development.
-  policy.connect_src :self, :https, "http://localhost:*", "ws://localhost:3035", "http://#{Rails.application.secrets.application_host}:*" if Rails.env.development?
+  policy.connect_src :self, :https, "http://localhost:*", "ws://localhost:*",
```
jessouyangbraven, line 19: Is this the only change needed to fix live reloading? If it is, can we pull this into a PR to review now and merge into staging? It'd be nice to get this working again. Ping me and I'll review ASAP!

app/helpers/lti_link_selection_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+# a Learners Record Store (LRS).
+# 
+# Note: The endpoint for the LRS is set to a the LRS_PATH on this server and 
+# requests are proxied through to the actual LRS using LrsXapiProxy
+#
+# See:
+# - https://articulate.com/support/article/Implementing-Tin-Can-API-to-Support-Articulate-Content#launching-public-content
+# - https://xapi.com/try-developer/
+# - https://learningpool.com/how-to-launch-elearning-content-using-xapi/
+module LtiLinkSelectionHelper
```
jessouyangbraven, line 11: I deleted `LtiLinkSelectionController` and all of its associated files, since we're just hitting the `LessonContentsController` endpoint for now. Looking around the existing `helpers/` directory, does it make sense for this to live in `helpers/lti_helper`?

app/controllers/lrs_xapi_proxy_controller.rb
```diff
@@ -0,0 +1,14 @@
+require "lrs_xapi_proxy"
+
+class LrsXapiProxyController < ApplicationController
+  skip_before_action :verify_authenticity_token
```
jessouyangbraven, line 4: Does this go away when we do: https://app.asana.com/0/1174274412967132/1184057808812010

If so, can we add a `TODO` with a link to the task so we can keep track?

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
...
+    });
+} catch (e) {
+    console.log("Failed to setup LRS object: ", e);
+    // TODO: do something with error, can't communicate with LRS
+}
+
+function sendStatement(e) {
+    const input = e.target;
+    const text = input.value;
+    const id = input.attributes['data-bz-retained'].value;
```
jessouyangbraven, line 22: I'm wondering if we can stop using retained data IDs entirely here. Can we use the HTML element's `id` attribute instead? They'll be unique per document and we prepend them with `window.location` information below. The pros: not relying on retained data. The cons: having to rewrite existing assignments to have IDs assigned.

app/javascript/packs/xapi.js
jessouyangbraven: Just to make sure I understand: we're adding the `window.location` data so we know which assignment/module we're in? 

'id_with_window_location` to be more descriptive when we see it below in the code? 

Also, just a heads I think I did something like this when we were prototyping Rise 360 + LRS and had a lot of trouble with getting this information out of the lesson because of cross-origin errors. We can deal with it when we do the integration with lessons later, though.

app/controllers/lrs_xapi_proxy_controller.rb
rshipp: No, this is so that you don't need to send a CSRF token to this endpoint. Since this is an API endpoint, not a regular controller, I don't see a need for that token - and requiring it would mean we'd have to strip it out before forwarding to the LRS. You do still have to be logged in to Platform to make requests against this endpoint, but I think that's the only protection we want/need.

app/javascript/packs/xapi.js
rshipp: In 360, the "name"/"description" fields would be the human-readable question name / description. For assignments, there is no such concept, because we have textareas in tables or whatever arbitrary configuration. There's no way to resolve a given textarea to a human-readable question. So, I just set name/description to the retained data ID. The `full_id` here follows the same pattern 360 uses of something that looks like a URL, followed by a `/#` and a unique ID. That's the only reason I'm using window.location here, for consistency across 360/assignments. The `full_id` becomes the object ID (so maybe `object_id` is a better name?), but I'm not sure what we'd ever do with that - maybe nothing. If we ever wanted to retrieve the info a student submitted for this, we'd query it by the statement ID, which we don't currently persist.

app/javascript/packs/xapi.js
rshipp: cc @sadleb bc this is a bigger decision.

I'm on board with using a different attribute name. I don't think `id` is the way to go - I'd rather have an explicit `data-object-id` or something that we only use for this purpose. Reasons:

* Reusing `id` feels "magic", and non-obvious.
* We do sometimes have multiple elements with the same data-bz-retained ID, when we want the same answer to show up in multiple places. That'd be impossible if we use ID, because they must be unique per-page.

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
...
+                        // TODO: do something with error, didn't save statement
+                        return;
+                    }
+
+                    console.log("Failed to save statement: " + err);
+                    // TODO: do something with error, didn't save statement
+                    return;
+                }
+
+                input.setAttribute('data-xapi-statement-id', statement.id);
```
rshipp, line 79: eventually, we'll need to save this statement ID in our db, to use for authorization - so people can only access the statements they created.

app/javascript/packs/xapi.js
sadleb: I think as we get into stripping down the Content Editor to only create assignments, we can revisit this. For now, since we're using the existing content editor I think this makes sense. I queued up a backlog task here: https://app.asana.com/0/1174274412967132/1185569091008473

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
+import tincan from 'tincanjs';
+
+// Connect to the LRS.
+var lrs;
+
+try {
+    lrs = new tincan.LRS({
+        // Note: auth is intentionally left blank, we set it server-side.
+        endpoint: `${window.location.origin}/data/xAPI`,
+        username: "",
```
sadleb, line 10: Should we put placeholders in there so that if somehow things get messed up and the server side stuff breaks, it will be clearer where it's coming from and what the intention is? E.g. `username: "LRS_USERNAME_TOKEN"` or `username: "USERNAME_TO_REPLACE"`

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
...
+        password: "",
+        allowFail: false
+    });
+} catch (e) {
+    console.log("Failed to setup LRS object: ", e);
+    // TODO: do something with error, can't communicate with LRS
```
sadleb, line 16: Ideally we send to honeycomb, but that's a whole other project. Maybe just mention that in the TODO?

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
+import tincan from 'tincanjs';
```
sadleb, line 1: Can we rename `app/javascript/packs/xapi.js` to `app/javascript/packs/xapi_assignment.js` or something since it's specific to an assignment and looks for `retained-data` (or whatever we eventually replace it with)?

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
...
+    // If the input is empty, return early.
+    if (!text) {
+        return;
+    }
+
+    // Form the statement.
+    var statement = new tincan.Statement({
+        actor: {
+            // Note: actor info is intentionally left blank, we set it server-side.
+            name: "",
```
sadleb, line 34: Same comment. Should we use placeholders to make it more obvious if failures happen somewhere in the stack?

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
...
+
+    // Send to LRS.
+    lrs.saveStatement(
+        statement,
+        {
+            callback: function (err, xhr) {
+                if (err !== null) {
+                    if (xhr !== null) {
+                        console.log("Failed to save statement: " + xhr.responseText + " (" + xhr.status + ")");
+                        // TODO: do something with error, didn't save statement
```
sadleb, line 70: Same as above. Just mention Honeycomb?

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
...
+                        return;
+                    }
+
+                    console.log("Failed to save statement: " + err);
+                    // TODO: do something with error, didn't save statement
```
sadleb, line 75: Ditto

app/javascript/packs/xapi.js
```diff
@@ -0,0 +1,93 @@
...
+                    return;
+                }
+
+                input.setAttribute('data-xapi-statement-id', statement.id);
+                console.log("Statement saved");
+                // TODO: do something with success (possibly ignore)
```
sadleb, line 81: I think we should ignore success. I don't see a reason to fill up the console log with this.

sadleb:
As for the question "Do we want different data in the xAPI payload?" I think we should cross that bridge once we know how we're going to consume this data. This is just the first step to be able to get it flowing. Very soon we'll want to close the loop on the overall flow by taking the submitted assignment and allowing TAs to see it somewhow.

app/javascript/packs/xapi.js
rshipp: sending to honeycomb makes sense, but I think we'll also want some in-browser feedback for some or all of these eventually. If a page just stops sending responses, students will want to know so they don't lose all their work.

app/javascript/packs/xapi.js
sadleb: Task for this: https://app.asana.com/0/1174274412967132/1185569091008475

spec/feature/lrs_xapi_proxy/assignment_xapi_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+include ERB::Util
+include Rack::Utils
+
+unless ENV['BZ_AUTH_SERVER'] # Only run these specs if on a server with local database authentication enabled
+
+RSpec.describe CourseContentsController, type: :feature do
+  let!(:valid_user) { create(:admin_user) }
+  let!(:assignment) { create(:course_content_assignment) }
+  let(:vcr_variables) {{ platform_server: "#{ENV['VCR_SSO_SERVER']}", user_email: valid_user.email }}
```
sadleb: I don't really understand what `VCR_SSO_SERVER` is. Do we set this differently in different envs?

sadleb: If so, I think you should add what the value in the dev env should be to `.env.example`

rshipp: Yes, but this is removed now, so it doesn't matter anymore.

app/javascript/packs/xapi.js
rshipp: followup https://app.asana.com/0/1174274412967132/1185571851360169/f

PR #278, opened 2020-07-15
Add an endpoint that acts as a reverse proxy for the LRS.

sadleb:
This allows us to hide the authentication token for the LRS.

#### Test Plan:
- A url like the following should successfully send xApi events to the LRS
  assuming you configure LRS_BASE_URI and LRS_AUTH_TOKEN properly and drop
  an unzipped Rise360 xApi package at this path: 03-unlock-your-hustle/index.html

https://platformweb/03-unlock-your-hustle/index.html?endpoint=https://platformweb/data/xAPI&actor=%7B%22name%22:%20%5B%22briantestS3_proxied3%22%5D,%20%22mbox%22:%20%5B%22mailto:brian.test.proxied3@bebraven.org%22%5D%7D


lib/lrs_xapi_proxy.rb
```diff
@@ -0,0 +1,41 @@
...
+    request = Rack::Request.new(env)
+
+    # E.g. '/data/xAPI'
+    if request.path =~ path_to_proxy_regex
+      new_uri = target_uri_for(request.fullpath)
+      Rails.logger.debug("  proxying request to #{new_uri}")
+      backend = new_uri
+      env["HTTP_HOST"] = backend.host
+      env['PATH_INFO'] = new_uri.path
+      env['HTTP_AUTHORIZATION'] = authentication_header
```
sadleb, line 19: This is the meat of this PR.

config/secrets.yml
```diff
@@ -10,7 +10,8 @@ development:
   devise_pepper: <%= ENV['DEVISE_PEPPER'] %>
   honeycomb_dataset: <%= ENV['HONEYCOMB_DATASET'] %>
   honeycomb_write_key: <%= ENV['HONEYCOMB_WRITE_KEY'] %>
-  lrs_uri: <%= ENV['LRS_URI'] %>
+  lrs_base_uri: <%= ENV['LRS_BASE_URI'] %>
+  lrs_path: <%= ENV.fetch('LRS_PATH') { '/data/xAPI' } %>
```
sadleb, line 14: Having the path split out makes it easier to setup the proxy. Basically, look at the path on this server and it it's this send it to the same path at the LRS server.

app/helpers/lti_link_selection_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+# a Learners Record Store (LRS).
+# 
+# Note: The endpoint for the LRS is set to a the LRS_PATH on this server and 
+# requests are proxied through to the actual LRS using LrsXapiProxy
+#
+# See:
+# - https://articulate.com/support/article/Implementing-Tin-Can-API-to-Support-Articulate-Content#launching-public-content
+# - https://xapi.com/try-developer/
+# - https://learningpool.com/how-to-launch-elearning-content-using-xapi/
+module LtiLinkSelectionHelper
```
sadleb, line 11: I moved the `lib/lrs_launch_config.rb` here and renamed it. Once we merge this with the code for the `LtiLinkSelectionController` then this will be available there. Note that we may want to move this to be a helper for LessonContent instead though. I just couldn't have it in `lib` because I needed the `root_url` for this server.

config/secrets.yml
sadleb: I sort of did this in a hurry though. If you think it’s cleaner to have a single ENV var, i can go back to a single URL and parse out the path in the proxy class. I was struggling with so many other things but I’m actually leaning that way now that I can focus on this piece. Thoughts?

rshipp: I don't think it matters too much, but I might lean a little towards the single variable.

lib/lrs_xapi_proxy.rb
```diff
@@ -0,0 +1,41 @@
...
+  def authentication_header
+    @@auth_header ||= "Basic #{Rails.application.secrets.lrs_auth_token}"
+  end
+
+  def path_to_proxy_regex
+    @@ptp ||= /^#{Regexp.escape(Rails.application.secrets.lrs_path)}/
+  end
+
+  def target_uri_for(fullpath)
+    URI("#{Rails.application.secrets.lrs_base_uri}#{fullpath}")
```
rshipp, line 38: I think we might want to check this against a static allow-list instead, otherwise anyone could craft URLs that let them fetch any data from our LRS.

lib/lrs_xapi_proxy.rb
```diff
@@ -0,0 +1,41 @@
...
+    if request.path =~ path_to_proxy_regex
+      new_uri = target_uri_for(request.fullpath)
+      Rails.logger.debug("  proxying request to #{new_uri}")
+      backend = new_uri
+      env["HTTP_HOST"] = backend.host
+      env['PATH_INFO'] = new_uri.path
+      env['HTTP_AUTHORIZATION'] = authentication_header
+      env['HTTP_COOKIE'] = ''
+      env["content-length"] = nil
+
```
rshipp, line 22: we also need to modify the JSON in-transit to add user context, see https://github.com/bebraven/platform/blob/lrs-assignment-xapi/app/javascript/packs/xapi.js#L33

sadleb: @rshipp I was envisioning that we actually do put the `actor` param in the query string and let the Rise360 package handle that part. See https://github.com/bebraven/platform/pull/278/files#diff-874bb97b048c4b7d671a801c5cddc2ffR18

lib/lrs_xapi_proxy.rb
sadleb: Good point

rshipp:
@sadleb I added the new stuff, does this look like it'd work with the 360 stuff and whatever else we're doing? If so I'll do some refactoring and then rebase my stuff on this.

lib/lrs_xapi_proxy.rb
```diff
@@ -1,33 +1,55 @@
...
+      body = request.body.string
+
+      # Rewrite body.
+      begin
+        data = JSON.parse body
+        if data['actor']
+          data['actor'] = {
+            # TODO: this name stuff is bad, decide how we want to do this for real.
+            name: "#{user.first_name} #{user.last_name}",
+            mbox: user.email
```
rshipp, line 35: note: this will also override whatever's in the 360 query params. we might want to to take the actor stuff out of the query string so that's not confusing.

lib/lrs_xapi_proxy.rb
```diff
@@ -1,33 +1,55 @@
...
-     
-      super(env)      
-    else
-      @app.call(env)
+  def self.request(request, path, user)
+    if request.method == 'GET'
+      begin
+        response = RestClient.get("#{Rails.application.secrets.lrs_url}/#{path}", {
+          authorization: LrsXapiProxy.authentication_header,
+          x_experience_api_version: '1.0.3',
```
rshipp, line 18: can't figure out how to get the existing headers tincanjs might pass in (like x-experience-api-version) and merge them with the auth header - so I'm just ignoring any headers that might already be there. We'll have to do some testing to make sure this is ok.

lib/lrs_xapi_proxy.rb
```diff
@@ -1,33 +1,55 @@
...
+          params: request.query_parameters
+        })
+        response
+      rescue RestClient::NotFound => e
+        e.response
+      end
+    elsif request.method == 'POST'
+      body = request.body.string
```
rshipp, line 26: i need to move this down inside the exception handling code below, whoops

app/controllers/lrs_xapi_proxy_controller.rb
```diff
@@ -0,0 +1,14 @@
+require "lrs_xapi_proxy"
+
+class LrsXapiProxyController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  # Non-standard controller without normal CRUD methods. Disable the convenience module.
+  def dry_crud_enabled?() false end
+
+  def xAPI
+    response = LrsXapiProxy.request(request, request.params['endpoint'], current_user)
```
sadleb, line 10: Not sure if this is different b/c of the `match` route, but we should use strong params to only accept the params we expect.  Here is how the `endpoint` param get's set https://guides.rubyonrails.org/routing.html#route-globbing-and-wildcard-segments

lib/lrs_xapi_proxy.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+    elsif request.method == 'POST'
+      begin
+        # Rewrite body.
+        body = request.body.string
+        data = JSON.parse body
+        if data['actor']
+          data['actor'] = {
+            # TODO: this name stuff is bad, decide how we want to do this for real.
+            name: "#{user.first_name} #{user.last_name}",
```
sadleb, line 33: There is a `user.full_name` helper method on the model.

lib/lrs_xapi_proxy.rb
```diff
@@ -0,0 +1,83 @@
...
+            mbox: user.email
+          }
+        end
+        body = data.to_json
+      rescue JSON::ParserError, NoMethodError
+        # Body is not valid JSON, or is empty.
```
sadleb, line 39: Log and throw? Return an appropriate response?

lib/lrs_xapi_proxy.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      begin
+        response = RestClient.post("#{Rails.application.secrets.lrs_url}/#{path}", body, {
+          authorization: LrsXapiProxy.authentication_header,
+          x_experience_api_version: '1.0.3'
+        })
+
+        response
+      rescue RestClient::NotFound => e
+        e.response
```
sadleb, line 50: Should we log that this happened?

lib/lrs_xapi_proxy.rb
```diff
@@ -0,0 +1,83 @@
...
+        end
+        body = data.to_json
+      rescue JSON::ParserError, NoMethodError
+        # Body is not valid JSON, or is empty.
+      end
+
+      begin
+        response = RestClient.post("#{Rails.application.secrets.lrs_url}/#{path}", body, {
+          authorization: LrsXapiProxy.authentication_header,
+          x_experience_api_version: '1.0.3'
```
sadleb, line 45: Use a constant for the version?

config/routes.rb
```diff
@@ -89,4 +89,7 @@
 
   get '/lti/assignment_selection/new', to: 'lti_assignment_selection#new'     # https://canvas.instructure.com/doc/api/file.assignment_selection_placement.html
   post '/lti/assignment_selection', to: 'lti_assignment_selection#create'     # https://canvas.instructure.com/doc/api/file.assignment_selection_placement.html
+
+  # Proxy xAPI messages to the LRS.
+  match '/data/xAPI/*endpoint', to: 'lrs_xapi_proxy#xAPI', via: :all
```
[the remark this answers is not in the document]
rshipp: why these three? does xapi never use options/delete/patch/etc?

lib/lrs_xapi_proxy.rb
rshipp: this isn't an error, sometimes the body is just empty, and we pass it along as-is.

config/routes.rb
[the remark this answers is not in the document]
rshipp: now that I think about it, I've only implemented GET and PUT in the proxy lib, so this should restrict to just those two.

app/helpers/lti_link_selection_helper.rb
sadleb: Delete this file from this PR. We'll implement this using the new LRS proxy stuff once we rebase that code in.

config/initializers/content_security_policy.rb
```diff
@@ -16,9 +16,13 @@
 
   # Customizable through .env
   # Allow plain http, webpack-dev-server, and livereload connections in development.
-  policy.connect_src :self, :https, "http://localhost:*", "ws://localhost:3035", "http://#{Rails.application.secrets.application_host}:*" if Rails.env.development?
+  policy.connect_src :self, :https, "http://localhost:*", "ws://localhost:*",
```
sadleb, line 19: @jessouyangbraven is going to pull this into it's own PR. We'll merge to staging and @rshipp you can delete from this PR and you'll get this changes when you rebase off staging.

lib/lrs_xapi_proxy.rb
```diff
@@ -0,0 +1,80 @@
...
+          x_experience_api_version: '1.0.3',
+          params: request.query_parameters
+        })
+        response
+      rescue RestClient::NotFound => e
+        e.response
+      end
+    elsif request.method == 'PUT'
+      # Rewrite body.
+      # TODO: huh???
```
sadleb, line 27: @rshipp TODO: write up more context about this in a comment and leave it at that.

spec/helpers/lti_link_selection_helper_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  describe '#to_query' do
+    let(:lrs_proxy_url) { 
+      uri = URI(root_url)
+      uri.path = '/data/xAPI'
+      uri
+    }
+
+    it 'creates the query with no activity_id or registration' do
```
sadleb, line 17: If those don't pass right now, make them pending and we'll reimplement with the real link selection stuff in a followup PR.

app/helpers/lti_link_selection_helper.rb
rshipp: deleting the spec alongside this, so here it is in case you need it later:

```diff
diff --git a/spec/helpers/lti_link_selection_helper_spec.rb b/spec/helpers/lti_link_selection_helper_spec.rb
deleted file mode 100644
index 9a47921..0000000
--- a/spec/helpers/lti_link_selection_helper_spec.rb
+++ /dev/null
@@ -1,28 +0,0 @@
-require 'rails_helper'
-
-RSpec.describe LtiLinkSelectionHelper, type: :helper do
-
-  before(:all) do
-    ENV['LRS_URL'] = 'https://example.lrs.com/data/xAPI'
-    ENV['LRS_AUTH_TOKEN'] = 'IFDJ_EXAMPLE_TOKENoasdfj'
-  end
-
-  describe '#to_query' do
-    let(:lrs_proxy_url) {
-      uri = URI(root_url)
-      uri.path = '/data/xAPI'
-      uri
-    }
-
-    it 'creates the query with no activity_id or registration' do
-      query = helper.launch_query('Some Name', 'someemail@example.com')
-      expect(query).to eq("endpoint=#{lrs_proxy_url}&actor=%7B%22name%22%3A%5B%22Some%20Name%22%5D%2C%20%22mbox%22%3A%5B%22mailto%3Asomeemail%40example.com%22%5D%7D")
-    end
-
-    it 'creates the query with activity_id and registration' do
-      query = helper.launch_query('Some Name', 'someemail@example.com', 'some_activity_id', 'some_registration')
-      expect(query).to eq("endpoint=#{lrs_proxy_url}&actor=%7B%22name%22%3A%5B%22Some%20Name%22%5D%2C%20%22mbox%22%3A%5B%22mailto%3Asomeemail%40example.com%22%5D%7D&activity_id=some_activity_id&registration=some_registration")
-    end
-
-  end
-end
```

PR #279, opened 2020-07-16
Implement uploading Rise 360 zip as lesson to Canvas module

jessouyangbraven:
**NOTES FOR PR**
- Ignore the individual commits. I'll squash for the actual PR. 
- `TODO` ~= Not going to be part of this PR (obviously open to discussion), has Asana task to track

**Task**: https://app.asana.com/0/1174274412967132/1135638232028169

**Description**: A Designer can add a Rise360 package as an item in a module

**Dev environment breaking changes**
- Update your `.env` file with the entries added in `env.example`
- Rebuild for new gems: `docker-compose/scripts/rebuild.sh`
- Migrate for `ActiveStorage`: `deva` then `bin/rails db:migrate`

**Test**
- You'll need to follow the *LTI Extension* set up steps here: https://github.com/bebraven/platform#lti-extension
- Download Rise 360 lesson as zip
- Add lesson to Canvas course > External Tool > LTI Link Extension > Select zip > Add Item
- Click on link and verify it loads
- Screenshots below show sequence of steps (I would have made a video, but the uploading was ~30 seconds of waiting that I didn't know how to edit out):

Screen Shot 2020-07-15 at 8 27 45 PM
Screen Shot 2020-07-15 at 8 27 49 PM
Screen Shot 2020-07-15 at 8 27 56 PM
Screen Shot 2020-07-15 at 8 28 01 PM
Screen Shot 2020-07-15 at 8 29 20 PM
Screen Shot 2020-07-15 at 8 35 40 PM
Screen Shot 2020-07-15 at 8 35 44 PM
Screen Shot 2020-07-15 at 8 35 52 PM

Additional ways you can access the course content:
- Clicking on the link you created in the module
- The public object URL for the index on S3: `https://s3.console.aws.amazon.com/s3/buckets/platform-dev-file-uploads/lessons/{lesson_content_zipfile.key}/index.html`
- Your local development endpoint `platformweb/lesson_contents/{lesson_contents_id}` will also redirect you to the above


app/controllers/lesson_contents_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    # TODO: Should we make state optional, e.g. allow creating lessons contents
+    # outside of Canvas's LTI link extension?
+    # https://app.asana.com/0/1174274412967132/1184800386160070
+    params.require([:state, :lesson_content_zipfile])
+    lti_launch = LtiLaunch.current(params[:state])
+
+    @lesson_content = LessonContent.new(lesson_content_create_params)
+    @lesson_content.save
+
+    # TODO: Extract this asynchronously
```
jessouyangbraven, line 23: It's slow, but not Gateway Timeout.

app/controllers/lesson_contents_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    # https://app.asana.com/0/1174274412967132/1184800386160057
+    @s3_index_url = @lesson_content.extract
+    lesson_url = lesson_content_url(@lesson_content)
+
+    # TODO: Decide whether it's better to use lesson_url that does a redirect
```
jessouyangbraven, line 28: I tried to use `s3_index_url`. For reasons unclear to me, it doesn't work, but the `lesson_url` does. 

@sadleb is this because of how the LTI launch flow works? Does the endpoint have to be on `platformweb`?

config/storage.yml
```diff
@@ -6,6 +6,15 @@ local:
   service: Disk
   root: <%= Rails.root.join("storage") %>
 
+# TODO: Hard-code in your access and secret keys for now, figure out how to 
+# add to encrypted secrets
+amazon:
```
jessouyangbraven, line 11: **ActiveStorage**: this where the S3 configuration lives

config/environments/production.rb
```diff
@@ -36,7 +36,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :local
+  config.active_storage.service = :amazon
```
jessouyangbraven, line 39: **ActiveStorage**: update to S3 in production

config/environments/development.rb
```diff
@@ -28,7 +28,7 @@
   end
 
   # Store uploaded files on the local file system (see config/storage.yml for options)
-  config.active_storage.service = :local
+  config.active_storage.service = :amazon
```
jessouyangbraven, line 31: **ActiveStorage**: upload to S3 in development

app/views/lesson_contents/new.html.erb
```diff
@@ -0,0 +1,7 @@
+<!-- should this be PUT?  -->
+<%= form_tag lesson_contents_path(@lesson_content), method: :post, multipart: true do %>
+  <%= label_tag :lesson_content_zipfile, 'Rise 360 zipfile' %>
+  <%= file_field_tag :lesson_content_zipfile, :accept => "application/zip" %>
```
jessouyangbraven, line 4: **ActiveStorage**: this is how you add the field in the form so it updates the model

app/models/lesson_content.rb
```diff
@@ -0,0 +1,59 @@
+# Represents the contents of a Canvas module, e.g. a Rise 360 course
+class LessonContent < ApplicationRecord
+  # Rise 360 lesson zipfile
+  has_one_attached :lesson_content_zipfile
```
jessouyangbraven, line 4: **ActiveStorage**: this is how you add an attachment to your model

app/models/lesson_content.rb
```diff
@@ -0,0 +1,59 @@
...
+    return get_index_url unless !bucket.object(s3_object_key(INDEX_FILE)).exists?
+
+    # TODO: Clean up after Brian pushes code
+    # https://app.asana.com/0/1174274412967132/1184800386160066
+    credentials = Aws::Credentials.new(ENV["AWS_ACCESS_KEY_ID"], ENV["AWS_SECRET_ACCESS_KEY"])
+    Aws.config.update({credentials: credentials})
+    bucket = Aws::S3::Resource.new(region: ENV["AWS_REGION"]).bucket(ENV["AWS_S3_BUCKET"])
+
+    # Extract the zipfile
+    self.lesson_content_zipfile.open do |file|
```
jessouyangbraven, line 34: **ActiveStorage**: this downloads the attachment from S3 to local `/tmp/`

app/models/lesson_content.rb
```diff
@@ -0,0 +1,59 @@
...
+      Zip::File.open(file.path) do |zip_file|
+        zip_file.each do |entry|
+          next unless entry.file?
```
jessouyangbraven, line 37: This has to be file and not directory (or symlink, which we don't have) because S3 buckets don't actually have directories, just objects to have `/` in the name that render as directories.

app/views/lesson_contents/new.html.erb
```diff
@@ -0,0 +1,7 @@
+<!-- should this be PUT?  -->
+<%= form_tag lesson_contents_path(@lesson_content), method: :post, multipart: true do %>
```
jessouyangbraven, line 2: **ActiveStorage**: this has to be lesson_content**s**_path -- not the singular. This was not obvious to me and I got weird errors. I've also used the plural when I needed the singular.

app/controllers/lesson_contents_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def new
+    @lesson_content = LessonContent.new
+  end
+
+  def create
+    # TODO: Should we make state optional, e.g. allow creating lessons contents
+    # outside of Canvas's LTI link extension?
+    # https://app.asana.com/0/1174274412967132/1184800386160070
+    params.require([:state, :lesson_content_zipfile])
```
jessouyangbraven, line 17: **ActiveStorage**: it's like any other parameter being passed into creation

db/migrate/20200710232731_create_active_storage_tables.active_storage.rb
```diff
@@ -0,0 +1,27 @@
+# This migration comes from active_storage (originally 20170806125915)
+class CreateActiveStorageTables < ActiveRecord::Migration[5.2]
```
jessouyangbraven, line 2: **ActiveStorage**: the DB migration you get after `bin/rails active_storage:install` that you'll have to `bin/rails db:migrate`

app/views/lesson_contents/new.html.erb
rshipp: i still dont understand all the path names, but i realized the other day if you go to the 404 page, like http://platformweb/notarealpage, the paths in the left column there are the same as the ones you use in helper tags.

app/controllers/lesson_contents_controller.rb
```diff
@@ -0,0 +1,42 @@
+require 'zip'
+
+class LessonContentsController < ApplicationController
+  # FIXME: Get rid of this, something weird is going on right now
```
sadleb, line 4: I think if you login to the platform in a separate tab, that will create a session and it will work. We have a task to fix this here: https://app.asana.com/0/1174274412967132/1184057808812010 -- if that works, let's remove these lines (keep the `verify_authenticity_token` one b/c we need that) and we'll do the whole authorization stuff in a follow-up PR.

app/controllers/lesson_contents_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  skip_before_action :authenticate_user!
+  skip_before_action :ensure_admin!
+  skip_before_action :verify_authenticity_token
+
+  def new
+    @lesson_content = LessonContent.new
```
sadleb, line 10: The `DryCrud` stuff should magically take care of this. I think you can remove this line and leave an empty `new` method and the `@lesson_content` variable will be accessible.

app/controllers/lti_link_selection_controller.rb
```diff
@@ -0,0 +1,4 @@
+class LtiLinkSelectionController < ApplicationController
```
sadleb, line 1: If we're not using this now, let's get rid of it and bring it back if we need it.

app/controllers/lesson_contents_controller.rb
sadleb: For now, let's leave it as being required. If we do a future iteration where there is a flow to create a lesson_content outside of an LTI launch, we'll reintroduce the LtiLinkSelectionController and refactor.

app/controllers/lesson_contents_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def create
+    # TODO: Should we make state optional, e.g. allow creating lessons contents
+    # outside of Canvas's LTI link extension?
+    # https://app.asana.com/0/1174274412967132/1184800386160070
+    params.require([:state, :lesson_content_zipfile])
+    lti_launch = LtiLaunch.current(params[:state])
+
+    @lesson_content = LessonContent.new(lesson_content_create_params)
+    @lesson_content.save
```
sadleb, line 21: We should get in the habit of calling `save!` which if it fails will throw an exception. This will fail silently.

app/controllers/lesson_contents_controller.rb
sadleb: `lesson_content_url(@lesson_content)` is a standard helper for getting URLs on the platform for a particular record. It looks like we're telling Canvas something like "go to `https://platformweb/lesson_contents/1` and then when the launch happens, we redirect to S3 in the show. That makes sense to me. I don't think we should put the S3 URLs in Canvas. It doesn't look like `@s3_index_url` is currently being used, so we should remove that line here.

app/controllers/lesson_contents_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    @s3_index_url = @lesson_content.extract
+    lesson_url = lesson_content_url(@lesson_content)
+
+    # TODO: Decide whether it's better to use lesson_url that does a redirect
+    # https://app.asana.com/0/1174274412967132/1184800386160059
+    @deep_link_return_url, @jwt_response = helpers.lti_deep_link_response_message(lti_launch, lesson_url)
+  end
+
+  def show
+    lesson_content = LessonContent.find(params[:id])
```
sadleb, line 34: `DryCrud` provides a `@lesson_content` variable magically. You can just redirect to `@lesson_content.get_index_url`

app/models/lesson_content.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  # Extract the lesson_content_zipfile and store it on S3
+  # Return index to access content
+  def extract
+    # TODO: Is there a more robust way to check?
+    # https://app.asana.com/0/1174274412967132/1184800386160068
+    return get_index_url unless !bucket.object(s3_object_key(INDEX_FILE)).exists?
+
+    # TODO: Clean up after Brian pushes code
```
sadleb, line 27: I think we should refactor the AWS stuff into a class in `lib` that abstracts this away.

app/models/lesson_content.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+      end
+
+    get_index_url
+  end
+
+  # TODO: Clean up after Brian pushes code
+  # https://app.asana.com/0/1174274412967132/1184800386160066
+  private
+  def bucket
```
sadleb, line 53: Ditto. We should put the aws zipping/unzipping/pushing back to the bucket stuff in a helper class in `lib`, not directly in the model.

app/views/lesson_contents/new.html.erb
sadleb: You can also run `rake routes` to get a list, but yeah, I do the same thing to see the helper path names.

config/routes.rb
```diff
@@ -1,4 +1,7 @@
 Rails.application.routes.draw do
+
+  resources :lesson_contents
```
sadleb, line 3: Two things with the routes. 1) I think we should restrict it to only what we need. E.g. `only: [:new, :show, :create]` and 2) I think `lesson_contents` should mimic `lesson_submissions`. E.g. remove this line and make the stuff below look like this:

```
  resources :lessons, only: [:index, :show] do
    resources :lesson_submissions, only: [:index, :show], :path => 'submissions'
    resources :lesson_contents, only: [:new, :show, :create], :path => 'contents'
  end

  resources :lesson_contents, only: [:new, :show, :create]
```

This will make it available at two routes, for example:
https://platformweb/lesson_contents/1
https://platformweb/lessons/1/contents/1

app/controllers/lesson_contents_controller.rb
jessouyangbraven: Hmm, this isn't working for me. Will debug in our sync today.

spec/factories/lti_link_launch_request.rb
```diff
@@ -59,11 +68,6 @@
         'https://purl.imsglobal.org/spec/lti/claim/version' => '1.3.0',
         'https://purl.imsglobal.org/spec/lti/claim/deployment_id' => '59:id_of_lti_deployment',
         'https://purl.imsglobal.org/spec/lti/claim/target_link_uri' => evaluator.target_link_uri,
-        'https://purl.imsglobal.org/spec/lti/claim/resource_link' => {
```
jessouyangbraven, line 62: @sadleb you might already be aware, but a heads up that this breaks the `./spec/models/lti_launch_spec.rb` test
```
Failures:

  1) LtiLaunch methods .authenticate when valid sets id_token_payload
     Failure/Error: @resource_link = payload.fetch(RESOURCE_LINK_CLAIM)['id']
     
     KeyError:
       key not found: "https://purl.imsglobal.org/spec/lti/claim/resource_link"
       Did you mean?  "https://purl.imsglobal.org/spec/lti/claim/roles"
                      "https://purl.imsglobal.org/spec/lti/claim/version"
                      "https://purl.imsglobal.org/spec/lti/claim/target_link_uri"
                      "https://purl.imsglobal.org/spec/lti/claim/context"
                      "https://purl.imsglobal.org/spec/lti/claim/custom"
                      "https://purl.imsglobal.org/spec/lti/claim/deployment_id"
                      "https://purl.imsglobal.org/spec/lti/claim/message_type"
                      "https://purl.imsglobal.org/spec/lti/claim/tool_platform"
                      "https://purl.imsglobal.org/spec/lti-ags/claim/endpoint"
     # ./lib/lti_resource_link_request_message.rb:15:in `fetch'
     # ./lib/lti_resource_link_request_message.rb:15:in `initialize'
     # ./spec/models/lti_launch_spec.rb:37:in `new'
     # ./spec/models/lti_launch_spec.rb:37:in `block (5 levels) in <top (required)>'

  2) LtiLaunch methods .authenticate when valid updates target_link_uri
     Failure/Error: @resource_link = payload.fetch(RESOURCE_LINK_CLAIM)['id']
     
     KeyError:
       key not found: "https://purl.imsglobal.org/spec/lti/claim/resource_link"
       Did you mean?  "https://purl.imsglobal.org/spec/lti/claim/roles"
                      "https://purl.imsglobal.org/spec/lti/claim/version"
                      "https://purl.imsglobal.org/spec/lti/claim/target_link_uri"
                      "https://purl.imsglobal.org/spec/lti/claim/context"
                      "https://purl.imsglobal.org/spec/lti/claim/custom"
                      "https://purl.imsglobal.org/spec/lti/claim/deployment_id"
                      "https://purl.imsglobal.org/spec/lti/claim/message_type"
                      "https://purl.imsglobal.org/spec/lti/claim/tool_platform"
                      "https://purl.imsglobal.org/spec/lti-ags/claim/endpoint"
     # ./lib/lti_resource_link_request_message.rb:15:in `fetch'
     # ./lib/lti_resource_link_request_message.rb:15:in `initialize'
     # ./spec/models/lti_launch_spec.rb:37:in `new'
     # ./spec/models/lti_launch_spec.rb:37:in `block (5 levels) in <top (required)>'
```

README.md
```diff
@@ -31,7 +37,7 @@ up. It will also install all the necessary Ruby gems and JavaScript libraries th
 
 Now create the needed databases:
 
-    docker-compose run platformweb bundle exec rake db:create db:schema:load db:dummies
+    docker-compose run platformweb bundle run rake db:create db:schema:load db:dummies
```
rshipp, line 40: why this change? if the `up -d` above worked, exec will run on the existing container.

README.md
```diff
@@ -64,7 +72,7 @@ Now build the Docker environment:
 
 And set up the database in your Join server's dev environment:
 
-    docker-compose exec joinweb bundle exec rake db:create && ./docker-compose/scripts/dbrefresh.sh
+    docker-compose exec joinweb bundle run rake db:create && ./docker-compose/scripts/dbrefresh.sh
```
rshipp, line 75: same q here

spec/factories/lti_link_launch_request.rb
sadleb: I pushed a fix for this. My bad.

README.md
sadleb: I think something changed from when we initially wrote this where on a fresh install with the DB not created, the container exits.

README.md
sadleb: Same answer. I think it's just safer to not assume the container won't exit with no DB created.

app/models/lesson_content.rb
sadleb: I'm not the biggest fan of this attribute name. Seems redundant. I opened a placeholder here to revisit: https://app.asana.com/0/1174274412967132/1185359100297593

sadleb: We decided to just leave this. If we every add another type of package in the future, we would want to store it in this column too. Can't think of a better name.

PR #284, opened 2020-07-21
Don't freeze the bundle install in dev.

sadleb:
This is more for production machine where a release / deploy freezes
the Gemfile.lock. In dev, we want to be able to add gems to Gemfile and
have bundle install just update Gemfile.lock to match.

### Test Plan:
- Add a gem to Gemfile. Rebuild the container. Gemfile.lock should get
  updated.


sadleb:
See https://bebraven.slack.com/archives/C01166ZLR1V/p1595356361016700?thread_ts=1595355359.013800&cid=C01166ZLR1V

PR #285, opened 2020-07-23
Fix CAS specs

rshipp:
Task: https://app.asana.com/0/1174274412967132/1166260321465780/f

CAS controller specs no longer reach out to real SSO servers. Removed all real SSO servers from the VCR ignored list. 

Some additional changes just to support the above.


docker-compose/scripts/rebuild.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 docker-compose down
 docker volume rm platform_vendor-bundle
+docker volume rm platform_node-modules
```
rshipp, line 4: unrelated change, but I had to rebuild several times and needed this.

spec/controllers/cas/login_spec.rb
```diff
@@ -40,32 +40,36 @@
     describe "/cas/login with service url" do
       let(:return_service) { 'http://braven/' }
       before(:each) do 
-        visit "/cas/login?service=#{url_encode(return_service)}"
-        fill_and_submit_login(username, password)
+        VCR.use_cassette("sso_ticket_invalid", :match_requests_on => [:path]) do
```
rshipp, line 43: note everywhere we use `sso_ticket_invalid` - I'm not sure of those tests are working as intended, but they do pass. Figuring out if the tests actually test what they're supposed to would be a bigger task, that I suggest we don't spend time on unless it becomes an issue.

spec/feature/course_contents/smoke_spec.rb
```diff
@@ -16,10 +16,12 @@
     describe "/course_contents/new loads ckeditor", :js do
       let(:return_service) { '/course_contents/new' }
       before(:each) do 
-        visit "/cas/login?service=#{url_encode(return_service)}"
         VCR.configure do |c|
           c.ignore_localhost = true
+          # Must ignore the Capybara host IFF we are running tests that have browser AJAX requests to that host.
```
rshipp, line 21: important! this is because I removed `platformweb` from the global VCR ignore list, which is required to get VCR to process the requests in the controller specs, where we're not using Selenium and therefore don't have a real server.

spec/feature/course_contents/smoke_spec.rb
```diff
@@ -16,10 +16,12 @@
...
+          c.ignore_hosts Capybara.server_host
         end
+        visit "/cas/login?service=#{url_encode(return_service)}"
```
rshipp, line 24: this needs to be below the configure, or it doesn't ignore the Capybara host.

spec/feature/lrs_xapi_proxy/assignment_xapi_spec.rb
```diff
@@ -17,11 +17,23 @@
...
 
+      after(:each) do
+        # Print JS console errors, just in case we need them.
+        # From https://stackoverflow.com/a/36774327/12432170.
+        errors = page.driver.browser.manage.logs.get(:browser)
+        if errors.present?
+          message = errors.map(&:message).join("\n")
+          puts message
+        end
+      end
```
rshipp, line 35: I added this in to troubleshoot an issue that ended up being a JS build failure. I figure it'll probably be useful to have later.

spec/support/vcr.rb
```diff
@@ -3,8 +3,9 @@
 VCR.configure do |config|
   config.cassette_library_dir = "spec/fixtures/vcr_cassettes"
   config.hook_into :webmock
-  config.ignore_hosts 'platformweb', 'chromedriver.storage.googleapis.com',
-                      'api.codacy.com', 'api.honeycomb.io'
   config.ignore_hosts ENV['SELENIUM_HOST'] if ENV['SELENIUM_HOST']
   config.ignore_localhost = true
+  # Only put third-party hosts we don't care about below. Things we do actually care about the
+  # responses should be VCR-recorded instead.
+  config.ignore_hosts 'chromedriver.storage.googleapis.com', 'api.codacy.com', 'api.honeycomb.io'
```
rshipp, line 10: Note there are no longer any platform hosts in here.

.env.example
```diff
@@ -40,6 +40,9 @@ SALESFORCE_HOST=test.salesforce.com
 SECRET_KEY_BASE=f843142daa4f4c69c7ac67bacd5837fd43f7b5390a922d9ac43692db24eb7a0f490229354db1720cd7e6f2cf1533a6560324a0584ebf87e2ba496c2d3e92fb9c
 SELENIUM_HOST=platform_chrome
 SELENIUM_PORT=4444
+# The host used to set up the Capybara server.
+# MUST resolve to localhost, but not be "localhost", "127.0.0.1", "::1", etc.
+# Nonstandard IPv4 loopbacks are acceptable, e.g. 127.11.11.11.
```
sadleb, line 45: https://bebraven.slack.com/archives/CMDSASG73/p1595530727007100

rshipp:
* application_host: Used by Rails, nothing here touches it
* selenium_host: The host where the selenium driver is running (in our case, localhost) - Capybara calls into this under the hood to control the headless chrome browser
* spec_host: The host where Capybara will start a Rails server (in our case, using Puma) with our application code.
* localhost: A hostname that always resolves to "127.0.0.1", which is one of a [very large number](https://en.wikipedia.org/wiki/Localhost) of possible "loopback" addresses. Loopback addresses always refer to the local computer.

PR #286, opened 2020-07-23
Lessons send xAPI statements to LRS

jessouyangbraven:
**Task**: https://app.asana.com/0/1174274412967132/1183556772940371

**Dev environment breaking changes**: 

- Rebuild: `rack-cors` gem added, you'll need to rebuild you Docker container.

**Test**:  

```
rspec spec/helpers/lesson_contents_helper.rb
rspec spec/controllers/lesson_contents_controller.spec.rb
```

**If you're running into login issues, you can serve a static lesson from local `platformweb` instead of AWS S3 using this [commit](https://github.com/bebraven/platform/commit/96540a7c833ed4e38a5ad3e623ad6a29ceab33d3), or by running**:
```
git cherry-pick -n 96540a7c833ed4e38a5ad3e623ad6a29ceab33d3
```
FYI, it'll cause the controller spec to fail, since we're altering the redirect. 

 - Login to `platformweb`
 - Upload a Rise 360 course (see PR #279 for instructions)
 - Open the link returned (something like, https://platformweb/lesson_contents/1) in a new browser tab
 - See that there are xAPI events logged in Learning Locker (https://lrs.braven.org/):
Screen Shot 2020-07-23 at 2 58 30 PM
 - You can also look at the network traffic in the Chrome developer console and verify that `statements/` work:
Screen Shot 2020-07-23 at 2 58 48 PM


lib/lrs_xapi_proxy.rb
```diff
@@ -19,87 +19,73 @@ def self.request(request, path, user)
...
-        # getting duplicate params here. Every param is passed through as expected at the root level,
-        # but then there's an additional key, 'lrs_xapi_proxy', that contains an exact copy of all
-        # those params. Since I have no idea why this happens, and the LRS complains if you pass in
-        # params it doesn't recognize, we're just passing in the contents of the duplicate key here.
-        # If anyone ever figures out what's going on here and how to get rid of that extra key, this
-        # should be updated to pass in just request_parameters.
-        data = request.request_parameters['lrs_xapi_proxy']
-        if data['actor']
-          data['actor'] = {
+      return unless request.method == 'GET' || request.method == 'PUT'
```
jessouyangbraven, line 22: I realized that troubleshooting yesterday was really slow because there were two places we needed to make changes because `GET` and `PUT` were handled in separate blocks. 

So I refactored to dedupe as much code as possible and hopefully it'll be easier to troubleshoot the xAPI messages going forward.

rshipp: nice, I couldnt figure out how to do this.

config/application.rb
```diff
@@ -16,6 +16,14 @@ class Application < Rails::Application
     # be backwards compatible with v5.2 and below who may have removed it from ApplicationController
     config.action_controller.default_protect_from_forgery = true
 
+    # Allows Rise 360 lesson running in AWS S3 to call into our LRS xAPI proxy
+    config.middleware.insert_before 0, Rack::Cors do
+      allow do
+        origins "https://#{Rails.application.secrets.aws_files_bucket}.s3.amazonaws.com"
```
sadleb, line 22: Can we do this in `config/initializers/content_security_policy.rb` or is there some reason why we need to use this new gem and add this middleware in?

app/helpers/lesson_contents_helper.rb
```diff
@@ -1,2 +1,25 @@
...
 module LessonContentsHelper
+  
+  def launch_query
+    lrs_proxy_url = URI(root_url)
+    lrs_proxy_url.path = LrsXapiProxy.lrs_path
+    {
+      :endpoint => lrs_proxy_url.to_s,
+      # Our LRS proxy will supply the correct values for these
+      # Send empty values to get the Rise 360 Tincan code won't error out on missing keys
+      :actor => '{"name":"", "mbox":["mailto:''"]}',
```
sadleb, line 22: I think we should add placeholders similar to what we do here: https://github.com/bebraven/platform/pull/277/files#diff-27356557a83de0f7a0f7bd3b4cebc88fR10

app/helpers/lesson_contents_helper.rb
```diff
@@ -1,2 +1,25 @@
...
+#
+# See:
+# - https://articulate.com/support/article/Implementing-Tin-Can-API-to-Support-Articulate-Content#launching-public-content
+# - https://xapi.com/try-developer/
+# - https://learningpool.com/how-to-launch-elearning-content-using-xapi/
 module LessonContentsHelper
+  
+  def launch_query
+    lrs_proxy_url = URI(root_url)
+    lrs_proxy_url.path = LrsXapiProxy.lrs_path
```
sadleb, line 17: Would it be cleaner to expose this URL as a helper method on `LrsXapiProxy` instead of building it here?

config/application.rb
rshipp: CSP and cors are diffrerent, I had to look up why https://stackoverflow.com/questions/39488241/what-is-the-difference-between-cors-and-csps

lib/lrs_xapi_proxy.rb
```diff
@@ -19,87 +19,73 @@ def self.request(request, path, user)
...
+
+      # Rewrite query string
+      params = request.query_parameters
+      if params['agent']
```
sadleb, line 26: This comment applies to the whole class. This stuff feels worthy of some specs.

rshipp: See fae4455

lib/lrs_xapi_proxy.rb
```diff
@@ -8,8 +8,22 @@ class LrsXapiProxy
...
+        :host => Rails.application.secrets.application_host,
+        :protocol => 'https',
+      ))
+      lrs_proxy_url.path = lrs_path
+      lrs_proxy_url
+    end
+  end
+
   def self.lrs_path
-    @@lrs_path ||= get_lrs_path
```
jessouyangbraven, line 12: Codacy didn't like this. The suggestion was to change class variables to class instance variables. This explains the issues you can run into with class variables: https://stackoverflow.com/questions/15773552/ruby-class-instance-variable-vs-class-variable

lib/lrs_xapi_proxy.rb
```diff
@@ -19,87 +33,72 @@ def self.request(request, path, user)
...
+        span.add_field('error', e.message)
+        span.add_field('http_code', e.http_code)
+        span.add_field('http_body', e.http_body)
+        raise
+      end
+    end
   end
 
-  def self.authentication_header
+  private_class_method def self.authentication_header
```
jessouyangbraven, line 101: Gotcha: `private_class_method` has to be on the same line as `def`. Unlike, say, `private` on instance method `def`.

lib/lrs_xapi_proxy.rb
```diff
@@ -19,87 +33,72 @@ def self.request(request, path, user)
...
 
-  def self.lrs_base_uri
-    @@lrs_base_uri ||= get_lrs_base_uri
+      begin
+        response = RestClient::Request.execute(
+          method: request.method,
+          url: "#{Rails.application.secrets.lrs_url}/#{path}",
+          payload: request.method == 'PUT' ? data.to_json : {},
+          headers: {
+            authorization: authentication_header,
```
jessouyangbraven, line 83: This gets lost in the changes because of the refactoring, but explicitly specifying `XApiLrs.authentication_header` now longer works when we make the class method private. You have to let Ruby resolve the call implicitly.

lib/lrs_xapi_proxy.rb
jessouyangbraven: Other ways to get the same effect: http://jakeyesbeck.com/2016/01/24/ruby-private-class-methods/

This one seemed the most explicit, least verbose.

app/helpers/lesson_contents_helper.rb
jessouyangbraven: I spent embarrassingly long on this. `root_url` seems to work magically (gives you the right proper protocol, URL, etc.) in controllers and things Rails associates with controllers (helpers, here). I couldn't figure out a straightforward way to get this to work in `lib/`, e.g. non-controller land in Rails. 

A couple of the things I tried in all sorts of combinations: 
https://stackoverflow.com/questions/34862410/how-to-use-root-url-in-the-activemodelserializer
https://stackoverflow.com/questions/36131936/missing-host-to-link-to-please-provide-the-host-parameter-for-rails-4/41336950

sadleb: Got it. Yeah, I forgot about that. `root_url` is a Rails helper which won't be available to `lib` classes. We could probably get it to work with an initializer, but maybe it's not worth it for now.

PR #287, opened 2020-07-24
Add LtiAdvantageAPI to help access those services.

sadleb:
There are a ton of links in the `lib/lti_advantage_api.rb` class for more context about LTI Advantage. High level, this API will allow us to send grades back to Canvas for Projects and Lessons. 

I'm still unclear about the difference between this API and just using the normal Canvas API to hit it's assignment and submissions endpoints. Maybe it's just about permissions where if we hit the Canvas API we rely on our code to make sure the permissions are correct but going this route the permissions are handled more on Canvas's end?

Test Plan:
- `bundle exec rspec spec/lib/lti_advantage_api_spec.rb`
- Go into `lti_launch_controller.rb` and put something like the following in to see that it actually works with Canvas and sets the grade:
image

### DIFF
--- a/app/models/lti_launch.rb
+++ b/app/models/lti_launch.rb
@@ -9,6 +9,7 @@ class LtiLaunch < ApplicationRecord
   # This is the Redirect URI that must be configured in the Developer Key for the LTI
   # and must match EXACTLY.
   LTI_LAUNCH_REDIRECT_URI="https://#{Rails.application.secrets.application_host}/lti/launch".freeze
+  BRAVEN_ISS="https://#{Rails.application.secrets.application_host}".freeze
 
   scope :authenticated, -> { where.not id_token_payload: nil }
   scope :unauthenticated, -> { where id_token_payload: nil }
@@ -34,6 +35,16 @@ class LtiLaunch < ApplicationRecord
 
     launch.save!
     launch
+  end
+
+  # Returns the parse request message payload of the launch
+  def request_message
+    @request_message ||= LtiIdToken.parse(id_token_payload)
+  end
+
+  # Returns a unique identifier for us when issuing and JWT
+  def braven_iss
+    BRAVEN_ISS
   end
 
   # Returns the parameters needed in Step 2 of the LTI Launch flow in order
--- a/lib/lti_advantage_api.rb
+++ b/lib/lti_advantage_api.rb
@@ -28,19 +28,24 @@ class LtiAdvantageAPI
     @iss = assignment_lti_launch.braven_iss
     @client_id = assignment_lti_launch.client_id 
     @scope = parse_scope(assignment_lti_launch) 
-    @line_items_url = parse_line_items(assignment_lti_launch)
-    @line_item_url = parse_line_item(assignment_lti_launch)
+    @line_items_url = assignment_lti_launch.request_message.line_items_url
+    @line_item_url = assignment_lti_launch.request_message.line_item_url
     @global_headers = JSON_HEADERS # get_access_token below needs this initialized
     @global_headers = @global_headers.merge(:Authorization => "Bearer #{get_access_token}") # Note: these expire about about an hour.
   end
 
-  # Use LtiScore.generate(...) to call this
-  # See: https://canvas.instructure.com/doc/api/score.html
+  # Sends a score / grade for this LTI resource (aka a Project or a Lesson) back
+  # to Canvas so it shows up in the gradebook. See: https://canvas.instructure.com/doc/api/score.html
+  #
+  # Usage:
+  # LtiAdvantageAPI.new(<the_launch>).create_score( LtiScore.generate(...) ) 
   def create_score(lti_score)
     response = post(@line_item_url + '/scores', lti_score)
     JSON.parse(response.body)
   end
 
+  # Gets a list of line items, aka columns, in the gradebook for this LTI launch context (aka the course)
+  # that scores can be sent to.
   # See: https://canvas.instructure.com/doc/api/line_items.html
   def get_line_items
     response = get(@line_items_url)
@@ -81,22 +86,14 @@ private
 
   def validate_param(assignment_lti_launch)
     req_msg = assignment_lti_launch.request_message
-    unless req_msg.is_a?(LtiResourceLinkRequestMessage) && req_msg.scope.present?
-      raise ArgumentError.new, 'LtiAdvantageAPI can only be used with the launch of an assignment created with the Assignment Selection Placement'
-    end
+    raise ArgumentError.new, 'Wrong LTI launch message type. Must be a launch of a resource.' unless req_msg.is_a?(LtiResourceLinkRequestMessage) 
+    raise ArgumentError.new, 'Missing scope. Needed to get an access token for LTI Advantage access' unless req_msg.scope.present?
+    raise ArgumentError.new, 'Missing lineitem. Needed to send a score back for this resource.' unless req_msg.line_item_url.present?
   end
 
   # Parses the scope array out of the launch message into the space delimited list expected in the access token grant message.
-  def parse_scope(lti_resource_link_launch)
-    lti_resource_link_launch.request_message.scope.join(' ')
-  end
-
-  def parse_line_items(lti_resource_link_launch)
-    lti_resource_link_launch.request_message.line_items_url
-  end
-
-  def parse_line_item(lti_resource_link_launch)
-    lti_resource_link_launch.request_message.line_item_url
+  def parse_scope(assignment_lti_launch)
+    assignment_lti_launch.request_message.scope.join(' ')
   end
 
   def get(target_url, headers={})
--- a/lib/lti_advantage_api.rb
+++ b/lib/lti_advantage_api.rb
@@ -17,7 +17,7 @@ require 'lti_resource_link_request_message'
 # - https://www.imsglobal.org/spec/security/v1p0/#using-oauth-2-0-client-credentials-grant
 class LtiAdvantageAPI
 
-  OAUTH_ACCESS_TOKEN_URL = "#{Rails.application.secrets.canvas_cloud_url}#{}/login/oauth2/token"
+  OAUTH_ACCESS_TOKEN_URL = "#{Rails.application.secrets.canvas_cloud_url}/login/oauth2/token"
   JSON_HEADERS = {content_type: :json, accept: :json}
 
   # The assignment_lti_launch should be an LtiLaunch of an already Deep Linked assignment
--- a/lib/lti_advantage_api.rb
+++ b/lib/lti_advantage_api.rb
@@ -1,0 +1,121 @@
+# frozen_string_literal: true
+require 'rest-client'
+require 'lti_resource_link_request_message'
+
+# Helps access the LTI Advatange services described in the following links. These
+# services are mostly used for passing grade data back to Canvas from our LTI extension.
+# Note: these are only available for LTI resources added using the Assignment Selection
+# placement. See the notes under "Limitations" here: https://canvas.instructure.com/doc/api/file.link_selection_placement.html
+#
+# See: 
+# - https://www.imsglobal.org/spec/lti-ags/v2p0/#introduction
+# - https://canvas.instructure.com/doc/api/line_items.html
+# - https://canvas.instructure.com/doc/api/score.html
+#
+# - For OAuth stuff to get an access token, see:
+# - https://canvas.instructure.com/doc/api/file.oauth.html#accessing-lti-advantage-services
+# - https://www.imsglobal.org/spec/security/v1p0/#using-oauth-2-0-client-credentials-grant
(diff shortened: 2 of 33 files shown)


app/models/lti_launch.rb
```diff
@@ -9,6 +9,7 @@ class LtiLaunch < ApplicationRecord
   # This is the Redirect URI that must be configured in the Developer Key for the LTI
   # and must match EXACTLY.
   LTI_LAUNCH_REDIRECT_URI="https://#{Rails.application.secrets.application_host}/lti/launch".freeze
+  BRAVEN_ISS="https://#{Rails.application.secrets.application_host}".freeze
```
sadleb, line 12: We need access to the URL for platform to use as the identifier of the "authority" issuing the JWT when requesting an access token. This just makes a way for the lib class to get this URL from an instance of an `LtiLaunch`

lib/lti_id_token.rb
```diff
@@ -26,14 +25,19 @@ class LtiIdToken
...
-    MESSAGE_TYPE_TO_CLASS[message_type].new(payload)
+    parse(payload)
   rescue => e
     Rails.logger.error("{\"Error\":\"#{e.message}\"}")
     Rails.logger.error(e.response.body) if e.is_a?(RestClient::Exception)
     raise
   end
 
+  # Parses the payload into the appropriate LtiRequestMessage
+  def self.parse(payload)
```
sadleb, line 36: Refactored the parsing into a public method so that we can get the parsed payload back out after the launch has been saved and expose it on the `LtiLaunch` as the `request_message`

lib/lti_score.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+# Represents an LTI score message sent to the API. See:
+# https://canvas.instructure.com/doc/api/score.html
+class LtiScore
+  
+  module ActivityProgress
```
sadleb, line 7: Just grouping some constants to use as a pseudo-enum. We should talk about if we want to do enums in a consistent way across the code base.

lib/lti_advantage_api.rb
```diff
@@ -0,0 +1,118 @@
...
+# - https://www.imsglobal.org/spec/lti-ags/v2p0/#introduction
+# - https://canvas.instructure.com/doc/api/line_items.html
+# - https://canvas.instructure.com/doc/api/score.html
+#
+# - For OAuth stuff to get an access token, see:
+# - https://canvas.instructure.com/doc/api/file.oauth.html#accessing-lti-advantage-services
+# - https://www.imsglobal.org/spec/security/v1p0/#using-oauth-2-0-client-credentials-grant
+class LtiAdvantageAPI
+
+  OAUTH_ACCESS_TOKEN_URL = "#{Rails.application.secrets.canvas_cloud_url}#{}/login/oauth2/token"
```
jessouyangbraven, line 20: Remove `#{}`? Unless it's a Ruby thing I'm missing.

lib/lti_advantage_api.rb
```diff
@@ -0,0 +1,118 @@
...
+    @line_item_url = assignment_lti_launch.request_message.line_item_url
+    @global_headers = JSON_HEADERS # get_access_token below needs this initialized
+    @global_headers = @global_headers.merge(:Authorization => "Bearer #{get_access_token}") # Note: these expire about about an hour.
+  end
+
+  # Sends a score / grade for this LTI resource (aka a Project or a Lesson) back
+  # to Canvas so it shows up in the gradebook. See: https://canvas.instructure.com/doc/api/score.html
+  #
+  # Usage:
+  # LtiAdvantageAPI.new(<the_launch>).create_score( LtiScore.generate(...) )
```
jessouyangbraven, line 41: This is probably too nit-picky, but the `LtiAdvantageAPI` naming made me think it'd only have class methods and be stateless. I expected the usage to something like `LtiAdvantageAPI.post(lti_launch, score)`. Essentially, that the API is stateless and I'd always be passing the `lti_launch` around like `request` for a REST client.

lib/lti_advantage_api.rb
```diff
@@ -0,0 +1,118 @@
...
+    @global_headers = @global_headers.merge(:Authorization => "Bearer #{get_access_token}") # Note: these expire about about an hour.
+  end
+
+  # Sends a score / grade for this LTI resource (aka a Project or a Lesson) back
+  # to Canvas so it shows up in the gradebook. See: https://canvas.instructure.com/doc/api/score.html
+  #
+  # Usage:
+  # LtiAdvantageAPI.new(<the_launch>).create_score( LtiScore.generate(...) ) 
+  def create_score(lti_score)
+    response = post(@line_item_url + '/scores', lti_score)
```
jessouyangbraven, line 43: This is a general question that I keep having and forgetting to ask: what's the canonical/correct/safe way of constructing URLs in Rails? I always feel like string concatenation is sketchy, but maybe it's just what you do in Rails?

lib/lti_score.rb
jessouyangbraven: I like consistency in coding conventions. :) It'd be good to figure out what works best for us and what's expected in Rails.

lib/lti_score.rb
```diff
@@ -0,0 +1,56 @@
...
+  #               present.
+  # comment: Comment visible to the student about this score.
+  # activityProgress: Indicate to Canvas the status of the user towards the activity's completion.
+  #                   Must be one of Initialized, Started, InProgress, Submitted, Completed
+  # gradingProgress:  Indicate to Canvas the status of the grading process. A value of
+  #                   PendingManual will require intervention by a grader. Values of NotReady,
+  #                   Failed, and Pending will cause the scoreGiven to be ignored. FullyGraded
+  #                   values will require no action. Possible values are NotReady, Failed, Pending,
+  #                   PendingManual, FullyGraded
+  def self.generate(userId, scoreGiven, scoreMaximum, activityProgress = COMPLETED, gradingProgress = FULLY_GRADED, comment = nil)
```
jessouyangbraven, line 40: Do the arguments here have to be camel case? If not, can we stick to the lower-case and underscore convention?

jessouyangbraven:
Still trying to wrap my head around LTI advantage, but this looks reasonable to me so far.

lib/lti_advantage_api.rb
rshipp: There's not a good way unless we use Addressable, which seems like overkill unless we start running into double-slash issues or something. I ran into the URI.join mess a while ago https://stackoverflow.com/questions/8900782/how-do-i-safely-join-relative-url-segments. 

Personally I always prefer string interpolation over appending, but only bc it's historically more efficient in Python, and more readable to me.

lib/lti_advantage_api.rb
```diff
@@ -0,0 +1,118 @@
...
+  def get_line_items
+    response = get(@line_items_url)
+    JSON.parse(response.body)
+  end
+
+private
+
+  def get_access_token
+    response = post(OAUTH_ACCESS_TOKEN_URL, client_credentials_grant_message)
+    JSON.parse(response.body)['access_token']
```
rshipp, line 59: we ok with this 500'ing if there's an upstream error and we don't get JSON back? (I'm guessing yes, but just want to double check.)

lib/lti_advantage_api.rb
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  # Parses the scope array out of the launch message into the space delimited list expected in the access token grant message.
+  def parse_scope(assignment_lti_launch)
+    assignment_lti_launch.request_message.scope.join(' ')
+  end
+
+  def get(target_url, headers={})
+    RestClient.get(target_url, @global_headers.merge(headers))
+  rescue => e
```
rshipp, line 101: can we be more specific with all these rescues? Not sure if this is the same in Ruby as Python, but bare excepts there catch issues you're not trying to, like typos. RestClient::Exception would catch all restclient errors.

lib/lti_advantage_api.rb
```diff
@@ -0,0 +1,118 @@
...
+    raise ArgumentError.new, 'Missing lineitem. Needed to send a score back for this resource.' unless req_msg.line_item_url.present?
+  end
+
+  # Parses the scope array out of the launch message into the space delimited list expected in the access token grant message.
+  def parse_scope(assignment_lti_launch)
+    assignment_lti_launch.request_message.scope.join(' ')
+  end
+
+  def get(target_url, headers={})
+    RestClient.get(target_url, @global_headers.merge(headers))
```
rshipp, line 100: let's make a codebase-wide rule to just always wrap RestClient calls in a Honeycomb span. more generally speaking, we should wrap anything "long-running".

lib/lti_advantage_api.rb
```diff
@@ -0,0 +1,118 @@
...
+  def post(target_url, body, headers={})
+    RestClient.post(target_url, body, @global_headers.merge(headers))
+  rescue => e
+    handle_rest_client_error(e)
+  end
+
+  def handle_rest_client_error(e)
+    Rails.logger.error("{\"Error\":\"#{e.message}\"}")
+    Rails.logger.error(e.response.body)
+    raise
```
rshipp, line 114: can we send this to Honeycomb too

lib/lti_advantage_api.rb
sadleb: I see what you're saying. The way I'm thinking about this is that that the `lti_launch` is the context that you have to use the API in. It feels more accurate to create a new API context and then call the API methods rather than posting a score to a context? There will likely be other types of posts that we do against the API so your stateless suggestion would be more like `LtiAdvantageAPI.create_score(lti_launch, score)` -- I could be convinced this is more intuitive but for the sake of the cost/benefit I'm going to leave it for now. But consider a refactor when/if we ever consolidate the core `get`, `post`, etc methods that all these RestClient based APIs in the lib dir have more or less copy/pasted.

lib/lti_advantage_api.rb
sadleb: lol, no clue how that got there. fixed.

lib/lti_advantage_api.rb
sadleb: I've been struggling with this too. The internal guidance I've landed on in my own head is 1) avoid `URI.join` (I've also gotten into a mess with it) 2) prefer string interpolation over appending (I broke my own internal rule here) b/c it's more readable 3) use URI() if you have to do naive stuff like get the query params out or add them on 4) use Addressable if you have to get fancy

lib/lti_advantage_api.rb
sadleb: This actually happened once during testing. I got a timeout. I didn't want to proactively try and write error handling code at this stage b/c I wasn't sure how to handle it since we don't have any clients of this written yet. We should think about how to handle errors as cleanly as possible once we actually start writing the consumers of this.

sadleb: E.g. do we need to write a retry mechanism?

lib/lti_score.rb
sadleb: Good call, i was copy/pasting from the message spec. Fixed.

lib/lti_advantage_api.rb
sadleb: I have a placeholder task to refactor all these almost verbatim copied get/post/etc wrappers from the various API classes https://app.asana.com/0/1174274412967132/1186032881622703 -- wondering if we set aside a tech debt pay-down day or two after next weeks demo to do some of that. I could see us "opening" the RestClient class and adding the honeycomb instrumentation and error logging to every call automatically.

sadleb: I added Honeycomb spans to this class for now with the hopes of doing something more DRY as part of that above task. https://github.com/bebraven/platform/pull/287/commits/9046cf9f99221f852cef39e93d9603b243b318d1

lib/lti_advantage_api.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+  end
+
+  def with_error_handling(method, target_url)
+    ret_val = nil
+    Honeycomb.start_span(name: "LtiAdvantageAPI.#{method}") do |span|
+      span.add_field('request.method', method)
+      span.add_field('url', target_url)
+      begin
+        ret_val = yield
```
sadleb, line 117: Executes the passed in block and sets the return value to whatever it returns. This `with_error_handling` method is a technique called "execute around" using blocks.

lib/lti_advantage_api.rb
sadleb: btw, i switched this to string interpolation

lib/lti_advantage_api.rb
sadleb: done

lib/lti_advantage_api.rb
rshipp: I'd avoid retries, that gets messy. We should just try to make sure it's obvious to the user when something breaks and if they should retry.

PR #289, opened 2020-07-27
VCR record and fix RSpec test for LessonContentsController

jessouyangbraven:
**Task**: N/A

**Test**:  `rspec spec/controllers/lesson_contents_controller.spec.rb` passes now

**Details**

- Match VCR recording to `:host` so we can reuse the recording between runs because `lesson_content.lesson_content_zipfile.key` changes
- ~~Need to stub out `LessonContentPublisher:: s3_object_key` because VCR recording relies on the `lesson_content.lesson_content_zipfile.key` being the same, since it's part of the params going through the AWS S3 network request.~~


spec/controllers/lesson_contents_controller.spec.rb
```diff
@@ -62,22 +60,29 @@
...
-      # From: https://www.dwightwatson.com/posts/testing-activestorage-uploads-in-rails-52
       it 'attaches uploaded zipfile' do
-        expect {
-          post :create, params: {state: state, lesson_content_zipfile: lesson_content_zipfile} , session: valid_session
-        }.to change(ActiveStorage::Attachment, :count).by(1)
-        expect(LessonContent.last.lesson_content_zipfile).to be_attached
+        # In order to record the network traffic with VCR, we have to ensure the object key
+        # we're writing to is the same every run, otherwise it changes with the attachment's
+        # active storage key and the recording fails.
+        allow(LessonContentPublisher).to receive(:s3_object_key).and_return('PLACEHOLDER')
```
jessouyangbraven, line 78: This was the tricky part to figure out to get VCR recording to work for all runs of the test.

spec/controllers/lesson_contents_controller.spec.rb
```diff
@@ -13,13 +13,7 @@
 
   # TODO: https://app.asana.com/0/1174274412967132/1184057808812010
   let(:valid_session) { {} } 
-
   let(:state) { SecureRandom.uuid }
-  let(:target_link_uri) { 'https://target/link' }
```
jessouyangbraven, line 18: Refactored this to tighten up the scope of variables a bit becauseI kept shooting myself in the foot with the wrong `lesson_content_*`.

spec/controllers/lesson_contents_controller.spec.rb
rshipp: probably too late now, but you can also change the request matching https://relishapp.com/vcr/vcr/v/5-1-0/docs/request-matching to something less strict like :host, if this is an issue with the URI changing

jessouyangbraven: Thanks @rshipp! Updated to match on `:host`.

PR #292, opened 2020-07-28
Add bundle no-deployment flag, to allow rebuilds

rshipp:
See: https://stackoverflow.com/questions/6300613/whats-the-difference-between-bundle-install-deployment-and-bundle-pack

This change allows the `rebuild.sh` script to work when you add new gems to the Gemfile. Hopefully without any unwanted side-effects.


rshipp:
What's production mode, like the rails env? I don't think that's related to this.

sadleb:
> What's production mode, like the rails env? I don't think that's related to this.

Yeah, changing the RAILS_ENV to prod.

sadleb:
I'm just going to merge this so I can start trying it out.

PR #293, opened 2020-07-28
Adds a reverse proxy to the AWS S3 hosted files for the LTI lessons.

sadleb:
Now we can give Canvas a URL to launch on our server for the Rise360
lesson content which is hosted on AWS S3 and our server will proxy the
request/response so that both Canvas and the browser think the files
live on our server which we can more easily control the trusting relationship
we've developed. We need this b/c the content tries to Ajax request xApi statements
to the LRS but since it's cross-origin things go awry. The above task has more 
context on the discussion: https://app.asana.com/0/1174274412967132/1186566037117865/f

Example:
https://platformweb/lesson_contents_proxy/lessons/ytec17h3ckbr92vcf7nklxmat4tc/index.html?actor=%7B%22name%22%3A%22LESSON_CONTENTS_USERNAME_REPLACE%22%2C%20%22mbox%22%3A%5B%22mailto%3ALESSON_CONTENTS_PASSWORD_REPLACE%22%5D%7D&endpoint=https%3A%2F%2Fplatformweb%2Fdata%2FxAPI

### Test Plan:
Load an existing or new Lesson and look at the network traffic. It should successfully send xApi statements to the LRS. Login to the LRS to verify. Note that in an incognito window you have to allow third party cookies for all this LTI stuff to work.

### DIFF
--- a/app/models/lesson_content.rb
+++ b/app/models/lesson_content.rb
@@ -8,10 +8,7 @@ class LessonContent < ApplicationRecord
   has_one_attached :lesson_content_zipfile
 
   def launch_url
-    # TODO: in a future iteration we'll want to generate pre-signed URLs with expirations on
-    # them so that our content doesn't get leaked for anyone to access. Right now the bucket is public
-    # but we'll want to lock it down.
-    LessonContentPublisher.launch_url(lesson_content_zipfile.key)
+   "#{LtiLessonContentsProxy.proxy_url}#{LessonContentPublisher.launch_path(lesson_content_zipfile.key)}"
   end
    
   private
--- a/config/application.rb
+++ b/config/application.rb
@@ -2,6 +2,8 @@ require_relative 'boot'
 
 require 'rails/all'
 
+require_relative '../lib/lti_lesson_contents_proxy'
+
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
@@ -16,13 +18,8 @@ module Platform
     # be backwards compatible with v5.2 and below who may have removed it from ApplicationController
     config.action_controller.default_protect_from_forgery = true
 
-    # Allows Rise 360 lesson running in AWS S3 to call into our LRS xAPI proxy
-    config.middleware.insert_before 0, Rack::Cors do
-      allow do
-        origins "https://#{Rails.application.secrets.aws_files_bucket}.s3.amazonaws.com"
-        resource '*', :headers => :any, :methods => [:get, :put,]
-      end
-    end
+    # Allows us to serve Rise360 static files hosted on AWS S3 through our server avoiding browser cross-origin issues.
+    config.middleware.use LtiLessonContentsProxy, backend: "https://#{Rails.application.secrets.aws_files_bucket}.s3.amazonaws.com"
 
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration can go into files in config/initializers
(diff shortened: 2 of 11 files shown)


app/models/lesson_content.rb
```diff
@@ -8,10 +8,7 @@ class LessonContent < ApplicationRecord
   has_one_attached :lesson_content_zipfile
 
   def launch_url
-    # TODO: in a future iteration we'll want to generate pre-signed URLs with expirations on
-    # them so that our content doesn't get leaked for anyone to access. Right now the bucket is public
-    # but we'll want to lock it down.
-    LessonContentPublisher.launch_url(lesson_content_zipfile.key)
```
sadleb, line 14: I just moved this comment closer to the source, it's still valid.

config/application.rb
```diff
@@ -16,13 +18,8 @@ class Application < Rails::Application
     # be backwards compatible with v5.2 and below who may have removed it from ApplicationController
     config.action_controller.default_protect_from_forgery = true
 
-    # Allows Rise 360 lesson running in AWS S3 to call into our LRS xAPI proxy
-    config.middleware.insert_before 0, Rack::Cors do
-      allow do
-        origins "https://#{Rails.application.secrets.aws_files_bucket}.s3.amazonaws.com"
-        resource '*', :headers => :any, :methods => [:get, :put,]
-      end
-    end
```
sadleb, line 25: We don't need this anymore b/c it's no longer cross origin from the browser's perspective, but let's keep this solution in mind in case we run into something else.

lib/lesson_content_publisher.rb
```diff
@@ -8,9 +8,13 @@ class LessonContentPublisher
   # Publicly accessible URL for the lesson
   # The "filekey" is the "key" attribute of an ActiveStorage zipfile.
   # Aka the name of the zipfile on S3.
```
sadleb, line 10: TODO: update this comment.

config/routes.rb
```diff
@@ -1,3 +1,6 @@
+# Needed to use the url_helpers outside of views and controller
+Rails.application.routes.default_url_options[:host] = Rails.application.secrets.application_host
```
sadleb, line 2: So I didn't end up using it b/c i needed https and not http, but fyi @jessouyangbraven this line makes it so that from `lib` classes going forward you can get access to the root_url using `Rails.application.routes.url_helpers.root_url`

config/routes.rb
```diff
@@ -98,4 +101,7 @@
 
   # Proxy xAPI messages to the LRS.
   match '/data/xAPI/*endpoint', to: 'lrs_xapi_proxy#xAPI', via: [:get, :put]
+
+  # There is a route similar to the commented out one below that doesn't show up here. See 'lib/lti_lesson_contents_proxy.rb' and 'config/application.rb'
```
jessouyangbraven, line 105: Thank you for adding this. I would have been hunting forever to figure out the proxying otherwise.

lib/lti_lesson_contents_proxy.rb
```diff
@@ -0,0 +1,39 @@
...
+class LtiLessonContentsProxy < Rack::Proxy
+
+  PROXIED_PATH='/lesson_contents_proxy'
+  PROXY_REGEX=/^#{Regexp.escape(PROXIED_PATH)}(\/.*)$/
+
+  # E.g. https://platformweb/lesson_contents_proxy
+  def self.proxy_url
+    # This is hard-coded to HTTPS b/c none of this stuff works over HTTP and in dev thats the default
+    # if we use a helper URL like: Rails.application.routes.url_helpers.root_url
+    @proxl_url ||= "https://#{Rails.application.secrets.application_host}#{PROXIED_PATH}"
```
jessouyangbraven, line 25: Should this be `@proxy_url`?

sadleb: typo king this week. fixed

PR #299, opened 2020-07-30
Fix livereloading controllers

rshipp:
Fixes the issue with controller changes not being detected by Guard and live-reloaded.


jessouyangbraven:
Thank you!!!!!

PR #300, opened 2020-07-30
Implement platform sign-in from LTI launch

rshipp:
Sign in users automatically when they come through a valid LTI launch.

Task: https://app.asana.com/0/1174274412967132/1184057808812010/f

This ended up being a lot easier than I thought it was. For future reference, all we have to do to sign in users that have already been authenticated is ... call `sign_in`.

Note: In order to see the new auto-login behavior, you will need a local account that has a canvas_id matching the ID of your canvas user. For example, my canvas user's ID is `81`, so I created a new local platformweb user with `canvas_id: 81`.


app/controllers/lti_launch_controller.rb
```diff
@@ -55,6 +57,24 @@ def launch
...
+
     redirect_to target_uri_with_params.to_s
   end
+
+  private
+
+  def sign_in_from_lti(ll)
+    # Grab the user email out of the payload and tell devise that they are authenticated for this session.
+    idt = LtiIdToken.parse(ll.id_token_payload)
+    # NOTE: Crash with KeyError if no user supplied.
```
rshipp, line 69: is this the correct behavior?

app/controllers/lti_launch_controller.rb
```diff
@@ -55,6 +57,24 @@ def launch
...
+    username = idt.custom.fetch('user_email')
+    user = User.where(email: username).first
+
+    if user
+      sign_in user
+      Rails.logger.debug("Done signing in LTI-authenticated user #{username}")
+    else
+      Rails.logger.debug("Invalid user came through LTI launch: #{username}")
```
rshipp, line 77: since there's no sign-in, the default behavior here is to take the user back to the login page. this shouldn't happen in a real setup, because we *should* have matching users/emails in Canvas and Platform.

app/controllers/lti_launch_controller.rb
sadleb: You can just do `ll.request_message.custom` instead of using LtiIdToken to parse it.

rshipp: context: `user_email` is a custom field supplied by Canvas. Presumably, the only way for it to not exist is if Canvas changes the way their LTI payload is set up, in which case this would start failing for all users. Us raising a 500 seems high-visibility.

sadleb: I would assume there would always be a user in the custom claim part of the message. Any clue what scenario it was missing?

rshipp: I saw the keyerror when I put the wrong custom field name (`canvas_email` instead of `user_email`)

app/controllers/lti_launch_controller.rb
```diff
@@ -55,6 +57,24 @@ def launch
...
     redirect_to target_uri_with_params.to_s
   end
+
+  private
+
+  def sign_in_from_lti(ll)
+    # Grab the user email out of the payload and tell devise that they are authenticated for this session.
+    idt = LtiIdToken.parse(ll.id_token_payload)
+    # NOTE: Crash with KeyError if no user supplied.
+    username = idt.custom.fetch('user_email')
```
sadleb, line 70: I feel like using the canvas user ID would be more reliable in case they add their own alternate emails to their account? We probably want to figure out a way to disable that if we're going to use SSO though since the alternate emails won't work for login, just notifications. Using the canvas ID makes dev a little more painful at the moment since our users aren't provisioned through the registration flow, so we would need to set it manually to match.

app/controllers/lti_launch_controller.rb
```diff
@@ -55,6 +57,24 @@ def launch
...
+    user = User.where(email: username).first
```
sadleb, line 71: Note that there is a `find_by!` that will just return one so you don't have to call first.

sadleb:
I agree that we should write tests before merging.

app/controllers/lti_launch_controller.rb
sadleb: Ahh, ok. Phew.

app/controllers/lti_launch_controller.rb
rshipp: Yeah, I think joining by canvas id makes sense here.

spec/factories/users.rb
```diff
@@ -25,6 +25,11 @@
       factory :admin_user do
         admin { true }
       end
+
+      factory :with_canvas_id_user do
```
rshipp, line 29: this name is awkward but i think it has to end with `_user`

spec/controllers/lti_launch_controller_spec.rb
```diff
@@ -29,20 +29,41 @@
...
+      end
+
+      it 'redirects to the target_link_uri with the state param' do
+        expect(response).to redirect_to(target_link_uri + "?state=#{state}")
+      end
+
+      it 'signs in the user with matching canvas_id' do
+        Rails.logger.warn(controller.current_user.inspect)
+        expect(controller.current_user).to eq lti_user
+      end
```
rshipp, line 53: this is the success test - the rest of the green above is because i had to put everything inside the `context`.

spec/controllers/lti_launch_controller_spec.rb
```diff
@@ -29,20 +29,41 @@
...
+        stub_request(:get, LtiIdToken::PUBLIC_JWKS_URL).to_return(body: public_jwks)
+        post :launch, params: lti_launch_params
+      end
 
-    it 'redirects to the target_link_uri with the state param' do
-      expect(response).to redirect_to(target_link_uri + "?state=#{state}")
-    end 
+      it 'does not sign in the user' do
+        expect(controller.current_user).to eq nil
+      end
```
rshipp, line 65: this is the failure test, rest is boilerplate

app/controllers/lti_launch_controller.rb
```diff
@@ -55,6 +55,21 @@ def launch
...
     redirect_to target_uri_with_params.to_s
   end
+
+  private
+
+  # Grab the user ID out of the payload and tell devise that they are authenticated for this session.
+  def sign_in_from_lti(ll)
+    # NOTE: Crash with KeyError if no user supplied.
+    canvas_id = ll.request_message.custom.fetch('user_id')
+    if user = User.find_by(canvas_id: canvas_id)
```
sadleb, line 68: Are you purposely doing `find_by` instead of `find_by!` to avoid it throwing an error and just redirecting you to login instead?

rshipp: yep!

spec/controllers/lti_launch_controller_spec.rb
```diff
@@ -29,20 +29,41 @@
...
+        lti_launch = LtiLaunch.current(state)
+        expect(lti_launch.id_token_payload).to eq(JSON.parse(id_token_payload))
+      end
+
+      it 'redirects to the target_link_uri with the state param' do
+        expect(response).to redirect_to(target_link_uri + "?state=#{state}")
+      end
+
+      it 'signs in the user with matching canvas_id' do
+        Rails.logger.warn(controller.current_user.inspect)
```
sadleb, line 51: Accidentally left in?

rshipp: 😞

spec/factories/users.rb
```diff
@@ -25,6 +25,11 @@
...
+        # Match custom.user_id in :lti_link_launch_request factory
```
sadleb, line 30: Ideally we should use a transient variable for this.... not a big deal though, i've done things like this out of laziness b/c the cost/benefit wasn't there.

rshipp: I want to ask about this later bc I don't know how to do this and it seems useful

app/controllers/lti_launch_controller.rb
```diff
@@ -55,6 +55,21 @@ def launch
...
+      sign_in user
+      Rails.logger.debug("Done signing in LTI-authenticated user #{user.email}")
+    else
+      Rails.logger.debug("Invalid user came through LTI launch: Canvas ID #{canvas_id}")
```
jessouyangbraven, line 72: Thanks for adding this debug line. I totally forgot to set my user's canvas_id.

PR #303, opened 2020-08-03
Add span around unzip/upload

rshipp:
Task: https://app.asana.com/0/1174274412967132/1185562642403857/f

Adds a span (so we get timing information) and a couple fields (zip file path & size) to Honeycomb when unzipping/uploading Rise360 zips. We already get filename information for all the files inside the zip from the s3 autoinstrumentation.


lib/lesson_content_publisher.rb
```diff
@@ -26,14 +26,19 @@ def self.publish(zipfile)
...
 
       zipfile.open do |file|
-        Zip::File.open(file.path) do |zip_file|
-          zip_file.each do |entry|
-            next unless entry.file?
-            s3_object = bucket.object(s3_object_key(zipfile.key, entry.name))
-            s3_object.put({
-                acl: "public-read",
-                body: entry.get_input_stream.read,
-            })
```
rshipp, line 36: (this is all just indented)

PR #305, opened 2020-08-04
Suspend callback on user admin update

skaex:
### Description
This PR suspends `do_account_registration` on user admin update. This was causing a salesforce error.

### Motivation and Context
This is was reported to me by @rshipp.

### Checklist:


app/controllers/admin/users_controller.rb
```diff
@@ -44,7 +44,12 @@ def edit; end
 
   def update
     filtered_user_params = user_params.reject { |_, v| v.blank? }
-    if @user.update_attributes(filtered_user_params)
```
sadleb, line 47: Would it be cleaner to use `https://apidock.com/rails/ActiveRecord/Persistence/update_columns` instead of setting and unsetting the callback? You'd probably want to manually pass in `updated_at` though.

Another option would be to loop through and use `update_attribute` (singular) but I don't think that's as nice. https://guides.rubyonrails.org/active_record_validations.html#skipping-validations

If you decide to stick with this approach, I would at least put the `set_callback` in an ensure block. It's highly unlikely, but just good practice for something like this. E.g. maybe there is a temporary database connection failure just at the moment this is run and `update_attributes` fails. Then would validations be disabled?

skaex: To answer your question: `update_attributes` calls the soft [`save`](https://apidock.com/rails/v5.2.3/ActiveRecord/Persistence/save) which has a `rescue` and it returns a boolean. I think that handles the scenario you're worried about. I don't like the first approach since it does not run validation when updating.

sadleb: kk, if `update_attributes` is guaranteed **not** to throw, then it's fine as is.

PR #308, opened 2020-08-04
Fix links to versions in project index

jessouyangbraven:
**Task** N/A

**Description**: There's no other way to preview versions of a course, so I fixed the existing one linked in the index. 
 - Put version number in the link
 - Render preview without inputs if we don't have `state` from an LTI launch

Screen Shot 2020-08-04 at 1 30 10 PM
Screen Shot 2020-08-04 at 1 30 13 PM

**Test**

- Also check that SpeedGrader view in Canvas still works and populates submitted fields:
Screen Shot 2020-08-04 at 1 34 21 PM


app/controllers/course_content_histories_controller.rb
```diff
@@ -17,6 +17,11 @@ def index
   # GET /course_contents/:id/versions/1
   # GET /course_contents/:id/versions/1.json
   def show
+    if !params[:state].present?
+      @body_with_user_inputs = @course_content_history.body
+      return
```
sadleb, line 22: I don't think it should hold up this getting merged, but I'm personally more of a fan of `if ... else ... end` than short circuiting out, at least for the main logic of a method. Short circuiting is fine imo when there are initial checks for things like valid state or authorization, but when the method logic is `if state` do this `else` do that `end` I think it's more readable.

jessouyangbraven: Fair. When I do short-circuit, it's avoid deep indentation and having to think about compound conditions. Not the case here, obviously. 

I'm hoping this all goes away when we separate out project submissions, though.

PR #310, opened 2020-08-05
Move xAPI to JS

rshipp:
Reimplementing xAPI assignment project logic in JS instead of Rails.

Task: https://app.asana.com/0/1174274412967132/1187336478107377/f

Main things to review:

* overall JS structure
  * the `lrs` var and how it's used
  * the `javascript_variables` not being available as a const
  * naming of the JS functions / callback
* user_override param
  * high-level, is this the best way to do this?
  * adding it in the JS with `lrs.extended`
  * parsing it in the controller (is this the right place?)
  * explicitly removing it in the lib
* tests
  * do we need selenium tests for loading existing answers and/or the TA view? or are the unit tests good enough?

## Screenshots
<details>
<summary>Student view</summary>
Screen Shot 2020-08-11 at 11 16 17 AM
</details>

<details>
<summary>TA view</summary>
Screen Shot 2020-08-11 at 11 26 09 AM
</details>


app/javascript/packs/xapi_assignment.js
```diff
@@ -89,10 +89,63 @@ function sendStatement(e) {
...
+            verb: new tincan.Verb({
+                id: "http://adlnet.gov/expapi/verbs/answered"
+            }),
+            agent: new tincan.Agent({
+                // Note: We overwrite this actor info server-side.
+                name: "JS_ACTOR_NAME_REPLACE",
+                mbox: "JS_ACTOR_MBOX_REPLACE"
+            }),
+            activity: activity_id
+        },
```
sadleb, line 137: FYI: In my implementation I passed the `ascending` param so that the last statement encountered when looping over them is the most recent.

rshipp: yeah this logic is the opposite. i set the first (most recent) as the value in the dom, then ignore all the rest.

app/javascript/packs/xapi_assignment.js
```diff
@@ -89,10 +109,64 @@ function sendStatement(e) {
...
-});
+// Fill out inputs with previously submitted answers, fetched from the LRS.
+function populatePreviousAnswers(callback) {
+    const javascript_variables = document.getElementById('javascript_variables').attributes;
+    const project_lti_id = javascript_variables['data-project-lti-id'].value;
+    const activity_id = project_lti_id; // e.g. https://braven.instructure.com/courses/48/assignments/158
+
+    if (javascript_variables['data-student-id']) {
+        // This is the TA view.
+        lrs.extended = {
```
jessouyangbraven, line 153: I couldn't find `extended` in the TinCan API. Is this wha sends over `user_override` as a query param to the `LrsXapiProxyController`?

app/javascript/packs/xapi_assignment.js
```diff
@@ -89,10 +109,64 @@ function sendStatement(e) {
...
+                id: "http://adlnet.gov/expapi/verbs/answered"
+            }),
+            agent: new tincan.Agent({
+                // Note: We overwrite this actor info server-side.
+                name: "JS_ACTOR_NAME_REPLACE",
+                mbox: "JS_ACTOR_MBOX_REPLACE"
+            }),
+            activity: activity_id
+        },
+        callback: callback
```
jessouyangbraven, line 170: Is there a reason why we need the argument instead of doing `callback: populateAnswersCallback` like we do in the `lrs.moreStatements()` call?

jessouyangbraven:
In the course of reviewing, I tried to find resources on best Javascript in Rails practice. There wasn't a definitive answer. 

It broke down into three main camps, none of which seemed like a good fit for what we're doing: 
- write jQuery/CoffeeScript (https://brandonhilkert.com/blog/page-specific-javascript-in-rails/, https://guides.rubyonrails.org/v4.2/working_with_javascript_in_rails.html)
- use (Turbo)links (https://backend-development.github.io/javascript_and_ajax.html), or 
- use the `data-behavior` attribute that I couldn't find documentation on (https://brandonhilkert.com/blog/organizing-javascript-in-rails-application-with-turbolinks/)

app/javascript/packs/xapi_assignment.js
rshipp: It is, and it's also completely undocumented. I should go add a link to where I found this in the source code.

app/javascript/packs/xapi_assignment.js
rshipp: I moved it into an argument while trying random things to get the tests working, we don't need it. I wasn't sure if it was better to pass it in or hardcode it.

app/javascript/packs/xapi_assignment.js
```diff
@@ -17,13 +20,30 @@ try {
...
+    const inputs = document.querySelectorAll('textarea,input[type="text"]');
+    const javascript_variables = document.getElementById('javascript_variables').attributes;
+
+    inputs.forEach(input => {
+        input.onblur = sendStatement;
+        if (javascript_variables['data-student-id']) {
+            // TA view - mark all inputs as read-only.
+            input.disabled = true;
+        }
+    });
```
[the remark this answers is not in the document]
rshipp: Good point, there's no reason to add the send handler in the TA view.

app/javascript/packs/xapi_assignment.js
```diff
@@ -17,13 +20,30 @@ try {
     // Send this to Honeycomb?
 }
 
+// Main page logic.
+document.addEventListener('DOMContentLoaded', () => {
+    // Attach the xAPI function to all appropriate inputs, and load any existing data.
+    const inputs = document.querySelectorAll('textarea,input[type="text"]');
+    const javascript_variables = document.getElementById('javascript_variables').attributes;
```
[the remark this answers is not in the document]
rshipp: This is something @sadleb introduced. They are static in the DOM, so we can use them as constants. I thought about making them a const at the top of the file, but technically they're not guaranteed to exist until inside the DOMContentLoaded event handler, so I think I'd have to make them a global var? This is one part I was extra unhappy about in the JS.

app/helpers/course_content_histories_helper.rb
```diff
@@ -1,73 +1,2 @@
-require 'base64'
-
 module CourseContentHistoriesHelper
```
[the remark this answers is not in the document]
rshipp: is that allowed? can I delete all the empty helpers?

app/javascript/packs/xapi_assignment.js
```diff
@@ -1,8 +1,11 @@
 import tincan from 'tincanjs';
 
-// Connect to the LRS.
-var lrs;
+const INPUT_ID_ATTR = 'data-bz-retained';
+
+export var lrs;
```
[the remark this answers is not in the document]
rshipp: is that possible inside a try/catch? I'll try it out and see what happens.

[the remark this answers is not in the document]
rshipp: This one didn't work, "export must be at the top level"

lib/lrs_xapi_proxy.rb
```diff
@@ -24,7 +24,8 @@ def self.request(request, path, user)
       return unless request.method == 'GET' || request.method == 'PUT'
 
       # Rewrite query string
-      params = request.query_parameters
+      # TODO: Hardcoding a single parameter?? Make this clearer at least.
+      params = request.query_parameters.except(:user_override)
```
rshipp, line 28: is there a better way to do this?

spec/feature/lrs_xapi_proxy/project_xapi_spec.rb
```diff
@@ -48,6 +51,26 @@
...
+       #  }
+
+       #  VCR.use_cassette('lrs_xapi_proxy_load_previous', :match_requests_on => [:path, :method], :erb => lrs_variables) do
+       #    visit return_service
+       #    expect(page).to have_selector("[data-bz-retained=\"#{question_id}\"]")
+       #    puts find_field('test-question-id').value
+       #    expect(page).to have_field('test-question-id')
+       #    expect(page).to have_field('test-question-id', with: unique_string)
+       #  end
+       #end
```
rshipp, line 72: can't tell why this test doesn't work. I thought it would be nice to have it, but maybe we don't bother writing selenium tests for everything? This already has unit tests.

spec/javascript/packs/xapi_assignment.test.js
```diff
@@ -0,0 +1,320 @@
...
+                            definition: {
+                                name: {
+                                    und: 'test-id'
+                                }
+                            }
+                        }
+                    }
+                ]
+            })
+        );
```
rshipp, line 38: these big json blobs are annoying, but the alternative is setting up our own fixture type thing for JS. overkill? maybe something to revisit later if we end up needing it more often.

app/controllers/lrs_xapi_proxy_controller.rb
```diff
@@ -7,8 +7,11 @@ class LrsXapiProxyController < ApplicationController
   def dry_crud_enabled?() false end
 
   def xAPI
-    response = LrsXapiProxy.request(request, request.params['endpoint'], current_user)
+    # TODO: Make sure current_user is allowed to view user_override's stuff.
+    # https://app.asana.com/0/1174274412967132/1185569091008475/f
+    user = params[:user_override] ? User.find(params[:user_override]) : current_user
```
rshipp, line 12: how do we feel about this user_override param I'm using? that's the main magic I'm doing in this PR- if there's a better way, I'm open to it.

rshipp:
> My biggest questions are (1) whether we're going to need to fetch statements at the granularity of submissions rather than input fields (I think Harry said this a while back) and (2) if we can straightforwardly adapt this code for that situation in the future.
I agree this is worth thinking about, and don't have an answer. The xAPI query api is pretty limited. We could maybe do something with timestamps, and use answers before a `submitted_at`?

app/javascript/packs/xapi_assignment.js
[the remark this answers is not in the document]
sadleb: Ditto. I'm not super happy about this but I couldn't figure out a better way that seemed cleaner / more maintainable / easier to read.

app/javascript/packs/xapi_assignment.js
```diff
@@ -89,10 +110,67 @@ function sendStatement(e) {
...
+        console.log("Failed to query statements: " + err);
+        // TODO: Do something with error, didn't get statements.
+        return;
+    }
 
-// Attach the xAPI function to all appropriate inputs.
-var inputs = document.querySelectorAll('textarea,input[type="text"]');
+    // Fill empty fields with the first (most recent) response.
+    // Ignore all previous responses once a field is non-empty.
+    // Note: If you empty a field after filling it, this will populate the last non-empty result.
```
sadleb, line 124: This is going to generate a "bug" at some point. E.g. TA is like "why did you type 'wtf?!?!?'" in there and the student will be like WTF?!? This seems like a pain to fix though so I queued up a task and we'll tackle later. https://app.asana.com/0/1174274412967132/1188841364384041

spec/feature/lrs_xapi_proxy/project_xapi_spec.rb
sadleb: Yeah, I don't think we need selenium tests for everything. Just the most common stuff and main code paths in the code base.

spec/javascript/packs/xapi_assignment.test.js
sadleb: I'm fine with the inline blobs for now and revisit later. It would be nice to break the 4 spaces indent rule for these b/c two makes it more readable imo. Wouldn't want to mess with the linter just to do that though.

app/controllers/lrs_xapi_proxy_controller.rb
sadleb: I'm good with it. Very similar to how canvas does stuff with tons of XXXXoverride objects. I did push a comment here making it clearer as to what is going on though.

lib/lrs_xapi_proxy.rb
sadleb: I can't think of one. I've had to do this in the past, and it's the purpose of the `except` method, where you use almost all the params for initializing the model but you have some others used for different purposes. I added a clarifying comment though.

app/javascript/packs/xapi_assignment.js
```diff
@@ -89,10 +110,67 @@ function sendStatement(e) {
...
+}
+
+// Fill out inputs with previously submitted answers, fetched from the LRS.
+function populatePreviousAnswers() {
+    const javascript_variables = document.getElementById('javascript_variables').attributes;
+    const project_lti_id = javascript_variables['data-project-lti-id'].value;
+    const activity_id = project_lti_id; // e.g. https://braven.instructure.com/courses/48/assignments/158
+
+    if (javascript_variables['data-user-id']) {
+        // This is the TA view.
```
sadleb, line 153: We want to fill out previously submitted answers for Fellows too. Does the JS do that?

sadleb:
> > My biggest questions are (1) whether we're going to need to fetch statements at the granularity of submissions rather than input fields (I think Harry said this a while back) and (2) if we can straightforwardly adapt this code for that situation in the future.
> 
> I agree this is worth thinking about, and don't have an answer. The xAPI query api is pretty limited. We could maybe do something with timestamps, and use answers before a `submitted_at`?

We will have this requirement and I agree it's going to need some though. The task is here: https://app.asana.com/0/1174274412967132/1187445581799819

app/javascript/packs/xapi_assignment.js
rshipp: Yes, that's what this function and the callback one above do. Line 129 is what actually fills the fields with the response value.

app/javascript/packs/xapi_assignment.js
```diff
@@ -15,15 +18,33 @@ try {
...
+    const modifyInputElementForViewer = javascript_variables['data-user-id']
+        // TA view
+        ? function(input) { input.disabled = true }
+        // Student view
+        : function(input) { input.onblur = sendStatement };
+
+    inputs.forEach(input => modifyInputElementForViewer(input));
+
+    populatePreviousAnswers(populateAnswersCallback);
+});
```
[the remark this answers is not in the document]
rshipp: I mostly just forgot about it. Do we want to use that class? I wasn't sure why we were turning all the fields blue.

PR #311, opened 2020-08-06
Refactor spec refactories to be more intuitive and allow overriding values

sadleb:
`lti_launch_assignment_message` and `lti_launch_assignment_selection_message`
are more intuitive than what we had before. Also, the base class now has only the
shared stuff and these classes now have what is pertinent to them. This will allow
us to more easily test LRS stuff that relies on these messages being correct and
knowing the values of their innards. E.g. activity_id for the xAPI messages.

Now you can do things like
```
let!(:lti_assignment_launch_msg) { FactoryBot.json(:lti_launch_assignment_message, assignment_id: 1234, course_id: 55) } 
```
to get a launch message with those values in the payload.
### Test Plan:
- bundle exec rspec


spec/factories/lti_launch.rb
```diff
@@ -1,10 +1,16 @@
+require 'rails_helper'
```
rshipp, line 1: this appears to be breaking the tests

sadleb: Hmm. I think I was getting local errors b/c I was editing some comments and stuff while the specs were running. Let's try it without this...

PR #316, opened 2020-08-10
Add Warden authentication strategy for LtiLaunch'ed requests.

sadleb:
Task: https://app.asana.com/0/1174274412967132/1186566037117867

This fixes the issue where you have to enable third party cookies in Chrome incognito windows (and it would fix any other browsers that don't allow third party cookies).

Note the loose end / TODO: https://app.asana.com/0/1174274412967132/1184057808812020

### Test Plan:
- Launch both a Lesson and a Project in Chrome incognito with third party cookies disabled (default setting). They should open and data should be sent to the LRS.
- Do the same with third party cookies enabled and notice how all the logs of `Authenticated user_id = XX using LtiAuthentication::WardenStrategy` go away.
- Try accessing a page with an invalid state query param. It should error.
- `bundle exec rspec`


spec/controllers/lti_launch_controller_spec.rb
```diff
@@ -1,13 +1,13 @@
...
 RSpec.describe LtiLaunchController, type: :controller do
-  let(:state) { SecureRandom.uuid }
+  let(:state) { LtiLaunchController.generate_state }
 
   describe 'POST #login' do
     let(:lti_launch_params) { build(:lti_launch_login_params) }
 
     before(:each) do
-      allow(SecureRandom).to receive(:uuid).and_return(state)
+      allow_any_instance_of(LtiLaunchController).to receive(:generate_state).and_return(state)
```
rshipp, line 10: This `allow_any_instance_of` function is [deprecated](https://relishapp.com/rspec/rspec-mocks/v/3-3/docs/working-with-legacy-code/any-instance), is there any way around using it here?

app/javascript/packs/xapi_assignment.js
```diff
@@ -7,8 +7,9 @@ try {
     lrs = new tincan.LRS({
         // Note: We overwrite this auth server-side.
         endpoint: `${window.location.origin}/data/xAPI`,
-        username: "JS_USERNAME_REPLACE",
-        password: "JS_PASSWORD_REPLACE",
+        // Note: this is the authorization to hit the platform server. It doesn't use this when connecting
+        // to the LRS. The LRS authorization header is set by the LrsXapiProxy.
+        auth: 'LtiState '+ document.getElementById('javascript_variables').attributes['data-lti-auth-state'].value,
```
rshipp, line 12: I think this will fail on the course_content_histories view, unless it's added in there too

config/initializers/lti_authentication.rb
```diff
@@ -0,0 +1,55 @@
...
+      @lti_state.present?
+    end
+
+    def authenticate!
+     # TODO: cache this for an hour. Or a day? It's hit alot.
+     # https://app.asana.com/0/1174274412967132/1188248367583275
+     if ll = LtiLaunch.is_valid?(@lti_state)
+       user = ll.user
+       if user
+         Rails.logger.debug("Authenticated user_id = #{user.id} using LtiAuthentication::WardenStrategy")
```
rshipp, line 21: can we add Honeycomb stuff for all this too? you can add info to the existing trace with `Honeycomb.add_field('interesting_thing', 'banana')`

config/initializers/devise.rb
```diff
@@ -292,11 +289,18 @@
...
+      # For requests made in the context of an authenticated LtiLaunch, use this
+      # authentication strategy to validate that the request is authenticated
+      # based on the LTI Launch handshake. Note that this only applies if the browser
+      # doesn't accept third party cookies. If it does, the lti_launch_controller will
+      # call sign_in user which stores the authentication in session and bypasses the need
+      # to check authentication.
+      manager.strategies.add(:lti_auth, LtiAuthentication::WardenStrategy)
+      manager.default_strategies(scope: :user).unshift :lti_auth
+      manager.failure_app = DeviseCasAuthenticatable::SingleSignOut::WardenFailureApp
+    end
```
jessouyangbraven, line 303: Nit: is this tabbed funky?

config/initializers/lti_authentication.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  class WardenStrategy < Warden::Strategies::Base
+    
+    # True if this strategy should be run for this request
+    def valid?
+      @lti_state = fetch_state
+      @lti_state.present?
+    end
+
+    def authenticate!
```
jessouyangbraven, line 15: Nit: can we structure this to fail fast and error out? E.g.: 

```
unless ll = LtiLaunch.is_valid?(@lti_state)
  Rails.logger.warn("LtiAuthentication::WardenStrategy couldn't find LtiLaunch with state = '#{@lti_state}'")
  return handle_failure(:forbidden)
end

unless user = ll.user
  Rails.logger.warn("LtiAuthentication::WardenStrategy couldn't find user with canvas_id = #{ll.request_message.canvas_user_id}")
  return handle_failure(:forbidden)
end

Rails.logger.debug("Authenticated user_id = #{user.id} using LtiAuthentication::WardenStrategy")
success(!user)
```

config/initializers/devise.rb
```diff
@@ -292,11 +289,18 @@
   # ==> Warden configuration
   # If you want to use other strategies, that are not supported by Devise, or
   # change the failure app, you can configure them inside the config.warden block.
-  #
-  # config.warden do |manager|
+  config.warden do |manager|
   #   manager.intercept_401 = false
```
jessouyangbraven, line 293: Nit: delete?

app/javascript/packs/xapi_assignment.js
```diff
@@ -7,8 +7,9 @@ try {
     lrs = new tincan.LRS({
         // Note: We overwrite this auth server-side.
```
jessouyangbraven, line 8: Nit: Delete comment. I think it's redundant/obsolete with the one below. If not, I'm confused with the many layers of authentication.

app/javascript/packs/xapi_assignment.js
jessouyangbraven, line 12: I'm worried that this is going to get out of sync with `LtiAuthentication::WardenStrategy::fetch_state` expects in it's regex, especially since this can't use `LtiAuthentication::LTI_AUTH_HEADER_PREFIX` in JS-land. 

Does it make sense to construct this server-side in `course_contents_controller#show` and pass the entire string over so it's opaque to JS?

config/initializers/lti_authentication.rb
jessouyangbraven: +1

@rshipp, do you have any naming conventions for the Honeycomb fields?

spec/controllers/lti_launch_controller_spec.rb
sadleb: I lucked out that this could be a class method. Changed.

app/javascript/packs/xapi_assignment.js
sadleb: Good call out. Project submission was completely broken.

sadleb: I thought about something like that but I don't think it actually makes it easier to read/maintain. I'll just add a comment on `LtiAuthentication::LTI_AUTH_HEADER_PREFIX` noting the existence of this.

config/initializers/lti_authentication.rb
rshipp: When I have to add a span around long-running code, I've been using `Class.function` or if it doesnt wrap the whole function, `Class.function.somethingspecific`. For field names, I try to use something consistent, like if I'm tracking `response.code`, I'll name the field `response.code`. For error messages, we seem to be using `error`. IIRC the beeline gem will auto-namespace things you put in an existing trace (but not a new span) with `app.your_field_name`.

rshipp: Another note that might be useful - the way Honeycomb is set up both technically and for billing, we're incentivized to make all events as wide as possible. Any data you think might be remotely useful, adding it to an event is free and can only add value. You can also add JSON blobs as field contents and have them automatically unpacked and namespaced, if you have a hash or something that you want to track.

PR #318, opened 2020-08-13
Use last published version of project

jessouyangbraven:
**Task**: N/A

**Description**:  *Bug*: A TA or Fellow can see and submit an unpublished version of a project. 

We were using `course_contents#show` as the URL for assignments in Canvas. That endpoint displays the project in its current saved state, not the version of the project when ti was inserted into Canvas.

This PR changes the URL used for assignments in Canvas to one that contains the version of the project (e.g., `course_content_histories#show`).

**Development-breaking changes**: Assignments added to Canvas before this PR won't work anymore. Insert new assignments so they go to the right endpoint.

**Test**

1. Create project in Content Editor and Save
Screen Shot 2020-08-13 at 1 26 11 PM

2. Add assignment to Canvas and view
3. Go back to Content Editor, edit the project, Save again
4. Refresh assignment in Canvas

**Before**:  shows unpublished version
Screen Shot 2020-08-13 at 1 28 22 PM

**After**: project doesn't change, still shows published/inserted version
Screen Shot 2020-08-13 at 1 28 47 PM

**Notes**:

Two types of objects are handled in `CourseContentHistoriesController#show`:

A. A published version of a project
B. A project submission, which is tied to a particular version of a project, a user viewing the submission page, and any responses that have been previously entered 

The latter will eventually be moved into `ProjectSubmissionsController`. 

These are the four scenarios #show has to handle: 

1. Showing a preview of the project when it's being inserted into Canvas assignments (A)
2. Showing a project version in platformweb (A)
3. Showing an editable project submission (B)
4. Showing a read-only project submission (B)

(3) is what a Fellow sees, which is the project version, any previous responses, editable fields, and a "Submit" button
(4) is the same as the above, except the fields are frozen and highlighted and there's no "Submit" button. This can be viewed by a TA grading the submission, or a Fellow viewing a previous submission for a project.


app/controllers/course_contents_controller.rb
```diff
@@ -74,6 +66,10 @@ def destroy
 
   # POST /course_contents/1/publish
   # POST /course_contents/1/publish.json
+  # This publish path only works for modules created in Content Editor,
+  # not assignments. Assignments are saved in the Content Editor, but
+  # published by inserting them into Canvas as assignments. See
+  # LtiAssignmentSelectionController.
```
jessouyangbraven, line 72: Added this comment because I got confused about how projects were being published/project versions were being created: https://bebraven.slack.com/archives/C01166ZLR1V/p1597351022020600

app/views/course_contents/show.html.erb
```diff
@@ -1,12 +1,3 @@
...
-  <!-- TODO: https://app.asana.com/0/1174274412967132/1186960110311121 -->
-  <%= form_with url: course_content_course_content_histories_path(@course_content) do |f| %>
-    <input id="state" name='state' type='hidden' value="<%= params[:state] %>" />
-    <!-- TODO: https://app.asana.com/0/1174274412967132/1187342807279865 -->
-  	<input type="submit" value="Submit" />
-  <% end %>
 </div>
-
-<%= content_tag :div, '', id: "javascript_variables", data: {project_lti_id: @project_lti_id, lti_auth_state: @lti_auth_state} %>
-<%= javascript_pack_tag 'xapi_assignment' %>
```
jessouyangbraven, line 12: The changes in this file break backwards compatibility with previous assignment placements. The old links to `course_contents#show` will just show the preview of the last saved version of the file. We'll have to insert new assignments to test after this is merged. I think this is fine, since no one else is using these and we won't get confused about it going forward.

app/controllers/lti_assignment_selection_controller.rb
```diff
@@ -16,7 +16,10 @@ def create
     cc = CourseContent.find(params[:assignment_id])
     cc.save_version!(current_user)
 
-    assignment_url = course_content_url(params[:assignment_id])
+    assignment_url = course_content_course_content_history_url(
+      params[:assignment_id],
+      cc.last_version.id,
+    )
```
jessouyangbraven, line 22: This is the biggest change for Canvas.

app/controllers/course_content_histories_controller.rb
```diff
@@ -20,32 +20,50 @@ def index
...
 
   # TODO: https://app.asana.com/0/1174274412967132/1186960110311121
   def create
-    params.require([:state, :course_content_id]) 
+    params.require([:state, :course_content_id, :version])
 
     submission_url = Addressable::URI.parse(course_content_course_content_history_url(
       @course_content,
-      @course_content.last_version,
+      params[:version],
```
jessouyangbraven, line 66: This is the last of it, I think. 

The version has to be specified by the form being submitted.

This is the scenario that breaks if we just use `@course_content.last_version` and try to load the submission:

1. Student opens project version 1 and starts filling it out
2. Designer edits project, saves, adds a new assignment (the project's last version is now 2)
3. Student submits project (still version 1 in browser), this code would previous use version 2 as the version in the submission URL and submit it to Canvas
4. TA goes to view the submission and it fails because we're trying to show project version 1 LRS values in project version 2's HTML

app/controllers/course_contents_controller.rb
rshipp: This comment isn't quite right. The publish path works for things on `*portal.bebraven.org`, but not for things on `braven.instructure.com`.

app/controllers/course_content_histories_controller.rb
```diff
@@ -20,32 +20,50 @@ def index
...
-    launch = LtiLaunch.current(params[:state])
-    @lti_auth_state =launch.state
-    @project_lti_id = launch.activity_id
+    params.require([:course_content_id])
+
+    # Viewed in platformweb as published version of project
+    unless params[:state]
+      # Shows a version of the project
+      return
+    end
```
rshipp, line 29: do we still need this? it seems like a debug helper that shouldn't be left in production.

app/views/course_content_histories/show.html.erb
```diff
@@ -1,9 +1,28 @@
...
+  <% end %>
 </div>
 
-<%= content_tag :div, '', id: "javascript_variables", data: {
-  user_id: @user.id,
-  lti_auth_state: @lti_auth_state,
-  project_lti_id: @project_lti_id } %>
-<%= javascript_pack_tag 'xapi_assignment' %>
+<!-- Load previous answers -->
+<% if @lti_auth_state && @project_lti_id %>
```
rshipp, line 22: related to my comment in course_content_histories_controller, I feel like these should be required params and we should get rid of the if. otherwise we're adding complexity that's only used in debugging.

app/views/course_content_histories/show.html.erb
```diff
@@ -1,9 +1,28 @@
...
+
+  <!-- Submit -->
+  <% unless @user_id %>
+    <%= form_with url: course_content_course_content_histories_path(@course_content) do |f| %>
+      <input id="state" name='state' type='hidden' value="<%= params[:state] %>" />
+      <input
+        id='version'
+        name='version'
+        type='hidden'
+        value="<%= @course_content_history.id %>"
```
rshipp, line 13: what's this used for? the form url should be specific to the version

rshipp:
I don't understand the form in the course_content_history show view, I'm guessing it'll be clearer in context with the submit PR?

app/controllers/course_contents_controller.rb
jessouyangbraven: Oof. Thanks for the clarification. I'll update the comment!

app/controllers/course_content_histories_controller.rb
jessouyangbraven: This lets us use the link on the `course_content_versions#index` page and render `https://platformweb/course_contents/1/versions/13` when there's no LTI `state` involved. 

I think it makes sense to support this, since it's the expected behavior for the endpoint? It's also helpful to me to be able to look at particular versions of a project rather than try to read the HTML in its `body`.

app/views/course_content_histories/show.html.erb
jessouyangbraven: This is probably a combination of me not totally understanding Rails and how we're using `course_content_histories` for project submissions. 

What the form does right now is `POST` to `/course_contents/1/versions`, which calls `course_content_histories#create`. 

I just naively tried changing this code to `POST` to `/course_contents/1/versions/22` and get a routing error. This makes sense to me given how we have the controller and route set up. 

Do we want to change the routing/controller for `course_content_histories` to allow the above instead of what I'm doing now?

jessouyangbraven:
> I don't understand the form in the course_content_history show view, I'm guessing it'll be clearer in context with the submit PR?

Um, probably not, unfortunately. I wrote comments in-line, but this is probably worth discussing in the standup.

app/views/course_content_histories/show.html.erb
jessouyangbraven: This is like `params[:state]` in the controller, so we can load `https://platformweb/course_contents/1/versions/13` without the JS erroring out.

app/controllers/course_content_histories_controller.rb
```diff
@@ -20,32 +20,50 @@ def index
...
+
+    # Viewed in Canvas as a project submission for this version of the project
+    # For a student, this will load the project, populate the student's
+    # most recent answers from teh LRS, and allow them to submit the project.
+    # For a TA, this will do the same as above, but in read-only mode.
+    # The answers showns are the most recent entered by the student, at
+    # the granularity of the input fields, not at the granularity of
+    # a submission.
+    # Specifying a user_id indicates a TA is viewing a student's submission
+    @user_id = params[:user_id] ? User.find(params[:user_id]).id : nil
```
sadleb, line 55: I originally called this `student_id` then changed it to `user_id` to be more consistent with the fact that it's a platform ID and not the canvas ID. But at this point, it's thoroughly confusing what this is. I think we should renamed this to `user_override_id` to be consistent with the naming in the JS. When we're one user, aka the TA, and we want to view the submission of another user, aka the Fellow, we send a `user_override_id` to do so. That feels sort of intuitive to me?

sadleb: Also, we don't need to find the user by ID and then get the ID, right? We could just do: `@user_override_id = params[:user_override_id]` -- if it's nil or blank than it's nil or blank, otherwise it's the ID

app/controllers/course_content_histories_controller.rb
sadleb: Now that we're showing a course_content_history version and submitting from there, can't we just use `@course_content_history.id` since that's the ID of the current version being shown?

app/controllers/lti_assignment_selection_controller.rb
sadleb: This is the URL we want access to when a Fellow is submitting an assignment. I think this will be in the LTI launch payload instead of using that "version" param. We should check.

app/views/course_content_histories/show.html.erb
sadleb: Yeah, I don't think you need all this. You're currently looking at a particular version, so in the controller you can just access `@course_content_history.id`.

sadleb: But it's a little non-standard and confusing for normal rails routes, so we should add a comment in there about how `create` isn't really creating a new history version. It's creating an LTI submission using the current history version.

lib/lti_score.rb
```diff
@@ -4,6 +4,8 @@
 # https://canvas.instructure.com/doc/api/score.html
 class LtiScore
   
+  LTI_SUBMISSION_URL = "https://canvas.instructure.com/lti/submission"
```
sadleb, line 7: I would rename this to `LTI_SCORE_SUBMISSION_URL_KEY`. When I look at this I immediately think "we shouldn't hardcode the URL we're submitting to" when really this is a static key used in the message.

app/controllers/course_content_histories_controller.rb
sadleb: I'm indifferent here. I can see an argument both ways. I suppose I lean towards allowing us to view it outside an LTI launch b/c it doesn't seem like much extra logic to support.

app/controllers/course_content_histories_controller.rb
sadleb: See my other comment. I think you can just use `@course_content_history.id` instead of passing it from the view in a param.

app/controllers/course_content_histories_controller.rb
```diff
@@ -20,32 +20,50 @@ def index
...
+    # submission for this user_id.
+    # Must be the student themselves or a TA or staff who has access.
+    # Need to use Canvas roles to check.
+
+    # TODO: https://app.asana.com/0/1174274412967132/1187445581799823
+    # We can also view this when adding an assignment in Canvas in
+    # the iframe preview mode:
+    # app/views/lti_assignment_selection/create.html.erb
+    # Clean this up so we don't try to communicate with the LRS in this
+    # case.
```
jessouyangbraven, line 42: Note to self, move this comment to above.

app/controllers/course_content_histories_controller.rb
jessouyangbraven: Move `course_content_histories_controller#create` to `project_submission_controller#create`

app/controllers/course_content_histories_controller.rb
sadleb: For posterity: we discussed offline and decided to leave this here b/c we need the ability to show a preview of the project and not have it talk to the LRS or be associated with a user. Plus, it's nice to be able to view versions in platform directly when troubleshooting. Finally, we may eventually write tools for designers to manage / see versions.

app/controllers/course_content_histories_controller.rb
```diff
@@ -20,42 +20,41 @@ def index
...
+    # Need to use Canvas roles to check.
+
+    # Viewed in Canvas as a project submission for this version of the project
+    # For a student, this will load the project, populate the student's
+    # most recent answers from teh LRS, and allow them to submit the project.
+    # For a TA, this will do the same as above, but in read-only mode.
+    # The answers showns are the most recent entered by the student, at
+    # the granularity of the input fields, not at the granularity of
+    # a submission.
+    # Specifying a user_id indicates a TA is viewing a student's submission
```
sadleb, line 51: s/user_id/user_override_id

app/controllers/project_submissions_controller.rb
```diff
@@ -1,16 +1,47 @@
...
   nested_resource_of Project
 
-  # GET /project_submissions
-  # GET /project_submissions.json
-  def index
-  end
+  before_action :set_project_version, only: [:create]
+
+  # Project submissions aren't recorded in our DB (e.g, we don't use the 
+  # ProjectSubmission table or model).
```
sadleb, line 16: Can you add a "yet" comment and link to the task about cutting over the whole course_contents/history to projects/project submissions?

app/views/course_content_histories/show.html.erb
```diff
@@ -1,9 +1,28 @@
 <div class="bz-assignment">
+  <!-- Project -->
   <%== @course_content_history.body %>
+
+  <!-- Submit -->
+  <% unless @user_override_id %>
+    <%= form_with url: project_project_submissions_path(@course_content) do |f| %>
+      <input id="state" name='state' type='hidden' value="<%= params[:state] %>" />
+      <input
```
sadleb, line 9: Small nit: is there a reason not to do this on one line? If this was going to be our coding style guidelines, this particular fragment would look like the following which is less readable to me than the two lines:
```
<input 
        id="state"
        name='state'
        type='hidden'
        value="<%= params[:state] %>" />
<input 
        id='version'
        name='version'
        type='hidden'
        value="<%= @course_content_history.id %>"

File: app/views/course_content_histories/show.html.erb [hunk 14]
```diff
@@ -1,9 +1,28 @@
...
+    <% end %>
+  <% end %>
 </div>
 
-<%= content_tag :div, '', id: "javascript_variables", data: {
-  user_id: @user.id,
-  lti_auth_state: @lti_auth_state,
-  project_lti_id: @project_lti_id } %>
-<%= javascript_pack_tag 'xapi_assignment' %>
+<!-- Load previous answers -->
```
sadleb, line 21: Small nit: if we move this comment inside the `if` then it wouldn't show up in source HTML randomly when the conditional is false.

app/views/course_contents/show.html.erb
sadleb: Maybe put this note in the PR description?

Side note: glad I procrastinated on filling out more of the demo course structure by actually inserting assignments!

app/views/course_content_histories/show.html.erb
jessouyangbraven: Yeah, I went back-and-forth on this. It's not as straightforward a case as an argument/param list.

I have my max-lines set to 80 and 125, the latter being what renders nicely as a single line in GitHub. My auto-wrapping starts kicking in and I start putting each on a new line once this happens. I find it makes the git blame cleaner going forward because it isolates changes. 

I can put it back on one line for now until we figure out what we want to do going forward. It might all move to React so this might be moot?

app/views/course_contents/show.html.erb
jessouyangbraven: Updated the **Development-breaking changes** part of the PR description.

PR #320, opened 2020-08-16
Cleanup LtiLaunch table on app startup.

sadleb:
Task: https://app.asana.com/0/1174274412967132/1184057808812020

By default, on app startup wait about 30 seconds to let the app do
it's normal stuff and then on a background thread delete all LtiLaunch
records that are more than 2 weeks old. This serves several purposes:
  - makes it so that a brute force attack trying to hack the "state"
    parameter in order to have a valid session without actually going through
    the Lti Launch flow is unlikely to succeed
  - keeps queries of this table lightning fast which is important b/c this
    table is hit a lot and would grow rather large over time.

### Test Plan
- Set the `LTI_LAUNCH_REMEMBER_FOR` ENV var to `'30.seconds'`, open a Project in
  Canvas that was added using the Assignment Selection LTI placement. Grab the
  URL of of the project on platform with the "state" param from the logs and
  open it in a new tab, it should show properly b/c the LtiLaunch is valid.
  Now restart the server, wait 30 more seconds, then refresh the tab with the
  direct platform URL for the project and it should not show (either error out
  or redirect to the login page).
- `bundle exec rspec spec/jobs/cleanup_lti_launch_table_job_spec.rb`


rshipp:
can we send the `logger` stuff to honeycomb too? Then we can set up alerts if this is ever failing, or track it alongside traffic if it's causing an issue.

app/jobs/cleanup_lti_launch_table_job.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def perform
+    # The value of 'lti_launch_remember_for' can be a Rails Time helper like 2.weeks or the 
+    # number of seconds, like 1209600 (this is also 2 weeks)
+    cutoff_time = Time.at(Time.now - eval(Rails.application.secrets.lti_launch_remember_for))
+    Rails.logger.info("Cleaning up LtiLaunches older than #{cutoff_time}")
+    records_to_delete = LtiLaunch.where("updated_at < ?", cutoff_time) # Use updated_at so that continuing to interact with the launch extends it's validity.
+    total_count = LtiLaunch.count
+    count = records_to_delete ? records_to_delete.count : 0
+    records_to_delete.delete_all
```
jessouyangbraven, line 27: Since we do a `nil` check above, do we need to do one where as well?

sadleb:
> can we send the `logger` stuff to honeycomb too? Then we can set up alerts if this is ever failing, or track it alongside traffic if it's causing an issue.

image

app/jobs/cleanup_lti_launch_table_job.rb
sadleb: That was left over stuff. I can just call `count` on the result. If there are 0, it works.

PR #321, opened 2020-08-17
Instrument RestClient calls with Honeycomb tracing and logging.

sadleb:
Task: https://app.asana.com/0/1174274412967132/1187503001085713

If you go under the covers of all RestClient requests, they end up calling
the instance method `execute` on the Request object. This commit prepends
a module that overrides that method to wrap it with a Honeycomb span and
to read the relevant info out of any exceptions that are thrown and trace/log
that as well.

### Test Plan

- `bundle exec rspec`
- Do a Project, submit it, make sure data flows to the LRS and is also
  sent to Honeycomb. Here is a sample query: https://ui.honeycomb.io/braven/datasets/stagingplatform.bebraven.org/result/aijLvyA7Fdo
- Change the payload of a message from some RestClient API usage so that
  it will throw an exception. Make sure the exception details come through
  in the logs and on Honeycomb


config/initializers/restclient_instrumentation.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    # I'm trying to mimic what normal Rails request's send:
+    # https://guides.rubyonrails.org/active_support_instrumentation.html#process-action-action-controller
+    Honeycomb.start_span(name: "RestClient.#{method}") do |span|
+      span.add_field('class_name', calling_class)
+      span.add_field('method', method)
+      span.add_field('url', redacted_url) # Strips out password if it's in there.
+      span.add_field('timestamp', DateTime.now)
+      redacted_headers = headers.dup
+      redacted_headers[:authorization] = 'Basic [REDACTED]' if redacted_headers[:authorization]
```
rshipp, line 18: this isn't necessarily `Basic` auth, maybe just use `[REDACTED]`?

lib/restclient_instrumentation.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    # I'm trying to mimic what normal Rails request's send:
+    # https://guides.rubyonrails.org/active_support_instrumentation.html#process-action-action-controller
+    Honeycomb.start_span(name: "RestClient.#{method}") do |span|
+      span.add_field('class_name', calling_class)
+      span.add_field('method', method)
+      span.add_field('url', redacted_url) # Strips out password if it's in there.
+      span.add_field('timestamp', DateTime.now)
+      redacted_headers = processed_headers.dup
+      redacted_headers['Authorization'] = '[REDACTED]' if redacted_headers['Authorization']
```
sadleb, line 24: This was broken before if the auth header was specified as `'Authorization'` when using `RestClient`. Using `processed_headers` normalizes it to that format and now lowercase and uppercase and symbols work.

sadleb:
I made the small change you requested, added specs, and fixed a small bug found while adding specs.

PR #322, opened 2020-08-17
Add new Projects CSS

rshipp:
Add new CSS to apply to Projects. 

Task: https://app.asana.com/0/1174274412967132/1189101944927283/f

### Screenshots

Screen Shot 2020-08-17 at 1 02 41 PM
Screen Shot 2020-08-17 at 1 03 06 PM

Mobile responsiveness

Screen Shot 2020-08-17 at 1 03 31 PM


rshipp:
> Should this be moved to course_content_histories.scss after #318 was merged?

I'm not sure why, but the histories view already includes the course_contents css, so it works fine. We don't want these changes to be limited to the `histories` views, because they also apply inside the editor.

> Also, update them to the .woff format once you convert so they are fully compatible with IE.

Mentioned this in slack but the converter refuses to convert these, we'll just have to leave as-is for now.

PR #324, opened 2020-08-18
Make selecting a Lesson or Project work in Chrome incognito

sadleb:
Task: https://app.asana.com/0/search/1189491188178881/1188539121871585

To make the `LtiAuthentication::WardenStrategy` work for anything accessed through an `LtiLaunch`, the `state` parameter must be passed. This PR adds it to two spots I missed in https://github.com/bebraven/platform/pull/316. Also, CSRF authenticity token validation only works when we have a session. The concept is to prevent an attacker from getting a session token and being able to piggyback off it so that only applies for session based auth. Only skip verification when we have to fall-back to `LtiAuthentication::WardenStrategy` auth.

Screenshot of where this now works. If you choose a Project or Lesson and insert it with Chrome in this mode, this PR causes that to work:
image


app/controllers/lesson_contents_controller.rb
```diff
@@ -25,4 +27,9 @@ def create_params
     params.require([:state, :lesson_content_zipfile])
     params.permit(:lesson_content_zipfile, :state, :commit, :authenticity_token)
   end
+
+  def is_sessionless_lti_launch?
+    @lti_launch = LtiLaunch.current(params[:state]) if params[:state]
```
sadleb, line 32: I'm not happy with setting this here instead of in a separate `set_lti_launch` before action, but this depends on that running first and for some reason this kept running first in certain scenarios. Maybe skip_verification stuff always runs first? I would need to read up on rails callback order and guarantees to split this out.

app/controllers/lti_assignment_selection_controller.rb
```diff
@@ -20,6 +25,13 @@ def create
       params[:assignment_id],
       cc.last_version.id,
     )
-    @deep_link_return_url, @jwt_response = helpers.lti_deep_link_response_message(lti_launch, assignment_url)
+    @deep_link_return_url, @jwt_response = helpers.lti_deep_link_response_message(@lti_launch, assignment_url)
+  end
+
+private
+
+  def is_sessionless_lti_launch?
```
sadleb, line 33: Now that this PR is up, I'll tinker with moving this to a shared helper.

config/initializers/devise.rb
```diff
@@ -297,7 +297,7 @@
     # call sign_in user which stores the authentication in session and bypasses the need
     # to check authentication.
     manager.strategies.add(:lti_auth, LtiAuthentication::WardenStrategy)
-    manager.default_strategies(scope: :user).unshift :lti_auth
+    manager.default_strategies(scope: :user) << :lti_auth
```
sadleb, line 300: This doesn't seem to have really had any impact in my testing, but logically we want to favor session based authentication over LtiLaunch.state based auth so put the strategy at the end of the array not the beginning.

config/initializers/lti_authentication.rb
```diff
@@ -40,18 +44,23 @@ def authenticate!
...
 
     def finish_authenticate(span, status, message, url = nil, canvas_id = nil, user = nil)
-      span.add_field('lti_auth.status', status)
+      span.add_field('status', status)
       span.add_field('lti_auth.message', message)
-      span.add_field('lti_auth.url', url) if url 
-      span.add_field('lti_auth.canvas_id', canvas_id) if canvas_id
+      span.add_field('url', url) if url 
+      span.add_field('canvas_id', canvas_id) if canvas_id
+      span.add_field('user_id', user.id) if user
```
sadleb, line 63: Just cleaning these keys up based on what we discussed as a norm and adding `user_id` in there.

spec/controllers/lesson_contents_controller_spec.rb
```diff
@@ -2,27 +2,20 @@
...
 
 RSpec.describe LessonContentsController, type: :controller do
-  # TODO: https://app.asana.com/0/1174274412967132/1184057808812010
   render_views
 
-  let(:user) { create :admin_user }
-
-  before(:each) do
-    sign_in user
-  end
```
sadleb, line 12: I contemplated creating two sets of tests, one for session based auth and one for LtiLaunch.state auth, but I don't think it's worth it in the controller so I opted to change this into the fallback which should always work which is LtiLaunch.state auth. Same for the lti_assignments_selection_controller.

app/views/lesson_contents/create.html.erb
```diff
@@ -5,5 +5,5 @@
     <input type='submit' value='Yes' />
   </form>
   <br/>
-  <iframe id="lesson-content-preview" src="<%= lesson_content_url(@lesson_content) %>"></iframe>
+  <iframe id="lesson-content-preview" src="<%= lesson_content_url(@lesson_content) %>?state=<%= params[:state] %>"></iframe>
```
rshipp, line 8: can we pass state as `params` in the `lesson_content_url` helper? i'm not sure if that exists but it seems like it should

sadleb: oh, yup. that's better. i'll make the change.

PR #329, opened 2020-08-19
Add lesson interaction tracking

rshipp:
Task: https://app.asana.com/0/0/1189791588110883/f
Task: https://app.asana.com/0/0/1189791588110884/f

### verb: progressed

Used to compute lesson engagement, as a percentage.

The value in the `verb` column will be `http://adlnet.gov/expapi/verbs/progressed`.

* For records that match the given user, where `activity_id = 'MY_ACTIVITY_ID'` if there are any with `new = true`, order by `progress` and select the *highest*.
* If there are none that are `new`, return. Note we can't select only records where `new = true` above, because the highest progress in `new` records is not necessarily the highest progress of all records.

That `progress` will be an integer from 0-100, representing % completion.

Form the API call for sending grades to canvas using the `canvas_course_id` and `canvas_assignment_id` associated with each record.

After sending the score to Canvas and getting a success back, set `new` to `false` for all records selected.

### verb: answered

Used to compute mastery quiz total score, as a percentage.

The value in the `verb` column will be `http://adlnet.gov/expapi/verbs/answered`.

For `answered` statements, the `activity_id` does NOT match `MY_ACTIVITY_ID`. (Note the `like` below.) Instead, if `MY_ACTIVITY_ID` is the same URL-like string, `activity_id` will look like

    MY_ACTIVITY_ID/some_unique_quiz_id/some_unique_question_id

* Select all records that match the given user, where `activity_id like 'MY_ACTIVITY_ID%'` and `new = true`. 
* Select the LessonContent record that has an `activity_id` that matches `MY_ACTIVITY_ID`.
* Total the unique "activity_id" strings matching this "Lesson", divide number of them with success:true by the number of total mastery questions in this "Lesson" (from `lesson_content.quiz_questions`) to get the score.

That score will be a float; multiply by 100 and truncate (or round) to get the final % score.

Form the API call for sending grades to canvas using the `canvas_course_id` and `canvas_assignment_id` associated with each record.

After sending the score to Canvas and getting a success back, set `new` to `false` for all records selected.


lib/lesson_content_util.rb
```diff
@@ -46,7 +51,26 @@ def self.publish(zipfile)
       launch_path(zipfile.key, bucket)
   end
 
+  # Pulls down the tincan.xml from S3 and updates the record with extra context.
+  def self.update_metadata(lesson_content, bucket = nil)
+    # TODO: in a future iteration we'll want to generate pre-signed URLs with expirations on
+    # them so that our content doesn't get leaked for anyone to access. Right now the bucket is public
+    # but we'll want to lock it down.
+    bucket = AwsS3Bucket.new unless bucket
+    tincan_xml = bucket.object(s3_object_key(lesson_content.lesson_content_zipfile.key, TINCAN_XML_FILE)).data
```
jessouyangbraven, line 60: Instead of `.data`, I had to use

```
  .get()['body'].read()
```

to get this to return a string that `Nokogiri::XML` can parse in the next line.

lib/lesson_content_util.rb
```diff
@@ -46,7 +51,26 @@ def self.publish(zipfile)
...
+
+    # Parse tincan.xml and extract what we need.
+    tincan_data = Nokogiri::XML(tincan_xml)
+    lesson_content.quiz_questions = tincan_data.xpath(QUIZ_QUESTION_XPATH).count
```
jessouyangbraven, line 64: I kept getting `[]` from all of the `xpath` calls, so I added

```
  tincan_data.remove_namespaces!
```

based on this: https://stackoverflow.com/questions/4690737/nokogiri-xpath-namespace-query

lib/lesson_content_util.rb
```diff
@@ -46,7 +51,26 @@ def self.publish(zipfile)
...
+    lesson_content.activity_id = tincan_data.xpath(COURSE_XPATH).first['id']
 
+    lesson_content.save!
+  catch => e
```
jessouyangbraven, line 68: Syntax error so I commented this out.

lib/lesson_content_util.rb
```diff
@@ -46,7 +51,26 @@ def self.publish(zipfile)
       launch_path(zipfile.key, bucket)
   end
 
+  # Pulls down the tincan.xml from S3 and updates the record with extra context.
+  def self.update_metadata(lesson_content, bucket = nil)
```
jessouyangbraven, line 55: `update_metadata!` to match `LessonContent.rb` call.

lib/lrs_xapi_proxy.rb
```diff
@@ -46,13 +46,17 @@ def self.request(request, endpoint, user)
...
-        if content_type == JSON_MIME_TYPE 
+        if content_type == JSON_MIME_TYPE
           data = JSON.parse(request.raw_post)
           data['actor'] = build_agent_hash(user) if data['actor']
           data = data.to_json
         end
       end
 
+      # Inspect statements for grading purposes.
+      save_interaction!(data, request.authorization) if data
```
jessouyangbraven, line 57: `data` could be a string at this point, so I used `JSON.parse(request.raw_post)`.

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +96,41 @@ def self.request(request, endpoint, user)
...
+    return unless authorization.start_with? LtiAuthentication::LTI_AUTH_HEADER_PREFIX
+
+    # Get verb and activity_id from the payload.
+    verb = data.dig('statement', 'verb', 'id')
+    activity_id = data['activities'].first
+
+    # Get course and assignment IDs from the auth header.
+    # Header looks like "<prefix> <state>".
+    state = authorization.split(LtiAuthentication::LTI_AUTH_HEADER_PREFIX).last.strip
+    ll = LtiLaunch.current(state)
```
jessouyangbraven, line 110: I passed in `params[:state]` instead of trying to pull the state here. For some reason, my authorization header only had the prefix in it.

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +96,41 @@ def self.request(request, endpoint, user)
   private_class_method def self.authentication_header
     @@auth_header ||= "Basic #{Rails.application.secrets.lrs_auth_token}"
   end
+
+  private_class_method def self.save_interaction!(data, authorization)
+    return unless authorization.start_with? LtiAuthentication::LTI_AUTH_HEADER_PREFIX
+
+    # Get verb and activity_id from the payload.
+    verb = data.dig('statement', 'verb', 'id')
```
jessouyangbraven, line 104: For all of the `data.dig()` calls, I removed `'statement'`. There's `statement` in the LRS DB records, but not in the request data for some reason.

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +96,41 @@ def self.request(request, endpoint, user)
...
+    activity_id = data['activities'].first
```
jessouyangbraven, line 105: Like `statement`, I found this in the LRS, but not in the request data. I used ` data.dig('object', 'id')`.

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +96,41 @@ def self.request(request, endpoint, user)
...
+
+    # Get course and assignment IDs from the auth header.
+    # Header looks like "<prefix> <state>".
+    state = authorization.split(LtiAuthentication::LTI_AUTH_HEADER_PREFIX).last.strip
+    ll = LtiLaunch.current(state)
+    parsed = ll.activity_id.match /\/courses\/(?<course>\d+)\/assignments\/(?<assignment>\d+)/
+
+    # Save to the db, raise an exception if any part of this fails.
+    # We want this to 500 out if there's anything unexpected.
+    if verb == LessonInteraction.PROGRESSED
```
jessouyangbraven, line 115: `LessonInteraction::PROGRESSED`

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +96,41 @@ def self.request(request, endpoint, user)
...
+      LessonInteraction.create!(
+        verb: verb,
+        user: user,
+        canvas_course_id: parsed[:course],
+        canvas_assignment_id: parsed[:assignment],
+        activity_id: activity_id,
+        progress: data.dig('statement', 'result', 'extensions',
```
jessouyangbraven, line 122: Ditto. No `'statement'` key.

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +96,41 @@ def self.request(request, endpoint, user)
...
+            'http://w3id.org/xapi/cmi5/result/extensions/progress')
+        )
+    elsif verb == LessonInteraction.ANSWERED
+      LessonInteraction.create!(
+        verb: verb,
+        user: user,
+        canvas_course_id: parsed[:course],
+        canvas_assignment_id: parsed[:assignment],
+        activity_id: activity_id,
+        success: data.dig('statement', 'result', 'success')
```
jessouyangbraven, line 132: Ditto. No `'statement'` key.

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +96,41 @@ def self.request(request, endpoint, user)
...
+      LessonInteraction.create!(
+        verb: verb,
+        user: user,
+        canvas_course_id: parsed[:course],
+        canvas_assignment_id: parsed[:assignment],
+        activity_id: activity_id,
+        progress: data.dig('statement', 'result', 'extensions',
+            'http://w3id.org/xapi/cmi5/result/extensions/progress')
+        )
+    elsif verb == LessonInteraction.ANSWERED
```
jessouyangbraven, line 125: `::`

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +96,41 @@ def self.request(request, endpoint, user)
   private_class_method def self.authentication_header
     @@auth_header ||= "Basic #{Rails.application.secrets.lrs_auth_token}"
   end
+
+  private_class_method def self.save_interaction!(data, authorization)
+    return unless authorization.start_with? LtiAuthentication::LTI_AUTH_HEADER_PREFIX
+
```
jessouyangbraven, line 102: I also did an early bail out on the `verb` in `data`, e.g.:

```
   verb = data.dig('verb', 'id')
    return unless (
      verb == LessonInteraction::PROGRESSED || 
      verb == LessonInteraction::ANSWERED
    )
```

I don't think it's necessary, but I was running into some issues with how the data was formatted for other verbs later on.

lib/lrs_xapi_proxy.rb
rshipp: this one is weird, if it comes up again we might need Brian to troubleshoot. it's my understanding that anything in the lti iframe should have that header set with a valid `state`.

lib/lrs_xapi_proxy.rb
rshipp: did this slightly different, hopefully it still works.

db/migrate/20200819202940_create_lesson_interactions.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.integer :progress
+      t.string :verb, null: false
+      t.integer :canvas_course_id, null: false
+      t.integer :canvas_assignment_id, null: false
+      t.boolean :new, default: true, null: false
+
+      t.timestamps
+    end
+
+    add_index(:lesson_interactions, [:new, :user_id, :activity_id, :verb], :name => 'index_lesson_interactions_1')
```
rshipp, line 16: index order is important - columns that narrow down the result set the most should be first. I think this is the index we want, but I can't be sure until we have the actual queries written. lmk if you think there's a better index, or more we should add.

db/migrate/20200820205947_add_metadata_to_lesson_content.rb
```diff
@@ -0,0 +1,7 @@
+class AddMetadataToLessonContent < ActiveRecord::Migration[6.0]
+  def change
+    add_column :lesson_contents, :activity_id, :string
+    add_column :lesson_contents, :quiz_questions, :integer
+    add_index :lesson_contents, [:activity_id]
```
rshipp, line 5: indexing on activity ID, to make that search faster.

lib/lrs_xapi_proxy.rb
```diff
@@ -46,13 +47,19 @@ def self.request(request, endpoint, user)
...
           data = JSON.parse(request.raw_post)
           data['actor'] = build_agent_hash(user) if data['actor']
           data = data.to_json
         end
       end
 
+      # Inspect statements for grading purposes.
+      if data.is_a? String and endpoint == XAPI_STATEMENTS_API_ENDPOINT
+        save_interaction!(data, request.authorization, user)
+      end
```
rshipp, line 60: This looked prettier as a 1-liner, but it got hard to read when i added more checks :(

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +99,48 @@ def self.request(request, endpoint, user)
   private_class_method def self.authentication_header
     @@auth_header ||= "Basic #{Rails.application.secrets.lrs_auth_token}"
   end
+
+  private_class_method def self.save_interaction!(data, authorization, user)
+    return unless authorization&.start_with? LtiAuthentication::LTI_AUTH_HEADER_PREFIX
```
rshipp, line 104: double-checking this is the right logic for `&.` ?

spec/fixtures/example_tincan.xml
```diff
@@ -0,0 +1,207 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
rshipp, line 1: I put this tincan.xml inside the zip too, but I didn't want to unzip it every time

spec/lib/lesson_content_util_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    context 'when valid Rise360 zipfile' do
+
+      it 'updates the record appropriately' do
+        expect(aws_object).to receive(:get).and_return({'body':
+            File.open("#{Rails.root}/spec/fixtures/example_tincan.xml")}.with_indifferent_access).at_least(:once)
+        lesson_content  # The creation of this object causes the callback to fire.
+        expect(lesson_content.quiz_questions).to eq(4)
+        expect(lesson_content.activity_id).to eq('http://OrDngAKqbvX4sCs0vBpsk-P1VXsst1vc_rise')
+      end
+
```
rshipp, line 48: this is the only new test in this file, the rest were renamed from the old util test

lib/lrs_xapi_proxy.rb
sadleb: I would need to look at this particular flow more deeply, but generally the `state` can be found in one of three places. See here for where: https://github.com/bebraven/platform/blob/caecf5b52b50486e8eccc57ecaa05397eaf05199/config/initializers/lti_authentication.rb#L70

lib/lrs_xapi_proxy.rb
sadleb: Yeah, that looks like the right syntax.

app/models/lesson_interaction.rb
```diff
@@ -0,0 +1,7 @@
+class LessonInteraction < ApplicationRecord
+  PROGRESSED = 'http://adlnet.gov/expapi/verbs/progressed'
+  ANSWERED = 'http://adlnet.gov/expapi/verbs/answered'
+
+  belongs_to :user
+  validates :user, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id, presence: true
```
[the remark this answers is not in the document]
rshipp: Yeah, it turns out you can't enforce presence on a boolean. So then I tried to enforce success in [true, false]... but realized that's what boolean means and that validation does nothing. The progress validation I took out because I couldn't figure out how to get the `should validate_presence_of` test to pass, and decided I was spending too much time on something that will probably never have any benefit.

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +99,48 @@ def self.request(request, endpoint, user)
...
+
+    # Get course and assignment IDs from the auth header.
+    # Header looks like "<prefix> <state>".
+    state = authorization.split(LtiAuthentication::LTI_AUTH_HEADER_PREFIX).last.strip
+    ll = LtiLaunch.current(state)
+    parsed = ll.activity_id.match /\/courses\/(?<course>\d+)\/assignments\/(?<assignment>\d+)/
+
+    # Save to the db, raise an exception if any part of this fails.
+    # We want this to 500 out if there's anything unexpected.
+    if verb == LessonInteraction::PROGRESSED
```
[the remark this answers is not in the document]
rshipp: don't think i've ever had a valid reason to use a case statement before!

lib/lrs_xapi_proxy.rb
```diff
@@ -92,4 +99,48 @@ def self.request(request, endpoint, user)
...
+    # Exit early if it's not a supported verb.
+    return unless (
+      verb == LessonInteraction::PROGRESSED or
+      verb == LessonInteraction::ANSWERED
+    )
+
+    # Get course and assignment IDs from the auth header.
+    # Header looks like "<prefix> <state>".
+    state = authorization.split(LtiAuthentication::LTI_AUTH_HEADER_PREFIX).last.strip
+    ll = LtiLaunch.current(state)
```
sadleb, line 120: Wondering if we should move this and the above check for the state auth header should be moved into a convenience method inside `LtiLaunch` similar to `current` but it takes a request and looks at `params[:state]`, `params[:auth]`, and the `auth` header, pulls the state out and does the lookup to return the current `LtiLaunch` or null if not found. Basically, move the `fetch_state` logic there and have both this and `lti_authentication` use that to look in the 3 places that the state param could be stored in a request in order to get the current launch.

lib/lesson_content_util.rb
```diff
@@ -46,7 +51,28 @@ def self.publish(zipfile)
...
+    tincan_data.remove_namespaces!
+    lesson_content.quiz_questions = tincan_data.xpath(QUIZ_QUESTION_XPATH).count
+    lesson_content.activity_id = tincan_data.xpath(COURSE_XPATH).first['id']
 
+    lesson_content.save!
+  rescue Exception => e
+    Rails.logger.error("Could not update LessonContent metadata: #{lesson_content}")
+    Rails.logger.error("  Error: #{e}")
+    Honeycomb.add_field('error', e)
+    Honeycomb.add_field('lesson_content', lesson_content)
```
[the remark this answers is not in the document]
rshipp: The part we talked about was for the proxy. You can't raise exceptions in `after_create_commit` callbacks, because they're not within the transaction, so here we just have to log.

lib/lrs_xapi_proxy.rb
rshipp: This sounds worthwhile but beyond the scope of this PR, since it's used in other places. Task https://app.asana.com/0/1174274412967132/1191053265938819/f

lib/lrs_xapi_proxy.rb
jessouyangbraven: FYI, I re-inserted the Rise360 zipfile as a Canvas assignment and stopped having issues with the LTI authentication header. I probably had something wonky going on on my dev branch at the time I inserted the one that was giving me issues.

PR #331, opened 2020-08-21
Add Honeycomb instrumentation for Javascript

sadleb:
Task: https://app.asana.com/0/1174274412967132/1188732532251833

**Note**: this PR should be reviewed alongside [this PR](https://github.com/bebraven/boomerang/pull/1). They are dependent on each other.

**Release Note**: just make sure and test that a basic trace comes through in prod since there is prod vs dev code.

Adds support for [Boomerang](https://github.com/akamai/boomerang) as a tool to auto-instrument our Javascript. Honeycomb doesn't directly support auto-instrumented browser JS, so we proxy the Boomerang "beacons" server side, turn them into "spans", do some translation of Boomerang fields, and send the events off the Honeycomb. See: https://docs.honeycomb.io/getting-data-in/javascript/browser-js/

#### Some notes:
1. We want the Honeycomb spans to show up in traces for the original controller that loaded the JS, not the proxy controller. We use a slightly undocumented `X-Honeycomb-Trace` header, set using middleware, to accomplish this.
2. The concept of a Honeycomb span doesn't directly translate to a Boomerang beacon. We add a wrapper class in JS to make it easier to ensure that Boomerang beacons translate into Honeycomb spans when we proxy them. 
3. The Boomerang fields are non-intuitive. We translate some standard ones into the Rails equivalent to make it easier to query for and analyze.
4. The loading of Boomerang should happen as early as possible but not interfere with the page load. This was hard to get right, but it turns out they have a [loader snippet](https://developer.akamai.com/tools/boomerang/docs/tutorial-loader-snippet.html) showing how to do it which we only needed to tweak slightly.
5. This depends on our custom built Boomerang JS. [See here for what we've customized](https://github.com/bebraven/boomerang/pull/1). We host the compiled version of this on AWS S3. This is how Boomerang recommends you do it. It can't be imported as an ES6 module b/c it doesn't comply with strict mode.

### Test Plan:
- Load a project and make sure the traces look good in Chrome, Chrome incognito, and Firefox (and IE 11 if you have time). On Honeycomb, just query for `trace.trace_id` output in the logs and then click the little icon shown in the second screenshot:
 

- Do something to cause the JS to throw an error and make sure the stacktrace comes through in the fields.

- Try it out both with the Boomerang JS cached by the browser and with it not cached. Make sure the traces look good. I use Chrome inspector to turn off browser caching to try this. The time it takes to download was giving me false positives on things working properly before I found their loader snippet.

- `yarn test`

- `bundle exec rspec`


app/controllers/lrs_xapi_proxy_controller.rb
```diff
@@ -23,5 +24,7 @@ def xAPI
       render plain: response_body, status: response.code, content_type: content_type and return if content_type == LrsXapiProxy::OCTET_STREAM_MIME_TYPE
       render json: response_body, status: response.code and return if content_type == LrsXapiProxy::JSON_MIME_TYPE
     end
+  rescue RestClient::Exception => e
```
sadleb, line 27: When we let the exception propagate, Rails created a HTML page response which was massive and blew up the content of the Honeycomb span sent from JS with the error details about the Ajax call that failed. Render as json since this is an API not a user rendered page.

app/controllers/honeycomb_js_controller.rb
```diff
@@ -0,0 +1,143 @@
...
+        span.add_field('request.query_string', pathinfo.query)
+      end
+
+      if params['http.initiator'] == 'xhr'
+        method = params['http.method'] || 'GET' # Annoyingly, 'http.method' may not be set for GET. If missing on XHR beacon, it's GET.
+        span.add_field('request.method', method)
+      end
+  end
+
+  def isPageLoadBeacon?
```
sadleb, line 89: Note: don't use Camelcase

app/controllers/application_controller.rb
```diff
@@ -66,4 +66,17 @@ def cas_login_url
...
+
+  # For LTI authentication (esp in a browser without access to cookies / session), the
+  # state param is the de-facto "authenticated session" identifier. It is used to look up
+  # the LtiLaunch for context on this "session"
+  def current_state_param
+    # TODO: the state param may be in one of three places, see: config/initializers/lti_authentication.rb 
+    # Abstract that out into a common utility / way for any controller or lib or service to be able to 
+    # access the state param of the current request. Perhaps using middleware? Like TracePropagation::Middleware does?
+    # https://app.asana.com/0/1174274412967132/1191053265938819
+    params[:state]
```
sadleb, line 78: I only instrumented `xapi_assignment` which runs in a controller where this is the only spot we need to look for state. But when we do the Asana task linked or if we instrument other JS that needs to work in a session-less browser, we'll need to look in the other two places in order to set this for the Boomerang plugin to read and include in it's request.

app/javascript/packs/boomerang.js
```diff
@@ -0,0 +1,225 @@
...
+        // Note that the following metadata must be set on this for the browser to handle it properly
+        // Content-Type: application/javascript
+        // Content-Encoding: gzip
+        window.BOOMR.url = "https://boomerangjs.s3.amazonaws.com/boomerang-1.0.0.min.js.gz";
+    }
+
+    // Note: I've altered this from the original snippet to insert the script after the
+    // <meta name="serialized-trace" ...> tag because our compiled Boomerang relies on that
+    // to propagate Honeycomb traces.
+    var where = document.querySelector('meta[name="serialized-trace"]'),
```
sadleb, line 45: This is what I customized. The rest is out of the box from their loader snippet.

app/javascript/packs/xapi_assignment.js
```diff
@@ -94,74 +103,79 @@ function sendStatement(e) {
...
+                }
+            });
+        });
+    
+        // Check for more after, so the first result for each field is always used.
+        // Note: The TinCanJS docs say to use sr.more !== null, but sr.more is sometimes
+        // the empty string "" when there are no more results.
+        if (sr.more) {
+            // Fetch additional page(s) of statements, recursively.
+            honey_span.addField('statement.query.get_more', 1);
```
sadleb, line 160: TODO: maybe count the number of recursions and add that? I never tested this piece btw.

spec/feature/project_submissions/project_submit_spec.rb
```diff
@@ -154,10 +154,12 @@
         # Make the LTI state ID invalid
         # Capybara's fill_in doesn't work because the input is hidden
         page.execute_script(
-          "document.getElementsByName('state')[0].value = 'invalidstate'",
+          "document.querySelector('input[type=\"hidden\"][name=\"state\"]').value = 'invalidstate'"
```
sadleb, line 157: There are now two elements named `state` on this page. One is the `<meta>` element and one is the hidden element to pass this through to the next controller on submit.

rshipp (approved):

PR #333, opened 2020-08-25
Compute lesson grade from LessonInteractions

jessouyangbraven:
**Task**: https://app.asana.com/0/1174274412967132/1189791588110887

**Test**: 

**Spec**
- Added spec tests for `LessonGradeCalculator` logic

**Manual**
- In Canvas, upload Rise360 zipfile using Assignments LTI, go through lessons, answer questions
- In LRS, check that the Tincan statements are being sent
- In  Rails console (devc):
```
> LessonInteractions.all # make sure stuff is getting saved to DB
> LessonGradeCalculator.compute_grade(user_id, activity_id)
```

**Details**
- This branch is based on `lrs-grading`, which I used to generate `LessonInteractions` in my development DB.


app/models/lesson_interaction.rb
```diff
@@ -1,7 +1,25 @@
...
 class LessonInteraction < ApplicationRecord
   PROGRESSED = 'http://adlnet.gov/expapi/verbs/progressed'
   ANSWERED = 'http://adlnet.gov/expapi/verbs/answered'
 
+  ACTIVITY_ID_SUFFIX = 'x_rise'
+
   belongs_to :user
   validates :user, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id, presence: true
+
+  # TODO: should this be a belongs to?
```
jessouyangbraven, line 12: I'm not sure the DB ramifications this would have. But it seems logical that a `LessonInteraction` should `belong_to` a `LessonContent`.

lib/lesson_grade_calculator.rb
```diff
@@ -0,0 +1,78 @@
...
+  def self.compute_grade(user_id, activity_id)
+    interactions = LessonInteraction.where(
+      "activity_id LIKE ? AND user_id = ? AND new IS ?", 
+      activity_id,
+      user_id,
+      true,
+    )
+
+    # No new interactions for user and lesson, don't need to recompute grade
+    return nil unless interactions.exists?
```
jessouyangbraven, line 24: This was the only place I was sure of the usage of the `new` column in `LessonInteraction`. 

In order to short-circuit the other grade calculations, we need to store the engagement and quiz grades separately. 

I think this would have to be in our DB, since Canvas's submission grading only lets you store a single value that's the overall score. 

Might make sense to discuss in our. next stand-up.

lib/lesson_grade_calculator.rb
```diff
@@ -0,0 +1,78 @@
...
+  def self.grade_weights
+    {
+      lesson_engagement: LESSON_ENGAGEMENT_WEIGHT,
+      mastery_quiz: 1.0 - LESSON_ENGAGEMENT_WEIGHT,
+    }
+  end
+
+  # Returns nil if we do't need to update the user's grade. 
+  # Returns [0, 100] for user's total grade for a lesson.
+  def self.compute_grade(user_id, activity_id)
```
jessouyangbraven, line 15: I think this is the right level of abstraction?

I'm assuming our Rake task is going to figure out the `user_id`s and `activity_id`s to call `LessonGradeCalculator` with. 

The alternative would be to have `LessonGradeCalculator` do it and Rake just does `LessonGradeCalculator.compute_all()` or something.

lib/lesson_grade_calculator.rb
```diff
@@ -0,0 +1,78 @@
...
+    correct_count = 0.0
+    latest_answers.each do |activity_id, most_recent|
+      answer = interactions
+        .where(
+          activity_id: activity_id,
+          created_at: most_recent,
+          success: true,
+        )
+      correct_count += 1.0 if answer.exists?
+    end
```
jessouyangbraven, line 74: For the life of me I couldn't figure out the `ActiveModel` query-only way to do this. If there's a better way to do this that you know of, I'd be glad to be rid of this loop.

spec/lib/lesson_grade_calculator_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+        )
+
+        # Wrong answer occurs later
+        LessonInteraction.create!(
+          verb: LessonInteraction::ANSWERED,
+          user: user,
+          canvas_course_id: 333,
+          canvas_assignment_id: 222,
+          activity_id: quiz_question_id,
+          success: false, # User later go the same question wrong
```
jessouyangbraven, line 224: This tests our grading if the user retries the mastery quiz. 

I know we're probably not going to allow mastery quiz retries from a design/Rise360 lesson perspective. 

But after talking to @rshipp I don't think the code rely on that being the case.

lib/lesson_grade_calculator.rb
rshipp: does `latest_answers.count.where(success: true)` work?

app/models/lesson_interaction.rb
```diff
@@ -3,6 +3,22 @@ class LessonInteraction < ApplicationRecord
...
   ANSWERED = 'http://adlnet.gov/expapi/verbs/answered'
 
+  ACTIVITY_ID_SUFFIX = 'x_rise'
+
   belongs_to :user
   validates :user, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id, presence: true
+
+  # TODO: should this be a belongs to? 
+  # Returns LessonContent where this interaction was generated
+  def lesson_content
```
rshipp, line 13: This seems nice to have here, but the fragility of the activity_id makes me think we should be doing it in the quiz_question query instead, that way we know 100% what the activity_id should be.

lib/lesson_grade_calculator.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    100 * (correct_count / interactions.first.lesson_content.quiz_questions)
```
rshipp, line 76: re: `.lesson_content`, since we know here the activity_id is `first_part/quiz_part/question_part`, we can split on the `/` and use `first_part` in the query.

lib/lesson_grade_calculator.rb
```diff
@@ -0,0 +1,78 @@
...
+  # questions in the lesson
+  def self.grade_mastery_quiz(user_id, activity_id)
+    interactions = LessonInteraction.where(
+      "activity_id LIKE ? AND user_id = ? AND verb = ?", 
+      activity_id + '%',
+      user_id,
+      LessonInteraction::ANSWERED,
+    )
+    
+    return 0 if interactions.empty?
```
rshipp, line 59: I think using `LessonInteraction.count.where` is faster than `where` and `empty?` - but I'm not actually sure.

lib/lesson_grade_calculator.rb
rshipp: I'd add another function to this lib that does all the grades, and calls `compute_grade`. It's the same code regardless, I just think it makes more sense to put it here than in the rake task.

lib/lesson_grade_calculator.rb
rshipp: this logic is different than what I imagined, but it might be better. I think we should talk about this one, I'm not good at thinking through the logic stuff alone.

lib/lesson_grade_calculator.rb
```diff
@@ -33,46 +40,30 @@ def self.compute_grade(user_id, activity_id)
...
-          activity_id: activity_id,
-          created_at: most_recent,
-          success: true,
-        )
-      correct_count += 1.0 if answer.exists?
-    end
+    # Number of correct answers from most recent interactions
+    correct_answers = interactions
+      .where(
+        created_at: most_recent,
```
rshipp, line 62: i can't follow this logic, so I'm going to assume its right unless you wanted someone to double check it

spec/lib/lesson_grade_calculator_spec.rb
```diff
@@ -0,0 +1,224 @@
...
+        interactions = LessonInteraction
+          .for_user_and_activity(user.id, activity_id)
+          .where(verb: LessonInteraction::ANSWERED)
+        grade = LessonGradeCalculator.grade_mastery_quiz(interactions, quiz_questions)
+        expect(grade).to eq(2.0/quiz_questions * 100)
+      end
+
+      it "uses success from most recent interaction" do
+        quiz_question_id = "#{activity_id}/somequizid/somequestionid"
+        timestamp = Time.now.to_i
```
jessouyangbraven, line 191: Added this so the test actually fails when I `group` naively on `activity_id` for `ANSWERED` interactions.

lib/lesson_grade_calculator.rb
```diff
@@ -0,0 +1,89 @@
...
+  def self.grade_lesson_engagement(interactions)
+    interactions.maximum(:progress) || 0
+  end
+
+  # Returns [0, 100] for number of questions user got right out of the
+  # questions in the lesson
+  private_class_method
+  def self.grade_mastery_quiz(interactions, total_questions)
+    most_recent = interactions
+      .group(activity_id_without_timestamp)
```
jessouyangbraven, line 53: This is all because there's `_timestamp` at the end of the activity_id.

lib/lesson_grade_calculator.rb
```diff
@@ -0,0 +1,89 @@
...
+    # This gives us the index of _ before <timestamp>
+    idx = "POSITION('_' IN #{activity_id})"
+
+    # This strips off _<timestamp> from the activity_id
+    substring = "SUBSTRING(#{activity_id}, #{idx} + 1)"
+
+    # Remember we were operating on the reversed activity_id, so flip it around
+    # to make it human readable again
+    "REVERSE(#{substring})"
+  end
```
jessouyangbraven, line 88: For context: https://bebraven.slack.com/archives/C01166ZLR1V/p1598562500011800

I couldn't figure out a cleaner way to do this, so I tried to make it really explicit in comments and code what SQL shenanigans I was getting up to.

app/models/lesson_interaction.rb
```diff
@@ -5,4 +5,12 @@ class LessonInteraction < ApplicationRecord
...
   belongs_to :user
   validates :user, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id, presence: true
+
+  scope :for_user_and_activity, -> (user_id, activity_id) do
+    where(
+      "activity_id LIKE ? AND user_id = ?",
+      "#{activity_id}%",
+      user_id,
+    )
+  end
```
jessouyangbraven, line 15: Codacy complained about not using a `lambda` here. I looked around for some examples and decided to ignore Codacy and leave it as `->` after reading: https://github.com/testdouble/standard/issues/21

PR #334, opened 2020-08-27
Grading rake task

rshipp:
Task: https://app.asana.com/0/0/1189791588110885/f

Summary: Using the existing lesson interactions, grade calculator, and canvas API, add a rake task that ties them all together and sends grades to canvas.

Screenshot with 50% on each of three mastery quizzes, plus 100% module completion, `(50+100)/200 = 75`:
Screen Shot 2020-09-03 at 10 24 57 AM

Testing:

1. Upload a new Rise 360 module as a Canvas assignment, worth 10 points.
2. Log into canvas as a student and complete the module.
3. Run the rake task, `docker-compose exec platformweb bundle exec rake grade:lessons`
4. Check the grade tab for the correct grade in that assignment.

Example honeycomb trace: https://ui.honeycomb.io/braven/datasets/development/result/ga8nKtLSvPQ/trace/xsPpFiMZZQy


lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,35 @@
...
+      |record| [ record.user, record.root_activity_id ]
+    }
+
+    # Compute.
+    Honeycomb.start_span(name: 'lesson_grade.compute_grades') do |span|
+      grades = Hash.new
+      filtered_records.each do |record|
+        grades[record.canvas_course_id] ||= Hash.new
+        grades[record.canvas_course_id][canvas_assignment_id] ||= Hash.new
+        grades[record.canvas_course_id][canvas_assignment_id][record.user.canvas_id] = LessonGradeCalculator.compute_grades(record.user, record.root_activity_id)
```
jessouyangbraven, line 18: `compute_grade` singular in PR #333.

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,35 @@
+namespace :grade do
+  desc "grade lessons"
+  task grade: :environment do
+    records = LessonInteractions.select(:user, :activity_id, :verb).where(new: true).group(:user, :activity_id)
+    return if records.empty?
+
+    # Filter duplicate quiz activity_ids.
+    filtered_records = records.uniq {
+      |record| [ record.user, record.root_activity_id ]
+    }
```
jessouyangbraven, line 10: I'm not sure what `uniq` does. Are we going to get the most recent record for each quiz question when we dedupe? 

There's logic in `LessonGradeCalculator` to do this get the most recent record for each quiz question, so this might be unnecessary?

rshipp: this was just to take the list from 

```
user1, http://activity1/quiz1/question1
user1, http://activity1/quiz1/question2
user1, http://activity1/quiz2/question1
```

to 

```
user1, http://activity1
```

Otherwise, we'd run `compute_grades` once for each quiz question plus once for the overall lesson, when we only want to run it once.

rshipp: This is where I was reading about `uniq` https://www.rubyguides.com/2019/08/ruby-uniq-method/

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,35 @@
...
+        grades[record.canvas_course_id][canvas_assignment_id][record.user.canvas_id] = LessonGradeCalculator.compute_grade(record.user, record.root_activity_id)
+      end
+    end
+
+    Honeycomb.start_span(name: 'lesson_grade.update_grades') do |span|
+      # Send in batches.
+      grades.keys.each do |canvas_course_id|
+        grades[canvas_course_id].keys.each do |canvas_assignment_id|
+          grades_by_user_id = grades[canvas_course_id][canvas_assignment_id]
+          CanvasAPI.update_lesson_grades(canvas_course_id, canvas_assignment_id, grades_by_user_id)
```
jessouyangbraven, line 27: This hash needs to be keyed on canvas user ID. Am I just missing it in the code? 

I can add code to `CanvasAPI.update_lesson_grades` to do the user ID => canvas user ID look up if that's easier than doing it in the Rake task.

rshipp: it uses the canvas user id as the key here https://github.com/bebraven/platform/pull/334/files#diff-4f15adc169ba550856b204b4df59ca08R18

rshipp: This hash looks like 

```ruby
{
  canvas_course_id: {
    canvas_assignment_id: {
      canvas_user_id: grade,
    },
  },
}
```

jessouyangbraven: Ohhhhh. You have the user object. That makes sense!

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,35 @@
...
+      |record| [ record.user, record.root_activity_id ]
+    }
+
+    # Compute.
+    Honeycomb.start_span(name: 'lesson_grade.compute_grades') do |span|
+      grades = Hash.new
+      filtered_records.each do |record|
+        grades[record.canvas_course_id] ||= Hash.new
+        grades[record.canvas_course_id][canvas_assignment_id] ||= Hash.new
+        grades[record.canvas_course_id][canvas_assignment_id][record.user.canvas_id] = LessonGradeCalculator.compute_grade(record.user, record.root_activity_id)
```
jessouyangbraven, line 18: I think this has to be `record.user.id`.

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,41 @@
...
+    exit if records.empty?
+
+    # Filter duplicate quiz activity_ids.
+    filtered_records = records.uniq {
+      |record| [ record.user_id, record.root_activity_id ]
+    }
+
+    # Compute.
+    grades = Hash.new
+    Honeycomb.start_span(name: 'lesson_grade.compute_grades') do |span|
```
rshipp, line 19: wasn't sure what to use for the span name, since I usually use Class.method and there are none here

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,41 @@
+require 'lesson_grade_calculator'
+
+namespace :grade do
+  desc "grade lessons"
+  task lessons: :environment do
+    records = LessonInteraction
+      .select(:user_id, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id)
+      .where(new: true)
+      .group(:user_id, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id)
```
rshipp, line 9: this query selects a bunch of things, but its the best I can do to limit the amount of records we're grabbing while still getting all the info we need to compute grades

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,41 @@
...
+    exit if records.empty?
+
+    # Filter duplicate quiz activity_ids.
+    filtered_records = records.uniq {
+      |record| [ record.user_id, record.root_activity_id ]
+    }
```
rshipp, line 15: here i filter down the query results to things that are unique by user id and root (without the quiz/question part) activity id

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,41 @@
...
+
+    Honeycomb.start_span(name: 'lesson_grade.update_grades') do |span|
+      # Send in batches.
+      grades.keys.each do |canvas_course_id|
+        grades[canvas_course_id].keys.each do |canvas_assignment_id|
+          grades_by_user_id = grades[canvas_course_id][canvas_assignment_id]
+          CanvasAPI.client.update_lesson_grades(canvas_course_id, canvas_assignment_id, grades_by_user_id)
+          LessonInteraction.where(new: true, canvas_course_id: canvas_course_id, canvas_assignment_id: canvas_assignment_id).each do |record|
+            record.update(new: false)
+          end
```
rshipp, line 36: I'd like extra eyes here. this query seems to work, but I wasn't sure if I was selecting the right things to update.

lib/tasks/grade_lessons.rake
rshipp, line 9: could also do this `group by` as a `distinct` instead, which would make the ruby cleaner, but `distinct` is usually waaaay slower.

lib/tasks/grade_lessons.rake
jessouyangbraven: I might be missing something here, but don't we have all the `LessonInteractions` in `records` from line 6 to set `new: false` on? 

We do a bunch of filtering and manipulation on `records` to figure out the Canvas course/assignment/user IDs to use with `LessonGradeCalculator` and `CanvsAPI`, but aren't they in effect all the records we're doing the grade computations over?

jessouyangbraven: This isn't something to address in this PR, but my mind went down the rabbit hole, so I'm going to record it here.

I don't think we can run multiple instances of this Rake task simultaneously. That's fine, but then I worry about atomicity violations even within a single run, since we're not using transactions anywhere and doing `LessonInteraction` `selects` multiple times throughout (here and `LessonGradeCalculator`). So it's possible we get `new` records between `LessonInteraction.where()`s. 

I think we're okay because we don't care about the `new` field when doing the actual grade computation. As long as we trigger computation at some point, we'll compute the right numerical value for the grade. 

The thing to go wrong here would be marking something as `new: false` before it was used to compute the grade and never triggering grade computation again because there are never any other `new` records that come in. 

All this to say, it's hard to reason about and I wouldn't want to be the one debugging why a bunch of students's grades are wrong some of the time.

jessouyangbraven (commented):
The only thing to resolve how we're setting `new: false`. Everything else is non-blocking.

lib/tasks/grade_lessons.rake
rshipp: the parent of all of this is called `rake:grade:lessons`, so I could use e.g. `rake:grade:lessons:compute`? i'm not sure how span names are actually used outside the waterfall view, maybe that's good enough.

lib/tasks/grade_lessons.rake
rshipp: > all the LessonInteractions in records from line 6 to set new: false on?

Unfortunately, since we don't select `:id`, we can't use those `records` to reference the original rows in the table. They're essentially read-only and disconnected from the database as soon as we select them.

The risk of records being added in between when we select and grade, and us marking them as old without ever reading them, is an issue though. The only way I can think of to fix that is to stop using `group` and `uniq` to filter, and maintaining all the full records in memory until we've processed them, so we can mark the exact records we used as old. I'm worried that doesn't scale well. We could batch the select with `limit`, loop as many times as we need to until there are no longer new records, and do one `update_lesson_grades` per loop? More network requests, more SQL queries, but guaranteed correct results (still limited to 1 instance of the rake task at once). Is there a better way?

lib/tasks/grade_lessons.rake
rshipp: I think the only worries are 1) the rake task lasts longer than a day and the heroku app restarts in the middle, and 2) we use up a bunch of cpu/memory on our heroku app and it autoscales and charges us more and/or makes the UI slower. Distinct scales poorly with large tables, even if you're only selecting a few records, to the point where I've seen `group by` take under a second while `distinct` takes minutes.

jessouyangbraven: Gotcha. Thanks!

lib/tasks/grade_lessons.rake
jessouyangbraven: I agree with the general approach you've outlined. And think we should guarantee correct results first and then work on improving any scaling issues. 

The only "better" way I can think of is multithreading the computation, since the grades are independent of each other. Going your approach of working out of local memory helps with this, since we won't have data races on the DB from the threads.

rshipp: note to self: select max(id) before the first select, and use that to limit marking things in the second select

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,46 @@
...
+    # be the same on each.
+    filtered_records = records.uniq {
+      |record| [ record.user_id, record.root_activity_id ]
+    }
+
+    # Compute.
+    grades = Hash.new
+    Honeycomb.start_span(name: 'rake:grade:lessons:compute') do |span|
+      filtered_records.each do |record|
+        user = User.find(record.user_id)
```
sadleb, line 25: If we add `user_canvas_id` to `LessonInteraction` we could avoid querying for the user when processing. Is that worth it? Are there other situations where we need the users's canvas_id when dealing with a `LessonInteraction`?

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,46 @@
+require 'lesson_grade_calculator'
+
+namespace :grade do
+  desc "grade lessons"
+  task lessons: :environment do
+    max_id = LessonInteraction.maximum(:id)
```
sadleb, line 6: I would add a comment about the "why" behind this. Aka, interactions could come in while we're processing and we don't want to set them to processed after we've done the computation.

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    Honeycomb.start_span(name: 'rake:grade:lessons:update') do |span|
+      # Send in batches.
+      grades.keys.each do |canvas_course_id|
+        grades[canvas_course_id].keys.each do |canvas_assignment_id|
+          grades_by_user_id = grades[canvas_course_id][canvas_assignment_id]
+          CanvasAPI.client.update_lesson_grades(canvas_course_id, canvas_assignment_id, grades_by_user_id)
+          LessonInteraction.where(new: true, canvas_course_id: canvas_course_id, canvas_assignment_id:
+              canvas_assignment_id).where('id <= ?', max_id).each do |record|
```
sadleb, line 39: I don't think we care about the `updated_at` timestamp being reflected when setting `new: false` so I think we could just do `.where('id <= ?', max_id).update_all(new: false)` which would issue a single SQL statement and skip all the validations / callbacks.

sadleb (changes requested):

lib/tasks/grade_lessons.rake
rshipp: We need the user id too, I don't think adding a column to duplicate the canvas user ID in this table would be worth it. Selecting by id is an inexpensive query.

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,46 @@
...
+      .select(:user_id, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id)
+      .where(new: true)
+      .group(:user_id, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id)
+    exit if records.empty?
+
+    # Filter duplicate quiz activity_ids, so we only compute grades once for each (user,activity) pair.
+    # It doesn't matter which record we pick when we discard these "duplicates", because the info we
+    # care about (canvas course id, canvas assignment id, user id, root activity id) will always
+    # be the same on each.
+    filtered_records = records.uniq {
```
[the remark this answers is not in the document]
rshipp: Is `records = nil` enough to let the garbage collector run, if it needs to? I don't think it'd break anything.

lib/tasks/grade_lessons.rake
rshipp: With our new max-id change, it's totally fault tolerant.

lib/tasks/grade_lessons.rake
```diff
@@ -0,0 +1,46 @@
+require 'lesson_grade_calculator'
+
+namespace :grade do
+  desc "grade lessons"
+  task lessons: :environment do
+    max_id = LessonInteraction.maximum(:id)
+    records = LessonInteraction
```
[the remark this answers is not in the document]
rshipp: It's already in a span, but I can add `records.length` and `max_id` as a field.

app/models/lesson_interaction.rb
```diff
@@ -13,4 +13,13 @@ class LessonInteraction < ApplicationRecord
       user_id,
     )
   end
+
+  def root_activity_id
+    if verb == ANSWERED
+      # Remove the /quiz_id/question_id part.
+      activity_id.split('/')[0..-3].join('/')
```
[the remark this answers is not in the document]
rshipp: I can, but the only part of the id that matters is the part in the comment there. `.split('/').[0..-3]` means "starting at the end, remove up to the second slash"

lib/tasks/grade_lessons.rake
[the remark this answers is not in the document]
rshipp: Can't find a good answer to this. https://stackoverflow.com/questions/19523039/how-to-destroy-ruby-object 

I set it to nil but added a comment noting it's not guaranteed to do anything.

app/models/lesson_interaction.rb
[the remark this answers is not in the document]
sadleb: The comment is super helpful! I was trying to understand the `[0..-3]` and was completely confused.

sadleb (approved):
lgtm

PR #336, opened 2020-09-01
PROTOTYPE - Add surveys (peer review and impact surveys)

rshipp:
Task: https://app.asana.com/0/1174274412967132/1190325701574061/f

Summary: Add the ability to create and use custom forms (for impact surveys), and autogenerated peer-review surveys.

## Peer reviews

Peer reviews are autogenerated based on the other students in the Canvas section(s) you are a member of. To test, go to your course > People > click the 3 dots on the right of one of the users, and edit sections. You must have at least one **student** account that is not the account you are testing with, in the same section as the account you are testing with. E.g. when testing with `Ryan Shipp`:

Screen Shot 2020-09-04 at 11 29 53 AM

Peer reviews are always at https://platformweb/form_submissions/peer_review, and must be accessed through an LTI launch. To test standalone: visit the page, it'll raise an exception and give you a rails console, run `LtiLaunch.last.state`, and add the output string to the URL in a `?state=` param.

When submitted, peer reviews POST to `/form_submissions`.

## Custom forms (aka surveys)

Custom forms are a new `CourseContent.content_type = 'form'`. The content_type is the only difference from an `assignment` or `wiki_page`. When viewing the `course_content_histories` `show` view, the HTML body of the CourseContent will be rendered inside a `form`, with a submit button at the bottom.

When submitted, custom forms POST to `/form_submissions`.

## FormKeyValues

The new `FormKeyValue` model, along with the `form_submissions` controller, is the backbone of both peer reviews and custom forms. When a form is submitted, instead of creating a single object, it creates many FormKeyValue objects - one per field in the form. There will only ever be one `user, key` pair in the `FormKeyValue` table, so you don't have to worry about selecting the "latest" value. If the same user submits the same form twice, old values are overwritten with new values.

## Future work

We've discussed this in the past - at some point, we should replace the autogenerated `data-bz-retained` IDs in CKEditor code with something database-backed, so we can actually do uniqueness correctly, and track what questions are in which pages. If we ever get that far, `FormKeyValues` is intended to be that backend.

## Screenshots

<details><summary>Peer review</summary>
Screen Shot 2020-09-03 at 11 39 15 AM
</details>

<details><summary>Example form/survey</summary>
Screen Shot 2020-09-03 at 3 25 04 PM
</details>

<details><summary>LTI insert dialog</summary>
Screen Shot 2020-09-03 at 3 09 41 PM
</details>

Placeholder for refactoring LTI code so that we're not just putting everything in `lti_assignment`: https://app.asana.com/0/1174274412967132/1192382771093473/f

### DIFF
--- a/app/controllers/form_submissions_controller.rb
+++ b/app/controllers/form_submissions_controller.rb
@@ -16,7 +16,7 @@ class FormSubmissionsController < ApplicationController
   def create
     ActiveRecord::Base.transaction do
       request.request_parameters.keys.each do |key|
-        next if ['authenticity_token', 'commit'].includes? key
+        next if ['authenticity_token', 'commit'].include? key
         FormKeyValue.create_with(value: request.request_parameters[key]).create_or_find_by!(key: key, user: current_user)
       end
     end
--- a/app/controllers/form_submissions_controller.rb
+++ b/app/controllers/form_submissions_controller.rb
@@ -2,11 +2,14 @@ require 'lti_advantage_api'
 
 class FormSubmissionsController < ApplicationController
   include LtiHelper
+  layout 'content_editor'
 
   # Non-standard controller without normal CRUD methods. Disable the convenience module.
   def dry_crud_enabled?
     false
   end
+
+  before_action :set_lti_launch, only: [:peer_review]
 
   # POST /form_submissions
   # POST /form_submissions.json
@@ -25,6 +28,6 @@ class FormSubmissionsController < ApplicationController
     @lti_launch.section_ids.each do |section_id|
       all_users += CanvasAPI.client.get_section_students(@lti_launch.course_id, section_id)
     end
-    @users = all_users.delete_if { |x| x['id'] == current_user.canvas_id }
+    @users = all_users.uniq.delete_if { |x| x['id'] == current_user.canvas_id }
   end
 end
--- a/app/controllers/form_submissions_controller.rb
+++ b/app/controllers/form_submissions_controller.rb
@@ -8,8 +8,8 @@ class FormSubmissionsController < ApplicationController
     false
   end
 
-  # POST /form_submission
-  # POST /form_submission.json
+  # POST /form_submissions
+  # POST /form_submissions.json
   def create
     ActiveRecord::Base.transaction do
       request.request_parameters.keys.each do |key|
--- a/app/controllers/form_submissions_controller.rb
+++ b/app/controllers/form_submissions_controller.rb
@@ -1,0 +1,30 @@
+require 'lti_advantage_api'
+
+class FormSubmissionsController < ApplicationController
+  include LtiHelper
+
+  # Non-standard controller without normal CRUD methods. Disable the convenience module.
+  def dry_crud_enabled?
+    false
+  end
+
+  # POST /form_submission
+  # POST /form_submission.json
+  def create
+    ActiveRecord::Base.transaction do
+      request.request_parameters.keys.each do |key|
+        next if ['authenticity_token', 'commit'].includes? key
+        FormKeyValue.create_with(value: request.request_parameters[key]).create_or_find_by!(key: key, user: current_user)
+      end
+    end
+  end
+
+  # GET /form_submissions/peer_review
+  def peer_review
+    all_users = []
+    @lti_launch.section_ids.each do |section_id|
+      all_users += CanvasAPI.client.get_section_students(@lti_launch.course_id, section_id)
+    end
+    @users = all_users.delete_if { |x| x['id'] == current_user.canvas_id }
+  end
+end
--- a/app/models/form_key_value.rb
+++ b/app/models/form_key_value.rb
@@ -1,0 +1,5 @@
+class FormKeyValue < ApplicationRecord
+  belongs_to :user
+
+  validates :user, :key, presence: true
+end
--- a/app/models/lti_launch.rb
+++ b/app/models/lti_launch.rb
@@ -72,6 +72,17 @@ class LtiLaunch < ApplicationRecord
     "#{Rails.application.secrets.canvas_cloud_url}/courses/#{cid}/assignments/#{aid}"
   end
 
+  # The course ID in this launch. Raises ArgumentError if not an int.
+  def course_id
+    Integer(request_message.custom['course_id'])
+  end
+
+  # The section IDs this user is added to, as a list.
+  def section_ids
+    # Original format looks like "section_ids": "49,200"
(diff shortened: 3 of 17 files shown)


app/controllers/form_submissions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    false
+  end
+
+  before_action :set_lti_launch, only: [:peer_review]
+
+  # POST /form_submissions
+  # POST /form_submissions.json
+  def create
+    ActiveRecord::Base.transaction do
+      request.request_parameters.keys.each do |key|
```
sadleb, line 18: Note that by default Rails controllers have a `wrap_parameters` feature where all the parameters are added inside a `form_submission` param. This is convenient for controllers that deal with models. For others, you can turn it off using `wrap_parameters false`

rshipp: I thought about doing that, but I don't think we'll ever use JSON here. Do you think it's worth doing anyway?

app/controllers/form_submissions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        next if ['authenticity_token', 'commit'].include? key
+        FormKeyValue.create_with(value: request.request_parameters[key]).create_or_find_by!(key: key, user: current_user)
```
sadleb, line 20: What's the difference b/n `request.request_parameters[key]` and `params[key]`?

app/controllers/form_submissions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  # Non-standard controller without normal CRUD methods. Disable the convenience module.
+  def dry_crud_enabled?
+    false
+  end
+
+  before_action :set_lti_launch, only: [:peer_review]
+
+  # POST /form_submissions
+  # POST /form_submissions.json
+  def create
```
sadleb, line 16: Is there anyway to not allow random keys coming through and to make sure that only those that we intend are allowed? The only thing I can think of is either manually writing a whitelist or dynamically scraping the HTML of the view looking for inputs and whitelisting those. Is there a security concern here?

app/controllers/form_submissions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        next if ['authenticity_token', 'commit'].include? key
+        FormKeyValue.create_with(value: request.request_parameters[key]).create_or_find_by!(key: key, user: current_user)
+      end
+    end
+  end
+
+  # GET /form_submissions/peer_review
+  def peer_review
+    all_users = []
+    @lti_launch.section_ids.each do |section_id|
```
sadleb, line 28: We may need to discuss this more. If they are in multiple sections, how do we decide which one is *the* cohort. For the prototype I think it's fine to not deal with this, but we should queue up a task to revisit how to ensure that we're coming up with the correct list of peers.

app/controllers/form_submissions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      all_users += CanvasAPI.client.get_section_students(@lti_launch.course_id, section_id)
```
sadleb, line 29: What's the difference b/n `+=` and `<<` when adding to an array? I usually use `<<`

app/models/form_key_value.rb
```diff
@@ -0,0 +1,5 @@
+class FormKeyValue < ApplicationRecord
+  belongs_to :user
+
+  validates :user, :key, presence: true
```
sadleb, line 4: I think we should give each form an ID and associate the key/value with the form as well.

app/models/lti_launch.rb
```diff
@@ -72,6 +72,17 @@ def activity_id
...
 
+  # The course ID in this launch. Raises ArgumentError if not an int.
+  def course_id
+    Integer(request_message.custom['course_id'])
+  end
+
+  # The section IDs this user is added to, as a list.
+  def section_ids
+    # Original format looks like "section_ids": "49,200"
+    request_message.custom['section_ids'].split(',')
```
sadleb, line 83: Edge case, if they are not in any sections what happens?

app/controllers/form_submissions_controller.rb
rshipp: In theory, nothing, but reading from a different thing than you're iterating over just to save a few characters seems confusing.

app/controllers/form_submissions_controller.rb
rshipp: The only concern is someone submitting a bunch of things, that'd still be tied to their account. There's no way to restrict keys, since this is by definition an arbitrary key:value store.

app/models/form_key_value.rb
rshipp: We can't do that here, because peer reviews don't have an ID. We'd have to make a `course_content_form_key_values` join table, and I don't want to add another table when we're not sure we'd ever need it ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)).

app/models/lti_launch.rb
rshipp: Not sure, I'll add this to my list to unit test.

app/controllers/form_submissions_controller.rb
rshipp: We're adding lists together, not adding single elements:

```
2.2.3 :013 > a = []
 => []
2.2.3 :014 > a << 1
 => [1]
2.2.3 :015 > a = []
 => []
2.2.3 :016 > a += [1]
 => [1]
```

app/controllers/form_submissions_controller.rb
rshipp: Is there an example of when that might happen? If the student is in two courses, with one section in each course, this code will actually raise a 500. My thought was that this is something we need to manage correctly on the Canvas side.

sadleb: Decision: have this throw if they are in more than one section. Setup an alert on Honeycomb to monitor for that error but don't turn it on. Throw a meeting on the calendar to turn it on in Jan time-frame when we're closer to launch and not testing random stuff.

app/models/form_key_value.rb
sadleb: Discussed verbally, there is a lot to figure out about how we want to architect this. We're going to leave this draft alone for the time being and move onto the next project that we need to figure out for the overall Highlander experience. Once we have more things fleshed out broadly, we'll have a view of the moving pieces and will be able to more concretely discuss how the architecture can cleanly support all the features.

PR #340, opened 2020-09-10
Implemented LinkedIn authorization flow

jessouyangbraven:
**Task**: https://app.asana.com/0/1174274412967132/1191740236188816

**Development environment breaking changes**
- You need to run a DB migration
- Add `LINKED_IN_CLIENT_{ID, SECRET}` to your `.env` file

**Test**:

- Go to `https://platformweb/linked_in/login` endpoint on desktop browsers (Chrome, Safari, Firefox)

Screen Shot 2020-09-14 at 7 01 12 PM

- Using `linked_in/signin` on Heroku test app on mobile browser (Safari)
IMG_1601

- Create a Project in Content Editor. Add Section, then iFrame, with URL `https://platformweb/linked_in/login`, save the Project, and add it as a Canvas assignment. 

Screen Shot 2020-09-15 at 2 31 55 PM

Screen Shot 2020-09-15 at 2 33 15 PM

Screen Shot 2020-09-15 at 2 33 24 PM

- Once you've accepted using one the above methods, look up the access token in `devdb`
```
select linked_in_access_token from users where id=<your user ID>;`
```

- Query the API and make sure you get a JSON blob of data back:
```
curl 'https://api.linkedin.com/v2/me' \
-X GET \
-H "Authorization: Bearer <linked_in_access_token>"
```

- Query the API using `LinkedInConnection` in `devc`:
> require 'linked_in_connection'
> LinkedInConnection.get_request('/v2/me', Users.find(your_user_id).linked_in_access_token)

Screen Shot 2020-09-15 at 3 17 01 PM

- You can also check your LinkedIn profile > Settings > Data Privacy > Other applications (https://www.linkedin.com/psettings/permitted-services) and see that Braven is listed. Note: to test this again with the same user, you'll need to "Remove" the Braven permissions. 

Screen Shot 2020-09-15 at 12 33 36 PM

**Details**

**Testing**
- Added `ci_exclude` tag for potentially flaky tests that talk to LinkedIn
- Updated `app.json` so Heroku CI skips `ci_exclude` tests
- Updated Testing section in `README`

**LinkedIn**
- We can't embed the LinkedIn authorization flow in our Canvas iframe (it violates the ancestor policy that LinkedIn has set)
- Need to whitelist redirect endpoint (`https://platformweb/linked_in/auth_redirect`) to Braven Accelerator app on LinkedIn (https://www.linkedin.com/developers/apps/5004876/settings)
- You'll also need to whitelist the Heroku Review app if you want to use that to test (just don't forget to delete it)
- The authorization code you get back from LinkedIn's redirect is valid for 30 minutes for getting the access token.

**New window**
- For window features, see: https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Window_features
- To get a new window (instead of a new tab), you need to specify `width` and `height` in the Javascript when creating the new window. 
- To get the window to center before it renders, you need to specify `left` and `top` in the window features. 

**React examples**
- https://medium.com/hackernoon/using-a-react-16-portal-to-do-something-cool-2a2d627b0202
- https://dev.to/dinkydani21/how-we-use-a-popup-for-google-and-outlook-oauth-oci

**Future work**
~~Content Editor output: https://app.asana.com/0/1174274412967132/1193832365458406~~
- Prevent multiple pop-up windows from opening: https://app.asana.com/0/1174274412967132/1194517634372073
- More UI feedback: https://app.asana.com/0/1174274412967132/1193832365458401


config/webpacker.yml
```diff
@@ -11,7 +11,9 @@ default: &default
 
   # Additional paths webpack should lookup modules
   # ['app/assets', 'engine/foo/app/assets']
-  resolved_paths: []
+  resolved_paths: [
+    'app/assets',
```
jessouyangbraven, line 15: This lets me access the LinkedIn button image in React components.

app/controllers/linked_in_authorization_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  def dry_crud_enabled?() false end
+
+  # iframe-able endpoint that renders a static LinkedIn button
+  def signin
+  end
+
+  # Genereates and renders the LinkedIn authorization URL in the current
+  # window after doing some server-side set-up
+  def launch
+    Honeycomb.start_span(name: 'LinkedInAuthorizationController.launch') do |span|
```
jessouyangbraven, line 21: I went back and forth on whether to put the `Honeycomb` instrumentation in the controller or the `LinkedInConnection`. 

I went with this controller in the end because we have more context here with respect to `params`, etc.

app/javascript/components/LinkedIn/LinkedInAuthorizationButton.js
```diff
@@ -0,0 +1,54 @@
...
+    window.open(
+      this.props.url,
+      '', // this window title gets overwritten by LinkedIn
+      windowFeaturesStr,
+    );
+  }
+
+  render() {
+    return (
+      <a id="linked-in-sign-in" onClick={this._handleClick}>
```
jessouyangbraven, line 47: FYI, Capybara only lets you find links by id or text, not by name. They're happy to use name for button, though, which is what I did in `project_submit_spec`, so I got confused for a bit.

app/javascript/components/LinkedIn/LinkedInAuthorizationButton.js
```diff
@@ -0,0 +1,54 @@
...
+const HEIGHT = 750;
+
+class LinkedInAuthorizationButton extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this._handleClick = this._handleClick.bind(this);
+  }
+
+  _handleClick(event) {
```
jessouyangbraven, line 19: I also tried putting `window.open()` in the `LinkedInAuthorizationWindow` component. 

The complication is that we need to hit the server at some point to generate the CSRF token and save it to `user.linked_in_state`. 

That meant that the `LinkedInAuthorizationWindow` component would have to do an AJAX request to generate `state`, get it back, and then redirect to LinkedIn with it.

It was a pain, so I had this click handler create the window to hit `linked_in/auth`, generate `state`.

config/routes.rb
```diff
@@ -97,6 +97,11 @@
   get '/cas/proxyValidate', to: 'cas#proxyValidate'
   get '/cas/proxy', to: 'cas#proxy'
 
+  # LinkedIn authorization routes
+  get '/linked_in/signin' => 'linked_in_authorization#signin'
```
jessouyangbraven, line 101: This is what we use in the iframe in Content Editor.

Is there a more intuitive name for it? This is the best I could come up with.

lib/linked_in_connection.rb
```diff
@@ -0,0 +1,58 @@
...
+require 'rest-client'
+require 'uri'
+
+# This class handles communicating with the LinkedIn's authorization flow:
+#  https://docs.microsoft.com/en-us/linkedin/shared/authentication/authorization-code-flow
+# and LinkedIn's API:
+#  https://docs.microsoft.com/en-us/linkedin/shared/api-guide/concepts
+# behalf fo the Braven's LinkedIn app: 
+#  https://www.linkedin.com/developers/apps/5004876
+class LinkedInConnection
```
jessouyangbraven, line 12: I called it `Connection` instead of `API` because that's how the portal code was. Also because it handles both authentication and API calls.

spec/feature/linked_in_authorization/linked_in_authorize_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    context "redirect" do
+      let (:path) { linked_in_signin_path }
+
+      it "closes the pop-up", js: true do
+        popup = window_opened_by { find('#linked-in-sign-in').click }
+
+        within_window popup do
+          # Don't sign in on LinkedIn
+          find("a", :text => "Cancel").click
```
jessouyangbraven, line 72: I'm still working on actually signing in and checking that we save the access_token in the DB.

spec/feature/linked_in_authorization/linked_in_authorize_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+ 
+        # Check pop-up
+        within_window popup do
+          # Redirected to LinkedIn
+          expect(current_url).to include('https://www.linkedin.com/')
+
+          # See initial sign-in page
+          # Note: this page title is specified by LinkedIn, if it changes
+          # this test will fail until we update the string.
+          expect(page).to have_title('LinkedIn Login, Sign in | LinkedIn')
```
rshipp, line 45: Is this something we can't use VCR for, like it didn't work with AJAX calls? Calling out to something on the real internet is exactly the kind of thing we'd normally want to mock out.

spec/feature/linked_in_authorization/linked_in_authorize_spec.rb
rshipp: isn't that impossible unless we mock it out? otherwise it'd need a real linkedin account and embedded credentials

jessouyangbraven: Update from our meeting today:

- the Selenium test doesn't know about these network requests
- we're testing that the access token gets saved in the controller spec
- there might be another testing framework (puffingbilly?) that would let us do end-to-end tests like this in the future

spec/feature/linked_in_authorization/linked_in_authorize_spec.rb
jessouyangbraven: Yeah, VCR won't work for this. There's a `window.location =...` in the React component that it won't detect. I modified this test and left it in so we can test that the pop-up is being created. I can change how we test it to `expect(popup.open?).to be true` if you think that's a better way to do it.

config/routes.rb
sadleb: See my comment above about `login` vs `signin` vs `sign-in`. I thought about trying to use a name that made it clear it was just rendering a button to start off the flow, but couldn't come up with anything better that made sense.

lib/linked_in_connection.rb
sadleb: Portal will be gone hopefully next year, so we shouldn't make decisions based off that. It's a bit odd to have `CanvasAPI`, `SalesforceAPI`, `LtiAdvantageAPI`, and `LinkedInConnection` when there isn't really any logic in here other than interacting with the LinkedIn API.

lib/linked_in_connection.rb
```diff
@@ -0,0 +1,58 @@
...
+      response_type: 'code',
+      state: state,
+    }.to_query
+    url.to_s
+  end
+
+  # Exchange authorization code for access token with LinkedIn
+  # See: https://docs.microsoft.com/en-us/linkedin/shared/authentication/authorization-code-flow#step-3-exchange-authorization-code-for-an-access-token
+  def self.exchange_code_for_token(redirect_url, authorization_code)
+    response = RestClient.post(
```
sadleb, line 34: fyi: as mentioned above, we've auto-instrumented RestClient calls with Honeycomb spans and error logging.

spec/feature/linked_in_authorization/linked_in_authorize_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        # Check pop-up
+        within_window popup do
+          expect(current_url).to include('https://www.linkedin.com/')
+
+          # Note: this page title is specified by LinkedIn, if it changes
+          # this test will fail until we update the string.
+          # We only check that the title ends to "| LinkedIn" because it can
+          # vary depending on whether the test runs on an endpoint that's
+          # whitelisted by Braven's LinkedIn app. 
+          expect(page.title).to match(/| LinkedIn$/)
```
sadleb, line 42: I'm still struggling to think about, from a higher level than this PR, if we want specs that hit external servers. I know we already do for some other things like Honeycomb, but it scares me to actually be hitting the linked in servers every-time we run our tests.

sadleb (commented):
None of my feedback would rise to the level of "we shouldn't merge this" but I didn't want to explicitly approve until I hear your thoughts.

spec/feature/linked_in_authorization/linked_in_authorize_spec.rb
rshipp: I don't think we're hitting honeycomb anywhere in tests, we explicitly ignore honeycomb requests through VCR.

app/controllers/linked_in_authorization_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+require 'linked_in_connection'
+
+class LinkedInAuthorizationController < ApplicationController
+  layout 'content_editor'
+
+  def dry_crud_enabled?() false end
+
+  # iframe-able endpoint that renders a static LinkedIn button
+  def signin
```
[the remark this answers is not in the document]
jessouyangbraven: I'm cool with `login` to keep it consistent with the other endpoints.

app/controllers/linked_in_authorization_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      @authorization_url = LinkedInConnection::authorize_url(
+        redirect_url,
+        state,
+      )
+    end
+  end
+
+  # Redirect endpoint for LinkedIn authorization flow
+  def redirect
```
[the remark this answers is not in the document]
jessouyangbraven: I agree `redirect` is confusing. 

But I also got confused with Portal's `oauth_success` because we can get an error back if the user cancels out of login/granting permissions.

I think I called this `oauth_redirect` in `routes.rb`. Is that any better than the other two?

app/javascript/components/LinkedIn/LinkedInAuthorizationButton.js
```diff
@@ -0,0 +1,54 @@
+// LinkedIn button rendered by linked_in_authorization#iframe.
+// This button launches the LinkedIn authorization flow in a pop-up window.
+
+import Rails from '@rails/ujs';
+import React from "react";
+
+import LinkedInButton from 'images/linked_in_button.png';
+
+const WIDTH = 500;
+const HEIGHT = 750;
```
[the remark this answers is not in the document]
jessouyangbraven: Empirically on my laptop to get all the LinkedIn stuff to fit without a scroll bar. I Google-ed around for a more rigorous way, but didn't turn up anything. I wish there were a pre-existing React Bootstrap component for this.

spec/feature/linked_in_authorization/linked_in_authorize_spec.rb
jessouyangbraven: I added `ci_exclude` that we can tag tests that we want skipped for continuous integration tests and updated `app.json`.

app/controllers/linked_in_authorization_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  # Genereates and renders the LinkedIn authorization URL in the current
+  # window after doing some server-side set-up
+  def launch
+    Honeycomb.start_span(name: 'LinkedInAuthorizationController.launch') do |span|
+      state = SecureRandom.hex
+      redirect_url = linked_in_auth_redirect_url
+
+      span.add_field('user.linked_in_state', state)
```
sadleb, line 25: You can just do Honeycomb.add_field() b/c the controller `launch` action is already auto-instrumented so this whole method would be in a span already.

sadleb (approved):
Lgtm.

rshipp (approved):

PR #341, opened 2020-09-11
Add course resources, restructure courses

rshipp:
# Summary

Task: https://app.asana.com/0/1174274412967132/1192413211463500/f
DB diagram: https://dbdiagram.io/d/5f5b8dba10a0a51c74d4ac40

What's done so far...

# Course resources

* Visiting /course_resources/new lets you create a new CourseResource, with zipfile
* Course and CourseTemplate forms let you attach a CourseResource
* Course lets you add a Canvas Course ID
* The static URL `/lti/course_resources` uses the current LTI launch to show the resource for the current course, or a 404 page if there isn't one configured
  * (Add the Course Navigation LTI placement in your developer key with URI `https://platformweb/lti/course_resources` and text `Resources`)

## Screenshots

<details><summary>Course Resources showing a Rise 360 module in Canvas</summary>Screen Shot 2020-09-14 at 4 00 29 PM</details>

# Restructuring

* Renamed `Program` to `BaseCourse`, removed `Course`
* Updated (most??) references to `Program` to point to `BaseCourse` - some may need to change again to reference `Course` instead
* Added new Course and CourseTemplate, which inherit from BaseCourse
* View all BaseCourses at /course_management
* Clicking on one takes you to either /courses/:id, or /course_templates/:id, depending on type
* Clicking 'edit' for each one takes you to either /courses/:id/edit, or /course_templates/:id/edit, depending on type
* Clicking 'delete' for each one deletes correctly, depending on type
* Success flashes use the correct humanized version of the name 'Course' or 'Course template', depending on type
* Edit forms update the object correctly
* Visiting /courses/new and /course_templates/new use the correct model for the form, and create correctly
* All Course/CourseTemplate objects use the same `base_course` views and controller
* Removed Organizations, Course `term`, and all related references
* Renamed `ProgramMembership` to `CourseMembership`, updated references

# Things to Review

* For things that reference `BaseCourse` or `Course`, are they referencing the right one? (Courses should have users and other things tied to a specific run of a course; BaseCourse should have lessons and other things that are reused in every run of the course.) Specifically, are GradeCategory and its friends Logistic and Section using the right things?


app/models/base_course.rb
```diff
@@ -1,31 +1,46 @@
...
+  # executing for a given semester at a given school.
 
-  validates :name, presence: true
+  validates :name, :type, presence: true
 
   def to_show
     attributes.slice('name')
   end
+
+  def humanized_type
```
rshipp, line 38: should this be here or in a helper?

app/models/course_membership.rb
```diff
@@ -1,7 +1,10 @@
-class ProgramMembership < ApplicationRecord
+class CourseMembership < ApplicationRecord
   belongs_to :user
-  belongs_to :program
   belongs_to :role
+  belongs_to :course, -> {
+    where(base_courses: { type: 'Course' })
+  }, foreign_key: :base_course_id
```
rshipp, line 6: what does this mean? and am I doing it right?

app/models/course_resource.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  after_create_commit :publish
+
+  # Rise 360 resource zipfile
+  has_one_attached :course_resource_zipfile
+
+  validates :name, presence: true
+
+  def launch_url
+   "#{LtiLessonContentsProxy.proxy_url}#{LessonContentUtil.launch_path(course_resource_zipfile.key)}"
```
rshipp, line 15: This is reusing lessoncontents stuff without renaming, should we rename it to be less specific now that we're using it in more places?

app/views/base_courses/_list.html.erb
```diff
@@ -0,0 +1,24 @@
...
+          <th></th>
+        </tr>
+      </thead>
+
+      <tbody>
+        <% base_courses.each do |base_course| %>
+          <tr>
+            <td><%= link_to base_course.name, base_course %></td>
+            <td class="text-right smaller">
+              <%= link_to 'Edit', edit_polymorphic_path(base_course) %> |
```
rshipp, line 16: polymorphic_path helpers resolve to the correct route for the object we pass in. in this case, /courses/1/edit or /course_templates/1/edit

app/views/base_courses/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for(:headline) do %>
+  Course Management
+<% end %>
+
+<%= success(notice) %>
+
+<h3>Course Templates</h3>
+<%= link_to 'New course template', new_course_template_path, class: 'btn btn-sm btn-outline-secondary' %>
+<%= render 'list', base_courses: @base_courses.where(type: 'CourseTemplate') %>
```
rshipp, line 9: I pulled the table out into a `_list` view to make it easier to change, and so we don't have two copies of it. I assume we'll heavily redesign this page at some point.

app/views/course_resources/show.html.erb
```diff
@@ -0,0 +1,4 @@
+<p>
+  <strong>You should have been redirected</strong>
+</p>
+
```
rshipp, line 4: not sure we still need this file

app/views/layouts/application.html.erb
```diff
@@ -22,6 +22,8 @@
   </head>
 
   <body>
+    <h2><%= content_for?(:headline) ? yield(:headline) : 'Platform' %></h2>
+    <%= content_for?(:headline_buttons) ? yield(:headline_buttons) : '' %>
```
rshipp, line 26: we were using these `:headline` and `:headline_buttons` things all over, but never showed them anywhere, so I just threw them in here.

config/routes.rb
```diff
@@ -23,7 +25,9 @@
   resources :locations, only: [:index, :show]
   resources :majors, except: [:show]
 
-  resources :programs
+  resources :base_courses, only: [:index], path: 'course_management'
+  resources :courses, controller: 'base_courses', type: 'Course'
+  resources :course_templates, controller: 'base_courses', type: 'CourseTemplate'
```
rshipp, line 30: note you can technically visit `/courses` and `/course_templates` and go to the base_courses #index view. i didn't want to type out every possible view except index for the `only` param on these other two. :/

config/routes.rb
```diff
@@ -107,6 +111,8 @@
   get '/lti/link_selection/new', to: 'lesson_contents#new' # https://canvas.instructure.com/doc/api/file.link_selection_placement.html
   post '/lti/link_selection', to: 'lesson_contents#create' # https://canvas.instructure.com/doc/api/file.link_selection_placement.html
 
+  get '/lti/course_resources', to: 'course_resources#lti_show'
```
rshipp, line 114: i'm shoving LTI stuff in the course_resources controller again. :/

this one's different from lesson_contents and lti_assignment_selection in that course_resources has views that are available outside of the LTI flow. We should revisit renaming lesson_contents and/or lti_assignment_selection when we update the lesson/project/survey model stuff.

db/migrate/20200914175834_rename_program_refs_to_base_course.rb
```diff
@@ -0,0 +1,12 @@
+class RenameProgramRefsToBaseCourse < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :grade_categories, :program_id, :base_course_id
+
+    remove_index :logistics, name: :index_logistics_on_day_of_week_and_time_of_day_and_program_id
+    rename_column :logistics, :program_id, :base_course_id
+    add_index :logistics, [:day_of_week, :time_of_day, :base_course_id], unique: true, name: :index_logistics_on_day_time_course
```
rshipp, line 7: normally rails handles indexes for us, but since `base_course` is slightly more characters than `program`, the new index name was over the character limit. so I had to manually remove it and add it back with a shorter name.

lib/sync_to_lms.rb
```diff
@@ -215,7 +215,7 @@ def get_section(course_id, section_name)
   # Loads the program info on the first call and caches it for future calls.
   def get_program(program_id)
     unless @programs[program_id]
-      p = Program.new
```
rshipp, line 218: note: I did not test sync_to_lms at all, I'm hoping our unit tests catch anything that would break here.

spec/factories/grade_category.rb
```diff
@@ -3,6 +3,6 @@
     sequence(:name) { |i| "Grade Category Name #{i}" }
     percent_of_grade { 0.25 }
 
-    program
+    association :base_course, factory: :course
```
rshipp, line 6: evidently the `program` thing is shorthand for something like `association :program, factory: :program`, so I can't use `course` here because it needs to set `base_course_id`, not `course_id`.

rshipp:
Forgot to add CourseResource tests, working on that now...

spec/factories/grade_category.rb
sadleb: Technically, I think this should use the `base_course` factory since a grade category applies to both course and course template. But I wouldn't go down that rabbit hole until we start working on something that actually uses grade category.

sadleb (changes requested):
To answer your Q in the PR about "does gradecategory/logistics/section" reference the right things? --> perhaps not in some spots, but I think we should fix that up when actually working on functionality that uses those instead of in this PR.

app/controllers/base_courses_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    @base_course.destroy
+    respond_to do |format|
+      format.html { redirect_to base_courses_url, notice: "#{set_type.humanize} was successfully deleted." }
+      format.json { head :no_content }
+    end
+  end
+
+  private
+
+  def set_type
```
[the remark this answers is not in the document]
rshipp: I just copied from this https://youtu.be/2fH_V91b0H4?t=416 (which has some more context for why it's there), idk if it's a convention or just something they made up on the spot. I do think it's strange that it's not setting anything, but I'm not sure what else to call it. The params[:type] input is set in `config/routes.rb` where you pass in the `type` param as a string to the controller, so it's not really related to a class name.

app/controllers/course_resources_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def create
+    @course_resource = CourseResource.create!(name: create_params[:name], course_resource_zipfile: create_params[:course_resource_zipfile])
+    redirect_to base_courses_path, notice: 'Course resource was successfully created.'
+  end
+
+  def lti_show
+    canvas_course_id = @lti_launch.request_message.custom['course_id']
+    base_course = Course.find_by(canvas_course_id: canvas_course_id)
+    if base_course&.course_resource
```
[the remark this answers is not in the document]
rshipp: hm, I hadn't seen that. the only comparison i see is [this](https://rubyinrails.com/2017/11/17/safe-navigation-operator-ampersand-dot-in-ruby/#try-vs-safe-navigation). my initial reaction here is that it's more readable to use a core language feature than a Rails method.

app/models/user.rb
```diff
@@ -62,36 +62,36 @@ def confirmed?
     !!confirmed_at
   end
 
-  def start_membership(program_id, role_id)
-    find_membership(program_id, role_id) ||
-      program_memberships.create(program_id: program_id, role_id: role_id, start_date: Date.today)
+  def start_membership(base_course_id, role_id)
```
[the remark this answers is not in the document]
rshipp: For things that expect specifically a `course`, I'd agree we should use `course` over `base_course`. But there's no such thing as a `course_id`, to me it seems more confusing to have `course_id` and `course_template_id`s than explicitly saying `base_course_id`. Maybe because I don't trust the Rails magic stuff and want a reminder of how this is reflected at the database level? In this specific case, if we renamed that variable, we'd have to pass in `base_course_id: course_id` to `create!`, which also seems confusing.

lib/sync_to_lms.rb
```diff
@@ -215,7 +215,7 @@ def get_section(course_id, section_name)
...
+      p = BaseCourse.new
```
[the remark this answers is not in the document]
rshipp: Yeah I assumed all the Salesforce stuff would stay - in our db diagram we had a `salesforce_program_id` which is totally separate from the now-removed `program_id`.

spec/factories/grade_category.rb
rshipp: We actually can't use the :base_course factory, since it would fail the `type` validation. That factory is only used in `models/base_course_spec.rb`, where it's called by `build`, not `create`, and so never uses the validations.

app/models/base_course.rb
```diff
@@ -1,31 +1,46 @@
...
-  belongs_to :organization
-  
-  has_many :program_memberships
-  has_many :users, through: :program_memberships
-  has_many :roles, through: :program_memberships
+class BaseCourse < ApplicationRecord
+  belongs_to :course_resource, optional: true
 
-  # TODO: putting these "virtual" attributes here for now. In the next iteration, we'll want to 
+  # from course
```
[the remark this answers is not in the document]
rshipp: yeah, this was a "note to self" comment that i forgot to remove

app/models/base_course.rb
```diff
@@ -1,31 +1,46 @@
-class Program < ApplicationRecord
-  belongs_to :organization
-  
-  has_many :program_memberships
-  has_many :users, through: :program_memberships
-  has_many :roles, through: :program_memberships
+class BaseCourse < ApplicationRecord
+  belongs_to :course_resource, optional: true
```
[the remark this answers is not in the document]
rshipp: I keep trying to understand what the difference is, but from what I can tell they aren't interchangeable? The way I understood it "the thing with the fkey reference gets a belongs_to", and in this case we have a course_resource_id in base_courses. (I've never seen a case where `belongs_to` actually makes sense as the name for this method, I wish it was called something else.)

spec/factories/grade_category.rb
rshipp: To put this another way bc I think this isn't very clear - factories are for making instances of objects, and `BaseCourse` should never have instances (it's basically an "abstract" class, if Ruby had those), so we must always choose a type if we want to test with an instance. If we need to test both, we should do both.

app/controllers/course_resources_controller.rb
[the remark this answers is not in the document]
sadleb: word. i have yet to use `try` since that thread so I'm down for trying to stick with the ruby operator

app/controllers/base_courses_controller.rb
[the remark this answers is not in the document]
sadleb: hmm, ok. maybe add a link to that video as a comment? i struggled to make heads from tails without having a deep understanding and experience with Single Table Inheritance and I'm sure future readers (including ourselves) will stumble on this again and be similarly confused.

app/models/base_course.rb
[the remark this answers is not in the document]
sadleb: Just went and re-read the stuff. This is the correct association. The name just throws me.

app/models/user.rb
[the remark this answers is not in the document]
sadleb: Word. I'm down with sticking with `base_course_id` --> I'll just have to get used to having the keep the Rails magic stuff in mind instead of trying to make things as intuitive and specific as possible.

sadleb (approved):
Assuming the specs pass after that last commit, lgtm now.

PR #344, opened 2020-09-14
Add Join User Creation

skaex:
### Description
This adds the functionality that create a user on the join server. 

### Motivation and Context
There are still some features of the old join server that we need to get working e.g Braven Network.

https://app.asana.com/0/1172786304330247/1189555546176041
https://app.asana.com/0/1172786304330247/1189555546176042

### Types of changes

### Checklist:


app/services/setup_portal_account.rb
```diff
@@ -12,8 +12,10 @@ def initialize(salesforce_contact_id:)
 
   def run
     find_or_create_portal_user!
+    join_user = find_or_create_join_user!
```
sadleb, line 15: Would it make more sense to have `UpdateJoinUsers` take a collection of users and then pass in the single user here and the collection from the maintenance task?

sadleb (changes requested):

app/services/update_join_users.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+  end
+
+  private
+
+  def find_or_create_join_user!(email:, first_name:, last_name:)
+    join_user = join_api_client.find_user_by(email: email)
+    return join_user unless join_user.nil?
+
+    join_api_client.create_user(email: email, first_name: first_name,
```
sadleb, line 22: If this fails will it be easy to troubleshoot and fix? E.g. if the Join server is in the middle of a restart?

app/services/setup_portal_account.rb
sadleb: Also, for Highlander we won't want to create Join users. So if the ENV var isn't set we should skip the Join user creation logic both here and in the sync.

app/services/update_join_users.rb
skaex: There's a rake that the will run periodically to cater for this edge case.

config/secrets.yml
```diff
@@ -32,6 +32,9 @@ development:
   smtp_port: <%= ENV.fetch('SMTP_PORT') { 587 } %>
   smtp_username: <%= ENV['SMTP_USERNAME'] %>
   smtp_password: <%= ENV['SMTP_PASSWORD'] %>
+  join_base_url: <%= ENV.fetch('JOIN_BASE_URL') { 'http://joinweb:3001' } %>
+  join_api_token: <%= ENV.fetch('JOIN_API_TOKEN') { 'some_random_key' } %>
+  create_join_user_on_sign_up: <%= ENV.fetch('CREATE_JOIN_USER_ON_SIGN_UP') { true } %>
```
sadleb, line 37: I don't think `CREATE_JOIN_USER_ON_SIGN_UP ` should be in the secrets file. We have other ENV vars that control behavior but aren't "secret" and the way we document their existence by adding them to `.env.example`. If you do that, then the code to check whether or not to run the "create on join" behavior can just check for `ENV["CREATE_JOIN_USER_ON_SIGN_UP"].present?` which I think it cleaner than looking for "true"

app/services/setup_portal_account.rb
```diff
@@ -33,13 +39,20 @@ def send_confirmation_notification(user)
...
-    user.update!(canvas_id: portal_user.id)
-    sf_client.update_contact(sf_contact_id, canvas_id: portal_user.id)
+  def update_user_references!(user, salesforce_id:, join_user_id:)
+    user.update!(canvas_id: portal_user.id, join_user_id: join_user_id)
+    sf_client.update_contact(salesforce_id, canvas_id: portal_user.id)
     user
   end
 
+  def find_or_create_join_user!(user)
+    if Rails.application.secrets.create_join_user_on_sign_up.eql?(true)
```
sadleb, line 49: This is where we can switch to `ENV["CREATE_JOIN_USER_ON_SIGN_UP"].present?` if we don't rely on `secrets`. You'd have to explicitly set this ENV var in your specs where you want to test this behavior.

spec/services/setup_portal_account_spec.rb
```diff
@@ -4,24 +4,27 @@
...
+      allow(JoinAPI).to receive(:client).and_return(join_api_client)
     end
 
     it 'creates a canvas user when it does not exist' do
       enrolled_participant = SalesforceAPI::SFParticipant.new(nil, nil, nil, nil, nil, nil, SalesforceAPI::ENROLLED)
       allow(sf_client).to receive(:find_participant).and_return(enrolled_participant)
       allow(canvas_client).to receive(:find_user_by).and_return(nil)
 
-      SetupPortalAccount.new(salesforce_contact_id: nil).run
+      described_class.new(salesforce_contact_id: nil).run
```
sadleb, line 27: We don't have an explicitly stated "style-guide" for the use of `described_class` Other than `keypair_spec.rb` which is a copy/paste from another codebase, there is only one spot that uses it, so I think the existing style of our specs are more in line with using `SetupPortalAccount` vs `described_class`. 

I think if you want to introduce `described_class` as the new style we should do it in it's own PR where we go cut everything over to use it instead.

I personally think `described_class` is less readable than being explicit, but that's a discussion outside the scope of this PR.

sadleb (changes requested):

app/services/setup_portal_account.rb
```diff
@@ -33,13 +39,20 @@ def send_confirmation_notification(user)
...
+      UpdateJoinUsers.new.run([user]).first
+    else
+      JoinAPI::JoinUser.new
```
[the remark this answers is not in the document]
skaex: It wouldn't make sense to return `nil` object because the caller is expecting a `JoinAPI::JoinUser.new` type object. We certainly can move the check upward to avoid it entirely.

spec/services/setup_portal_account_spec.rb
skaex: It is better to use `described_class` in my opinion. For example, it becomes easier to refactor. However, I'll revert till we have a general consensus on this. PS: My linter will be screaming at me: https://rubocop.readthedocs.io/projects/rspec/en/read-the-docs/cops_rspec/#rspecdescribedclass

sadleb (approved):
I don't see a spec to test that if `CREATE_JOIN_USER_ON_SIGN_UP` isn't set, the `expect(join_api_client).not_to have_received(:create_user)` passes, but I assume you tested it out? If it's not too much trouble, I think you should add that spec. Either way, PR looks good enough to merge.

PR #346, opened 2020-09-18
Add sync to join

skaex:
### Description
Allow for running sync to join

### Motivation and Context
We need to have all participants to have join users

### Types of changes

### Checklist:

### DIFF
--- a/app/views/admin/join/init_sync_to_join.html.erb
+++ b/app/views/admin/join/init_sync_to_join.html.erb
@@ -9,7 +9,7 @@
 <div class="container">
   <div class="row">
     <div class="col-sm">
-      <legend>Sync to Join</legend>
+      <legend>Sync to Join (Sync Program Participants to Join Server)</legend>
       <%= render 'sync_to_join_form' %>
     </div>
     <div class="col-sm"></div>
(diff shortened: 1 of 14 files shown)


app/views/admin/join/init_sync_to_join.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<div class="container">
+  <div class="row">
+    <div class="col-sm">
+      <legend>Sync to Join</legend>
+      <%= render 'sync_to_join_form' %>
+    </div>
+    <div class="col-sm"></div>
+  </div>
+</div>
+
```
sadleb, line 18: Can you just add some text here giving context about what "sync to join" is and why you would use it?

sadleb (approved):
Approving this without looking into too much detail b/c this is an admin tool with a very limited use case and impact, but please do add some context on the admin page about what that button should be used for before merging.

app/views/admin/join/init_sync_to_join.html.erb
skaex: I've added some explanation on the view. I add more here:
The reason we need the sync to join is to have participants for a program all have join user accounts. This is important for backfilling participants who have already signed up.

PR #349, opened 2020-09-22
PROTOTYPE - Waivers

sadleb:
First pass at a FormAssembly driven waivers form.
    
DO NOT MERGE. This is hacked up as a proof of concept. If you insert
the "waivers" when selecting an LTI Assignment in Canvas, when you launch
this it will demonstrate the flow as a hard-coded test participant (and hardcoded test form).

Note: the crazy icons after you hit Sign and Submit depends on this fix:
https://github.com/beyond-z/canvas-lms-js-css/pull/320

Note: none of the actual submit part of the flow is implemented. In the `create`
action we need to use the LtiAdvantageAPI to send a "score" back to Canvas for this
assignment (even through it doesn't count towards your grade).
    
### Test Plan:
    - Create an assignment in Canvas, choose External Tool as the submission type.
      Select "Insert Waivers".
    - Login as a test student and open the assignment. It should allow you to sign and submit and direct
      you back to Brians Playground course in Canvas at the end.

### Checklist:

### DIFF
--- a/app/views/layouts/lti_placement.html.erb
+++ b/app/views/layouts/lti_placement.html.erb
@@ -1,9 +1,5 @@
 <% content_for :title, 'LTI Placement' %>
 
-<% content_for :headline do %>
-  <!-- LTI Placement -->
-<% end %>
-
 <% content_for :head do %>
     <%= portal_stylesheet_link_tag %>
     <%= stylesheet_link_tag 'lti_placement', media: 'all', 'data-turbolinks-track': 'reload' %>
--- a/app/views/layouts/lti_placement.html.erb
+++ b/app/views/layouts/lti_placement.html.erb
@@ -1,5 +1,9 @@
 <% content_for :title, 'LTI Placement' %>
 
+<% content_for :headline do %>
+  <!-- LTI Placement -->
+<% end %>
+
 <% content_for :head do %>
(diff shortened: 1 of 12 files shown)


app/views/layouts/lti_placement.html.erb
```diff
@@ -1,5 +1,9 @@
 <% content_for :title, 'LTI Placement' %>
 
+<% content_for :headline do %>
+  <!-- LTI Placement -->
+<% end %>
```
sadleb, line 5: @rshipp just pinging you about this. You may have noticed that "Platform" shows up at the top of all our LTI launched pages after adding in the `:headline` to the application layout. I did this in this waivers branch to hide that. Do you have a better solution? If not, I think i'll pull this into a PR of it's own so that folks don't see this in production when exploring the prototype.

Note: I tried `<% content_for :headline, '' %>` and variations of using an empty string to no avail. Even dug into the Rails source code and can't see why it wasn't working so I just punted and put an actual block of markup that wouldn't show anything to the end user.

rshipp: yeah that's fine.

rshipp: Actually, this goes away with https://github.com/bebraven/platform/pull/353

PR #350, opened 2020-09-23
Add authorization/role framework

rshipp:
### Description

Task: https://app.asana.com/0/1174274412967132/1187445581799827/f

Summary: Add [Pundit](https://github.com/varvet/pundit) authorization library and [Rolify](https://github.com/RolifyCommunity/rolify) role library.

Also removed a large amount of old code, so that we wouldn't have to maintain those controllers/specs and write new policies for them.

The following have been removed:

* industries
* interests
* locations
* majors
* postal_codes
* validations (controller, data_validator lib, and rake task)
* address (model only)
* email (model only)
* phone (model only)
* course_memberships
* user.admin (just the admin column)

The following have been moved outside of the `admin` namespace, which was also removed:

* users_controller
* join_controller
* salesforce_controller

See the PR comments below for more detailed changes.

### Motivation and Context

We don't want to allow all pages on Platform to be viewed by all users. Most pages are admin-only, but some allow all logged in users, or even anonymous users. We also need tight control over the xAPI proxy data, so students can't see or alter other students' answers.

### Types of changes

### Checklist:


app/controllers/application_controller.rb
```diff
@@ -5,24 +5,24 @@
...
   before_action :authenticate_user!
-  before_action :ensure_admin!
   after_action :remove_x_frame_options
 
+  # This callback is a development helper that complains if an action has not explicitly
+  # called `authorize`. It is *not* a fallback mechanism, and should not be relied upon
+  # for added security. Its only purpose is to remind us to call `authorize` in each action.
+  # The `unless` param excludes CasSessionsController from this check, since we have no
+  # reason to call authorize in devise or CAS controller acitons.
+  after_action :verify_authorized, unless: :devise_controller?
```
rshipp, line 18: Note: This will cause a `Pundit::AuthorizationNotPerformedError` on any controller action where `authorize` is not called. If you see that error, it's because we haven't added authorization there, and we should.

app/controllers/base_courses_controller.rb
```diff
@@ -1,25 +1,31 @@
 class BaseCoursesController < ApplicationController
+  layout 'admin'
```
rshipp, line 2: note course management uses the admin layout now.

app/controllers/cas_controller.rb
```diff
@@ -6,7 +6,9 @@
 class CasController < ApplicationController
   layout 'accounts'
   skip_before_action :authenticate_user!
-  skip_before_action :ensure_admin!
+  # Skip the verify_authorized callback for this controller, since we won't be logged in
+  # and thus get no benefit from authorization.
+  skip_after_action :verify_authorized
```
rshipp, line 11: It doesn't make sense to have authorization in controllers where we're doing authentication. We exclude CasController from the verify_authorized check because we don't want to use `authorize` here. Another way to do this would be to write a `CasPolicy` and just return `true` from everything. I'm doing it this way instead because we skip the checks in the Devise controllers, and this seemed more consistent.

app/controllers/course_contents_controller.rb
```diff
@@ -1,33 +1,38 @@
 class CourseContentsController < ApplicationController
   layout 'content_editor'
-  before_action :set_course_content, only: [:show, :edit, :update, :destroy, :publish]
+  before_action :set_model_instance, only: [:show, :edit, :update, :destroy, :publish]
```
rshipp, line 3: refactored this to not duplicate code from `dry_crud`.

app/controllers/home_controller.rb
```diff
@@ -2,5 +2,6 @@ class HomeController < ApplicationController
   layout 'admin'
 
   def welcome
+    authorize :application, :index?
```
rshipp, line 5: we could also define a `HomePolicy` with a `welcome?` method. But this seems just as good, and less code.

app/controllers/join_controller.rb
```diff
@@ -1,12 +1,16 @@
 # frozen_string_literal: true
 
 # Join controller
-class Admin::JoinController < ApplicationController
+class JoinController < ApplicationController
   layout 'admin'
 
-  def init_sync_to_join; end
+  def init_sync_to_join
+    authorize :application, :index?
```
rshipp, line 8: again, we could have another policy, but this is shorter. push back here if you think it's clearer to define individual policies for these, I don't have a strong opinion.

app/controllers/lrs_xapi_proxy_controller.rb
```diff
@@ -15,6 +15,14 @@ def xAPI
     # submission for a Student/Fellow. The current_user would be the TA but we really
     # want to be querying the LRS as the Student.
     user = params[:user_override_id] ? User.find(params[:user_override_id]) : current_user
+
+    # Only allow if `current_user` should have access to `user`'s data.
+    if request.method == 'GET'
+      authorize user, :xAPI_read?, policy_class: LrsXapiProxyPolicy
+    else
+      authorize user, :xAPI_write?, policy_class: LrsXapiProxyPolicy
+    end
```
rshipp, line 24: This action is a little different from the rest, but this seems ok? I'll do a little testing here to make sure this actually works. Would appreciate more eyes here too.

app/controllers/users_controller.rb
```diff
@@ -1,17 +1,88 @@
+# frozen_string_literal: true
```
rshipp, line 1: most of this file is just moved from admin/users_controller.rb

app/models/course.rb
```diff
@@ -1,5 +1,2 @@
 class Course < BaseCourse
-  has_many :course_memberships
-  has_many :users, through: :course_memberships
-  has_many :roles, through: :course_memberships
```
rshipp, line 4: removed because course_memberships is gone. in the future we'd use `course.sections` and `section.users` to find all the users in all the sections in a given course.

app/models/role.rb
```diff
@@ -1,11 +1,14 @@
...
   
-  def to_show
-    attributes.slice('name')
-  end
+
+  validates :resource_type,
+            :inclusion => { :in => Rolify.resource_types },
+            :allow_nil => true
+
+  scopify
```
rshipp, line 13: this is just replacing our manual role class with the one Rolify generates

app/models/section.rb
```diff
@@ -1,4 +1,5 @@
 class Section < ApplicationRecord
+  resourcify
```
rshipp, line 2: this `resourcify` is what lets us do things like `user.add_role :student, section`

app/models/section.rb
```diff
@@ -7,18 +8,14 @@ class Section < ApplicationRecord
...
-      merge(UserSection.assists)
+  # All users, with any role, in this section.
+  # This is a function just because I don't know how to write it as an association.
+  def users
+    all_users = []
+    roles.distinct.map { |r| r.name }.each do |role_name|
+      all_users += User.with_role(role_name, self)
     end
+    all_users.uniq
   end
```
rshipp, line 19: This is the association stuff I brought up in that one meeting.

app/models/user.rb
```diff
@@ -28,27 +27,7 @@ class User < ApplicationRecord
...
-
-    def as_lc
-      merge(UserSection.facillitates)
-    end
 
-    def as_ta
-      merge(UserSection.assists)
-    end
-  end
-
```
rshipp, line 49: all gone, replaced by Rolify functions

app/models/user.rb
```diff
@@ -57,37 +36,32 @@ def full_name
...
+  # All sections where this user has any role.
+  # This is a function just because I don't know how to write it as an association.
+  def sections
+    Section.with_roles(roles.distinct.map { |r| r.name }, self).distinct
+  end
+
+  # All sections with a specific role.
+  def sections_with_role(role_name)
+    Section.with_roles(role_name, self).distinct
+  end
```
rshipp, line 48: More not-association stuff. I added `sections_with_role` just because it's much more efficient and this is called frequently.

app/models/user.rb
```diff
@@ -57,37 +36,32 @@ def full_name
...
   def confirmed?
     !!confirmed_at
   end
 
-  def start_membership(base_course_id, role_id)
-    find_membership(base_course_id, role_id) ||
-      course_memberships.create(base_course_id: base_course_id, role_id: role_id, start_date: Date.today)
+  def admin?
+    has_role? :admin
   end
```
rshipp, line 57: A shortcut, just because we use this often in policies.

app/models/user.rb
```diff
@@ -57,37 +36,32 @@ def full_name
...
-      course_membership.update! end_date: Date.yesterday
-    else
-      return false
+
+  # True if this user is a TA in the same section where target_user is a student.
+  def ta_for?(target_user)
+    sections_with_role(:ta).each do |section|
+      return true if target_user.has_role? :student, section
     end
-  end
```
rshipp, line 76: Used in the policy. Relatively optimized for the queries this runs.

app/policies/application_policy.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      scope.all
+    end
+  end
```
rshipp, line 54: we don't use scopes anywhere yet, this is just the default autogenerated code. see pundit docs for more on scopes.

app/views/users/edit.html.erb
```diff
@@ -47,10 +45,18 @@
...
+        <div class="row">
+          <div class="form-group col">
+            <p>Global roles</p>
+            <%= f.collection_check_boxes :role_ids, Role.global, :id, :name do |b| %>
+              <div class="form-check">
+                <%= b.check_box(class: "form-check-input") %>
+                <%= b.label(class: "form-check-label") %>
+              </div>
+            <% end %>
+          </div>
```
rshipp, line 57: note `Role.global` - only shows global roles, not roles on individual sections.

app/views/users/show.html.erb
```diff
@@ -1,37 +1,28 @@
...
-  <%= link_to 'Back', postal_codes_path, class: 'btn btn-sm btn-outline-secondary' %>
-  <%= link_to 'JSON', user_path(@user, format: 'json'), class: 'btn btn-sm btn-secondary', target: '_blank' %>
-<% end %>
-
-<div class="row">
-  <div class="col-sm-4">
-    <div class="card">
-      <div class="card-header">
-        <span data-feather="shield"></span>
-        Course Memberships
```
rshipp, line 15: this is gone because memberships is gone. we could add something similar back later if we wanted.

db/migrate/20200923143315_remove_admin_from_user.rb
```diff
@@ -0,0 +1,10 @@
+class RemoveAdminFromUser < ActiveRecord::Migration[6.0]
+  def change
+    # Migrate all current admins to new role system.
+    User.where(admin: true).each do |user|
+      user.add_role :admin
+    end
```
rshipp, line 6: actually migrates admin users to the new role system, so we don't have to do it manually

lib/dry_crud.rb
```diff
@@ -5,14 +5,14 @@ module DryCrud
...
-    # DRY's up the index, show, edit, update, destroy, new, create and publish methods
+    # DRY's up the index, show, edit, update, destroy, new, and create methods
     # by setting the appropriate instance variables for the corresponding model
     #
     # NOTE: you can alway override this behavior by defining whatever you want in the 
     # associated controller's index, show, edit, etc methods.
     included do
       before_action :set_models_instance, only: [:index]
-      before_action :set_model_instance, only: [:show, :edit, :update, :destroy, :publish]
+      before_action :set_model_instance, only: [:show, :edit, :update, :destroy]
```
rshipp, line 15: I removed `publish` because it's not part of the standard CRUD, it was just in course contents.

spec/controllers/cas/login_join_spec.rb
```diff
@@ -99,13 +99,3 @@
...
-
-def fill_and_submit_login(username, password)
-  # Input login credentials
-  fill_in "username", :with => username
-  fill_in "password", :with => password
-  
-  # Make sure there is a login button that can be clicked
-  # Capybara with Selenium can have problems with javascript
-  find("#login-submit").click
-end
```
rshipp, line 111: This was causing some weirdness in the tests, now all the tests use `fill_and_submit_login` from `platform_helper` instead.

spec/controllers/cas/proxy_validate_join_spec.rb
```diff
@@ -1,6 +1,5 @@
 require "rails_helper"
 require "capybara_helper"
-require "platform_helper"
```
rshipp, line 3: platform_helper is included in rails_helper, so we don't need to include it in individual test files

spec/controllers/cas/validate_spec.rb
```diff
@@ -1,16 +1,20 @@
...
-RSpec.describe CasController, type: :routing do
+RSpec.describe CasController, type: :controller do
+  render_views
+
   describe "RubyCAS routing" do
     let!(:valid_user) { create(:registered_user) }
     let(:valid_user_creds) {{ email: valid_user.email, password: valid_user.password }}
-    let(:return_service) { 'http://braven/' }
+    # Must be a page a non-admin user can access. The host is *ignored*, even if explicitly set.
+    let(:return_service) { 'http://braven/cas/login' }
```
rshipp, line 17: took me way too long to figure out

spec/controllers/honeycomb_js_controller_spec.rb
```diff
@@ -4,7 +4,7 @@
   let(:trace_id) { 'example-trace-id' }
   let(:serialized_trace) { "1;dataset=example-dataset,trace_id=#{trace_id},parent_id=example-parent-id,context=e30=" }
 
-  let(:user) { create :registered_user, admin: true } # TODO: bug where you have to be an admin. Remove admin once that's fixed.
```
rshipp, line 7: hooray for getting rid of all of these TODOs finally

spec/controllers/users_controller_spec.rb
```diff
@@ -54,22 +54,4 @@
...
-      end
-
-      it "allows access token via headers" do
-        request.headers.merge!('Access-Key' => access_token.key)
-        
-        get :index, params: {}, session: valid_session, format: :json
-        expect(response).to be_successful
-      end
-    end
-  end
```
rshipp, line 74: this was for the old non-admin users controller. note we don't actually have specs for the users controller.

spec/factories/users.rb
```diff
@@ -16,18 +16,28 @@
...
         canvas_id { '1234' }
+        after :create do |user|
+          user.add_role :student
+        end
+      end
+
+      factory :ta_user do
+        after :create do |user|
+          user.add_role :ta
+        end
```
rshipp, line 34: these would eventually be replaced with something that sets e.g. `user.add_role :student, section`, since `:student` and `:ta` aren't especially meaningful as global roles.

spec/feature/course_contents/smoke_spec.rb
```diff
@@ -21,22 +19,23 @@
           # Must ignore the Capybara host IFF we are running tests that have browser AJAX requests to that host.
           c.ignore_hosts Capybara.server_host
         end
-        visit "/cas/login?service=#{url_encode(return_service)}"
+        #visit "/cas/login?service=#{url_encode(return_service)}"
+        visit return_service
```
rshipp, line 23: See https://app.asana.com/0/1174274412967132/1196188630628489/f

spec/policies/lrs_xapi_proxy_policy_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    it "disallows TA to read data for a user in another section" do
+      user1 = create(:registered_user)
+      user2 = create(:registered_user)
+      section1 = create(:section)
+      section2 = create(:section)
+      user1.add_role :ta, section1
+      user2.add_role :student, section2
+      expect(subject).not_to permit(user1, user2)
+    end
```
rshipp, line 44: 🙃

app/controllers/application_controller.rb
jessouyangbraven: Add to `README`?

app/controllers/join_controller.rb
jessouyangbraven: I could go either way on this. 

I like that there's less code/fewer policies to maintain in this implementation.

If I were to push for defining `JoinPolicy` (and `HomePolicy`) separately, it would be so that you don't have to worry about other controllers when you're modifying your controller's policy. 

I think I'm good with this as long as we remember we can refactor going forward if we end up needing different authentication for these controllers (instead of making `ApplicationPolicy` complex to support many controllers).

app/controllers/lrs_xapi_proxy_controller.rb
jessouyangbraven: This looks correct to me. 

Random question because I just noticed at this point: is there a reason why we don't use parentheses when we call `authorize`? Is this a `pundit` convention?

app/models/role.rb
jessouyangbraven: Thank you for explaining. This was very mysterious to me in the first pass.

jessouyangbraven (approved):
Thank you so much for calling out things inline with comments! 

This looks good to me and everything in my comments is something I think we can do as needed going forward.

app/controllers/lrs_xapi_proxy_controller.rb
rshipp: It is how they do it in the Pundit readme. I'm not sure what the rails conventions are for when to use parens in function calls. I try to only use them if it makes it clearer, like if I'm method chaining or something... but I don't always remember to be consistent there.

spec/factories/users.rb
rshipp: See also https://app.asana.com/0/1174274412967132/1196321172031701/f

README.md
```diff
@@ -356,6 +360,62 @@ that matches how the data-central project was structured. See [USAGE here](lib/g
...
+the class itself:
+
+    autorize NewThing
+
+This will call `NewThingPolicy.index?` with the `NewThing` class as the scoped record. For more information,
+check the Pundit README and API docs.
+
+For an example policy for anonymous (no login) access, see `KeypairPolicy`. For an example controller
+with no authorization, see `CasController`. For an example admin-only policy, see `BaseCoursePolicy`.
+For an example policy allowing any logged-in user, see `HoneycombJsPolicy`.
```
sadleb, line 390: Super helpful

README.md
```diff
@@ -356,6 +360,62 @@ that matches how the data-central project was structured. See [USAGE here](lib/g
...
+
+There is a global `after_action` callback in `ApplicationPolicy`, `verify_authorized` thats only purpose is
```
sadleb, line 392: Smart. I wonder if eventually we want to write a module that uses meta-programming to default everything to just the current instance or class unless you add your own authorize similar to what `DryCrud` does. I'm not sure if that would be better or not since we do want to force explicit authorization decisions, but this sure is a lot of repetitive code.

README.md
```diff
@@ -356,6 +360,62 @@ that matches how the data-central project was structured. See [USAGE here](lib/g
...
+    section = Section.new
+    user = User.new
+    user.add_role :student, section
+
+You must instead use:
+
+    # Good - do this instead!
+    section = create(:section)
+    user = create(:registered_user)
+    user.add_role :student, section
```
sadleb, line 415: Could we override `add_role` only in the test env and detect this?

app/controllers/course_contents_controller.rb
```diff
@@ -1,33 +1,38 @@
...
   # GET /course_contents/1
   # GET /course_contents/1.json
   def show
+    authorize @course_content
   end
 
   # GET /course_contents/new
   def new
-    @course_contents = CourseContent.all
     @course_content = CourseContent.new
```
sadleb, line 20: Do you get this one for free from `DryCrud` and we just need to do `@course_contents` explicitly?

app/controllers/course_contents_controller.rb
```diff
@@ -1,33 +1,38 @@
...
+    authorize @course_content
+    @course_contents = CourseContent.all
```
sadleb, line 22: Doesn't `set_course_contents` already do this?

app/controllers/course_contents_controller.rb
```diff
@@ -1,33 +1,38 @@
...
   end
 
   # GET /course_contents/1/edit
   def edit
+    authorize @course_content
     @course_contents = CourseContent.all
```
sadleb, line 28: Doesn't `set_course_contents` already do this?

app/controllers/course_contents_controller.rb
```diff
@@ -87,10 +95,6 @@ def publish
   end
 
   private
-    def set_course_content
-      @course_content = CourseContent.find(params[:id])
-    end
-
     def set_course_contents
       @course_contents = CourseContent.all
```
sadleb, line 99: Outside the scope of this PR, but if we eventually end up with 100's or 1000's of course contents, will this load all the HTML blobs into memory?

app/controllers/join_controller.rb
sadleb: Ditto.

sadleb, line 8: I think this can end up being called from a cronjob. Will this work from a cronjob since there is no user? The default policy is to only allow admins, so do we need to do something for cronjob related actions to bypass needing a user (or somehow execute them in the context of a user)? This comment applies to a handful of places, so if something has to change here and you need help identifying the cronjob accessible stuff lmk.

app/controllers/linked_in_authorization_controller.rb
```diff
@@ -13,11 +13,14 @@ def dry_crud_enabled?() false end
 
   # iframe-able endpoint that renders a static LinkedIn button
   def login
+    authorize :LinkedInAuthorization
```
sadleb, line 16: I assume this is some Pundit magic that I would understand if I read the docs? E.g. `:application` is for ApplicationController with no model and `:LinkedInAuthorization` is for this controller with no model?

app/policies/project_submission_policy.rb
```diff
@@ -0,0 +1,5 @@
+class ProjectSubmissionPolicy < ApplicationPolicy
+  def create?
+    user&.has_role? :student
```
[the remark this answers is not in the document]
sadleb: I lean towards `!!user` for now. I don't see a reason to make testing and dev harder by making sure someone is a student vs letting TAs and admin's submit projects even if that doesn't really make sense in a course.

app/models/course.rb
sadleb: I'm definitely going to have to revisit this to fully understand, but I'll leave that for when we get back to something that uses the concept of a course_membership, like whether the waivers have been signed by that user for that course.

README.md
```diff
@@ -356,6 +360,62 @@ that matches how the data-central project was structured. See [USAGE here](lib/g
...
+
+    autorize NewThing
+
+This will call `NewThingPolicy.index?` with the `NewThing` class as the scoped record. For more information,
+check the Pundit README and API docs.
+
+For an example policy for anonymous (no login) access, see `KeypairPolicy`. For an example controller
+with no authorization, see `CasController`. For an example admin-only policy, see `BaseCoursePolicy`.
+For an example policy allowing any logged-in user, see `HoneycombJsPolicy`.
+
```
sadleb, line 391: Can you also add a blurb about how we `resourcify` the section and why? Or point out some other place in the code where more context about that is given?

app/models/section.rb
sadleb: I'm good with this for now. We can revisit in the future if this ends up slow, hard to maintain / modify, etc.

spec/controllers/cas/validate_spec.rb
sadleb: oh wow, ok. good to know in the future.

sadleb (approved):
I don't see anything in here that would make me say we shouldn't merge until it's change, but I did leave some comments and questions. The main one I'm concerned about is stuff that runs from a cronjob. Will this stuff work when run in that context or do we need to identify cronjob called code and do something special?

README.md
rshipp: I couldn't find an explanation of their reasoning, but the Pundit readme says "Pundit encourages you to add the authorize call manually to each controller action". I actually had some of the controllers set up with `before_action` callbacks at first, but took that out and did everything manually instead. I think manual just forces us to be super intentional with how we're calling `authorize`.

README.md
rshipp: That seems dangerous. This behavior caught me off guard, but it makes sense when I think about how the roles stuff works. The only reason I was using `User.new` to begin with was because that's what the pundit generator puts in the spec files. As long as we base new specs off the existing ones, all of which I've updated to the correct syntax, we shouldn't have any problems.

app/controllers/course_contents_controller.rb
rshipp: Yeah you're right, I wasn't paying much attention to the code here. I'll leave this for another PR since it'll definitely conflict with #359. https://app.asana.com/0/1174274412967132/1196545422806329/f

app/controllers/course_contents_controller.rb
rshipp: Probably, though I have no idea what this implementation looks like under the hood. If we ever got to that point we'd have to start paginating things. I don't think it's worth doing now though.

app/controllers/join_controller.rb
rshipp: This is just a controller action, it's not called anywhere in the code. If rake tasks ever called controller actions we might have to think about this, but I don't think that's likely unless we're doing something really weird in rake tasks.

app/controllers/linked_in_authorization_controller.rb
rshipp: The `authorize` function takes any object as its first param, and tries to figure out what policy it maps to. Generally we pass an individual record or a class, but for controllers with no associated class you can just pass in a symbol of the policy name and it'll find it that way. You could also write this as `authorize :linked_in_authorization`, or `authorize nil, policy_class: LinkedInAuthorizationPolicy`, and it'd do the same thing.

README.md
rshipp: added a note to the readme explaining roles a little. lmk if you want something more

PR #351, opened 2020-09-24
Make Course Resources work for Course Templates.

sadleb:
I added a Course Resource Rise360 package to a course template but it wouldn't show. This PR makes the resources show for templates too.

### Test Plan:
- bundle exec rspec spec/controllers/course_resources_controller_spec.rb

### Checklist:


app/helpers/lti_helper.rb
```diff
@@ -1,5 +1,6 @@
 require 'lti_deep_linking_response_message'
 require 'lti_deep_linking_request_message'
+require 'lrs_xapi_proxy'
```
sadleb, line 3: Things generally work without this if you do something else that loads it. But if you go directly to something that uses lib stuff before using a controller that has loaded this (like run a spec directly) it won't be loaded already.

spec/controllers/course_resources_controller_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'rise360_util'
```
sadleb, line 2: Ditto. If you run all the specs something else will likely have loaded this. But if you run this spec directly, the lib stuff isn't auto-loaded from what I can tell.

rshipp (approved):

PR #356, opened 2020-09-29
Fix open redirect

rshipp:
### Description

Task: https://app.asana.com/0/1174274412967132/1196136462683257/f

### Types of changes

### Checklist:


app/helpers/cas_helper.rb
```diff
@@ -15,4 +24,24 @@ def devise_mapping
...
+      end
+    else
+      # Service is a relative url, so it's safe.
+      return service
+    end
+
+    # If we get here, the specified service is unsafe.
+    # Default to nil, same as if this param was unset.
+    nil
+  end
```
rshipp, line 45: This logic seems overly complex, is there any way to simplify/railsify it?

rshipp:
not ready yet, whoops

app/helpers/cas_helper.rb
```diff
@@ -1,5 +1,20 @@
...
+    /^.*\.?braven\.org$/,
+    /^braven.instructure.com$/,
+    /^braven$/,
+    /^#{ENV['SPEC_HOST']}$/,
+    /^localhost$/,
+    /^127.0.0.1$/,
+    /^platformweb$/,
+    /^boosterplatformweb$/,
+    /^canvasweb$/,
+  ]
```
rshipp, line 16: does this list look correct? is there anything else we want to add now? anything that redirects you to the platform login page must match against this list.

jessouyangbraven: I'm bad at regexes. Does this cover `staging` for booster and platform?

jessouyangbraven:
This looks reasonable to me. Going to let @sadleb review, though, because I don't feel solid on the CAS/security stuff.

app/helpers/cas_helper.rb
rshipp: oh whoops. it does cover staging, but it also allows things like `evilbraven.org`... I'll have to tighten the regex

rshipp: ok I updated the regex and logic. I was also allowing `service=.evil.com`, which would have allowed redirects to e.g. `platform.bebraven.org.evil.com`.

app/helpers/cas_helper.rb
```diff
@@ -15,4 +30,28 @@ def devise_mapping
...
+    if parsed_service&.host
+      ALLOWED_SERVICES.each do |regex|
+        if regex.match parsed_service.host
+          return service
+        end
+      end
+    elsif service&.start_with? '.'
+      # `redirect_to '.evil.com'` is valid in Rails, so we need to hard-exclude
+      # anything that starts with a period.
+      return nil
```
rshipp, line 46: new logic

app/helpers/cas_helper.rb
```diff
@@ -1,5 +1,20 @@
 module CasHelper
 
+  # We don't want to redirect to arbitrary service URLs, since that's a dangerous open redirect.
+  # We check each of these regexes against the host portion of the service URL.
+  ALLOWED_SERVICES = [
+    /^(.+\.)?bebraven\.org$/,
+    /^(.+\.)?braven\.org$/,
```
rshipp, line 7: tighter regex to allow `*.bebraven.org` or `bebraven.org` but not `evilbebraven.org`

app/helpers/cas_helper.rb
```diff
@@ -15,4 +30,28 @@ def devise_mapping
     @devise_mapping ||= Devise.mappings[resource_name]
   end
 
+  # Only allow redirects to services we own.
+  # See ALLOWED_SERVICES above.
+  def safe_service_url(service)
+    parsed_service = Addressable::URI.parse(service)
```
sadleb, line 36: Does this throw if `params['service']` is nil? Is that what we want?

sadleb (changes requested):
I'm just unclear what happens if there is no `params['service']` or if that can happen.

app/controllers/cas_controller.rb
```diff
@@ -374,7 +375,8 @@ def set_settings
   end
 
   def set_params
-    @service = Utils.clean_service_url(params['service']) if params['service']
+    safe_service_param = helpers.safe_service_url(params['service'])
```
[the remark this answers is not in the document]
rshipp: the `if safe_service_param` below serves the same purpose.

app/helpers/cas_helper.rb
rshipp: nope
```
2.2.3 :041 > Addressable::URI.parse(nil)
 => nil
```
This is tested in the specs:
https://github.com/bebraven/platform/pull/356/files#diff-0b15deb88055ebbbc10eef66184751f4R40
If the param is nil, this function returns nil.

rshipp:
just found another issue. i thought only things starting with `.` were appended to the host, but that's not true, you can do it with anything. `service=test` redirects to `platform.bebraven.orgtest`. Rails' redirect_to is... really something.

rshipp:
OK fixed that too

sadleb (approved):
lgtm. Didn't realize the `Addressable::URI.parse(service)` behavior and didn't look closely enough at the specs to see the call with `nil` at the end.

PR #360, opened 2020-09-30
First attempt at a working setup script

a-willis:
A script that will build docker images, run db:create, and bring up the web server
This will be used by the development repository 

Test plan: run ./setup.sh. When done, visit platformweb:3020 in browser


setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/zsh
+
+cp .env.example .env
+cp .env.database.example .env.database
+docker-compose build
+docker-compose up -d platformdb 
+docker-compose up platformweb
```
hcli-work, line 7: how does this even work? wouldn't it just hang here?

hcli-work: If this works and is really necessary, would you mind just putting in a small comment?

hcli-work (changes requested):

setup.sh
a-willis: The db:create command won't work without trying to run platformweb first. I'll add a comment to explain more

hcli-work (approved):
lgtm!

setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/zsh
+
+cp .env.example .env
+cp .env.database.example .env.database
+docker-compose build
+docker-compose up -d platformdb 
+docker-compose up platformweb #this is a hack. it will exit with an error but needed for db:create
+docker-compose run platformweb bundle exec rake db:create db:schema:load db:migrate db:seed
```
sadleb, line 8: The steps in the README don't work? E.g. running
```
docker-compose up -d
docker-compose run platformweb bundle exec rake db:create db:schema:load db:migrate db:seed
```
And then restarting.

PR #361, opened 2020-10-01
Refactor project submission code

jessouyangbraven:
**Task**: https://app.asana.com/0/1174274412967132/1186960110311121

### Description

- Move project submission logic out of `custom_content_versions_controller`
- This finishes building out the data model here: https://dbdiagram.io/d/5f6e41177da1ea736e2f6328
- And it cleans up the tech debt described here: https://github.com/bebraven/platform/wiki/Projects#tech-debt

This PR is meant to make the code less confusing and will also let us build things like viewing multiple project submissions from a single Fellow correctly.

**Project Submissions**

- when a Fellow submits an assignment in Canvas, we create a `ProjectSubmission` in our database now
- this allows us to use `project_submissions#show` to render previous submissions
- when we send Canvas the URL to view the submission, we're using the `project_submissions#show` URL instead of `custom_content_versions#show`

### Test

- Submission URL when creating the Canvas Assignment:
Screen Shot 2020-09-30 at 5 22 32 PM

- Submitting a project as a Fellow:
Screen Shot 2020-09-30 at 5 23 27 PM

- Viewing submission as a TA:
Screen Shot 2020-09-30 at 5 23 34 PM

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #359, opened 2020-09-30
Migrate CourseContents to CustomContents

jessouyangbraven:
**Task**: https://app.asana.com/0/1174274412967132/1186960110311121

### Description

- This is the new data model for project submissions: https://dbdiagram.io/d/5f6e41177da1ea736e2f6328
- This PR specifically migrates `CourseContents` to `CustomContents`
- It depends on PR #358 and the `migrate-coursecontenthistories` branch
- This is a separate PR because this impacts publicly accessible (via Booster) `course_contents/` endpoints. A separate PR/commit to staging/production will make it easier for us to revert this if anything goes wrong
- This PR also migrates the `custom_content_version:course_content_id` column to `:custom_content_id`

### Motivation and Context

See PR #358 for more details. 

### Test

- Public-facing endpoints to `/course_contents/` still work:
Screen Shot 2020-09-29 at 5 11 44 PM
Screen Shot 2020-09-29 at 5 11 45 PM
Screen Shot 2020-09-29 at 5 11 47 PM
Screen Shot 2020-09-29 at 5 11 52 PM

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #358, opened 2020-09-29
Migrate CourseContentHitories to CustomContentVersions

jessouyangbraven:
**Task**: https://app.asana.com/0/1174274412967132/1186960110311121

### Description

- This is the new data model for project submissions: https://dbdiagram.io/d/5f6e41177da1ea736e2f6328
- This PR specifically migrates and renames `CourseContentHistory` to `CustomContentVersion`
- This PR shouldn't have any user-facing impact, since we only interact with this object behind-the-scene when Publishing from Content Editor to Portal, or inserting a project through Canvas's LTI assignment placement.

### Motivation and Context

For expedience, we put project submission logic in `CourseContentHistories`. The reasoning behind this is because a Fellow's project submission is always tied to a particular version of the project (e.g., a `CourseContentHistory` object). 

Overloading the `CourseContentHistoriesController` is getting confusing and making it difficult for us to implement things like showing multiple project submissions per fellow correctly. 

The first step in implementing the remaining project submission user stories is to refactor the code. 

### Test

- Run `rspec`
- Go to `platformweb/course_contents/#/versions` and make sure versions still render
Screen Shot 2020-09-29 at 4 10 47 PM

- Go to Canvas and create a new Assignment with a project saved in the Content Editor, have student submit project
Screen Shot 2020-09-29 at 4 11 04 PM

- Make sure the submission still renders in the SpeedGrader
Screen Shot 2020-09-29 at 4 10 54 PM

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #362, opened 2020-10-02
Refactor project code

jessouyangbraven:
**Task**: https://app.asana.com/0/1174274412967132/1186960110311121

### Description

- PR #361 got too big, so I've isolated the `Project` changes to this separate PR.
- Move project creation out of `lti_assignment_placement_controller#create` -> `projects_controller#create` (update views, specs to reflect this)
- To create `Project`s in the DB we need a few DB migrations: (1) make the existing columns `null`able, (2) add a `custom_content_version` to the table, (3) link back to the `project` from `custom_content_version`. This lets us do:
`Project.new` and `@project.custom_content_version`

### Motivation and Context

- We're moving to the data model here: https://dbdiagram.io/d/5f6e41177da1ea736e2f6328
- This will allow us to do things like view previous project submission answers correctly
- See PR #358 and #359 for additional details

### Test

- rspec
- Insert an Assignment in Canvas
Screen Shot 2020-10-01 at 5 31 42 PM
- This now creates a `Project` tied to a `CustomContentVersion`, viewable at `platformweb/projects/<project id>/`
Screen Shot 2020-10-01 at 5 31 51 PM
- Fellow can submit project like before
Screen Shot 2020-10-01 at 5 32 27 PM
- TA can view project submission in SpeedGrader like before
Screen Shot 2020-10-01 at 5 35 01 PM

### Types of changes

### Checklist:


app/controllers/projects_controller.rb
```diff
@@ -1,16 +1,51 @@
...
-  end
+  before_action :set_lti_launch, only: [:create]
+  before_action :set_custom_content, only: [:create]
+
+  skip_before_action :verify_authenticity_token, only: [:create], if: :is_sessionless_lti_launch?
 
-  # GET /projects/1
-  # GET /projects/1.json
   def show
+    authorize :application, :show?
```
jessouyangbraven, line 15: This is all very `admin`-y so I used `ApplicationPolicy`. I could add `ProjectPolicy`, but it'd be empty. I don't feel strongly, so let me know which is preferable.

app/controllers/projects_controller.rb
```diff
@@ -1,16 +1,51 @@
...
+    @custom_content_version = @project.custom_content_version
+  end
+
+  # Create a Project for an LTI assignment placement
+  def create
+    authorize :application, :create?
+
+    params.require([:state])
+    authorize @lti_launch
```
jessouyangbraven, line 24: Too much authorization? Should I pick one?

app/controllers/projects_controller.rb
```diff
@@ -1,16 +1,51 @@
...
+
+    # Create a new version for the project
+    @project = Project.new
+    @custom_content.save_version!(current_user, @project)
+    @project.custom_content_version = @custom_content.last_version
+    @project.save!
```
jessouyangbraven, line 30: I'd appreciate feedback here. It seems like I'm doing something wrong with the Rails ORM if I have to manually link the project and version. Is there a better way to do this?

app/controllers/projects_controller.rb
```diff
@@ -1,16 +1,51 @@
...
+    # Create a project submission URL tied to this project
+    # TODO: https://app.asana.com/0/1174274412967132/1186960110311121
+    # Use the project submission URL, e.g.:
+    # project_submission_url = new_project_project_submission_url(
+    #   project_id: @project.id,
+    # )
+    project_submission_url = custom_content_custom_content_version_url(
+      @project.custom_content_version.custom_content,
+      @project.custom_content_version,
+    )
```
jessouyangbraven, line 41: Still backward compatible here and still using `custom_content_versions` and not `project_submissions`. 

I swear I'm going to actually get to `ProjectSubmission` eventually.

Hopefully it'll be the next PR, aka #361 without all of this stuff in it.

app/controllers/projects_controller.rb
```diff
@@ -1,16 +1,51 @@
...
 
-  # GET /projects/1
-  # GET /projects/1.json
   def show
+    authorize :application, :show?
+    @custom_content_version = @project.custom_content_version
+  end
+
+  # Create a Project for an LTI assignment placement
+  def create
```
jessouyangbraven, line 20: This is mostly lifted from `lti_assignment_placement_controller#create`, with changes to introduce `Project`.

app/models/custom_content_version.rb
```diff
@@ -1,4 +1,5 @@
 class CustomContentVersion < ApplicationRecord
   belongs_to :custom_content
   belongs_to :user
+  belongs_to :project, required: false
```
jessouyangbraven, line 4: I couldn't do `project.custom_content_version` without this. See `project_controller.rb:26-30`'s comment.

app/views/projects/_upload.html.erb
```diff
@@ -0,0 +1,10 @@
+  <%= form_tag projects_path do -%>
```
jessouyangbraven, line 1: From `view/lti_assignment_placement/`

app/views/projects/create.html.erb
```diff
@@ -0,0 +1,8 @@
+<div id="assignment-create">
```
jessouyangbraven, line 1: From `view/lti_assignment_placement/`

config/routes.rb
```diff
@@ -98,7 +98,6 @@
   post '/lti/launch', to: 'lti_launch#launch'
 
   get '/lti/assignment_selection/new', to: 'lti_assignment_selection#new'     # https://canvas.instructure.com/doc/api/file.assignment_selection_placement.html
-  post '/lti/assignment_selection', to: 'lti_assignment_selection#create'     # https://canvas.instructure.com/doc/api/file.assignment_selection_placement.html
```
jessouyangbraven, line 101: We use `projects#create` now.

db/migrate/20201001232554_add_project_to_custom_content_version.rb
```diff
@@ -0,0 +1,5 @@
+class AddProjectToCustomContentVersion < ActiveRecord::Migration[6.0]
```
jessouyangbraven, line 1: Ditto what I said in the `custom_content_version` model. I needed to add this reference back to project, but am still confused why.

spec/controllers/projects_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  let!(:user) { create :admin_user, canvas_id: lti_launch.request_message.canvas_user_id}
+
+  describe "GET #show" do
+    it "returns a success response" do
+      get :show, params: { id: project.id, state: state }
+      expect(response).to be_successful
+    end
+  end
+
+  describe "POST #create" do
```
jessouyangbraven, line 19: These tests came from `lti_assigment_placement_controller_spec`.

app/controllers/projects_controller.rb
rshipp: I would say an empty projectpolicy is better, only because this controller does have an associated model to reference. That's the logic i used putting application policy in HomeController and JoinController, but empty policy for BaseCourseController.

app/controllers/projects_controller.rb
rshipp: Yeah, and `authorize @lti_launch` will actually use `LtiLaunchPolicy.create?` which I don't think is the permission you intend to check here.

app/controllers/projects_controller.rb
sadleb: I think you can create the version first and call `Project.new` passing the version object as a parameter. Something `Project.new(custom_content_version: the_version)`

config/routes.rb
```diff
@@ -44,7 +44,7 @@
     resources :lessons, only: [:index, :show]
   end
 
-  resources :projects, only: [:index, :show] do
+  resources :projects, only: [:show, :create] do
     resources :project_submissions, :path => 'submissions', only: [:create]
```
sadleb, line 48: We'll want to add the `show` routes to a submission, but maybe that's the next PR.

app/controllers/projects_controller.rb
rshipp: I'm confused by the double reference, and we're in the team call so I can't focus rn. It looks like project has a custom_content_version_id and custom_content_version has a project_id and only one of those should be true.

app/controllers/projects_controller.rb
rshipp: > authorize using the launch instance. That way, if we write more specific policies we'll have access to the launch in order to write the logic.

@sadleb just to clarify here, if you wanted to pass in the lti launch record to ProjectPolicy, you'd have to do `authorize @lti_launch, policy_class: ProjectPolicy`, which translates to `ProjectPolicy.new?`, or 

    can I create a new Project if I have this lti_launch?

different from `authorize @lti_launch`, which translates to `LtiLaunchPolicy.new?` or

    can I create a new LtiLaunch record?

app/controllers/projects_controller.rb
jessouyangbraven: From today's meeting, I have the `belongs_to`, `has_one` flipped in the models.

Ryan's mnemonic: the model who's table has the `id` in it gets the `belongs_to`.

app/controllers/projects_controller.rb
jessouyangbraven: @sadleb, they'll use the `project_submissions#show`, which is the next PR.

The `projects#show` looks like this:

Screen Shot 2020-10-02 at 12 32 20 PM

It's an auto-generated Rails view, so I assume only we/admins in `platformweb` would be looking at it.

rshipp (approved):

db/schema.rb
```diff
@@ -200,14 +200,16 @@
...
-    t.boolean "grades_muted", default: false, null: false
-    t.datetime "grades_published_at"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.datetime "grades_published_at"
+    t.boolean "grades_muted"
+    t.float "percent_of_grade_category"
+    t.integer "points_possible"
+    t.string "name"
+    t.bigint "grade_category_id"
```
sadleb, line 210: Note for future readers, we decided to do this as a hack for now since we're not sure if we'll need those other models yet. If we end up using them, we should put the non-null constraints back in place. If not, we'll end up deleting them.

sadleb (approved):
lgtm

PR #363, opened 2020-10-02
Implement user->project policy

rshipp:
### Description

Task: https://app.asana.com/0/1174274412967132/1196321172031701/f

Implement and call a project policy, such that a user can only see projects that are tied to a course in which they are enrolled.

#### Testing

First set up some test data

```ruby
user = User.find(<some user id>)
project = Project.find(<some project id>)

course = Course.create!(name: "Test course")
section = Section.create!(name: "Test section", base_course_id: course.id)
course_project = BaseCourseProject.create!(base_course_id: course.id, project_id: project.id)
user.add_role STUDENT_ENROLLMENT, section
```

Then visit the project (that you've already set up) in Canvas. It should let you view the page. If you log in with a different account, or change the connections in the test data (e.g. `user.remove_role STUDENT_ENROLLMENT, section`), it should show a not authorized error.

### Motivation and Context

Proof of concept for using our data model to perform authorization.

### Types of changes

### Checklist:


app/controllers/project_submissions_controller.rb
```diff
@@ -15,6 +15,7 @@ class ProjectSubmissionsController < ApplicationController
   skip_before_action :verify_authenticity_token, only: [:create], if: :is_sessionless_lti_launch?
 
   def show
+    authorize @project, :show?
     authorize @project_submission
```
rshipp, line 19: note we're calling authorize twice, once to check permission to show the project, and once to check permission to show the submission. For now, since the submission policy is "any logged in user", this doesn't accomplish much. We could take out the submission check, and remember to add it back in later if we make that policy more complex (i.e. if we drop LRS in favor of local db), or we could just leave it in now.

app/models/course.rb
```diff
@@ -1,2 +1,3 @@
 class Course < BaseCourse
+  has_many :sections, foreign_key: :base_course_id
```
rshipp, line 2: Since coursetemplates don't have users, the sections relation belongs on the course model. The `foreign_key` param is needed here because of the single table inheritance stuff, otherwise it tries to use `sections.course_id`.

app/models/project.rb
```diff
@@ -14,6 +14,12 @@ class Project < ApplicationRecord
 
   has_many :project_submissions
   has_many :users, :through => :project_submissions
+
+  has_many :base_course_projects
+  has_many :base_courses, through: :base_course_projects
+  has_many :courses, -> { courses }, through: :base_course_projects, source: :base_course, class_name: 'Course'
+  has_many :course_templates, -> { course_templates }, through: :base_course_projects, source: :base_course, class_name: 'CourseTemplate'
```
rshipp, line 21: these took me a while to figure out, but they make writing the policies much easier.

app/models/section.rb
```diff
@@ -1,9 +1,7 @@
 class Section < ApplicationRecord
   resourcify
 
-  belongs_to :course, -> {
-    where(base_courses: { type: 'Course' })
-  }, foreign_key: :base_course_id
+  belongs_to :course, -> { courses }, foreign_key: :base_course_id
```
rshipp, line 4: refactored to use the new `courses` scope.

app/models/user.rb
```diff
@@ -58,8 +58,8 @@ def admin?
 
   # True if this user is a TA in the same section where target_user is a student.
   def ta_for?(target_user)
-    sections_with_role(:ta).each do |section|
-      return true if target_user.has_role? :student, section
+    sections_with_role(TA_ENROLLMENT).each do |section|
+      return true if target_user.has_role? STUDENT_ENROLLMENT, section
```
rshipp, line 62: note any existing roles you have set up will break with this change, since `:ta` is now `:TaEnrollment` and `:student` is now `:StudentEnrollment`.

app/policies/project_policy.rb
```diff
@@ -1,2 +1,21 @@
...
+    # If no record, can't show it.
+    # Note: record is a Project.
+    return false unless record
+
+    # Users can see projects attached to a course they are enrolled in.
+    record.courses.each do |course|
+      course.sections.each do |section|
+        SECTION_ROLES.each do |role|
+          return true if user.has_role? role, section
+        end
```
rshipp, line 15: Check each role, because we want to allow students, TAs, and in the future designers, etc.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -31,13 +31,13 @@ def run
   def add_enrollment!
     case sf_participant.role
     when SalesforceAPI::LEADERSHIP_COACH
-      sync_enrollment(sf_program.fellow_course_id, CanvasAPI::TA_ENROLLMENT, 
+      sync_enrollment(sf_program.fellow_course_id, TA_ENROLLMENT,
                       course_section_name)
       sync_enrollment(sf_program.leadership_coach_course_id,
-                      CanvasAPI::STUDENT_ENROLLMENT,
+                      STUDENT_ENROLLMENT,
```
rshipp, line 37: the value for these constants has not changed, I just moved them out of canvas lib.

config/initializers/constants.rb
```diff
@@ -4,3 +4,10 @@
 # Roles
 TA_ENROLLMENT = :TaEnrollment
 STUDENT_ENROLLMENT = :StudentEnrollment
+
+# Add all section-level roles to this list.
+# I.e. anything that can be used like `user.add_role XXX_ENROLLMENT, section`.
+SECTION_ROLES = [
+  TA_ENROLLMENT,
+  STUDENT_ENROLLMENT,
```
rshipp, line 12: we'd add DESIGNER_ENROLLMENT, etc here once we add those roles.

db/migrate/20201006180732_create_base_course_project.rb
```diff
@@ -0,0 +1,9 @@
+class CreateBaseCourseProject < ActiveRecord::Migration[6.0]
+  def change
+    create_table :base_course_projects do |t|
+      t.references :base_course, null: false, foreign_key: true
+      t.references :project, null: false, foreign_key: true
+      t.integer :canvas_assignment_id
+    end
```
rshipp, line 7: We're using this table for "has_many: through" relations, as opposed to "has_and_belongs_to_many (habtm)" relations, because we want to be able to store additional data like the canvas assignment id in these records. this is the distinction I found online.

spec/controllers/project_submissions_controller_spec.rb
```diff
@@ -5,7 +5,11 @@
   render_views
 
   describe 'GET #show' do
-    let(:project_submission) { create :project_submission }
+    let(:project) { create :project }
+    let(:user) { create :fellow_user, section: section }
+    let(:project_submission) { create :project_submission, project_id: project.id, user: user }
+    let(:course_project) { create :course_project, project_id: project.id }
+    let(:section) { create :section, base_course_id: course_project.base_course_id }
```
rshipp, line 12: have to set up all of this stuff now, because of the policy. alternatively, we could make the user an admin, but I thought this made a more realistic test.

spec/controllers/project_submissions_controller_spec.rb
```diff
@@ -31,7 +35,7 @@
     end
 
     context "as a TA" do
-      let(:user) { create :ta_user }
+      let(:user) { create :ta_user, section: section }
```
rshipp, line 38: note when you create `:fellow_user` or `:ta_user` factories now, you can pass in a section to set that role on. if you don't pass in a section, one will be created, but you won't have access to it - so if you need to link it to a specific course, etc, you should pass your own in.

spec/factories/users.rb
```diff
@@ -22,15 +22,22 @@
...
 
       factory :fellow_user do
+        transient do
+          section { build(:section) }
+        end
         canvas_user_id { '1234' }
-        after :create do |user|
-          user.add_role :student
+        after :create do |user, options|
+          user.add_role STUDENT_ENROLLMENT, options.section
```
rshipp, line 30: these now add roles tied to a specific section, instead of globally.

app/controllers/project_submissions_controller.rb
jessouyangbraven: Or, we could call the `ProjectPolicy` under the hood in `ProjectSubmissionPolicy` if that's not too much of a pain?

app/models/base_course_project.rb
```diff
@@ -0,0 +1,4 @@
+class BaseCourseProject < ApplicationRecord
+  belongs_to :base_course
+  belongs_to :project
+end
```
jessouyangbraven, line 4: Is this a join table? I remember your DB diagram, but not why we needed the join tables. Also, you don't have to type out the explanation here, I just want to remember to ask in tomorrow's engineering meeting.

spec/controllers/project_submissions_controller_spec.rb
jessouyangbraven: Not for this PR, but is there a way to seed the test DB with something like `test_dummies.rb` that would do all of this setup? Or is that what the `spec/factories` are meant to do?

jessouyangbraven (commented):
Initial pass at reviewing this, publishing my comments for now since my brain is fried from trying to understand the scopes and DB changes.

spec/controllers/project_submissions_controller_spec.rb
sadleb: Factories are meant to do that in a way where you can isolate tests. If you seed the database and rely on that, then tests can interfere with each other.

sadleb (approved):

app/controllers/project_submissions_controller.rb
rshipp: I like that idea, let me look around and see if I can find any examples of other people calling one policy from another policy.

app/policies/project_policy.rb
rshipp: The only context we have here is `current_user` and `@project`, nothing from the lti_launch. This policy answers the question "can user X view project Y?".

spec/controllers/project_submissions_controller_spec.rb
rshipp: There might be a way to do it with the factories, something like how :fellow_user creates a default section - I tried to figure it out for a while and didn't get it working, and decided it was more trouble than it was worth. We do have to go through this whole setup in multiple test files though.

app/policies/project_policy.rb
rshipp: More info on this, in case we do end up needing to make more context-sensitive policies https://github.com/varvet/pundit#additional-context

app/controllers/project_submissions_controller.rb
rshipp: Found this https://stackoverflow.com/a/26553299/12432170, so I'll refactor like that

PR #369, opened 2020-10-08
Sync SF enrollments to local DB

rshipp:
### Description

Store sections in local db as well as syncing to LMS. Add "production_dummies" fallback.

### Motivation and Context

Prep for #363.

### Checklist:


app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -7,6 +7,7 @@ def initialize(portal_user:, salesforce_participant:, salesforce_program:)
     @portal_user = portal_user
     @sf_participant = salesforce_participant
     @sf_program = salesforce_program
+    @user = User.find_by!(email: sf_participant.email)
```
rshipp, line 10: note the `!` - this is relying on the assumption that the user already exists locally.

app/views/layouts/projects.html.erb
```diff
@@ -2,6 +2,7 @@
 
 <% content_for :head do %>
     <%= portal_stylesheet_link_tag %>
+    <%= stylesheet_link_tag 'custom_contents', media: 'all', 'data-turbolinks-track': 'reload' %>
```
rshipp, line 5: not related to the sync, I just added our CSS into this layout.

config/initializers/constants.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+# Global constants
+
+# Roles
+TA_ENROLLMENT = :TaEnrollment
+STUDENT_ENROLLMENT = :StudentEnrollment
```
jessouyangbraven, line 6: Are there going to be so few global constants that we only need a single `constants.rb` file? Or should we have `constants/roles.rb`? 

Not blocking. I can just see this file getting out of hand in the future.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -67,49 +68,71 @@ def complete_enrollment!
...
-    enrollment = find_user_enrollment(course_id)
+    section = find_or_create_section(canvas_course_id, section_name)
+    enrollment = find_user_enrollment(canvas_course_id)
     if enrollment.nil?
-      enroll_user(course_id, role, section)
+      enroll_user(canvas_course_id, role, section)
     elsif !enrollment.section_id.eql?(section.id) || !enrollment.type.eql?(role)
       canvas_client.delete_enrollment(enrollment: enrollment)
-      enroll_user(course_id, role, section)
+      enroll_user(canvas_course_id, role, section)
```
jessouyangbraven, line 95: Should we `drop_course_enrollment` here to remove the user from the old section? It looks like we're removing the enrollment from Canvas here, but not from `sections` in our DB.

jessouyangbraven (commented):
Just waiting on an answer to the `sync_enrollment` logic. Otherwise, I think this looks reasonable to merge.

app.json
```diff
@@ -16,7 +16,7 @@
         { "url": "heroku/ruby" }
       ],
       "scripts": {
-        "postdeploy": "DISABLE_DATABASE_ENVIRONMENT_CHECK=1 bundle exec rake db:schema:load db:seed"
+        "postdeploy": "DISABLE_DATABASE_ENVIRONMENT_CHECK=1 bundle exec rake db:schema:load db:seed db:production_dummies"
```
sadleb, line 19: Note that this is only helpful for Review Apps and any other new Heroku app we setup from this repo. From [the docs](https://devcenter.heroku.com/articles/app-json-schema#scripts) "The postdeploy script is run once, after the app is created and not on subsequent deploys to the app." Note that this file is missing something to scale the dyno which is why when we spun off the Highlander app the dyno wasn't automatically formed.

If we want something run on each deploy, we need to add it to the Procfile. 

I think this is fine, we just need to make sure and run this manually when we deploy this PR. I added this to the "deploy to prod" [Asana task](https://app.asana.com/0/1174274412967132/1198173495668881).

app/services/sync_portal_enrollments_for_program.rb
```diff
@@ -14,7 +14,7 @@ def run
         student_id: participant.student_id
       )
       if portal_user.nil?
-        # log skip no account yet
+        Rails.logger.warn("no portal account yet for '#{participant.email}'; skipping")
```
sadleb, line 17: This is actually an expected code path. This runs as part of the nightly sync. Every "Fully Confirmed" participant in Salesforce will be processed to see if we need to adjust their enrollment, but there may be quite a few who just haven't registered yet. I would either drop this down to a debug log and/or send a Honeycomb event so we can more easily answer the question "Fellow X isn't showing up in our reports but they are fully confirmed, why? Have they created their account yet?"

config/initializers/constants.rb
sadleb: I have a similar concern. I lean towards a single constants file but wrapping them in classes or structs so we can more easily group them. E.g.
```
class Role
  TA_ENROLLMENT = :TaEnrollment
  STUDENT_ENROLLMENT = :StudentEnrollment
end
```
and then callers would just use
`Role::TA_ENROLLMENT`

Note that b/c Ruby has open classes, if there is another `Role` class defined elsewhere, this just adds to it or vice versa.

spec/lib/canvas_api_spec.rb
```diff
@@ -71,6 +71,16 @@
         .with(headers: {'Authorization'=>'Bearer test-token'}).once
       expect(response).to eq(JSON.parse(response_user))
     end
+
+    it 'does not escape special characters' do
+      request_url = "#{CANVAS_API_URL}/accounts/1/users?include[]=email&search_term=test%2Bterm"
```
sadleb, line 76: `test%2Bterm` sure looks escaped to me! I'm confused here. If the `+` is escaped as `%2B` it works but `.CGI.Escape` was escaping it another way which didn't work? If so, maybe change the name of the example to "escapes a plus sign in an email correctly"

sadleb (approved):
I have a couple of small nit change requests but am approving anyway b/c they wouldn't prevent this from being merged.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -67,49 +68,71 @@ def complete_enrollment!
...
+    # TODO: Remove this fallback code once Highlander is in production.
+    # https://app.asana.com/0/1174274412967132/1197893935338145/f
+    base_course = BaseCourse.find_by(canvas_course_id: canvas_course_id)
+    if base_course.nil?
+      base_course = BaseCourse.find_by(name: 'Production Dummy')
+    end
+    Section.find_or_create_by!(
+      base_course_id: base_course.id,
+      name: section_name,
+      canvas_section_id: canvas_section.id
```
sadleb, line 125: Ok, so we're just going to have a dummy course but keep adding the new sections to it instead of detecting that we're using the dummy course and just using a single dummy section. I think that's fine. We may end up with hundreds of these for Spring b/c it will be across all regions, but that seems fine.

app/services/sync_portal_enrollments_for_program.rb
rshipp: Ah I wasn't even paying attention to the log level, i can make it debug. I just needed to see this on the console while i was debugging.

spec/lib/canvas_api_spec.rb
rshipp: Oh yeah I'll change the test name. I'm still confused about this CGI escape thing, I think my initial assumption of why it was breaking was wrong.

config/initializers/constants.rb
rshipp: Interesting. I can make that change. I'll do a new PR for it to avoid merge conflicts.

app/services/sync_portal_enrollment_for_account.rb
rshipp: Oh good catch, somehow I read this as doing exactly that, even though it's not.

app/services/sync_portal_enrollment_for_account.rb
rshipp: Yeah, I thought this was would be easier to fix later.

PR #372, opened 2020-10-13
Fix Add User functionality

jessouyangbraven:
**Task**: https://app.asana.com/0/1174274412967132/1198168803811457

### Description

Creating users from `platformweb/users/new` was broken. This PR fixes it. 

### Motivation and Context

Before this PR, `Add User` was generating the following Devise error:

```
Filter chain halted as :require_no_authentication rendered or redirected
```
When we're creating a user from `platformweb/users/new`, we shouldn't be going through the Devise route. We're really just offering a UI for what we do via Rails console. 

But because we moved the `post users#create` route out of the `admin` namespace in #350, Devise is redirecting us to `Users::RegistrationsController#create`. 

For more details on the Devise error (which wasn't necessary to fix since we re-routed): based on [this bug report](https://github.com/heartcombo/devise/issues/4524), Devise's default behavior is to prevent creating new users (e.g., user registration) when you are already signed in. It does this [here](https://github.com/heartcombo/devise/blob/e3a00b27d19ba995891d7dd92394fe2900a789c2/app/controllers/devise/registrations_controller.rb#L4), in the `Devise::RegistrationsController` (which we inherit from in `Devise::RegistrationsController`).

### Test

- Sign in to `platformweb`, click **Users**, then **Add User**
- Fill out form, click **Add User**
Screen Shot 2020-10-13 at 3 02 34 PM

- See that you're redirected to `platformweb/users` and see that the new user is in the list
Screen Shot 2020-10-15 at 12 00 48 PM

### Types of changes

### Checklist:


app/controllers/users/registrations_controller.rb
```diff
@@ -65,7 +67,12 @@ def find_user_by_salesforce_id
 
   # If you have extra params to permit, append them to the sanitizer.
   def configure_sign_up_params
-    devise_parameter_sanitizer.permit(:sign_up, keys: [:salesforce_id])
+    devise_parameter_sanitizer.permit(:sign_up, keys: [
+      :first_name,
+      :last_name,
+      :salesforce_id,
+      :canvas_user_id,
+    ])
```
jessouyangbraven, line 75: One the above error goes away, we get a new error that's fixed by whitelisting the form inputs.

app/services/account_creator.rb
```diff
@@ -33,13 +30,18 @@ def create_platform_user!
...
-    @password_params
-      .merge({
-               email: salesforce_contact.email,
-               first_name: salesforce_contact.first_name,
-               last_name: salesforce_contact.last_name,
-               salesforce_id: salesforce_contact.id
-             })
+    unless salesforce_contact
+      # Use the signup parameters from registration controller
+      return @sign_up_params
```
jessouyangbraven, line 35: This lets us create users without setting up Salesforce in our development environment.

app/services/account_creator.rb
```diff
@@ -33,13 +30,18 @@ def create_platform_user!
...
+    else
+      # Only overwrite non-nil values from Salesforce
+      return @sign_up_params.merge!({
+        email: salesforce_contact.email,
+        first_name: salesforce_contact.first_name,
+        last_name: salesforce_contact.last_name,
+        salesforce_id: salesforce_contact.id,
+      }) { |key, v1, v2| v2.nil? ? v1 : v2 }
```
jessouyangbraven, line 43: This lets us create test users without having to provision them in Salesforce.

jessouyangbraven:
A couple of notes for reviewers: 
- Don't look at the commits, they're non-sensical and just show my debug process. I can `git rebase -i` to clean it up if it's distracting. 
- This is the minimal/hot-fix-ish set of changes to get stuff working. I'm open to writing specs in a follow-up PR or cutting an Asana task for it. It'll depend on what we prioritize for Highlander this week.

sadleb (commented):
My understanding is that in `app/views/users/new.html.erb` when you submit, it does a POST to `users_path` which is routed to `UsersController::create`. There seems to be some Devise magic that is causing that post to go to `RegistrationsController::create` instead. While this does seem to work, it's a bit odd and doesn't match how the `app/views/users/new.html.erb` form appears to be supposed to work. I feel like the proper solution is to figure out why the routing is wrong so that Add User actually does post to `UsersController::create` and then redirects back to the users admin index view and flashes the message "User was added successfully".

I wonder if this would work in `config/routes`:
```
devise_scope :user do
  post '/users', to: "users#create", as: "users_path"
end
```

app/views/users/new.html.erb
```diff
@@ -8,7 +8,15 @@
...
     <div class="col-sm">
-      <%= form_for(@user, url: users_path) do |f| %>
+      <!-- 
+        Note: This uses admin_users_path, **not** users_path.
+        This lets bypass Devise's redirect to RegistrationsController#create
+        and post to our intended endpoint, UsersControllers#create.
+        See routes.rb for how this is configured.
+        TODO: https://app.asana.com/0/1174274412967132/1198306929320757
+        Write a spec for this.
+      -->
```
sadleb, line 18: This comment here is good/fine b/c it's an admin page, but something to keep in mind for end-user facing stuff is that that HTML comments are viewable by them so we should either have them be ruby erb comments, e.g. `<%# some comment %>` or move them to the model or controller.

sadleb (approved):

app/views/users/new.html.erb
jessouyangbraven: That's the difference between `<!-- -->` and `<%# %>`! Thanks for pointing this out.

PR #373, opened 2020-10-13
Implement course launch

rshipp:
### Description

Task: https://app.asana.com/0/1174274412967132/1196321172031698/f

Results of a running this, to double check it's correct:

* Fellow: https://braven.instructure.com/courses/215/
* LC: https://braven.instructure.com/courses/214

Main changes:

* Add CanvasAPI methods for creating a course and copying a course
* Add "launch" actions to the base_courses controller
* Put all the logic for launching a program in a new service
* Hook up those actions in the routes
* Add a view for the new form (see screenshot)
* Add policy rules for above
* Add specs for above

Supplemental changes:

* Whitespace :(
* Fixed some view layout annoyances
* Fixed base_course edit view so you can edit canvas_course_id for templates
* Fixed base_course edit view so you can select a nil course_resource

Not changed here:

* Anything to do with salesforce, except the program ID in the form.

New button on index page: 

Screen Shot 2020-10-15 at 2 32 22 PM

Launch program form:

Screen Shot 2020-10-15 at 2 33 11 PM

### Types of changes

### Checklist:


app/views/base_courses/launch_new.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <div class="row">
+    <div class="col-sm-4">
+      <div class="form-group">
+        <%= form.label :name, 'New course name' %>
+        <%= form.text_field :name, class: 'form-control', placeholder: 'Ex: 2020 Fall Braven Accelerator - SJSU' %>
+      </div>
+
+      <div class="form-group">
+        <%= form.label :salesforce_program_id, 'Salesforce program ID' %>
+        <%= form.text_field :salesforce_program_id, class: 'form-control', placeholder: 'Ex: a2Y11000001HY5xEAG' %>
```
rshipp, line 15: I don't love these placeholders, but when I tried it without the `Ex: `, it was hard to tell the fields weren't already filled in.

app/controllers/base_courses_controller.rb
```diff
@@ -61,6 +61,50 @@ def destroy
...
+          lc_course_template_id: lc_course_template_id,
+          lc_course_name: lc_course_name,
+      ).run
+    rescue RestClient::Exception => e
+      redirect_to course_management_launch_path, alert: "Canvas API error: #{e.message}" and return
+    rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotSaved => e
+      redirect_to course_management_launch_path, alert: "Error creating course: #{e.message}" and return
+    end
+
+    redirect_to base_courses_path, notice: "Program successfully launched."
```
sadleb, line 105: Is it ready immediately or do they have to wait? If they have to wait, should we add something like "The course should be ready to use in a couple of minutes."?

app/controllers/base_courses_controller.rb
```diff
@@ -61,6 +61,50 @@ def destroy
...
+  # POST /course_management/launch
+  def launch_create
+    authorize BaseCourse
+
+    # Validate form.
+    begin
+      salesforce_program_id, lc_course_template_id, lc_course_name, fellow_course_template_id, fellow_course_name = params.require([
+        :salesforce_program_id,
+        :fellow_course_template_id,
+        :fellow_course_name,
```
sadleb, line 81: In a future iteration, we probably want to pull the names from the Salesforce Program in order to have everything be consistent.

app/controllers/base_courses_controller.rb
```diff
@@ -61,6 +61,50 @@ def destroy
     end
   end
 
+  # Nonstandard actions related to course management.
+
+  # GET /course_management/launch
+  def launch_new
+    authorize BaseCourse
+    @course_templates = CourseTemplate.where.not(canvas_course_id: nil)
```
sadleb, line 69: I know you mentioned this in our standup, but I'm still confused. In what situation is the `canvas_course_id` nil? Or is this a temp hack b/c we don't already have templates populated to match what is in Canvas? If temp, let's queue up an asana task to remove this once we're guaranteed to have a Canvas course for every template, which is what we want I think.

app/services/launch_program.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class LaunchProgram
+  def initialize(salesforce_program_id:, fellow_course_template_id:, fellow_course_name:, lc_course_template_id:, lc_course_name:)
+    @salfesforce_program_id = salesforce_program_id
+    @fellow_course_template = CourseTemplate.find(fellow_course_template_id)
+    @fellow_course = Course.create!(
```
sadleb, line 7: Should we delay doing this until after the Canvas course is created in case there are failures so that we don't end up with orphaned `Course`s in our database?

sadleb (approved):
I left some comments about things you could consider changing but nothing rises to the level of blocking a merge so I'm approving.

app/controllers/base_courses_controller.rb
sadleb: https://app.asana.com/0/1174274412967132/1198495273642981

app/controllers/base_courses_controller.rb
```diff
@@ -61,6 +61,50 @@ def destroy
...
+        :lc_course_name,
+      ])
+      raise ActionController::BadRequest.new("Can't use the same template for Fellow and LC course") if fellow_course_template_id == lc_course_template_id
+    rescue ActionController::ParameterMissing, ActionController::BadRequest => e
+      redirect_to course_management_launch_path, alert: "Error: #{e.message}" and return
+    end
+
+    # Launch courses.
+    begin
+      LaunchProgram.new(
```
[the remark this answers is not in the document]
rshipp: I think that's the only one that's named that way, all the rest are imperative like this.

app/controllers/base_courses_controller.rb
rshipp: We don't have any validations to enforce presence of a canvas ID. In dev and test environments, we're mostly using courses without a canvas ID, so adding that validation would require a bunch of test changes. I don't think we can make this change until we make more decisions about how we're going to do course management.

I added this filter to the selector here because a course template without a canvas id is the worst case scenario as far as error handling goes - it means when you try to launch, you'll get a local course, an empty canvas course, and a 404 from the content migration API.

app/controllers/base_courses_controller.rb
rshipp: The content migrations might take longer, but the course links will work immediately. I added some more words to this, but realistically I don't think anyone ever reads alerts. If we get into a situation where people are confused by empty courses, better design would be to add progress indicators in our UI.

app/controllers/base_courses_controller.rb
```diff
@@ -61,6 +61,50 @@ def destroy
...
+          salesforce_program_id: salesforce_program_id,
+          fellow_course_template_id: fellow_course_template_id,
+          fellow_course_name: fellow_course_name,
+          lc_course_template_id: lc_course_template_id,
+          lc_course_name: lc_course_name,
+      ).run
+    rescue RestClient::Exception => e
+      redirect_to course_management_launch_path, alert: "Canvas API error: #{e.message}" and return
```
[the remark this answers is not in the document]
rshipp: Just look in honeycomb probably. This message shows to the end user, and they don't need to see all the error details.

app/services/launch_program.rb
rshipp: This is a consistent issue in this launch process - we're making changes all over on separate systems, and we don't have a good way to ensure consistency. For example, the canvas course create might succeed, and the content migration might fail, leaving us with empty courses in Canvas. In a reliable ideal, we'd wrap the whole thing in error handling that acted like a transaction, rolling back any changes in all systems if one of them failed. Realistically, I think that adds more complexity than the value it provides us. Each error handling code branch means more tests and higher maintenance cost.

For the scenario you describe, we have to choose between broken local courses (if the canvas course create fails) and something breaking later on in the process (like course copy, or salesforce) but with a local course record that tells us the canvas course ID and might help us troubleshoot.

lib/canvas_api.rb
```diff
@@ -281,6 +281,43 @@ def upload_file_to_course(file, original_filename, content_type, course_id=Conte
...
+      'course[name]': name,
+      'offer': publish,
+    })
+
+    JSON.parse(response.body)
+  end
+
+  # See: https://canvas.instructure.com/doc/api/content_migrations.html#method.content_migrations.create
+  # You probably don't want to call this directly; consider using copy_course or another appropriate
+  # method instead.
```
jessouyangbraven, line 298: Should this be `private`?

lib/canvas_api.rb
```diff
@@ -281,6 +281,43 @@ def upload_file_to_course(file, original_filename, content_type, course_id=Conte
 
   end
 
+  # See: https://canvas.instructure.com/doc/api/courses.html#method.courses.create
+  # Returns course Hash on success: https://canvas.instructure.com/doc/api/courses.html#Course
+  # Set publish:false to leave the course in the unpublished state.
+  def create_course(name, account_id=1, publish=true)
```
jessouyangbraven, line 287: I'm guessing the account_id `1` is the one associated with admin@bebraven.org.

Can we always guarantee it exists? And should this be a constant in `CanvasAPI`?

jessouyangbraven (approved):
A couple non-blocking comments. This looks good to me!

lib/canvas_api.rb
rshipp: I almost made it private, but looking at the other lower-level methods in this lib like `get` and `post`, I thought leaving it public would be more consistent.

lib/canvas_api.rb
rshipp: Pulled it out into a constant. Accounts in Canvas are different from users, right now account 1 is "Braven" (https://braven.instructure.com/accounts/1/). I'm not sure how we'll be using accounts in the future, at one point I think @sadleb mentioned something about an account per region.

lib/canvas_api.rb
```diff
@@ -281,6 +284,43 @@ def upload_file_to_course(file, original_filename, content_type, course_id=Conte
...
+  # Returns course Hash on success: https://canvas.instructure.com/doc/api/courses.html#Course
+  # Set publish:false to leave the course in the unpublished state.
+  def create_course(name, account_id=DefaultAccountID, publish=true)
+    response = post("/accounts/#{account_id}/courses", {
+      'course[name]': name,
+      'offer': publish,
+    })
+
+    JSON.parse(response.body)
+  end
```
sadleb, line 297: I created a task to revisit the account that we launch it under here: https://app.asana.com/0/1174274412967132/1198515006189725

PR #374, opened 2020-10-15
Fix Project Submission button CSS in prod

sadleb:
Task: https://app.asana.com/0/1174274412967132/1190416361215585

The project submission button uses Bootstrap CSS from the bootstrap node
package. It's in the application webpack and is only currently used for the admin layout.
This PR adds it to the project layout as well. 

Eventually, we need to clean this up and include bootstrap everywhere most likely, but for
now just error on the safe side.

Note that this works in dev b/c locally there is the webpacker-dev-server that compiles it:
https://stackoverflow.com/questions/58328026/css-and-bootstrap-not-loading-in-rails-6-app-when-deployed-to-heroku

### Test Plan:
- Open an LTI project as a student and scroll to the bottom when the app
  is built and running in production mode. The button should be full width
  and have the `.btn-block` CSS class applied to it.

image

### Checklist:


rshipp (approved):
seems like application stylesheet should be linked in the application layout, but I don't know what's in there and don't want to mess up the login page again, so i think this is better for now.

sadleb:
> seems like application stylesheet should be linked in the application layout, but I don't know what's in there and don't want to mess up the login page again, so i think this is better for now.

That was my thought as well. If we move it there, we need to go through all pages and make sure it doesn’t mess with the styling. This is a short term solution.

PR #381, opened 2020-10-21
Update portal ID for booster differently

skaex:
### Description
This allows for portal ID for booster to be updated differently on SF.

### Motivation and Context
This was required from the data team

https://app.asana.com/0/1178906757852153/1195445524446161

### Types of changes

### Checklist:

### DIFF
--- a/app/services/setup_portal_account.rb
+++ b/app/services/setup_portal_account.rb
@@ -41,7 +41,11 @@ class SetupPortalAccount
 
   def update_user_references!(user, salesforce_id:, join_user_id:)
     user.update!(canvas_user_id: portal_user.id, join_user_id: join_user_id)
-    sf_client.update_contact(salesforce_id, canvas_user_id: portal_user.id)
+    sf_client.update_contact(
+      salesforce_id,
+      canvas_user_id: portal_user.id,
+      is_booster_instance: booster_instance?
+    )
     user
   end
 
@@ -51,6 +55,10 @@ class SetupPortalAccount
 
   def should_create_join_user?
     ENV['CREATE_JOIN_USER_ON_SIGN_UP'].present?
+  end
+
+  def booster_instance?
+    ENV['BOOSTER_INSTANCE'].present?
   end
 
   def find_or_create_portal_user!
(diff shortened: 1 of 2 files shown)


app/services/setup_portal_account.rb
```diff
@@ -53,6 +57,10 @@ def should_create_join_user?
     ENV['CREATE_JOIN_USER_ON_SIGN_UP'].present?
   end
 
+  def booster_instance?
+    ENV['BOOSTER_INSTANCE'].present?
```
rshipp, line 61: should this just be `ENV['BOOSTER_INSTANCE']`? I can't find any reference to what `present?` is

skaex: Here's one on api dock: https://apidock.com/rails/v6.0.0/Object/present%3F . From the method signature, I want a boolean indicating if this is a booster instance or not.

rshipp: Ahh ok, thanks for the link!

sadleb (changes requested):
We already have an APPLICATION_HOST ENV var. Would it be cleaner and less maintainance to use that to determine which Salesforce field to set?

skaex:
> We already have an APPLICATION_HOST ENV var. Would it be cleaner and less maintainance to use that to determine which Salesforce field to set?

I thought of this but decided against it. My reasoning was that this would couple this implementation to the booster url

```ruby
def booster_instance?
    ENV['APPLICATION_HOST'].eql?('boosterplatform.braven.org')
end
```

Whenever the variable changes, we have to change the code too.

sadleb:
> > We already have an APPLICATION_HOST ENV var. Would it be cleaner and less maintainance to use that to determine which Salesforce field to set?
> 
> I thought of this but decided against it. My reasoning was that this would couple this implementation to the booster url
> 
> ```ruby
> def booster_instance?
>     ENV['APPLICATION_HOST'].eql?('boosterplatform.braven.org')
> end
> ```
> 
> Whenever the variable changes, we have to change the code too.

I can’t imagine us changing it, but I’m fine with this since the plan is for both Booster and Portal to go away by next fall so it’s not worth discussing in too much more depth.

sadleb (approved):

PR #383, opened 2020-10-22
Associate project version with course when assignment is created

jessouyangbraven:
**Dev environment breaking changes**
- `bundle exec rails db:migrate`
- create new assignments in Canvas (because the submission URL has changed)

### Description

**Tasks**
- https://app.asana.com/0/1174274412967132/1198222893347557
- https://app.asana.com/0/1174274412967132/1198201256763533

- DB migrations to implement this: https://dbdiagram.io/d/5f6e41177da1ea736e2f6328
- The data in the join table `base_course_custom_content_versions` has the data necessary for creating a course by copying course template data.

### Motivation and Context

The join table `base_course_custom_content_versions` tells us which project versions (`CustomContentVersion`) are in which Canvas courses (`BaseCourse`). This lets us copy course templates to create new courses. 

### Test

- `rspec`
- You'll need to manually add your Canvas Playground course to your local database. See `db/dummies.rb` for the SJSU example.
- As an admin, create a new Assignment in that Canvas course. You will need to manually set `type: 'Project'` on the projects you created in the content editor in order for them to appear in the drop-down menu. 
- As a student, view and submit the assignment through Canvas. You will need to manually add the role `StudentEnrollment` for a section in the course for this user in order to see the project submission page.
- As an admin, view the project submission in Speedgrader. You will need to manually add the role `TAEnrollment` for the section that the student is in to this user in order to see the student's responses populated in the project submission.

### Types of changes

### Checklist:


app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class BaseCourseCustomContentVersionsController < ApplicationController
```
jessouyangbraven, line 3: This is mostly copied from `ProjectsController` with `project` renamed to `base_course_custom_content_version`.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  before_action :set_lti_launch, only: [:create]
+  before_action :set_custom_content, only: [:create]
+  before_action :set_course, only: [:create]
+
+  skip_before_action :verify_authenticity_token, only: [:create], if: :is_sessionless_lti_launch?
+
+  def show
+    authorize @base_course_custom_content_version
+    @project = @base_course_custom_content_version
```
jessouyangbraven, line 17: **TODO**: Rename and update views.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  # Create a Project for an LTI assignment placement
+  def create
+    # Create new version of the content
+    authorize @custom_content
+    @custom_content.save_version!(current_user)
```
jessouyangbraven, line 24: **TODO**: Update `save_version` to return the `CustomContentVersion`.

app/controllers/project_submissions_controller.rb
```diff
@@ -8,28 +8,35 @@ class ProjectSubmissionsController < ApplicationController
...
     @project_lti_id = @lti_launch.activity_id
   end
 
   def new
+    # Update the join table entry for the Project because the submission page
+    # preview is the first time we have access to the Canvas assignment ID
+    unless @course_content_version.canvas_assignment_id
+      @course_content_version.canvas_assignment_id = @lti_launch.request_message.custom['assignment_id']
+      @course_content_version.save!
+    end
```
jessouyangbraven, line 30: @sadleb Do we still want to save this or should I get rid of this until we need it?

app/models/base_course_custom_content_version.rb
```diff
@@ -0,0 +1,12 @@
+class BaseCourseCustomContentVersion < ApplicationRecord
```
jessouyangbraven, line 1: This replaces `BaseCourseProject`.

app/models/project.rb
```diff
@@ -4,24 +4,7 @@
...
-  has_many :users, :through => :project_submissions
-
-  has_many :base_course_projects
-  has_many :base_courses, through: :base_course_projects
-  has_many :courses, -> { courses }, through: :base_course_projects, source: :base_course, class_name: 'Course'
-  has_many :course_templates, -> { course_templates }, through: :base_course_projects, source: :base_course, class_name: 'CourseTemplate'
- 
-  alias_attribute :submissions, :project_submissions
-
+class Project < CustomContent
```
jessouyangbraven, line 7: This is the model change for single-table inheritance. `Project` comes from the `custom_contents` table where the `type` column is set to `Project`.

app/models/project_submission.rb
```diff
@@ -1,10 +1,16 @@
 class ProjectSubmission < ApplicationRecord
```
jessouyangbraven, line 1: I left the `ProjectSubmission` name the same. I'm open to renaming it in a later PR, but this one was getting unwieldy.

config/routes.rb
```diff
@@ -47,10 +47,12 @@
     resources :lessons, only: [:index, :show]
   end
 
-  resources :projects, only: [:show, :create] do
+  resources :base_course_custom_content_versions, only: [:show, :create] do
     resources :project_submissions, :path => 'submissions', only: [:show, :new, :create]
   end
 
+  resources :project_submissions, only: [:show]
```
jessouyangbraven, line 54: Rails creates `ProjectSubmission`s with unique IDs. This lets us to `platformweb/project_submissions/{id}` and use `project_submission_{path, url}` directly, without having to specify `base_course_custom_content_version`.

db/migrate/20201019185309_drop_projects.rb
```diff
@@ -0,0 +1,8 @@
+class DropProjects < ActiveRecord::Migration[6.0]
```
jessouyangbraven, line 1: Because we're using single-table inheritance, the `Projects` model uses the `custom_contents` table.

db/migrate/20201019185309_drop_projects.rb
```diff
@@ -0,0 +1,8 @@
...
+  def change
+    remove_reference :rubrics, :project
+    remove_reference :project_submissions, :project
+    drop_table :base_course_projects, {}
```
jessouyangbraven, line 5: All of these things depend on the `projects` table, so we need to get rid of them before we drop it.

db/migrate/20201020175816_rename_base_course_project.rb
```diff
@@ -0,0 +1,9 @@
+class RenameBaseCourseProject < ActiveRecord::Migration[6.0]
```
jessouyangbraven, line 1: We actually create a new table here because had to drop `base_course_projects` above to get rid of `projects`.

db/migrate/20201020195702_add_base_course_custom_content_version_to_project_submissions.rb
```diff
@@ -0,0 +1,6 @@
+class AddBaseCourseCustomContentVersionToProjectSubmissions < ActiveRecord::Migration[6.0]
+  def change
+    ProjectSubmission.delete_all
```
jessouyangbraven, line 3: If you had stuff in `ProjectSubmissions`, it might fail to validate, so delete everything in it and then re-create the reference.

app/policies/base_course_custom_content_version_policy.rb
```diff
@@ -0,0 +1,21 @@
+class BaseCourseCustomContentVersionPolicy < ApplicationPolicy
```
jessouyangbraven, line 1: **TODO**: Ask @rshipp for help with the policy stuff.

app/policies/project_submission_policy.rb
```diff
@@ -15,20 +15,23 @@ def initialize(user, record)
   # When showing a submission, you also must have permission to show the
   # associated Project, so we check that too.
   def show?
-    ProjectPolicy.new(user, Project.find(record.project_id)).show?
+    true
+    # BaseCourseCustomContentVersionPolicy.new(user, Project.find(record.project_id)).show?
```
jessouyangbraven, line 19: **TODO**:  ditto on policy stuff

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  def set_custom_content
+    params.require(:custom_content_id)
+    @custom_content = CustomContent.find(params[:custom_content_id])
+  end
+
+  def set_course
+    @course = BaseCourse.find_by!(
+      canvas_course_id: @lti_launch.request_message.custom['course_id'],
+    )
+  end
```
jessouyangbraven, line 49: **TODO**: This is causing the controller spec to fail because I haven't figured out how to pass in the course ID.

jessouyangbraven:
Test failures:

- spec/policies/**project_submission_policy_spec.rb** because I need to actually fix the policies
- spec/controllers/**base_course_custom_content_versions_controller_spec.rb** because I need `course_id` stubbed out in the LTI launch request message custom
- spec/models/**base_course_spec.rb** because of single-table inheritance and model returning `Project` instead of `CustomContent`

app/models/base_course.rb
```diff
@@ -37,4 +38,12 @@ class BaseCourse < ApplicationRecord
...
     attributes.slice('name')
   end
+
+  def custom_contents
+    custom_content_versions.map { |v| v.custom_content }
+  end
+
+  def projects
+    project_versions.map { |v| v.project }
+  end
```
rshipp, line 48: any of these that i wrote as functions instead of associations, it's just because they're joining 4 tables instead of 3 so we can't do a `has_x through:`

app/models/project_submission.rb
```diff
@@ -1,10 +1,16 @@
...
   belongs_to :user
-  belongs_to :project
+  belongs_to :base_course_custom_content_version
   has_one :rubric_grade
 
+  has_one :course, through: :base_course_custom_content_versions, source: :base_course, class_name: 'Course'
+  has_one :project_version, through: :base_course_custom_content_versions, source: :custom_content_version, class_name: 'ProjectVersion'
```
rshipp: these are has_one because there's only one bcccv

app/controllers/lti_assignment_selection_controller.rb
```diff
@@ -12,6 +12,6 @@ def new
     params.require([:state])
     authorize @lti_launch
 
-    @assignments = CustomContent.where(content_type: 'assignment')
+    @assignments = Project.all
```
jessouyangbraven, line 15: This isn't being created by the Content Editor yet. You have to manually set `type` to `Project` in the console for now for it to show up in the drop-down.

app/controllers/project_submissions_controller.rb
```diff
@@ -55,17 +55,14 @@ def create
...
+      project_submission_url(@project_submission),
     )
     LtiAdvantageAPI.new(@lti_launch).create_score(lti_score)
   end
 
   private
-  def set_project
-    params.require([:project_id])
-    @project = Project.find(params[:project_id])
+  def set_course_content_version
```
jessouyangbraven, line 64: `course_content_version` is shorthand for `base_course_custom_content_version`. If this is confusing, I can stop shortening it and just use the whole thing.

app/policies/base_course_custom_content_version_policy.rb
```diff
@@ -0,0 +1,2 @@
...
```
jessouyangbraven, line 1: Admin-only policy

app/policies/project_policy.rb
```diff
@@ -1,21 +1,2 @@
 class ProjectPolicy < ApplicationPolicy
```
jessouyangbraven, line 1: Admin-only policy (because it's a `CustomContent`)

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+class BaseCourseCustomContentVersionsController < ApplicationController
```
sadleb, line 3: I think it would be helpful to add a comment on this class giving more context about what it is. We've been discussing it, so we're on the same page but if we come back 6 months from now this is going to be very confusing to be able to immediately get into and understand what this represents.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  skip_before_action :verify_authenticity_token, only: [:create], if: :is_sessionless_lti_launch?
+
+  # Create a Project for an LTI assignment placement
+  def create
+    # If we end up adding a designer role, remember to authorize `ProjectVersion.create?`.
+    authorize BaseCourseCustomContentVersion
+    # Create new version of the content
+    @custom_content.save_version!(current_user)
+
+    # Create join table entry
```
sadleb, line 22: I would mention in the comment that ideally we would get and set the Canvas assignment_id, but we don't have access to it so we rely on different code to initialize / sync / reconcile the assignment IDs.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    @course_content_version = BaseCourseCustomContentVersion.create!(
+      base_course: @base_course,
+      custom_content_version: @custom_content.last_version,
+    )
+
+    # Create a submission URL for this course and version
+    submission_url = new_base_course_custom_content_version_project_submission_url(
```
sadleb, line 29: Note that this works when run from the actual browser in our LTI dev env config, but when done from the console it's in `http` mode and leads to Canvas not being able to load it. Is there ever a reason to create the `http` url? If not, I think we should add `:protocol => 'https'` as a parameter to this method so that it always generates the https URL, even from the console. Considering that this returns LTI stuff and is meant for running in an LTI context, I can't think of a reason to allow `http` urls... thoughts?

app/controllers/project_submissions_controller.rb
sadleb: I like the shortened version. I briefly spoke with Ryan about making an alias so we can start using a shortened version in the URL but we decided to discuss that more once we got into the survey stuff so we had a better idea for how to handle it when comparing how we'll have to use this for both surveys and projects.

app/models/base_course_custom_content_version.rb
sadleb: Similar to above, but asking more generally. How should we document this stuff so folks just getting into it, or even if we come back to it 6 months down the road and forgot all the details. Add a class comment giving more context? Or maybe we write up a wiki and reference that?

db/migrate/20201020175816_rename_base_course_project.rb
```diff
@@ -0,0 +1,9 @@
...
+  def change
+    create_table :base_course_custom_content_versions do |t|
```
sadleb, line 3: I think we should add a unique ID for (base_course, canvas_assignment_id). We shouldn't be able to create two of these for the same course / canvas_assignment.

db/migrate/20201020175816_rename_base_course_project.rb
```diff
@@ -0,0 +1,9 @@
...
+      t.references :base_course, null: false, foreign_key: true
```
sadleb, line 4: Should we add an index on base_course?

db/migrate/20201020195702_add_base_course_custom_content_version_to_project_submissions.rb
sadleb: Should we avoid doing things like this in the migration and just run it from the console before we deploy? I think this breaks rollbacks. I know that Harry and Ryan discussed this in another situation where he did something similar. In the dev env, it's useful to go forward and backwards in time to see the state of the DB so we should avoid revertible migrations as much as possible.

db/migrate/20201020175816_rename_base_course_project.rb
rshipp: We can, because canvas_assignment_id is nullable and default null. We could unique on base_course and custom_content_version instead?

db/migrate/20201020175816_rename_base_course_project.rb
rshipp: foreign keys always have autogenerated indexes

db/migrate/20201020195702_add_base_course_custom_content_version_to_project_submissions.rb
rshipp: As-is, this allows rollbacks, but can't recover deleted data. If we'd like to prevent rollbacks, we can delete_all in the `up` and raise an [IrreversibleMigration](https://api.rubyonrails.org/classes/ActiveRecord/IrreversibleMigration.html) in the `down`, but I don't see a reason to do that here. We could probably write this migration in a more complicated way to actually migrate that data, but it's definitely not worth the time in this case. Having manual steps outside the migration means we have to communicate them to all devs, which is more effort.

db/migrate/20201020175816_rename_base_course_project.rb
sadleb: Hmmm. I can’t think of a reason to publish something in two places but maybe there is a use case? I would say let’s unique on that to prevent bad data and remove it / figure out a different solution if we run into an actual use case?

rshipp: I think there's a Rails validation for "doesn't match any existing row unless it's null", if that matches what you're trying to do better

PR #385, opened 2020-10-22
Refactor usersviews

jessouyangbraven:
### Description

Refactor the shared code in `new.html.erb` and `edit.html.erb` in user views. They shared:

 - `_form.html.erb`: form you fill out with user information
 - `_errors.html.erb`: errors encountered from **Save**

### Motivation and Context

Related to #384, I started doing this while working on this [bug](https://app.asana.com/0/1174274412967132/1198168803811457).

This reduces code duplication and it also fixes the issue where the `admin` checkbox doesn't appear in `users/new`. 

### Test

- rspec from #384

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #391, opened 2020-10-23
Refactor usersviews

jessouyangbraven:
See https://github.com/bebraven/platform/pull/385


jessouyangbraven:
I somehow merged #385 into a branch other than `staging` after it was approved, so I went ahead and made a new PR and force-merged as an admin.

PR #392, opened 2020-10-24
Using latest Heroku Platform DB snapshot

hcli-work:
Added a script to download latest Heroku database snapshot from the braven-platform app.
Another script to restore from that snapshot and change all the passwords to our standard default one.

Test Plan: brought up all the repos and then attempted to login


docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
       # rspec server host port. The web server that rspec listens on when running tests. http://SPEC_HOST:SPEC_PORT
       - "49883:49883"
     volumes:
-      - .:/app
+      - ./latest.dump:/latest.dump
```
rshipp, line 22: did you mean to do this?

docker-compose.yml
```diff
@@ -52,10 +50,12 @@ services:
...
   platformdb:
-    image: postgres
-    env_file:
-      - .env.database
+    image: postgres:11.8
+    environment:
+      POSTGRES_USER: user
+      POSTGRES_PASSWORD: password
     volumes:
+      - ./latest.dump:/latest.dump
```
rshipp, line 58: what will happen if the dump doesn't exist?

rshipp:
i'm not sure we want setup.sh to automatically pull a database snapshot. I worry that will make it too easy to e.g. accidentally publish modules to production. what's the usecase for this? note we try to keep db:seed / db:dummies such that you can run those and get reasonable test data in your local db.

docker-compose.yml
hcli-work: definitely not. thanks for the catch. I thought I was editing  a different file at the time. ugh.

docker-compose.yml
hcli-work: then it creates a latest.dump directory and things get really weird. it took me a while to figure out what was happening. this is the reason that in the download script there's a `rm -rf latest.dump*` before actually downloading.

hcli-work:
ready for re-review

docker-compose.yml
```diff
@@ -19,6 +19,7 @@ services:
       # rspec server host port. The web server that rspec listens on when running tests. http://SPEC_HOST:SPEC_PORT
       - "49883:49883"
     volumes:
+      - ./latest.dump:/latest.dump
```
sadleb, line 22: I personally would put this under the `db` dir instead of at the root level, but it's not worth figuring out if it's a pain.

docker-compose/scripts/dblatest_restore.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+echo "Refreshing your local dev database from the latest Heroku snapshot db"
+
+docker-compose exec platformdb pg_restore --verbose --clean --jobs 6 --no-acl --no-owner -U user -d platform_development /latest.dump
+
+
+# This is our normal test password for everyone. I got this by loading a staging db with the security keys 
+# and stuff we use, then setting someones password to the test one in the rails console.
+ENCRYPTED_TEST_PASS='$2a$10$C2W0hszrbmpk8tkw0ViLFOXVFH1Sj6HAiMyGah6vdEoRUj7GK1KzO'
```
sadleb, line 9: If we have a Booster and Highlander refresh option, this will need to be different.

docker-compose/scripts/dblatest_restore.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+echo "Refreshing your local dev database from the latest Heroku snapshot db"
+
+docker-compose exec platformdb pg_restore --verbose --clean --jobs 6 --no-acl --no-owner -U user -d platform_development /latest.dump
```
sadleb, line 4: Most of us don't have 6 cores and even if we do we haven't allocated that many to the Docker environment. I think we should just use `2` jobs here (if any), otherwise, it's just going to lead to contention. For example: https://dba.stackexchange.com/questions/213734/why-might-a-parallel-pg-restore-command-take-hours-longer-to-complete-its-non-pa

setup.sh
```diff
@@ -1,12 +1,16 @@
...
-docker-compose build
+docker-compose down
+./docker-compose/scripts/dblatest_download.sh
+
+docker volume rm platform_db-platform
 docker-compose up -d platformdb 
-docker-compose up platformweb #this is a hack. it will exit with an error but needed for db:create
+sleep 5
+docker-compose up platformweb # this is a hack. it will exit with an error but needed for db:create
 docker-compose run platformweb bundle exec rake db:create db:schema:load db:migrate db:seed
```
sadleb, line 11: Why do we need to run `db:schema:load db:migrate db:seed` if we're doing a restore from a snapshot?

setup.sh
```diff
@@ -1,12 +1,16 @@
...
+
+./docker-compose/scripts/dblatest_restore.sh
+./docker-compose/scripts/dblatest_restore.sh  # the first time has a couple of errors.
```
sadleb, line 14: wut?!? This is just a `pg_restore` followed by a SQL command. What are the errors that you're seeing? I'm confused why running this again would fix anything...

setup.sh
```diff
@@ -1,12 +1,16 @@
 #!/bin/zsh
 
 cp .env.example .env
-cp .env.database.example .env.database
-docker-compose build
+docker-compose down
+./docker-compose/scripts/dblatest_download.sh
```
sadleb, line 5: Seems like it's important where you run this script from. If you try and manually run it from the `./docker-compose/scripts` dir instead of as part of the `setup.sh` the dump will end up in the wrong dir.

Also, what's the ongoing development guidance (which we should also add to teh README). Do we just run `setup` once and then direction run `dbrefresh` occasionally when we want a fresh DB? Or is the plan to re-run setup each time you want a fresh DB?

setup.sh
```diff
@@ -1,12 +1,16 @@
 #!/bin/zsh
```
sadleb, line 1: Do we expect everyone to have `.zsh` or should we just use `sh`?

setup.sh
```diff
@@ -1,12 +1,16 @@
...
 
 cp .env.example .env
-cp .env.database.example .env.database
-docker-compose build
+docker-compose down
```
sadleb, line 4: What about `/etc/hosts` ? The original setup script run from the `development` repo sets this up and the dev env pretty much requires you do it and add `platformweb` (and other related apps that it needs to connect to locally if applicable, like join and canvas).

.gitignore
```diff
@@ -51,3 +51,4 @@ yarn-debug.log*
 # Mac stuf
 .DS_Store
 **/.DS_Store
+*.dump
```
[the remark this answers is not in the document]
hcli-work: wow. didn't even know it would get built into the container image. that's frightening.

docker-compose.yml
hcli-work: oh whoops! this shouldn't even be here. I had originally thought about doing the pg_restore from the web container, but then figured why not do it from within the db container instead.

docker-compose.yml
```diff
@@ -52,10 +51,12 @@ services:
     command: ./docker-compose/scripts/docker_compose_run.sh
 
   platformdb:
-    image: postgres
-    env_file:
-      - .env.database
+    image: postgres:11.8
+    environment:
+      POSTGRES_USER: user
+      POSTGRES_PASSWORD: password
```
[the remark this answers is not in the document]
hcli-work: when would someone change the database password?

the only dependency i'm aware of is that the password is stored both in the docker-compose.yml and also in .env.example. previously, the dependency would have existed between .env.example and .env.database.example. where else did I miss something?

docker-compose/scripts/dblatest_restore.sh
hcli-work: that's fair. I'm going to try and see if I can dust off some shell scripting things and dynamically figure out the number of cores and then divide by two.

docker-compose/scripts/dblatest_restore.sh
hcli-work: why would this need to be different? won't the encrypted password still be stored in the same place on platform regardless of deployment?

i thought only the database snapshot download would be diffferent

docker-compose/scripts/dblatest_download.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+echo "Downloading the latest Heroku database snapshot"
+
+rm -rf latest.dump*
+heroku pg:backups:download --app braven-platform
```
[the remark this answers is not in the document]
hcli-work: I think we'll need to add a parameter to the script so that it goes to appropriate heroku app to grab the download

setup.sh
hcli-work: /etc/hosts, AWS, Heroku, cloning repos is done from development scripts. that felt like the right place instead of here.

setup.sh
hcli-work: I think using zsh is fine as that's the standard for Macs going forward (Lexi's laptop didn't have bash on it). I'm the only crazy one using Linux Mint. I installed zsh so that I could use this stuff.

setup.sh
hcli-work: good point. i hadn't considered being anywhere except in the root of the repo when running `setup.sh` Fix now or fix later?

in terms of guidance for dbrefresh, what's your take? my flow in the past has always been to grab a snapshot manually and then do the restores. that's only because it was occasionally useful for me to be able to pick a particular snapshot that I wanted to use so I'd download a couple and then restore from whichever one i wanted.

i renamed the old one because i didn't want to lose it in case I f'd this up.   :)

docker-compose.yml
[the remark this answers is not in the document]
sadleb: I don't see a reason to change it. I'm just going off the README. We should just remove the stuff in there about the DB password, hardcode it here and in `.env.example`, and remove `.env.database`

docker-compose/scripts/dblatest_restore.sh
sadleb: `SECRET_KEY_BASE` is different for each database which causes the same password to be hashed differently. It's in the same spot, but the value will be different.

setup.sh
sadleb: I just installed `zsh` about a year ago and always used `bash` before. I think it's fine. If we ever hire a contractor or someone who refuses to use a Mac and does everything on Linux with some evangelical devotion to some other niche shell we'll let them fix all these up :)

setup.sh
sadleb: My take:

1. Account for where you run the script later. Not worth fixing now. It's fine to assume it's run from the root for now. 
2. I've pretty much never had an occasion to need a particular snapshot. Usually, if I'm doing a dbrefresh it's b/c I'm trying to troubleshoot some active production bug and just want my dev env to match prod or I fubar'd my local dev env so badly that I just need to start over. My vote is to keep the concept of a having a `dbrefresh` in each repo that lets you just blow away the database on demand and get it looking like prod without having to run the entire setup script. Setup feels like it should just be first time (and if you really want to nuke everything including your local git dirs and start completely from scratch)

docker-compose/scripts/dblatest_restore.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 echo "Refreshing your local dev database from the latest Heroku snapshot db"
 
-docker-compose exec platformdb pg_restore --verbose --clean --jobs 6 --no-acl --no-owner -U user -d platform_development /latest.dump
-
+cpucores=`cat /proc/cpuinfo | grep "cpu cores" | uniq | awk -F'cpu cores\t:' '{print $2}'`
```
sadleb, line 4: Getting fancy!

setup.sh
rshipp: let's not write anything complex in zsh, but this seems fine. i don't know zsh at all, and it's not posix compliant by default.

docker-compose.yml
rshipp: I don't personally ever want a db dump. Will this still let me bring up the app without one?

docker-compose/scripts/dblatest_restore.sh
Rowantatai: Unfortunately, this doesn't work on a mac.

docker-compose.yml
hcli-work: it will. you just would have to bring things up manually yourself. what's the concern about not wanting a db dump?

docker-compose.yml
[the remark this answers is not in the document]
hcli-work: will do

docker-compose/scripts/dblatest_restore.sh
hcli-work: oh boy. okay. let me get this working for Accelerator/BravenX first.

setup.sh
hcli-work: so maybe change dbrefresh to essentially just run the latest download and then run restore?

hcli-work:
alright. I have rebased and brought back the dbrefresh.sh script which just invokes the download and restore scripts

docker-compose.yml
rshipp: the data in production is not useful for local dev, and if I load the production db i'd have to set up my local db all over again.

setup.sh
sadleb: I think we can remove. Those are for setting up a database using rails stuff. For restoring from a dump, you just need to have the database and be able to connect to it but the restore should add the tables and data.

setup.sh
sadleb: My hunch is that the default heroku dump wasn’t done using the “clean” flag meaning that there are no “drop table” commands in there. I’m not sure what a pg_restore clean does without the dump itself being done with that flag, but maybe it just truncates instead of dropping tables? What we really want is a completely fresh db to restore into so rather than running it twice I feel like we should just drop the database and recreate it before restoring into the fresh db. I feel like that’s the approach I’ve taken in the past when I’m really trying to blow a db away and restore a freshy

setup.sh
sadleb: Yup

hcli-work (commented):
realized i never submitted my comment

docker-compose/scripts/dblatest_download.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+echo "Downloading the latest Heroku database snapshot"
+
+rm -rf latest.dump*
+app=${1:-braven-platform}  # could be braven-platform-booster or braven-platform-highlander, also
+heroku pg:backups:download --app $app
```
sadleb, line 6: Does this default to storing it in the `latest.dump` file?

docker-compose/scripts/dblatest_restore.sh
```diff
@@ -0,0 +1,17 @@
...
+
+
+docker-compose down
+docker volume rm platform_db-platform
+docker-compose run platformweb bundle exec rake db:create
+docker-compose up -d
+sleep 5
+#cpucores=`cat /proc/cpuinfo | grep "cpu cores" | uniq | awk -F'cpu cores\t:' '{print $2}'`
+#jobs=`expr $cpucores / 2`
+docker-compose exec platformdb pg_restore --verbose --clean --jobs 2 --no-acl --no-owner -U user -d platform_development /latest.dump
```
sadleb, line 12: Is the `/latest.dump` b/c it reads from the file inside the container? Should it be `./latest.dump` or is it really at the root of the filesystem in there? Would it be better to just pipe it from outside the container using `cat latest.dump | docker-compose exec -T platformdb pg_restore --verbose --clean --jobs 2 --no-acl --no-owner -U user -d platform_development`

docker-compose/scripts/dblatest_restore.sh
sadleb: Did this actually work for Accel/BravenX? Seems like you added other stuff to switch apps but didn't update this?

sadleb (changes requested):
Main change i'm unclear about is the encrypted password part.

docker-compose/scripts/dblatest_download.sh
hcli-work: it does

docker-compose/scripts/dblatest_restore.sh
hcli-work: yes. it's at the filesystem root via a volume mounted in the docker-compose.yml
i can try piping it in. i'm not sure if it'll work though since these are binary dumps

docker-compose/scripts/dblatest_restore.sh
hcli-work: It did! I was kinda surprised, too, but was able to login as booster.admin@bebraven.org after refreshing from the  snapshots for braven-booster and braven-platform-booster.

docker-compose/scripts/dblatest_restore.sh
hcli-work: confirmed it doesn't work from stdin unfortunately when there is parallelism set

hcli-work (commented):
Determined that piping a dump file is not compatible with parallelism while restoring.
Re-confirmed that grabbing snapshots from both Accel and Booster work.

sadleb (approved):
lgtm

docker-compose.yml
rshipp: yes, there's nothing in the production dbs except whatever projects Helen's working on at the time. local dev dbs have users in sections in courses, with projects and modules, everything we use to develop with.

PR #396, opened 2020-10-26
Remove custom_contents:content_type

jessouyangbraven:
### Description

This PR removes references to `content_type` in the code and removes the column from the `custom_contents` table. 

### Motivation and Context

We've branched `highlander` from `staging` and no longer need to support Booster content. (As far as we know, the Content Editor was only used to create `module`/`wiki_page`/`page` and never used to created `assignment`/`project`.)

We're no longer using the `content_type` column. Instead, we're `type` for single-table inheritance, where `type` are be `"Project"` or `"Survey"` (specifically, an impact survey).

This is because both of these content types have the same DB storage needs (e.g., an HTML `body`) and are authored using our custom Content Editor.

Other content, like modules and resources are authored using Articulate Rise 360. Pre- and post-accelerator are also authored outside of our Content Editor (interface TBD). 

### Test

- rspec
- make sure that Content Editor `custom_contents#new` loads
- make sure existing `CustomContent` with `type: nil` load

### Types of changes

### Checklist:


sadleb (changes requested):
I was going to ask whether we should replace the drop down to select the content type with project and set the `type` so that single table inheritance works, but I see you did that in Follow-Up PRs. So the only change request here is to merge to the `Highlander` branch instead of staging

jessouyangbraven:
@sadleb, oof. Good catch! Updated base branch to `highlander` now. I also checked the other PRs to make sure they're correct.

rshipp (approved):

PR #397, opened 2020-10-26
Fix routes for projects

jessouyangbraven:
### Description

This PR configures the routes so `Project`s use the `CustomContentsController`. The route automatically passes in the `type: "Project"` as a parameter.

This fixes the bug where the Content Editor won't load if you manually set `type: "Project"` to a `CustomContent` in our local development DB.

### Motivation and Context

The Content Editor only creates `CustomContent` (`custom_contents:type: nil`) objects currently . 

To create a `Project`, you need to use the development console (`devc`) and manually set `type: "Project"` on an existing `CustomContent` or a new `CustomContent`.

This is currently causing the `form_with` and `link_to` helpers to crash. See [this](https://guides.rubyonrails.org/form_helpers.html#relying-on-record-identification) for more details. 

### Test

- `platformweb/{course_contents, projects}/new` load

- **Save** works (note that `type: nil` here. Setting it to `"Project"` will be in a later PR)
Screen Shot 2020-10-26 at 3 44 29 PM

- `platformweb/custom_contents` renders **all** `CustomContent`s
Screen Shot 2020-10-26 at 3 56 58 PM

- `platformweb/projects` renders **only** `Project`s
Screen Shot 2020-10-26 at 3 43 28 PM

- check that `#show` renders for `type: nil` and `type: "Project"` (note that the latter has to be set in the dev console)

### Types of changes

### Checklist:


config/routes.rb
```diff
@@ -10,6 +10,8 @@
     resources :custom_content_versions, path: 'versions', only: [:index, :show]
   end
 
+  resources :projects, :controller => "custom_contents", :type => "Project"
```
jessouyangbraven, line 13: This is the key fix.

app/controllers/custom_contents_controller.rb
```diff
@@ -17,7 +18,7 @@ def show
 
   # GET /custom_contents/new
   def new
-    @custom_content = CustomContent.new
+    @custom_content = CustomContent.new(type: params[:type])
```
rshipp, line 21: can we use `set_type` here like we do in base_courses_controller? otherwise it feels like code injection

jessouyangbraven: I tried using `set_type` from `BaseCoursesController` here and couldn't get around this error:
Screen Shot 2020-10-27 at 11 01 26 AM

I think it's because we hard-code the inputs to `name="custom_content[...]"` in `ContentEditor.js`.

If we need to validate `params[:type]`, is there another way to go about it? 

I was thinking that we could conditionally `permit` `:type` in `custom_content_params`? Something like:

```
  def new_params
    params.permit(:type) if set_type
  end

  def custom_content_params
    optional_params = [
      :title,
      :body,
      :published_at,
      :course_id,
      :course_name,
      :secondary_id,
    ]
   optional_params << :type if set_type
   params.require(:custom_content).permit(optional_params)
  end

  def set_type
    case params[:type]
    when 'Project'
      :project
   end
  end
```

app/controllers/custom_contents_controller.rb
```diff
@@ -94,11 +94,38 @@ def publish
...
   end
 
-  # Never trust parameters from the scary internet, only allow the white list through.
+  # Always use the following to create a CustomContent object:
+  #   custom_content_class.new
+  # **never** use unsafe `type` from the parameters, e.g.:
+  #   CustomContent.{new, update}(type: params[:type]) # BAD!
+  def custom_content_class
+    # Prefer `type` specified by form over the one set in route parameters
+    type = params[:custom_content_type] ? params[:custom_content][:type] : params[:type]
```
jessouyangbraven, line 106: @rshipp, I did try the nested parameters in `routes.rb` to get rid of the type being in two places. 

You can specify nested parameters like:

  resources :projects, controller: "custom_contents", custom_content: { type: "Project" }

But that unfortunately clobbers all of the `custom_content[...]` form data from the Content Editor, so you can never save the title/body/etc.

app/controllers/custom_contents_controller.rb
jessouyangbraven: @rshipp and I talked over Slack about this. 

The key thing we want to ensure is that we don't use a value for `"type"` passed in from the parameters that we haven't sanitized. 

There are a number of ways to do this, including:
1. validation on the model
2. the `set_type` solution used in `BaseCoursesController` and described in details here: https://youtu.be/2fH_V91b0H4?t=416
3. what I ended up doing in this PR, which is checking the parameter value and converting it into a model class name: https://stackoverflow.com/questions/9229554/how-to-run-validations-of-sub-class-in-single-table-inheritance/9230522#9230522

I went with (3) because (1) means we always have to maintain a list of subclasses in the `CustomContent` model, which seemed less maintainable. And (2) wasn't straightforward because of the error from the screenshot above.

app/controllers/custom_contents_controller.rb
rshipp: ah, unfortunate

rshipp (approved):
lgtm

sadleb (approved):
Tricky. So timely that this literally came up this morning when I was prepping for the demo. Thanks for getting ahead of it!

sadleb:
fyi: I went ahead and merged this @jessouyangbraven b/c I need it for what I'm working on.

PR #406, opened 2020-10-28
Add SurveyVersion < CustomContentVersion

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1198925268705965

This PR builds on #402, which added Surveys. In order to submit Impact Surveys, we need a `SurveyVersion`.

This PR adds the `SurveyVersion`, which uses single-table inheritance from `CustomContentVersion` (like `ProjectVersion`). It also updates the pundit policies so both `Project` and `SurveyVersionPolicy` delegate to `CustomContentVersion` policy. 

Survey version uses single-table inheritance, storing its data with project versions in the table `custom_content_version`:
Screen Shot 2020-10-29 at 1 49 19 PM

### Motivation and Context

When a Fellow submit an Impact Survey, we need that submission to be tied to a particular `SurveyVersion`. This is so that we can continue to render the responses in the form even if we later update the contents of the `Survey`. 

This is analogous to what we do for `Project`, `ProjectVersion` and `ProjectSubmission`, described [here](https://github.com/bebraven/platform/wiki/Project-Submissions).

### Test

- Added/updated specs
- Create a new Survey in the Content Editor and **Save**:
Screen Shot 2020-10-28 at 3 03 56 PM

- Replace `/edit` with **`/versions`** in the URL and see that there aren't any versions yet:
Screen Shot 2020-10-28 at 3 04 01 PM

- Using the Rails console, create a new version of that survey:

```
me = User.find(9) # any admin-user will work
survey = Survey.last 
survey.save_version!(me)
```

- Refresh the page and see that the new version appears:
Screen Shot 2020-10-28 at 3 05 46 PM

- Click the version number and see the survey version:

Screen Shot 2020-10-28 at 3 05 50 PM

### Types of changes

### Checklist:


app/controllers/custom_content_versions_controller.rb
```diff
@@ -10,12 +10,43 @@ class CustomContentVersionsController < ApplicationController
   # GET /custom_contents/:id/versions
   # GET /custom_contents/:id/versions.json
   def index
-    authorize CustomContentVersion
+    authorize custom_content_version_class
+    # Overwrite the object automatically created because the parameter for the
+    # CustomContent ID depends on the path name. 
+    @custom_content_versions = CustomContent.find(set_custom_content_id).versions
```
jessouyangbraven, line 16: This annoys me, but I could figure out how else to get this to work. Other ideas that seemed like better ways to do this, but I couldn't figure out:

  - Some configuration in the `routes.rb` that forces that first id to always be `custom_content_id` instead of `project_id`/`survey_id` when it's `versions/` nested in `projects/surveys`.
 - Override whatever is setting `@custom_content_versions` without this line of code (somewhere in `ApplicationController` or something included from `DryCrud::Controllers::Nestable`, but I couldn't find it)?

config/routes.rb
```diff
@@ -10,8 +10,13 @@
...
 
-  resources :projects, controller: "custom_contents", type: "Project"
-  resources :surveys, controller: "custom_contents", type: "Survey"
+  resources :projects, controller: "custom_contents", type: "Project" do
+    resources :project_versions, path: 'versions', controller: 'custom_content_versions', type: "ProjectVersion", only: [:index, :show]
+  end
+
+  resources :surveys, controller: "custom_contents", type: "Survey" do
+    resources :survey_versions, path: 'versions', controller: 'custom_content_versions', type: "SurveyVersion", only: [:index, :show]
+  end
```
jessouyangbraven, line 19: The helpers like `link_to` generate paths like `survey_survey_version_path` and `project_project_version_path` because they don't know about single-table inheritance and just have the object's class name. 

This error doesn't show up until I click around `platformweb` manually. I already the content editor with `Projects` and fixed it in #397. 

Is there a way to automate testing this? rspec views?

app/controllers/custom_content_versions_controller.rb
```diff
@@ -10,12 +10,43 @@ class CustomContentVersionsController < ApplicationController
...
+  def set_custom_content_id
+    case params[:type]
+    when 'ProjectVersion'
+      params[:project_id]
+    when 'SurveyVersion'
+      params[:survey_id]
+    when nil
+      params[:custom_content_id]
+    else
+      raise StandardError.new "Unknown CustomContentVersion type: #{params[:type]}"
```
sadleb, line 35: I think there is a `TypeError` that may make more sense to raise.

app/controllers/custom_content_versions_controller.rb
```diff
@@ -10,12 +10,43 @@ class CustomContentVersionsController < ApplicationController
...
+  def custom_content_version_class
+    case params[:type]
+    when nil
+      CustomContentVersion
+    when 'ProjectVersion'
+      ProjectVersion
+    when 'SurveyVersion'
+      SurveyVersion
+    else
+      raise StandardError.new "Unknown CustomContentVersion type: #{type}"
```
sadleb, line 49: Ditto.

app/models/custom_content.rb
```diff
@@ -52,6 +53,8 @@ def set_version_type
     case type
     when 'Project'
       'ProjectVersion'
+    when 'Survey'
+      'SurveyVersion'
     end
```
sadleb, line 58: Do we want to stop allowing a CustomContent to be stored without a type? That seems like left-over from before we did all this and now it should throw if it's unrecognized and not allow `nil` types into the DB?

app/policies/custom_content_version_policy.rb
```diff
@@ -1,5 +1,21 @@
 class CustomContentVersionPolicy < ApplicationPolicy
   def show?
-    !user.nil?
+    # Admins can always see project versions.
+    return true if user&.admin?
+
+    # If no record, can't show it.
+    # Note: record is a ProjectVersion.
```
sadleb, line 7: `ProjectVersion` or `SurveyVersion` * ?

app/policies/custom_content_version_policy.rb
```diff
@@ -1,5 +1,21 @@
...
+    return false unless record
+
+    # Users can see project versions attached to a course they are enrolled in.
```
sadleb, line 10: Update comment?

config/routes.rb
sadleb: @rshipp should weigh in here, but I think we should just be using the `polymorphic_path` stuff. E.g. `edit_polymorphic_path(@custom_content)` would go to the `edit` path of the correct one depending on the type.

sadleb (approved):
None of my feedback would block a merge so I'm approving.

config/routes.rb
rshipp: > This error doesn't show up until I click around platformweb manually.

What error? I'm lost here, what are we trying to test?

jessouyangbraven: The error is that Rails can't find `survey_survey_version_path` (or `project_project_version_path`) when it's generating the `link_to` endpoints until I explicitly set the routes in `routes.rb`, which I forget to do a lot and break things. 

This might just be me not using Rails as intended and not something we need to test

 I'll check out `polymorphic_path` that @sadleb suggested and see if that means we no longer have to set the routes explicitly.

rshipp: lmk if the polymorphic path thing doesn't help. I'm not sure i fully understand but it does seem like we should be able to catch broken link_tos from tests.

app/models/custom_content.rb
jessouyangbraven: I think not allowing `'nil` on the single-table inheritance `type` column makes sense going forward, especially since we're not dealing with Booster content on this branch.

This would effectively make `CustomContent` and `CustomContentVersion` an abstract/interface class. 

I'd like to discuss whether we should enforce this at the table/model/controller level.

Can we tackle this in a separate PR? I can queue up an Asana task.

config/routes.rb
jessouyangbraven: I looked into the polymorphic path stuff. From what I understand, it does help us out in views/controllers, but we would still need to map the routes in `routes.rb`. And I think we'd catch unmapped routes if we wrote specs for the views.

app/models/custom_content.rb
jessouyangbraven: https://app.asana.com/0/1174274412967132/1198972743465983

PR #408, opened 2020-10-29
Add SurveySubmissionAnswer model, table

jessouyangbraven:
### Description

This PR adds the `SurveySubmissionAnswer` model and table.

This is a key-value store meant specifically for the input name-value pairs generated by the form presented for a `SurveySubmission`.

This is **not** meant to be a general-purpose key-value store. 

### Motivation and Context

This allows us to store impact survey responses in our database. This allows us to fetch use the data in Periscope for analysis. 

See the DB diagram here: https://dbdiagram.io/d/5f98e3c03a78976d7b797320 (screenshot below) for how the `survey_submission_answers` table relates to other Survey Submission data:

Screen Shot 2020-10-29 at 1 49 19 PM

### Test

- rspec
- Use Rails console to create `SurveySubmissionAnswer` and attach it to a `SurveySubmission`

### Types of changes

### Checklist:


db/migrate/20201028233656_create_survey_submission_responses.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSurveySubmissionResponses < ActiveRecord::Migration[6.0]
+  def change
+    create_table :survey_submission_responses do |t|
+      t.belongs_to :survey_submission, foreign_key: true, null: false
+      t.string :input_name, null: false
```
sadleb, line 5: This comment is going to be annoying b/c I could have brought it up when we were brainstorming, but should we call it `input_key` instead of `input_name`?

db/schema.rb
```diff
@@ -295,6 +295,15 @@
     t.integer "ticket_granting_ticket_id"
   end
 
+  create_table "survey_submission_responses", force: :cascade do |t|
+    t.bigint "survey_submission_id", null: false
+    t.string "input_name", null: false
```
sadleb, line 300: We may want to index either this or the (id, name) pair depending on how we end up querying for these, but imo it's not worth adding now since we haven't implemented the consumer of these yet. I think the current plan is just to point Periscope at this table and do the querying over there. Periscope will just be incrementally pulling new records I think or even just pulling the entire table in one go, so I think we're good.

app/models/survey_submission_response.rb
```diff
@@ -0,0 +1,9 @@
+# This it our custom key-value (e.g., input name and input value) model for 
+# storing responses to impact survey submissions. 
+# This is **not** a general-purpose key-value store.
+# Please do not introduce the bz-retained-data nightmare.
+class SurveySubmissionResponse < ApplicationRecord
```
sadleb, line 5: Alternate naming proposals:

* `SurveySubmissionInput`
* `SurveySubmissionUserInput`
* `SurveySubmissionAnswer`

I'm not the biggest fan of `SurveySubmissionResponse` b/c it makes me think of a request / response, so if I was just coming into this codebase my first though would be "is this the response returned when asking what the overall submission of the survey was?" In other words, response makes me think of the collection of all the inputs returned as the response when viewing a survey submission.

sadleb (commented):
Not explicitly approving or requesting changes. I'd like to hear your responses to my feedback first.

Also, ditto this comment on the other PR about the PR description: "I think we should get in the habit of both linking to the dbdiagram and dropping a screenshot in here. I'm imaging years from now when the account is deleted or dbdiagram shuts down, it would still be nice to have the screenshot to make things more concrete even if the link breaks."

db/migrate/20201028233656_create_survey_submission_responses.rb
rshipp: I kinda like that `input_name` makes it explicit that this matches `<input name=???>` in the HTML

app/models/survey_submission_response.rb
jessouyangbraven: Use `SurveySubmissionAnswer`.

sadleb (approved):
lgtm

PR #412, opened 2020-10-29
Course Management UI for publishing Projects and Surveys

sadleb:
This is part 1 of a set of PRs (next one [here](https://github.com/bebraven/platform/pull/413)). It's the UI portion needed to move creating, updating, and deleting Projects/ Surveys in Canvas to the platform Course Management UI. We're moving in this direction for a few reasons:
1. We need the Canvas assignment ID for a Project or Survey in the platform database to support some upcoming features we're building. 
2. We also need to be able to identify that Project or Survey in Canvas after a course is cloned and the Canvas assignment IDs change.
3. It's hard satisfying these requirements using the LTI Assignment Selection Placement. These tasks have more context [1](https://app.asana.com/0/1174274412967132/1198900743766613), [2](https://app.asana.com/0/1174274412967132/1198500729774997), [3](https://app.asana.com/0/1174274412967132/1198500729774999)
4. A long term goal after we launch Highlander which is meant to consolidate and streamlining our tech systems, is to streamline execution and operations of running the Accelerator. Having things done programmatically from the platform is a step in that direction.

The next PR will be to implement the core logic of the UI described below.

### Course Template Edit Page
1. Adds a list of Projects and Surveys to the BaseCourse edit page.
<details><summary>Screenshot</summary>
</details>

2. Clicking on the name of an existing project shows the published version contents in Canvas.
3. Clicking "Publish Latest" on a project is routed to the `update` action on the existing BaseCourseContentVersion record (not implemented yet)
4. Clicking "Delete" on a project is routed to the `destroy` action on the existing BaseCourseContentVersion record (not implemented yet)

5. Clicking "Add Project" or "Add Survey" shows the list of Projects or Surveys to choose from. This is the `new` action for a BaseCourseContentVersion
<details><summary>Screenshot</summary>
</details>

6. Clicking "Create" on the `new` page above calls the `create` action on base_course_custom_content_versions (not implemented yet)

### Main Page
1. Only the CourseTemplate section shows the button to Edit
<details><summary>Screenshot</summary>
</details>

2. Clicking on the name of a Course or CourseTemplate from the main page shows a read-only view of everything. There is a button to Edit if it's a CourseTemplate.
<details><summary>Screenshot</summary>
</details>

3. If you URL hack the edit routes for a Course, an exception is thrown.

### Screenshots (if appropriate):

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
[...]

PR #413, opened 2020-10-30
Implement update and delete of a Project or Survey

sadleb:
This builds on the [Course Managment UI for Projects/Surveys](https://github.com/bebraven/platform/pull/412) to actually implement publishing the latest version of a Project or Survey and to be able to delete one from both Canvas and the local platform database.

### Checklist:


app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -57,16 +57,30 @@ def create
   def update
     authorize @base_course_custom_content_version
 
-    raise NotImplementedError, "TODO: save a new custom_content_version from the latest and re-associated this BaseCourseCustomContentVersion[#{@base_course_custom_content_version.inspect}] with it. Canvas assignment doesn't need to change."
+    custom_content = @base_course_custom_content_version.custom_content_version.parent
```
rshipp, line 60: i think this is clearer as `@base_course_custom_content_version.custom_content`

rshipp (approved):

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -57,16 +57,30 @@ def create
   def update
     authorize @base_course_custom_content_version
 
-    raise NotImplementedError, "TODO: save a new custom_content_version from the latest and re-associated this BaseCourseCustomContentVersion[#{@base_course_custom_content_version.inspect}] with it. Canvas assignment doesn't need to change."
+    custom_content = @base_course_custom_content_version.custom_content_version.custom_content
+    new_custom_content_version = custom_content.save_version!(current_user)
+    @base_course_custom_content_version.custom_content_version = new_custom_content_version
+    @base_course_custom_content_version.save!
```
jessouyangbraven, line 63: Do we need a transaction for this? I think it's okay if we have versions not tied to any courses/templates floating around, so maybe not?

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -57,16 +57,30 @@ def create
...
 
   # Deletes a Project, Survey, etc (aka CustomContent) from the Canvas course that this
   # BaseCourseCustomContentVersion join model represents and then deletes this record locally.
   def destroy
     authorize @base_course_custom_content_version
+    name = @base_course_custom_content_version.custom_content_version.title
 
-    raise NotImplementedError, "TODO: delete both the Canvas assignment and this BaseCourseCustomContentVersion[#{@base_course_custom_content_version.inspect}]"
+    CanvasAPI.client.delete_assignment(@base_course.canvas_course_id, @base_course_custom_content_version.canvas_assignment_id)
+    @base_course_custom_content_version.destroy
```
jessouyangbraven, line 78: Does this need to be in a transaction? It'd be bad to have Canvas/our DB out of sync because only part of this worked.

Does the API throw/return an error code that we need cam check?

lib/canvas_api.rb
```diff
@@ -247,6 +248,11 @@ def update_assignment_lti_launch_url(course_id, assignment_id, new_url)
     JSON.parse(response.body)
   end
 
+  def delete_assignment(course_id, assignment_id)
+    response = delete("/courses/#{course_id}/assignments/#{assignment_id}")
+    JSON.parse(response.body)
+  end
```
jessouyangbraven, line 254: Related to the controller's `#destroy` action comment: should this be `delete_assignment!` where we throw if the response code isn't success?

app/controllers/base_course_custom_content_versions_controller.rb
sadleb: It couldn't hurt if it were the model, but transactions are a model thing. I'd have to move this to the model to make it work and I don't think it's worth it since it's fine to have orphaned versions.

app/controllers/base_course_custom_content_versions_controller.rb
sadleb: https://app.asana.com/0/1174274412967132/1198984932600565

sadleb: I added a tech debt item. Not worth trying to figure out now imo.

lib/canvas_api.rb
sadleb: RestClient will throw exceptions if it's anything other than 2XX or 3XX codes. None of the methods in this lib class have a `!` to indicate they can throw. I think it's sort of implied that API calls can throw, but either way, not worth changing all the methods.

PR #414, opened 2020-10-30
Implement publishing a new Project to Canvas

sadleb:
This builds on the [Course Managment UI for Projects/Surveys](https://github.com/bebraven/platform/pull/412) to actually implement publishing a new Project to Canvas and to initialize the local platform database with the associated records. It removes the Project publishing dropdown from the LTI Assignment Selection placement UI since we'll be doing all publishing through the platform UI.

Getting this working for Survey's will be done in a follow-up PR. See the [Asana task](https://app.asana.com/0/1174274412967132/1198990873749397).

TODO: when we squash and merge, remove all the lines for commits that somehow show up here but where we rebased on top of them already merged so they shouldn't. E.g. everything before "First pass of working Project creation logic from Course Mgmt page"

### Checklist:


app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -9,47 +9,55 @@ class BaseCourseCustomContentVersionsController < ApplicationController
...
-
     # If we end up adding a designer role, remember to authorize `ProjectVersion.create?`.
     authorize BaseCourseCustomContentVersion
-    # Create new version of the content
-    @custom_content.save_version!(current_user)
 
-    # Create join table entry
+    # We need to create the Canvas assignment to get the ID in order to setup the join table
+    # here in platform
+    ca = CanvasAPI.client.create_lti_assignment(@base_course.canvas_course_id, @custom_content.title)
```
jessouyangbraven, line 33: I think we should be using @course_content_version.title here, so we'd have to do `BaseCourseCustomContentVersion.create!` before the `create_lti_assignment`.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -9,47 +9,55 @@ class BaseCourseCustomContentVersionsController < ApplicationController
...
+  
+    # Create a submission URL for this course and content version
 
-    # Create a submission URL for this course and version
     submission_url = new_base_course_custom_content_version_project_submission_url(
       base_course_custom_content_version_id: @course_content_version.id,
     )
-    @deep_link_return_url, @jwt_response = helpers.lti_deep_link_response_message(@lti_launch, submission_url)
+  
+    CanvasAPI.client.update_assignment_lti_launch_url(@base_course.canvas_course_id, ca['id'], submission_url)
```
jessouyangbraven, line 51: I'm guessing we do this in `update` rather than `create` because we need a Canvas assignment before we can add LTI launch things to it?

jessouyangbraven (approved):
A couple of inline comments, but nothing that would prevent merging.

app/controllers/base_course_custom_content_versions_controller.rb
sadleb: I did it in this order b/c I think it's more likely that the API call would break (aka Canvas is down or our token become invalid) than it would be for the local database save to fail. If the local DB save is failing, it'll be b/c of a bug within our control. So I think doing the thing most likely to fail will lead to less chances for inconsistency. Though I do try to clean things up if they fail.

app/controllers/base_course_custom_content_versions_controller.rb
sadleb: Yup. Catch22 type of situation. I want to add database constraints and validation to prevent saving a `BaseCourseCustomContentVersion` -- task: https://app.asana.com/0/1174274412967132/1198965066699365

PR #418, opened 2020-11-01
Enhance DryCrud::Controllers::Nestable to set parent instance variable

sadleb:
The original implementation only took care of filtering the list of models
shown by the index to be those who are associated with the parent_id. This
enhances DryCrud to set the instance variable. E.g. if we have the following
URL:

    https://platformweb/base_course_custom_content_version/1/project_submissions/new

then a `@base_course_custom_content_version` instance variable will automatically
be set with the parent BaseCourseCustomContentVersion[id=1]

Test Plan:
- bundle exec rspec

### Checklist:


spec/feature/lrs_xapi_proxy/project_xapi_spec.rb
```diff
@@ -70,7 +70,7 @@
     # "sends data to the LRS". 
     # It might be that the shared LRS and reusing the question_id is causing 
     # the value to get overwritten?
-    it "fetches the most recent data from the lrs", ci_exclude: true do
+    it "fetches the most recent data from the lrs - FLAKY", ci_exclude: true do
```
sadleb, line 73: Just sticking the word `FLAKY` in here so I don't have to go dig every time I run the full suite to know whether I broke something or this was one of the flaky tests excluded from CI.

jessouyangbraven (approved):
Thanks for doing this! This is something I run into too in controllers for nested resources.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -9,7 +9,7 @@ class BaseCourseCustomContentVersionsController < ApplicationController
   nested_resource_of BaseCourse
```
jessouyangbraven, line 9: Just to double-check my understanding, this `nested_resource_of` is what makes the `DryCrud` `:set_parent` stuff automatically work?

sadleb: Yup. It tells DryCrud which class (or list of classes) to look for the parent using `params[:<parent_class>_id]`

PR #419, opened 2020-11-02
Filter new Project list to only those not already on BaseCourse.

sadleb:
When publishing a new Project to a Canvas course using the Course Management page,
don't show already published Projects in the list of those to publish. See: https://app.asana.com/0/1174274412967132/1198965066699369

### Already published projects:
------

### List of projects to publish (excludes the above):
------

## Test Plan:
- `bundle exec rspec spec/controllers/base_course_custom_content_versions_controller.rb`

### Checklist:


app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -16,11 +16,8 @@ def new
     # If we end up adding a designer role, remember to authorize `ProjectVersion.create?`.
     authorize @base_course_custom_content_version
 
-   # TODO: if this is a Survey, show the survey's list.
-
-    # TODO: exclude those already on this BaseCourse.
-    # https://app.asana.com/0/1174274412967132/1198965066699369
-    @projects = Project.all
+    # TODO: if this is a Survey, show the survey's list.
+    @new_custom_contents = Project.all - @base_course.custom_contents
```
jessouyangbraven, line 20: A UI nit that we shouldn't do in this PR, but will it confuse people if a project name is missing? Should we keep the titles in the drop-down and disable them instead?

jessouyangbraven (approved):
LGTM!

app/controllers/base_course_custom_content_versions_controller.rb
sadleb: I didn't realize you could create a Project with no name. I think we should require a name if that's the case. But yeah, not for this PR.

PR #421, opened 2020-11-03
Add the ability to associate a Canvas rubric with a Project when publishing it

sadleb:
### Description

See: https://app.asana.com/0/1174274412967132/1191666190028013

We need the ability for a TA to be able to grade a Project using a Canvas rubric. This PR
adds the ability to choose an existing Canvas rubric that exists in the Course Template
(aka it was created using the Canvas UI on the Canvas course that the Course Template
points to) and associate it with a new Project being published.

Notes:
* Once a rubric is associated with an assignment in Canvas (aka with a Project)
  we don't allow new Projects to be published using that same rubric b/c then Designers cannot
  edit the rubric. A single rubric associated with two Canvas assignments disables that.
* When a new program is launched and the course is cloned, the rubrics are also cloned and
   new rubric_id's are mapped to the new Projects, so we're good there

#### One Project already associated with a rubric viewed as a TA in the Speedgrader:
image

### Types of changes

### Checklist:


jessouyangbraven (changes requested):
I know this fulfills the user story, but I feel like there are a lot of loose ends here.

1. There's no information about rubrics in our database. 
2. Rubrics aren't copied from a course template to a course.
3. There doesn't seem to be a way to tell that a rubric is associated with a project except in Canvas. 
4. There's no way to switch the rubric for a project, except for deleting the project and adding it again with a new rubric. 

These are in order of how much I think they should be part of this sprint. 

I feel pretty strongly about (1). Is it possible to at least add the `Rubric` model and table? I don't have a ton of context on rubrics, but I guessed it would looks something like: https://dbdiagram.io/d/5fa1a2253a78976d7b7a5eb0

I think (2) is something we should always keep in our minds as we add surveys/rubrics/attendance/peer evaluations and that we'd ideally implement at the same time. (It seems like less context switching since you know what needs to be copied in the data model at the time and it would also tell us immediately if we're missing data that's preventing us from cloning and mitigates not being able to clone a template after we add all this stuff.)

(3) and (4) I'm good to punt on because it's something we can work around with process. 

I'm happy to discuss more in today's stand-up and I'm happy to do the follow-up PRs since I'm familiar with this code from doing the survey stuff.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -86,7 +94,13 @@ def destroy
     end
   end
 
-  private
+  def for_project?
```
rshipp, line 97: to me this is less clear than just using `custom_content.is_a? Project`, since I have to look up this definition to see what the function does

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -96,6 +110,25 @@ def set_custom_content
     @custom_content = CustomContent.find(params[:custom_content_id])
   end
 
+  def set_custom_contents
+    @new_custom_contents = custom_content_class.all - @base_course.custom_contents
```
rshipp, line 114: can you make the method name and the variable name match, just for consistency?

sadleb:
> I feel pretty strongly about (1). Is it possible to at least add the `Rubric` model and table? I don't have a ton of context on rubrics, but I guessed it would looks something like: https://dbdiagram.io/d/5fa1a2253a78976d7b7a5eb0

We have already modeled out a decent portion of rubrics in the platform code before we had actually implemented anything and now I'm starting to think it was all wasted work. See: https://dbdiagram.io/d/5e44e8749e76504e0ef1636a and `ls app/models/rubric*` for what we had/have. Not storing the rubric info locally in this sprint for this user story is taking the lesson learned there and only implementing something when we need it. Paragraphs two and three of the user story linked at the beginning of the PR link to the followup's to do what you're saying, if needed, which is still unclear at the moment @jessouyangbraven

rshipp:
I'm having trouble wrapping my head around the whole rubric flow, so I'd appreciate talking this over in standup too. 

For point 2 @jessouyangbraven, as I understand it - rubrics are copied from a course template to a course, because the content migration on the Canvas side does copy rubrics and assign the new ones to the new assignments.

app/controllers/base_course_custom_content_versions_controller.rb
sadleb: It's used in the `new` view so we don't have a `custom_content` yet. We need to find out from the `type` if we're working on a new Project or Survey. I think @jessouyangbraven opened a task to try and clean up the polymorphic stuff a bit here. E.g. we could add a `BaseCourseProjectVersion` and `BaseCourseSurveyVersion` model (not database backed) and route things properly to be able to tell what we're working with from the `new` view.

sadleb: Also, it's used here `before_action :set_rubrics, only: [:new], if: :for_project?` and I think it needs to be a method passed to the `if:` parameter of `before_action`

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -96,6 +110,25 @@ def set_custom_content
...
+  def rubrics_already_associated_with_a_project
+
+    # TODO: a better way to do this will be to store rubric_ids on the join model, but that's more work
+    # and we need to handle course clone, so this is a hack for now.  Task to do this properly: 
+    # https://app.asana.com/0/1174274412967132/1198996949946468 
+    CanvasAPI.client.get_assignments(@base_course.canvas_course_id).map do |ca| 
+      assoc_rubric = ca['rubric_settings']
+      CanvasAPI::LMSRubric.new(assoc_rubric['id'], assoc_rubric['title']) if assoc_rubric
+    end
+  end
```
jessouyangbraven, line 130: Can we put the filtering in `CanvasAPI` by adding a flag to `get_rubrics(filter_already_attached: false)`? Or a separate function `get_available_rubrics()`?

sadleb: I like that idea. I'll make a change.

For posterity, we discussed the rest of the above comments in standup and we're good with merging it after this change and removing some of the TODOs (except this one b/c the hack) b/c we're tracking it in Asana and it clutters the code.

PR #427, opened 2020-11-04
Stop nesting base_course_custom_content_versions under base_courses

jessouyangbraven:
### Description

We made `base_course_custom_content_version` used to be a route nested under `base_course` when building Course Management functionality because we were always creating a `base_course_custom_content_version` from the context of a `CourseTemplate`.

### Motivation and Context

The routes were confusing and not necessary, since we can pass the necessary data as query parameters or form data.

### Test

- rspec
- Load and click around in the Course Template editing view to make sure existing functionality works:
Screen Shot 2020-11-03 at 4 11 55 PM

### Types of changes

### Checklist:


app/views/base_courses/_form.html.erb
```diff
@@ -40,13 +40,23 @@
...
     </div>
   </div>
+<% end %>
 
-  <br/>
-  <h3>Edit Projects</h3>
-  <%= link_to 'Add Project', new_base_course_base_course_custom_content_version_path(base_course, type: 'Project'), class: 'btn btn-sm btn-outline-secondary' %>
-  <%= render :partial => 'base_course_custom_content_versions/list', locals: {base_course_custom_content_versions: base_course.base_course_project_versions} %>
+<h3>Edit Projects</h3>
+<%= link_to 'Add Project', new_base_course_custom_content_version_path(base_course_id: base_course.id, type: 'Project'), class: 'btn btn-sm btn-outline-secondary' %>
```
jessouyangbraven, line 46: In un-nested version of `new...path()`, we explicitly pass `base_course_id` as a query parameter now instead of having  `new_base_course_base_course_custom_content_version_path(base_course,...` do it automatically.

app/views/base_course_custom_content_versions/new.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="col-sm-4">
 
       <h4>Select a <%= params[:type] %> to create in Canvas</h4>
-      <%= form_with url: base_course_base_course_custom_content_versions_path do |form| %>
+      <%= form_with url: base_course_custom_content_versions_path do |form| %>
```
jessouyangbraven, line 6: We didn't need the nested path to the `#create` method even before this PR, since we're passing the `base_course_id` and `custom_content_id` are inputs that get submitted with the form.

app/views/base_course_custom_content_versions/_list.html.erb
```diff
@@ -14,8 +14,8 @@
             <td><%= link_to bcccv.custom_content_version.title, bcccv.canvas_url %></td>
             <td class="text-right smaller">
               <% unless local_assigns[:read_only]  %>
-                <%= link_to 'Publish Latest', base_course_base_course_custom_content_version_path(bcccv.base_course_id, bcccv.id), method: :put %> |
-                <%= link_to 'Delete', base_course_base_course_custom_content_version_path(bcccv.base_course_id, bcccv.id), method: :delete,
+                <%= link_to 'Publish Latest', base_course_custom_content_version_path(bcccv.id), method: :put %> |
+                <%= link_to 'Delete', base_course_custom_content_version_path(bcccv.id), method: :delete,
```
jessouyangbraven, line 18: We didn't need the nested paths to `#update` and `#delete` because we need a pre-existing `bcccv` object to do those actions, so passing its ID suffices.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -1,20 +1,17 @@
...
 class BaseCourseCustomContentVersionsController < ApplicationController
-  include DryCrud::Controllers::Nestable
   include LtiHelper
 
   layout 'admin'
 
-  nested_resource_of BaseCourse
-
-  before_action :set_custom_content, only: [:create]
+  before_action :set_model_instances, except: [:new]
```
jessouyangbraven, line 8: This is necessary because I removed the `Nestable` `nested_resource_of` that automatically set `@base_course` for us.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -96,27 +92,32 @@ def for_project?
...
-    @rubrics = CanvasAPI.client.get_rubrics(@base_course.canvas_course_id, filter_already_associated_rubrics)
-  end
-
-  def verify_can_edit!
-    @base_course.verify_can_edit!
+    @base_course = BaseCourse.find(params[:base_course_id])
+    @rubrics = CanvasAPI.client.get_rubrics(
+      @base_course.canvas_course_id,
+      true, # filter_already_associated_rubrics
+    )
```
jessouyangbraven, line 121: These are `before_action`s specifically for `#new`. 

Currently, we always create a BCCCV in the context of a `CourseTemplate` and this controller relies pretty heavily on that assumption for the `#new` action.  So we can currently expect that `params[:base_course_id]` is going to be set. I could add a `params.require()` so it fails more obviously, but it'll fails on `.find(nil)` too, so I wasn't sure which is preferable. 

I don't know how much flexibility we should support on the `#new` action in general for this controller.
- :base_course_id - currently supported, I'm going to add a new 
- :custom_content_id - not supported, the custom content would be fixed, you get to select the course template from a drop down (if we ever wanted to support Publish from the content editor for example)
-  neither - also not supported, we would show two drop downs and you get to decide which template and which content (it'd be hard to filter the drop-downs dynamically, so we'd have to make sure that #create or the model checks that we don't add a customcontent to a template twice)

rshipp (approved):
tentatively approving because I don't see anything that stands out as bad but i don't understand all of

jessouyangbraven:
> tentatively approving because I don't see anything that stands out as bad but i don't understand all of

Let's talk about it in the standup. This PR only makes sense in the context of making things more intuitive to developers, so I'd like to get feedback on what's confusing here!

config/routes.rb
```diff
@@ -39,10 +39,9 @@
   resources :course_templates, controller: 'base_courses', type: 'CourseTemplate'
 
   resources :base_courses, only: [:index], path: 'course_management' do
-    resources :base_course_custom_content_versions, only: [:new, :create, :update, :destroy]
   end
 
-  resources :base_course_custom_content_versions, only: [] do
```
jessouyangbraven, line 45: Put back `:new` under `:base_courses`. Add comment on why this is nested.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -1,17 +1,13 @@
...
 
   layout 'admin'
 
-  nested_resource_of BaseCourse
-
-  before_action :set_custom_content, only: [:create]
+  before_action :set_model_instances, except: [:new]
   before_action :set_new_custom_contents, only: [:new]
   before_action :set_rubrics, only: [:new], if: :for_project?
-  before_action :verify_can_edit!
```
sadleb, line 14: I personally like this as a separate `before_action` but I'm fine either way.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -64,13 +60,12 @@ def create
   def update
     authorize @base_course_custom_content_version
 
-    custom_content = @base_course_custom_content_version.custom_content_version.custom_content
-    new_custom_content_version = custom_content.save_version!(current_user)
-    @base_course_custom_content_version.custom_content_version = new_custom_content_version
-    @base_course_custom_content_version.save!
+    @base_course_custom_content_version.update!(
```
sadleb, line 63: Nice, this is cleaner. Generally, ruby encourages lots of small methods and stringing them together to make things readable.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -96,27 +91,32 @@ def for_project?
   end
   helper_method :for_project?
 
-private
-
+  private
```
sadleb, line 94: I feel like we're wrestling with moving the indentation and line break of the `private` between these two options. Do you think it's more readable this way?

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -96,27 +91,32 @@ def for_project?
...
   def custom_content_class
     CustomContentsController.class_from_type(params[:type])
   end
 
-  def set_custom_content
-    params.require(:custom_content_id)
-    @custom_content = CustomContent.find(params[:custom_content_id])
+  def set_model_instances
```
sadleb, line 99: The other option would be to have `set_base_course` and `set_custom_content` with each being responsible for something like
```
def set_base_course
  @base_course = @base_course_custom_content_version&.base_course
  @base_course = BaseCourse.find(params[:base_course_id]) if params[:base_course_id]
end
```
so that the controller would have
```
before_action :set_base_course
before_action :set_custom_content
before_action :verify_can_edit!
```
Which is more readable / maintainable? Separate or a big `set_model_instances` that does all the logic?

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -96,27 +91,32 @@ def for_project?
...
+      @custom_content = @base_course_custom_content_version.custom_content_version.custom_content
+    else
+      @base_course = BaseCourse.find(params[:base_course_id])
+      @custom_content = CustomContent.find(params[:custom_content_id])
+    end
+    @base_course.verify_can_edit!
   end
 
   def set_new_custom_contents
+    @base_course = BaseCourse.find(params[:base_course_id])
```
sadleb, line 111: If we separate out the `set_model_instances` you wouldn't have to call this here.

sadleb (changes requested):
I'm thinking it's cleaner to split out the logic into separate before actions. Thoughts?

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -96,27 +91,32 @@ def for_project?
...
   end
 
   def set_rubrics
-    filter_already_associated_rubrics = true
-    @rubrics = CanvasAPI.client.get_rubrics(@base_course.canvas_course_id, filter_already_associated_rubrics)
-  end
-
-  def verify_can_edit!
-    @base_course.verify_can_edit!
+    @base_course = BaseCourse.find(params[:base_course_id])
```
sadleb, line 116: Ditto on the above about splitting up `set_model_instances`

app/controllers/base_course_custom_content_versions_controller.rb
jessouyangbraven: I generally find the `private` rules in Ruby confusing and haven't found any style convention that makes it clear what's going on.

I was under the impression that everything after `private` was made private. So the line break makes sense to me, because it doesn't just apply to the one `def`. We could indent everything under the private over one, but that might be overkill? 

I'm agnostic; I'm happy to do whatever so long as we're consistent.

app/controllers/base_course_custom_content_versions_controller.rb
jessouyangbraven: Good call on splitting up `set_model_instances`.

app/controllers/base_course_custom_content_versions_controller.rb
jessouyangbraven: Sorry! Missed this one because I was scanning the right-hand column. I'm okay either way. I'll either put it back in `verify_can_edit!` on its own or change the other action to `:set_base_course!` since it can throw now. Will update PR after I get through some assignment submissions stuff.

app/controllers/base_course_custom_content_versions_controller.rb
sadleb: My vote is for sticking with no indentation, aka it applies to the whole class, and leaving a line break. So what it was before this diff. We should prob add or edit the codacy linter rules but I’m not inclined to take that on at the moment and have admittedly been very lax about checking codacy results over this past month or two.

sadleb (approved):
Lgtm!

app/controllers/base_course_custom_content_versions_controller.rb
jessouyangbraven: Re: Codacy, @rshipp mentioned it's been down for a while. Which explains why I haven't gotten those emails in a while.

PR #428, opened 2020-11-04
Fix submission URL in launching course

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199090006665741

This fixes the submission URL that gets copied from a course template's to a course's assignment to work for both projects and surveys.

### Motivation and Context

To launch a course, we're cloning from a course template. This requires us to copy all over the canvas assignments, and add a new `base_course_custom_content_version` for the course and the custom content version. This means the submissions URL for the assignment has to be updated, too. `base_course_custom_content_version` covers both survey and project submissions but these submissions are handled by different controllers, so we need to use a polymorphic URL to distinguish between them.

### Test

- rspec
- `grep` for `project_submissions_{url, path}`

### Types of changes

### Checklist:


jessouyangbraven:
From stand-up: don't manually test; update launch program spec to test both project and survey URLs.

app/models/base_course_custom_content_version.rb
```diff
@@ -15,6 +15,15 @@ class BaseCourseCustomContentVersion < ApplicationRecord
...
 
+  def submission_type
+    case custom_content_version.type
+    when 'ProjectVersion'
+      ProjectSubmission
+    when 'SurveyVersion'
+      SurveySubmission
+    end
+  end
+
```
jessouyangbraven, line 26: I needed `"#{@custom_content.class}Submission".constantize.new` in a multiple places, so I decided to refactor some of it into `BaseCourseCustomContentVersion`, since that's what submissions are tied ot.

app/services/initialize_new_course.rb
```diff
@@ -100,11 +102,13 @@ def create_new_base_course_custom_content_version(old_course_custom_content_vers
...
-      canvas_assignment_id: canvas_assignment_id
+      canvas_assignment_id: canvas_assignment_id,
     )
 
-    new_launch_url = Rails.application.routes.url_helpers
-      .new_base_course_custom_content_version_project_submission_url(new_cccv, :protocol => 'https')
+    new_launch_url = new_polymorphic_url(
+      [new_cccv, new_cccv.submission_type.new],
+      protocol: 'https',
+    )
```
jessouyangbraven, line 111: And now that I'm looking at this, maybe we should have `BaseCourseCustomContentVersion.new_submission_url()` so we don't accidentally forget to set `protocol` on one of these calls.

spec/services/launch_program_spec.rb
```diff
@@ -3,6 +3,9 @@
 require 'rails_helper'
 
 RSpec.describe LaunchProgram do
+
+  include Rails.application.routes.url_helpers
```
jessouyangbraven, line 7: This is unfortunate and all over the place now. We could centralize it in the BCCCV model if we went the `BCCCV.new_submission_url()` route.

spec/services/launch_program_spec.rb
```diff
@@ -63,6 +66,53 @@
     end
   end
 
+  shared_examples 'template assignment is copied to course assignment' do
```
jessouyangbraven, line 69: We do the same test for both surveys and projects, and I didn't want to copy-pasta the code. `shared_example` and `scenario` is what this https://www.netguru.com/codestories/eliminate-duplicated-code-rspec-capybara told me to do.

spec/services/launch_program_spec.rb
```diff
@@ -126,24 +176,23 @@
...
       end
 
+      context 'projects' do
+        let(:course_template_custom_content_version) { create(
+          :course_project_version,
+          base_course: fellow_course_template,
+        ) }
+
+        it_behaves_like 'template assignment is copied to course assignment'
+      end
```
jessouyangbraven, line 195: The test is the same, it's just the custom content version type that's different for projects and surveys.

jessouyangbraven (commented):
Updating the spec took a bit more refactoring than I anticipated, and I could do a bit more as mentioned in the comments. But it's a good thing we added a spec because my innocuous one-line change didn't actually run.

spec/services/launch_program_spec.rb
```diff
@@ -63,6 +63,41 @@
     end
   end
 
+  shared_examples 'template assignment is copied to course assignment' do
+    scenario 'it updates the LTI launch/submission URL' do
+      template_url = course_template_custom_content_version.new_submission_url
+
+      # This mimics an assignment that was cloned to the new course with the old launch URL for the template
+      course_assignment = build(
```
sadleb, line 71: Nit: `canvas_assignment` ? When I see `course` in the code I think of our local `BaseCourse` and `Course` vs the Canvas Assignment object from their API which is what this is.

spec/services/launch_program_spec.rb
```diff
@@ -63,6 +63,41 @@
...
+      allow(canvas_client)
+        .to receive(:get_assignments)
+        .and_return([course_assignment])
+
+      launch_program.run
+
+      course_custom_content_version = BaseCourseCustomContentVersion.find_by(
+        canvas_assignment_id: course_assignment['id'],
+      )
+      course_url = course_custom_content_version.new_submission_url
```
sadleb, line 86: `submission_url` ?

spec/services/launch_program_spec.rb
```diff
@@ -63,6 +63,41 @@
...
   end
 
+  shared_examples 'template assignment is copied to course assignment' do
+    scenario 'it updates the LTI launch/submission URL' do
+      template_url = course_template_custom_content_version.new_submission_url
+
+      # This mimics an assignment that was cloned to the new course with the old launch URL for the template
+      course_assignment = build(
+        :canvas_assignment,
+        course_id: new_fellow_course_id,
```
sadleb, line 73: This (and the LC one) should prob be called `new_fellow_canvas_course_id` but that was my bad naming from before.

spec/services/launch_program_spec.rb
sadleb: This is uncharted territory for me. I like it!

sadleb (approved):
Small nits, but lgtm.

PR #430, opened 2020-11-05
Slack and Zoom automation for Accelerator & Booster

Rowantatai:
**Related Asana Tasks:** 
Zoom URL logistics should be built into tools running in the cloud (not laptop) https://app.asana.com/0/1179330853755533/1196450538044072/f
Slack logistics should be built into tools running in the cloud (not laptop) https://app.asana.com/0/1179330853755533/1196450538044074/f
PLUS automations around Zoom & Slack management (i.e., Adding/Removing)

**Overview on what will now be accessible on platform:**
Generating unique Zoom (webinar) links for all program participants & canceling webinar registration for all drops 
Generate unique Zoom (webinar) links for anyone via a csv
For Accelerator, slack automation to create cohort-specific channels & add/move/remove slack users through cohort-specific channels
For Booster, accessibility for staff to add new Booster invites to cohorts


Gemfile
```diff
@@ -110,7 +110,7 @@ gem 'platform-api', require: false
 # To develop locally, you can mount your local rowan_bot code as a volume in docker-compose.yml
 # and then change the following to this:
 # gem 'rowan_bot', path: '/app/rowan_bot'
-gem 'rowan_bot', github: 'bebraven/rowan_bot', ref: 'b078c0f'
+gem 'rowan_bot', github: 'bebraven/rowan_bot', ref: '60a6f6f'
```
hcli-work, line 113: seems dangerous/clunky. we should get this code moved into platform instead of relying on a separate gem outside of our normal review process

app/controllers/webinar_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# Webinar controller
```
hcli-work, line 3: this comment is unnecessary.

app/controllers/webinar_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+class WebinarController < ApplicationController
+  layout 'admin'
+
+  def init_sync_to_webinar
```
hcli-work, line 7: what does this do? what is this for?

app/controllers/webinar_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    authorize :application, :index?
```
hcli-work, line 8: consider adding a before_action?

```
before_action :authorize_index
...

private
  def authorize_index
    authorize :application, :index?
  end
```

app/views/home/welcome.html.erb
```diff
@@ -7,5 +7,7 @@
   <%= link_to 'Course Management', base_courses_path, type: "button", class: "list-group-item list-group-item-action" %>
   <%= link_to 'Sync To LMS', sync_to_lms_path, type: "button", class: "list-group-item list-group-item-action" %>
   <%= link_to 'Sync To Join', sync_to_join_path, type: "button", class: "list-group-item list-group-item-action" %>
+  <%= link_to 'Sync To Slack', sync_to_join_path, type: "button", class: "list-group-item list-group-item-action" %>
```
hcli-work, line 10: in this list of links, there's a lot that feels repeated. the only one that's different is the Portal link. can we DRY this up using some embedded Ruby and iterating over a Hash from title => path ?

app/views/layouts/admin.html.erb
```diff
@@ -24,7 +24,7 @@
       </div>
     <% end %>
     <%= render 'layouts/admin_navbar' %>
-    <h2><%= content_for?(:headline) ? yield(:headline) : '' %></h2>
+    <h2 class="ml-4"><%= content_for?(:headline) ? yield(:headline) : '' %></h2>
```
hcli-work, line 27: wtf?

app/views/webinar/init_sync_to_webinar.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <ol class="breadcrumb">
+    <li class="breadcrumb-item active" aria-current="page">Update Webinars</li>
+  </ol>
+</nav>
+
+<div class="container">
+  <div class="row">
+    <div class="col-sm">
+      <legend>Update Program Participants Webinars</legend>
+      <%= render 'update_webinar' %>
```
hcli-work, line 11: this feels like prematurely DRYing up the code. why not just insert the form directly here?

app/controllers/webinar_controller.rb
Rowantatai: Fancy editors don't like that

Gemfile
Rowantatai: We'll do that in the future

app/controllers/webinar_controller.rb
Rowantatai: Done

app/controllers/webinar_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def sync_to_webinar
+    authorize :application, :index?
+    program_id = params[:program_id]
+    email = params[:email]
+    force_update = params[:force_update].eql?("1")
+    SyncToWebinarJob.perform_later(program_id, email, force_update)
```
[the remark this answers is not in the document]
Rowantatai: For readability purposes, leaving as is

app/jobs/sync_to_webinar_job.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Update webinar job
```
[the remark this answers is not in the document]
Rowantatai: My linter doesn't like classes without comments

app/controllers/slack_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def sync_to_booster_slack
+    redirect_to root_path, notice: 'Nothing happened! Run this on booster platform instead' and return unless booster_instance?
+
+    email = params[:email]
+    emails = params[:emails]
+
+    SyncToBoosterSlackJob.perform_later(emails, email)
+
+    redirect_to root_path, notice: 'The sync process was started. Watch out for an email'
```
hcli-work, line 21: to me, this feels hard to understand. that first line is hard to parse. how about
```
the_notice = nil
if booster_instance?
  SyncToBoosterSlackJob.perform_later(...)
  the_notice = 'The sync process was started. Watch out for an email'
else
  the_notice = 'Nothing happened! Run this on booster platform instead'
end

redirect_to_root_path(notice: the_notice)
```

app/controllers/webinar_controller.rb
hcli-work: your editor is too fancy then. you need to dumb it down.

app/controllers/webinar_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    participants = CSV.read(participants_file.path, headers: true).map(&:to_h)
+    GenerateWebinarJob.perform_later(meeting_id, email, participants)
+
+    redirect_to root_path, notice: 'The generation process was started. Watch out for an email'
+  end
+
+  def sync_to_webinar
+    program_id = params[:program_id]
+    email = params[:email]
+    force_update = params[:force_update].eql?('1')
```
hcli-work, line 25: we have different definitions of readability

app/jobs/sync_to_webinar_job.rb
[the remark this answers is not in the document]
hcli-work: turn off your linter then

app/mailers/generate_webinar_mailer.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def failure_email
+    mail(to: recipient, subject: 'Webinar Generation Failed')
+  end
+
+  private
+
+  def recipient
+    params[:email]
+  end
```
hcli-work, line 18: you could have done this trick in the SlackController to not write the redundant line `email = params[:email]`

app/services/generate_webinar_for_participants.rb
```diff
@@ -0,0 +1,24 @@
...
+    @meeting_id = meeting_id
+    @participants = participants
+  end
+
+  def run
+    tasks = RowanBot::Tasks.new(RowanBot::ZoomAPI.new)
+    registrants = tasks.add_participants_to_meetings(meeting_id, participants)
+    CSV.generate do |csv|
+      csv << registrants.first.keys
+      registrants.each { |registrant| csv << registrant.values }
```
hcli-work, line 17: I don't know what `csv` is, but could you make this more Ruby-ish, by doing
`csv << registrants.map(&:values).flatten`

app/services/sync_booster_slack_for_emails.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def run
+    tasks = RowanBot::Tasks.new
+    salesforce_api = RowanBot::SalesforceAPI.new
+    tasks.salesforce_api = salesforce_api
+    tasks.slack_api = RowanBot::SlackAPI.new
+
+    admin_emails = ENV.fetch('SLACK_ADMIN_EMAILS', '').split(',').map(&:strip)
+    emails_a = emails.split(',').map(&:strip)
```
hcli-work, line 17: I would suggest making these private methods so that this split, map, strip stuff is kept out of the body of the `run` method for clarity

app/services/sync_slack_for_program.rb
```diff
@@ -0,0 +1,27 @@
...
+    tasks = RowanBot::Tasks.new
+    salesforce_api = RowanBot::SalesforceAPI.new
+    tasks.salesforce_api = salesforce_api
+    program = salesforce_api.find_program_by_id(sf_program_id)
+    tasks.slack_api = RowanBot::SlackAPI.new({ slack_url: program.slack_url,
+                                               slack_token: program.slack_token,
+                                               slack_user: program.slack_user,
+                                               slack_password: program.slack_password,
+                                               slack_admin_emails: program.slack_admin_emails })
+    slack_admins = program.slack_admin_emails.split(',').map { |email| { email: email.strip } }
```
hcli-work, line 20: change to `map(&:strip)` for consistency

app/services/sync_slack_for_program.rb
```diff
@@ -0,0 +1,27 @@
...
+  def run
+    tasks = RowanBot::Tasks.new
+    salesforce_api = RowanBot::SalesforceAPI.new
+    tasks.salesforce_api = salesforce_api
+    program = salesforce_api.find_program_by_id(sf_program_id)
+    tasks.slack_api = RowanBot::SlackAPI.new({ slack_url: program.slack_url,
+                                               slack_token: program.slack_token,
+                                               slack_user: program.slack_user,
+                                               slack_password: program.slack_password,
+                                               slack_admin_emails: program.slack_admin_emails })
```
hcli-work, line 19: Rowan, this would be an example of something that is so DRY to the point of being unreadable. it is fewer lines of code and is easy to add stuff to, but....ouch.
```
names = [:slack_url, :slack_token, :slack_user, :slack_password, :slack_admin_emails]
args = names.reduce( Hash.new, { |hsh, name| hsh[name] = program.send(name); hsh }
tasks.slack_api = RowanBot::SlackAPI.new(args)
```

app/views/slack/init_sync_to_booster_slack.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <ol class="breadcrumb">
+    <li class="breadcrumb-item active" aria-current="page">Update Slack</li>
+  </ol>
+</nav>
+
+<div class="container">
+  <div class="row">
+    <div class="col-sm">
+      <legend>Update Booster Program Participants to Cohort Channel on Slack</legend>
+      <%= form_with(url: sync_to_booster_slack_path, method: "post", local: true) do |form| %>
```
hcli-work, line 11: use of parens across this code needs to be more consistent

hcli-work (approved):
I'm okay with this so that you can push. However, I don't want the feedback ignored and spread over several weeks to be dealt with. You should work on the feedback immediately after you merge this.

app/controllers/webinar_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+# Webinar controller
+class WebinarController < ApplicationController
```
sadleb, line 6: Ditto on disabling `DryCrud`. There is no model so this is non-standard.

app/jobs/sync_to_booster_slack_job.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# Update booster slack job
+class SyncToBoosterSlackJob < ApplicationJob
+  queue_as :default
+
+  def perform(emails, email)
+    SyncBoosterSlackForEmails.new(emails: emails).run
+    SyncSalesforceToLmsMailer.with(email: email).success_email.deliver_now
```
sadleb, line 9: `Lms` is Canvas. This isn't sync'ing to Canvas. You should create a new mailer specifically for this sync job called something like `SyncSalesforceToSlackMailer` or if it doesn't make sense to be able to tailor the messages in each of these notification emails to the actual sync job being run, then rename `SyncSalesforceToLmsMailer` to be more generic, like `BackgroundSyncJobMailer`

app/controllers/webinar_controller.rb
sadleb, line 25: Checking against `1` is an old C or C++ thing. In Rails, don't use 0 vs 1 to mean true vs false. Either use true or false directly or in this case use something like `params[:force_update].present?`

app/services/sync_webinar_links_for_program.rb
```diff
@@ -0,0 +1,22 @@
...
+    tasks = RowanBot::Tasks.new
+    tasks.salesforce_api = RowanBot::SalesforceAPI.new
+    tasks.zoom_api = RowanBot::ZoomAPI.new
+
+    tasks.sync_zoom_links_for_program(sf_program_id, force_update)
+  end
+
+  private
+
+  attr_reader :sf_program_id, :force_update
```
sadleb, line 21: What does these do? Can you remove them? If we need them, them move to the top and don't make them private. `attr_reader`'s are supposed to be public (or maybe protected, but mostly public)

app/services/generate_webinar_for_participants.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+  end
+
+  private
+
+  attr_reader :meeting_id, :participants
```
sadleb, line 23: Ditto, do you need this? If so, make public.

app/services/sync_booster_slack_for_emails.rb
```diff
@@ -0,0 +1,26 @@
...
+    @emails = emails
+  end
+
+  def run
+    tasks = RowanBot::Tasks.new
+    salesforce_api = RowanBot::SalesforceAPI.new
+    tasks.salesforce_api = salesforce_api
+    tasks.slack_api = RowanBot::SlackAPI.new
+
+    admin_emails = ENV.fetch('SLACK_ADMIN_EMAILS', '').split(',').map(&:strip)
```
sadleb, line 16: You should sync up with @hcli-work and @skaex about writing specs. If this stuff is not going into Highlander, I'm fine with not writing specs, but if we'll move this into the Highlander codebase, we need specs before we merge.

app/services/sync_booster_slack_for_emails.rb
```diff
@@ -0,0 +1,26 @@
...
+    emails_a = emails.split(',').map(&:strip)
+    salesforce_api.find_participants_by_emails(emails_a, 'Booster_Student')
```
sadleb, line 18: I think the `salesforce_api` should be responsible for knowing and managing the `Booster_Student` string, not this sync service.  I would make a `find_booster_participants_by_emails` method.

app/services/sync_booster_slack_for_emails.rb
```diff
@@ -0,0 +1,26 @@
...
+    tasks.assign_to_booster_run_channels_in_slack(emails_a, admin_emails)
+    tasks.assign_to_booster_peer_group_channel_in_slack(emails_a, admin_emails)
+  end
+
+  private
+
+  attr_reader :emails
```
sadleb, line 25: Another place of confusion for me about where these are used and why they're private.

app/views/slack/init_sync_to_booster_slack.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        <div class="form-group">
+          <%= form.label :emails, "Emails:" %>
+          <%= form.text_area :emails, class: "form-control", required: true, placeholder: "you@example.com, me@example.com", autocomplete: "emails" %>
+        </div>
+
+        <div class="form-group">
+          <%= form.label :email, "Email:" %>
```
sadleb, line 18: Can we label this `Email to notify`?

app/views/slack/init_sync_to_booster_slack.html.erb
```diff
@@ -0,0 +1,29 @@
...
+</nav>
+
+<div class="container">
+  <div class="row">
+    <div class="col-sm">
+      <legend>Update Booster Program Participants to Cohort Channel on Slack</legend>
+      <%= form_with(url: sync_to_booster_slack_path, method: "post", local: true) do |form| %>
+        <div class="form-group">
+          <%= form.label :emails, "Emails:" %>
+          <%= form.text_area :emails, class: "form-control", required: true, placeholder: "you@example.com, me@example.com", autocomplete: "emails" %>
```
sadleb, line 14: `autocomplete: "emails"` - what does this do?

app/views/slack/init_sync_to_booster_slack.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        </div>
+
+        <div class="form-group">
+          <%= form.label :email, "Email:" %>
+          <%= form.email_field :email, class: "form-control", required: true, placeholder: "you@example.com", autocomplete: "email" %>
```
sadleb, line 19: Ditto. What does `autocomplete: email` do?

sadleb:
Can you edit the PR name and description to give more info? Maybe "Slack and Zoom automation for Booster" as the title? And then link to the Asana tasks in the description as well as type out an overview of what is going on here. Screenshots would be helpful in the PR description as well.

Gemfile
Rowantatai: https://app.asana.com/0/1179330853755533/1199129462241589/f

app/views/home/welcome.html.erb
Rowantatai: Will address next week https://app.asana.com/0/1179330853755533/1199129462241591/f

app/views/layouts/admin.html.erb
Rowantatai: It's to align the header (ex: "Dashboard") with the rest of the layout on platform

app/controllers/slack_controller.rb
Rowantatai: Done

app/mailers/generate_webinar_mailer.rb
Rowantatai: Noted for the future

app/controllers/slack_controller.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+# Slack controller
```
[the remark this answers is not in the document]
Rowantatai: Got it. Working on creating a `NonStandardController` class that `SlackController` will inherit from. I think it'll make it clear and DRY up other cases. Thoughts?

app/controllers/slack_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    @show_booster_slack_sync = booster_instance?
+  end
+
+  def sync_to_booster_slack
+    redirect_to root_path, notice: 'Nothing happened! Run this on booster platform instead' and return unless booster_instance?
+
+    email = params[:email]
+    emails = params[:emails]
+
+    SyncToBoosterSlackJob.perform_later(emails, email)
```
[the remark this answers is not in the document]
Rowantatai: Great recommendation!

app/controllers/webinar_controller.rb
Rowantatai: What about `!!params[:force_update]`?

app/jobs/sync_to_booster_slack_job.rb
Rowantatai: Agreed!

app/services/generate_webinar_for_participants.rb
Rowantatai: Gotcha -- looks like `csv << registrants.map(&:values)` is what need

app/services/generate_webinar_for_participants.rb
Rowantatai: We just need the attributes readers within the class

app/services/sync_booster_slack_for_emails.rb
Rowantatai: Gothca & done

app/services/sync_booster_slack_for_emails.rb
Rowantatai: We only needed for this class

app/services/sync_slack_for_program.rb
Rowantatai: The final output we want here is a hash and not a string

app/services/sync_webinar_links_for_program.rb
Rowantatai: These are private. We only need them in this class.

app/controllers/slack_controller.rb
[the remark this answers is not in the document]
sadleb: I wouldn’t do that. I have a task queued up somewhere on a backlog to change DryCrud itself so that it detects when there is no model and short circuits it’s logic instead of having to explicitly disable it. When I originally implemented it, it was taking too much time to figure out so I just stuck that in as a temporary way to opt out.

app/jobs/sync_to_booster_slack_job.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  rescue_from(StandardError) do |exception|
+    SyncSalesforceToLmsMailer.with(email: arguments.second).failure_email.deliver_now
+    raise exception
```
sadleb, line 14: You should just call `raise` here. If you raise the exception I think it messes up the stacktrace and makes it harder to debug.

app/controllers/webinar_controller.rb
sadleb: Ahh, I see. The default values for a checkbox are 0 and 1. I think if you change it to something like the following in the view, true and false will come through: `<%= form.check_box :force_update, {class: "form-check-input", id: "force_update_check"}, true, false %>`

sadleb: The method definition for a checkbox is: `check_box(method, options = {}, checked_value = "1", unchecked_value = "0")` - notice how those last two parameters have a default of 1 and 0 if you don’t specify them

app/services/sync_booster_slack_for_emails.rb
Rowantatai: A [follow up task](https://app.asana.com/0/1179330853755533/1199160539824581/f) has been created for when/if that day comes

app/controllers/webinar_controller.rb
Rowantatai: I see...we're trying to return a boolean, and when testing the recommendation it returns `true` String  (for checked) and ` `  NilClass (for unchecked)

Rowantatai: Update -- all resolved now.

app/views/slack/init_sync_to_booster_slack.html.erb
Rowantatai: Improves user experience. When a user goes to this form, it'll provide suggestions of what they've filled before.

app/views/slack/init_sync_to_booster_slack.html.erb
Rowantatai: Ditto.

app/jobs/generate_webinar_job.rb
```diff
@@ -0,0 +1,16 @@
...
+  queue_as :default
+
+  def perform(meeting_id, email, participants)
+    csv = GenerateWebinarForParticipants.new(meeting_id: meeting_id, participants: participants).run
+    GenerateWebinarMailer.with(email: email, csv: csv).success_email.deliver_now
+  end
+
+  rescue_from(StandardError) do |exception|
+    GenerateWebinarMailer.with(email: arguments.second).failure_email.deliver_now
+    raise exception
```
sadleb, line 14: Ditto on using `raise` instead of `raise exception` so as not to mess with the stacktrace.

app/jobs/sync_to_slack_job.rb
```diff
@@ -0,0 +1,16 @@
...
+  queue_as :default
+
+  def perform(program_id, email)
+    SyncSlackForProgram.new(salesforce_program_id: program_id).run
+    BackgroundSyncJobMailer.with(email: email).success_email.deliver_now
+  end
+
+  rescue_from(StandardError) do |exception|
+    BackgroundSyncJobMailer.with(email: arguments.second).failure_email.deliver_now
+    raise exception
```
sadleb, line 14: Ditto.

app/jobs/sync_to_webinar_job.rb
```diff
@@ -0,0 +1,16 @@
...
+  queue_as :default
+
+  def perform(program_id, email, force_update)
+    SyncWebinarLinksForProgram.new(salesforce_program_id: program_id, force_update: force_update).run
+    BackgroundSyncJobMailer.with(email: email).success_email.deliver_now
+  end
+
+  rescue_from(StandardError) do |exception|
+    BackgroundSyncJobMailer.with(email: arguments.second).failure_email.deliver_now
+    raise exception
```
sadleb, line 14: DItto

app/services/generate_webinar_for_participants.rb
```diff
@@ -0,0 +1,24 @@
...
+    registrants = tasks.add_participants_to_meetings(meeting_id, participants)
+    CSV.generate do |csv|
+      csv << registrants.first.keys
+      csv << registrants.map(&:values)
+    end
+  end
+
+  private
+
+  attr_reader :meeting_id, :participants
```
sadleb, line 23: I brought this up before, but I still don't get why you're doing these private `attr_reader`s. Are these somehow used by the view or somewhere else outside this class that can somehow still access them? If the only place you're using them is here: 
```
registrants = tasks.add_participants_to_meetings(meeting_id, participants)
```
then just do
```
registrants = tasks.add_participants_to_meetings(@meeting_id, @participants)
```

This goes for all the files where you use this anti-pattern. The `attr_reader`, `attr_writer` etc methods are convenience methods to expose instance variables to some external class.

app/controllers/slack_controller.rb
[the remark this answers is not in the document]
sadleb: I got inspired to finally go implement this for Highlander. It's not worth backporting to Portal and Booster code, but I wanted you to be aware: https://github.com/bebraven/platform/pull/452

app/controllers/slack_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# A non standard controller used to initiate syncs from Salesforce into Slack cohorts
+class SlackController < ApplicationController
+  layout 'admin'
+
+  # Non-standard controller without normal CRUD methods. Disable the convenience module.
+  def dry_crud_enabled?() 
+    false 
+  end
+  before_action :authorize_index
```
hcli-work, line 11: convention is to put the all `before_action` calls before defining any methods.

app/services/generate_webinar_for_participants.rb
hcli-work: I had the same question. Abdul mentioned that there is a pattern of using private getters. The main advantage is that you get  a nice exception if there's a typo. Otherwise, just using a mistyped instance variable give you nice fat unhelpful `nil`

app/jobs/generate_webinar_job.rb
hcli-work: agree

app/controllers/webinar_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+# and generates webinar links from Salesforce
+class WebinarController < ApplicationController
+  layout 'admin'
+
+  # Non-standard controller without normal CRUD methods. Disable the convenience module.
+  def dry_crud_enabled?() 
+    false 
+  end
+
+  before_action :authorize_index
```
hcli-work, line 15: same

app/controllers/webinar_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+require 'csv'
+
+# Non-standard controller used to generate webinar links independently
+# and generates webinar links from Salesforce
+class WebinarController < ApplicationController
+  layout 'admin'
+
+  # Non-standard controller without normal CRUD methods. Disable the convenience module.
+  def dry_crud_enabled?()
```
hcli-work, line 11: does this have to be public instead of private?

app/services/generate_webinar_for_participants.rb
skaex: I think I should weigh in here since I started this 'anti pattern'. One of the key reason for using private getters in ruby for me can be explained in this scenario:


```ruby
class Hello
  def initialize(foo:, bar:)
    @foo = foo
    @bar = bar
  end

  def test_pattern
    p @foo
    p @foa
  end

  def test_antipattern
    p foo
    p foa
  end

  private

  attr_reader :foo, :bar
end

hello = Hello.new(foo: 'bar', bar: 'bar')
hello.test_pattern
hello.test_antipattern
```

Evaluate what the outputs for the two method call will be and which is better. Please share resources with me on why you think private getters and setters are anti-patterns.

skaex: Okay @hcli-work got here before me so I'll retract my response. However do share with me resources on why this is anti-pattern.

hcli-work (approved):
I'm good with these changes. Please just move the before_action to the normal spot before squashing and merging

app/services/generate_webinar_for_participants.rb
sadleb: Maybe anti-pattern is too strong of a way to put it @skaex but imo this pattern and the other pattern I've noticed you introducing of required keyword args (e.g. `def foo(id:, :name)`, feels like an attempt to introduce strong typing into a language that is consciously and unabashedly not strongly typed. In ruby, we should be writing tons and tons of specs and the scenario of a typo not getting caught until runtime just means you didn't specs. Imo both of these patterns are adding cruft and making it harder to read for the purpose of trying to move towards strongly typing stuff. 

Basically, I agree with this article: https://sikac.hu/private-attribute-accessor-in-ruby-is-an-artifact-from-other-languages-9d7f4438bc9b cc @hcli-work @rshipp

hcli-work: It's a good point regarding specs catching the typo and not relying on a private accessor. Brian is the A in the RACI here, so it seems clear this should be changed.

Regarding the required keyword args, that feels like a conversation for a different space. I never thought about that as having anything to do with typed args.

skaex: On the second pattern, I assume it is sometimes clearer (especially when the arguments are more than 3) to pass keyword arguments (`kwargs`) since it'll help the caller know what argument is passed. For example:
```ruby
# Definitions
def hello_no_kwargs(foo, bar, fizz, buzz)
end

def hello_kwargs(foo:, bar:, fizz:, buzz:)
end

# Callers
hello_no_kwargs('buzz', 'fizz', 'foo', 'bar')
# vs
hello_kwargs(foo: 'foo', bar: 'bar', fizz: 'fizz', buzz: 'buzz')

```

I never thought of it as having something to do with typing. If I wanted typing there are tools like [Sorbet](https://sorbet.org) for this.

I don't believe it is a good idea to solely depend on specs for guarantees. If the language has features to help with some guarantees I don't see why the features should be avoided.

@sadleb BTW I agree with the comments on the article you shared.

sadleb: Yeah, let's change it. The other concern here is that this doesn't match the rest of the codebase / style. If we did decide to use private `attr_reader` for everything, we should go refactor all the controllers do use this style. It's way worse imo to have two separate styles where depending on who wrote the particular class, you have to adjust to the patterns used and understand the ins and outs.

sadleb: I'm going to queue up a convo on the require keyword args thing at some point soon-ish.

sadleb: imo
```
hello_no_kwargs('buzz', 'fizz', 'foo', 'bar')
```
is more readable than
```
hello_kwargs(foo: 'foo', bar: 'bar', fizz: 'fizz', buzz: 'buzz')
```

We're almost always calling a method by passing it variables instead of strings. Variables should be named nicely so we know what they represent.

sadleb: imo, keyword args are useful when there are a lot of params and they're mostly optional so you choose the ones you want to pass by name instead of having to put a bunch of nil's into a positional params list. But required keyword args just makes it redundant.

app/controllers/webinar_controller.rb
sadleb: I personally wouldn’t spend time looking into this. This stuff is going away and the rest of the codebase just sticks this at the top after the before actions

sadleb (approved):
Lookin solid!

PR #437, opened 2020-11-08
Add BaseCourseProjectVersion and BaseCourseSurveyVersion

sadleb:
### Description

**NOTE:** this breaks previously published projects and surveys.

We're usually dealing with either a Project or a Survey (or a list of them),
so this adds the ability to refer to the join table between BaseCourse
and a ProjectVersion or SurveyVersion using `BaseCourseProjectVersion`
and `BaseCourseSurveyVersion` instead of `BaseCourseCustomContentVersion`

### Motivation and Context

This reduces the cognitive load required to think about both
Projects and Surveys at once in many places and let's you focus on
project vs survey specific logic / concerns b/c you know what it is.

It also makes the URLs a bit cleaner. E.g.
- https://platformweb/base_course_survey_versions/1/survey_submissions/new
- https://platformweb/base_course_project_versions/1/project_submissions/new

vs
- https://platformweb/base_course_custom_content_versions/1/survey_submissions/new
- https://platformweb/base_course_custom_content_versions/1/project_submissions/new          

### Types of changes

### Checklist:


app/controllers/custom_contents_controller.rb
```diff
@@ -94,7 +82,20 @@ def set_custom_contents
   def custom_content_class
     # Prefer `type` specified by form over the one set in route parameters
     type = params[:custom_content] ? params[:custom_content][:type] : params[:type]
-    CustomContentsController.class_from_type(type)
+    class_from_type(type)
+  end
+
+ def class_from_type(type)
```
sadleb, line 88: Just moving the above from a class method to an instance method and making it private since we don't need this anymore from the join model.

app/controllers/project_submissions_controller.rb
```diff
@@ -8,29 +8,29 @@ class ProjectSubmissionsController < ApplicationController
...
 
   def new
     @project_submission = ProjectSubmission.new(
       user: current_user,
-      base_course_custom_content_version: @base_course_custom_content_version,
+      base_course_project_version: @base_course_project_version,
     )
     authorize @project_submission
 
-    @has_previous_submission = ProjectSubmission.for_custom_content_version_and_user(
```
sadleb, line 31: `ProjectSubmission.for_project_version_and_user(project, user)` didn't feel any cleaner than `ProjectSubmission.where(project, user)`

config/routes.rb
```diff
@@ -41,12 +41,18 @@
...
     # a course template, e.g. through the Course Management page.
-    resources :base_course_custom_content_versions, only: [:new]
+    resources :base_course_project_versions, controller: 'base_course_custom_content_versions', type: 'BaseCourseProjectVersion', only: [:new]
+    resources :base_course_survey_versions, controller: 'base_course_custom_content_versions', type: 'BaseCourseSurveyVersion', only: [:new]
   end
 
-  resources :base_course_custom_content_versions, only: [:create, :update, :destroy] do
-    resources :project_submissions, only: [:show, :new, :create]
-    resources :survey_submissions, only: [:show, :new, :create]
+  resources :base_course_custom_content_versions, only: [:create, :update, :destroy]
```
sadleb, line 48: You can still create/update/destroy in a polymorphic way as long as you know the `BaseCourseCustomContentVersion.id` but when dealing with the nested submission, this splits it up so you nest it under the path to match the submission type.

config/routes.rb
```diff
@@ -41,12 +41,18 @@
...
+    resources :base_course_survey_versions, controller: 'base_course_custom_content_versions', type: 'BaseCourseSurveyVersion', only: [:new]
   end
 
-  resources :base_course_custom_content_versions, only: [:create, :update, :destroy] do
-    resources :project_submissions, only: [:show, :new, :create]
-    resources :survey_submissions, only: [:show, :new, :create]
+  resources :base_course_custom_content_versions, only: [:create, :update, :destroy] 
+
+  resources :base_course_project_versions, controller: 'base_course_custom_content_versions', type: 'BaseCourseProjectVersion', only: [] do
+    resources :project_submissions, only: [:new, :create]
```
sadleb, line 51: Note: I'm leaving the `create` route in here b/c where the creation happens from React, it was easier to let the nested route magic deal with setting the base_course properly. We could refactor this to match how only the `new` method is nested for `base_course_project_versions`, but it was blowing up the scope of this PR to get into that.

spec/controllers/survey_submissions_controller_spec.rb
```diff
@@ -3,19 +3,10 @@
...
+  let(:section) { create :section, course: course_survey_version.base_course }
   let(:user) { create :fellow_user, section: section }
-
-  # This puts the survey in the course
-  let(:survey_version) { create :survey_version }
-  let(:course_survey_version) { create(
-    :base_course_custom_content_version, 
-    base_course: course,
-    custom_content_version: survey_version,
-  )}
```
sadleb, line 17: Something was failing so I just made this match how `project_submissions_controller_spec` was setting things up and the failure went away. I didn't dig further.

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -84,14 +84,42 @@ def destroy
     end
   end
 
+  def humanized_custom_content_type 
+    custom_content_class.name
```
rshipp, line 88: i think this only works because both our types are one word. could you use `.model_name.human` instead?

app/controllers/base_course_custom_content_versions_controller.rb
```diff
@@ -84,14 +84,42 @@ def destroy
...
+    when nil
+      BaseCourseCustomContentVersion
+    when 'BaseCourseProjectVersion'
+      BaseCourseProjectVersion
+    when 'BaseCourseSurveyVersion'
+      BaseCourseSurveyVersion
+    else
+      raise TypeError.new "Unknown BaseCourseCustomContentVersion type: #{type}"
+    end
   end
```
rshipp, line 123: these appear to be the same, am i missing something?

sadleb: One is the class for `CustomContent` (aka Project vs Survey). The other is the `BaseCourse{Project|Survey}Version` class based on the passed in type, which we need for the `new` action since we don't already have an instance to work with.

rshipp: ohhh i was staring at the strings and missed the class names, got it

rshipp (approved):

jessouyangbraven:
This looks good to me! In general, I like having the most specifically typed object as possible because then I have fewer scenarios to reason through. 

A side note, if we're really interested in how the public-facing URLs look, I think we can eventually get the submission URLs to something like:

https://platformweb/course_project_versions/{id}/submissions/new`
https://platformweb/course_survey_versions/{id}/submissions/new`

Because submission are only valid on courses (not templates) and because the project/survey submission is redundant to the prior part of the URL. It'd be nice to eventually have `Course{Template}{Project, Survey}Version`. Definitely not something that needs to happen in this PR, though.

jessouyangbraven (approved):

PR #442, opened 2020-11-10
Require canvas_assignment_id in join table entries

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1198965066699365

- DB migration to make `canvas_assignment_id` non-nullable
- Refactors the `base_course_custom_content_versions#create` logic into a model transaction so we can clean up if the DB validation fails

### Motivation and Context

It doesn't make sense to have this join table if it doesn't tell us the Canvas assignment ID, so we're adding stricter, db-level checks for this.

### Checklist:


app/models/base_course_custom_content_version.rb
```diff
@@ -19,6 +19,55 @@ class BaseCourseCustomContentVersion < ApplicationRecord
   scope :base_course_project_versions, -> { where type: 'BaseCourseProjectVersion' }
   scope :base_course_survey_versions, -> { where type: 'BaseCourseSurveyVersion' }
 
+  # Create! a new record and its corresponding Canvas assignment
+  def self.publish!(base_course, custom_content, user, rubric_id=nil)
+    canvas_assignment = nil
+
+    begin
+      transaction do
```
sadleb, line 27: It makes sense to move this logic out of the controller so that we can put it in a transaction. I also think we're entering "fat model" territory at this point where the model has become responsible for the business logic instead of simple database reads/writes/updates/destroys. Not for this PR, but in the future (and in general) we should consider moving the business logic of publishing (new and update) and deleting to services in `app/services`. We can still run the service in a transaction so that all database writes are rolled back if it fails, but it would make the model easier to understand and maintain if this method was just something like:
```
def self.publish!(...)
  transaction do
    PublishCustomContent.new(...).run
  end
end
```

The reason I don't think we should do this for this PR is that it doesn't seem more readable/maintainable to move the `publish_latest` and `remove` methods to their own services since they are basically just an extra line of code and I can't come up with a clean service name responsible for all three actions other than `CustomContentPublisher` which would have a different API and naming convention from the rest of the services we already have which merits a bigger discussion.

app/models/base_course_custom_content_version.rb
```diff
@@ -19,6 +19,55 @@ class BaseCourseCustomContentVersion < ApplicationRecord
...
+        )
+
+        # Update the assignment ID in the record
+        # Note: the create! can fail, so clean up things not handled by the
+        # model transaction, e.g. any Canvas state, before this point
+        base_course_custom_content_version = create!(
+          base_course: base_course,
+          custom_content_version: custom_content_version,
+          canvas_assignment_id: canvas_assignment['id'],
+          type: "BaseCourse#{custom_content.type}Version",
```
sadleb, line 43: I feel like you and I go back and forth where one of us always refactors method calls to smoosh the params into a single line or as few as possible (me) and the other expands all the params to their own line (you). We should probably have a broader discussion about what coding style we want in this codebase, agree to it, and start being consistent going forward.

I lean towards not putting params on their own lines b/c when I'm reading code and trying to troubleshoot I'm mostly focused on the method names and what methods are called in which order to understand what is happening. When I get to the method that isn't doing what I expect, then I dig into the params and the guts to understand why. It's more readable for me personally to see at a glance something like the following:
```
begin
  transaction do
    custom_content_version  = custom_content.save_version!(...)
    canvas_assignment = CanvasAPI.client.create_lti_assignment(...)
    base_course_custom_content_version = create!(...)
    CanvasAPI.client.update_assignment_lti_launch_url(...)
    CanvasAPI.client.add_rubric_to_assignment(...)
    return base_course_custom_content_version
  end
rescue
  CanvasAPI.client.delete_assignment(...)
end
```
than it is to have to scroll through and force my eyes to skim past all the params in order get the full picture of all the methods this method calls in order to understand the logic happening here.

spec/models/base_course_custom_content_version_spec.rb
```diff
@@ -3,9 +3,10 @@
 RSpec.describe BaseCourseCustomContentVersion, type: :model do
 
   let(:course) { create :course_with_canvas_id }
-  let(:custom_content_version) { create :custom_content_version }
-  let(:base_course_custom_content_version) { create :base_course_custom_content_version, base_course: course, custom_content_version: custom_content_version }
-  
+  let(:project) { create :project }
+  let(:project_version) { create :project_version, custom_content: project }
```
sadleb, line 7: Not for this PR, but now that we've finished refactoring most CustomContent stuff into Project vs Survey, I think we should also refactor the specs so that in the generic CustomContent spec it only tests the generic logic that works on both but we move any logic that requires either a project or survey reference into their own specs, like `base_course_project_version_spec.rb`. 

The `base_course_custom_content_version_controller_spec` also comes to mind!

app/models/base_course_custom_content_version.rb
jessouyangbraven: Yeah, we never discussed nor came to a consensus on the coding style we'd use. I've mostly been following: https://github.com/airbnb/ruby#align-function-params and Codacy when it was running.

The argument-on-each-line I do for a couple reasons:
1. At some point, the line's just way too long and it's going to wrap in the editor/GitHub and the auto-wrap's random break is harder to read than an intentional line break at a logical point in the code.
2. It makes the blame really clean, you can see the commit that changed individual parameters rather than having to figure out what substring in a long line changed. (This is also why I'm obsessed with trailing commas.)

I think a good start would be if we choose set a standard line-length and stuck to it. 

But I think this all hinges on having Codacy running to enforce it (and us having a norm around fixing those errors before merging, or even strong, adding a branch rule to prevent merging until you fix the issues?). @rshipp's really convinced me that not having automated checks means maintaining coding style is not sustainable thing and devolves into noisy style-nit PR reviews.

sadleb: Yup, I hear all of that. Thank you for bringing up the trailing comma thing! I just assumed it was ... actually I am had no clue why it was so frequent but didn’t think it worth mentioning. Now I know there is a rhyme to the reason!

Completely agree on the codacy route to address this vs nit code review comments too!

sadleb (approved):
Lgtm

PR #446, opened 2020-11-11
Handle ID attributes for headings (and sometimes other things too)

rshipp:
### Description

Add general element ID handling, and custom heading-specific ID handling.

This does not check uniqueness, it just generates UUIDs and relies on them being unique.

Task: https://app.asana.com/0/1174274412967132/1195096839523739/f

### Testing

1. Open the editor with the inspector (/custom_contents/new?debug)
1. Type something in a document
2. From the block toolbar (left **P** button), change the Paragraph dropdown to any Heading.
3. See that there's an ID attribute on the newly inserted heading.
4. Try to break things (toggle code tab, back to design tab, press enter after the heading, change the heading type from the dropdown, ???), notice the IDs remain unchanged
5. Change the heading to a paragraph, notice the ID is removed
5. Copy/paste the heading (might have to mess with it so that the whole heading element is inside the selection), notice the ID changes
6. Click on the heading, notice the right panel with copyable (but read-only) heading anchor link.

### Known issues

* Press Enter before the heading (heading with duplicate ID created)
* Press Enter in the middle of the heading, splitting it in two (heading with duplicate ID created)

Screen Shot 2020-11-11 at 3 19 28 PM
Screen Shot 2020-11-11 at 3 25 47 PM

### Types of changes

### Checklist:


app/javascript/components/ContentEditor.js
```diff
@@ -146,12 +150,12 @@ BalloonEditor.defaultConfig = {
...
-            { model: 'heading2', view: 'h3', title: 'Heading 2', class: 'ck-heading_h3', converterPriority: 'low' },
-            { model: 'heading3', view: 'h4', title: 'Heading 3', class: 'ck-heading_h4', converterPriority: 'low' },
-            { model: 'heading4', view: 'h5', title: 'Heading 4', class: 'ck-heading_h5', converterPriority: 'low' },
-            { model: 'heading5', view: 'h6', title: 'Heading 5', class: 'ck-heading_h6', converterPriority: 'low' },
+            { model: 'paragraph', title: 'Paragraph', class: 'ck-heading_paragraph' },
+            { model: 'heading1', view: 'h2', title: 'Heading 1', class: 'ck-heading_h2' },
+            { model: 'heading2', view: 'h3', title: 'Heading 2', class: 'ck-heading_h3' },
+            { model: 'heading3', view: 'h4', title: 'Heading 3', class: 'ck-heading_h4' },
+            { model: 'heading4', view: 'h5', title: 'Heading 4', class: 'ck-heading_h5' },
+            { model: 'heading5', view: 'h6', title: 'Heading 5', class: 'ck-heading_h6' },
```
rshipp, line 158: the converterpriority stuff was to keep this from conflicting with our questionTitle/etc, which no longer exist.

app/javascript/components/ContentEditor.js
```diff
@@ -375,6 +374,23 @@ class ContentEditor extends Component {
                                 <label htmlFor='input-placeholder'>Placeholder</label>
                             </>
                         );
+                    } else if ( modelElement.startsWith('heading') ) {
+                        return (
+                            <>
+                                <h4>Heading</h4>
+                                <p>To link to this heading, copy the following ID tag and use it as the link URL.</p>
```
rshipp, line 381: wording?

app/javascript/ckeditor/attributeediting.js
```diff
@@ -0,0 +1,14 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import SetAttributesCommand from './setattributescommand';
+import RemoveAttributeCommand from './removeattributecommand';
+
+export default class AttributeEditing extends Plugin {
```
rshipp, line 5: i broke this out of contentcommonediting because it's a dependency for headingidediting and I didn't want to depend on all of contentcommonediting

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -3,18 +3,16 @@ import { enablePlaceholder } from '@ckeditor/ckeditor5-engine/src/view/placehold
...
 import InsertTextInputCommand from './inserttextinputcommand';
 import InsertFileUploadCommand from './insertfileuploadcommand';
 import InsertTextAreaQuestionCommand from './inserttextareaquestioncommand';
 import InsertTextAreaCommand from './inserttextareacommand';
-import SetAttributesCommand from './setattributescommand';
 import { getNamedChildOrSibling } from './utils';
 
 export default class ContentCommonEditing extends Plugin {
     static get requires() {
-        return [ Widget, RetainedData, List ];
```
rshipp, line 17: at a high level: we no longer add unique ids from converters. we only do it in headingidediting, and eventually in *commands* like insertradiocommand.

app/javascript/ckeditor/elementidediting.js
```diff
@@ -0,0 +1,80 @@
...
+            const operation = args[0];
+            // When you choose "Paragraph" or a heading on the dropdown, it applies a RenameOperation.
+            if ( operation.type == 'rename' ) {
+                if ( operation.oldName == 'paragraph' && operation.newName.startsWith( 'heading' ) ) {
+                    // paragraph -> heading; add an ID.
+                    // Note: We have to add *an* ID here, but this ID is immediately replaced by the
+                    // one generated in the addAttribute handling below. Unfortunately I can't think of
+                    // a better way to do this.
+                    const headingElement = operation.position.nodeAfter;
+                    this.editor.execute( 'setAttributes', { 'id': this._getNewId() }, headingElement );
```
rshipp, line 63: it feels a little weird to call commands from inside an event listener, but I also know these commands are safe. while we *can* directly modify the model with e.g. `headingElement._setAttribute`, I'm pretty sure that immediately puts us into undefined behavior territory.

app/javascript/components/ContentEditor.js
jessouyangbraven: To link to this heading:
- Add a new link (or edit an existing one)
- Copy **Heading ID** displayed here and use it as the URL in the text input

I dunno if that's better. But using terminology consistent with what we display in `<label>` might help?

app/javascript/ckeditor/elementidediting.js
jessouyangbraven, line 63: I don't know if this is better, but we could use a dummy value instead of the final/real UUID here. We'll reduce the number of UUIDs we generate, maybe reducing the possibility of collisions? Maybe it helps performance? 

But it does feel safer to always assign a unique ID, like you have it now, so we're never in a bad intermediate state with our HTML.

I could go either way, honestly.

jessouyangbraven (approved):
This looks good to me. Thanks for tackling this!

As a side note, I'm super stoked that we're chipping away at our retained data dependence.

app/javascript/components/ContentEditor.js
sadleb: One more option:

Use the following Heading ID as the URL when creating a link to this heading.

sadleb (approved):
Gave another option on wording. I have nothing constructive on the CKeditor side of things.

app/javascript/ckeditor/elementidediting.js
rshipp: yeah I thought about making it an obvious string like "REPLACE_ID" or something, but I think the cost of generating an id is lower than the cost of fixing a bunch of stuff if something changes/breaks and we end up with the fake ID instead of a real one.

PR #449, opened 2020-11-11
Fix adding project without a rubric

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1198965066699365

I introduced a bug in #442 and now you can't add a project without a rubric. This PR fixes this by checking for empty strings as the `rubric_id`. I also fixed the `delete_assignment` call that actually got hit when I ran into an error so it works and deletes the Canvas assignment if the transaction fails.

### Motivation and Context

In Rails, only `nil` and `false` evaluate to `false`. We're getting an empty string for the `rubric_id` and trying to add an association for it with the previous check. `.present?` also checks for empty strings (and arrays).

### Test

- Can add projects from the Course Management page again
- Added specs to explicitly test `nil` and empty string cases don't try to add rubric to project

### Types of changes

### Checklist:


sadleb (approved):
lgtm

PR #451, opened 2020-11-12
Allow hard-coding <select> and <option> in Content Editor HTML

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1198222893411568

This implements hard-coding `<select>` and its `<option>`s for the Impact Survey's 1-10 scale in the Content Editor HTML view.

### Motivation and Context

This is what the Impact Surveys currently look like:
Screen Shot 2020-10-26 at 12 30 22 PM

The slider components we were using aren't accessible, so we're switching to another component, while trying to balance the amount of effort spent building the authoring experience (e.g., making it configurable vs. hard-coding it) and making the surveys easy to fill out for the fellows.

This PR is a first step in that it lets us create Impact Surveys even if we don't have an authoring experience in the Content Editor's **Design** tab built for launch. 

Note that in this PR, we're doing a lot of manual HTML manipulation. In addition to adding `<select>` and `<option>` manually, we need to set the `name` attribute on `<select>`, so we have a key for our `SurveySubmissionAnswers` key-value store.

@rshipp is adding ID generation in #446 that we can probably piggy-back off of in a future PR to generate the `name`s.

### Test

- Everywhere we want a drop-down, paste the following HTML in the **Code** tab:
```
<select>
<option value="">&nbsp;</option>
<option value="1">1</option>
<option value="2">2</option>
<option value="3">3</option>
<option value="4">4</option>
<option value="5">5</option>
<option value="6">6</option>
<option value="7">7</option>
<option value="8">8</option>
<option value="9">9</option>
<option value="10">10</option>
</select>
```
Screen Shot 2020-11-11 at 5 00 54 PM

- Toggle to **Code** tab and see that the selector appears:
Screen Shot 2020-11-11 at 5 00 58 PM

- **Save** as a _Survey_ and preview it, checking that the values are set in the HTML:
Screen Shot 2020-11-11 at 5 01 38 PM

- I made some content edits so it looked more like our Impact Survey and published using our Course Management page:
Screen Shot 2020-11-11 at 5 00 37 PM

- Manually change the `name` attribute for each `<select>` in the HTML tab:
Screen Shot 2020-11-11 at 5 05 32 PM

- Save and Publish to Canvas through the Course Management page:
Screen Shot 2020-11-11 at 5 02 23 PM

- **Submit** as a Fellow after selecting some values:
Screen Shot 2020-11-11 at 4 24 43 PM

- The only way to see the answers is in our DB right now, so I used the console to check the `SurveySubmission.last.answers`, note that the names and values for the inputs are set:
Screen Shot 2020-11-11 at 4 38 07 PM

### Types of changes

### Checklist:


sadleb (approved):

app/controllers/survey_submissions_controller.rb
```diff
@@ -35,12 +35,11 @@ def create
     redirect_to @survey_submission
   end
 
-  private
-
+private
   # Get the "name" attributes from all the <input> elements in this survey 
   # version. These are the parameters we permit from the form being submitted.
   def survey_answer_params
```
rshipp, line 41: i thought we got rid of this? I still don't see a reason to try to verify params here, there's so many ways to stuff the database already.

jessouyangbraven: I got rid of the validation in the model, but I left this on the controller because our own form data sends params like `commit: Submit` (for the submit button) and `authenticity_token` that we would otherwise write to the DB.

I guess I could blacklist things like authenticity tokens, LTI state, etc., that we don't want saved to the DB (for security reasons). But I also know I'd always forget to do that when adding new features. Whereas if I have to white-listing allowable input types, I'm going to notice when I'm testing it out.

I don't know if this is actually the case, but I feel like the data team would have an easier time with a table that doesn't have extraneous stuff in it?

Is the concern here that the whitelist here is difficult to maintain in comparison to what we get out of it?

rshipp: i missed this comment but it's coming up again as I do the peer review stuff so I want to bring this up in standup today

rshipp: related: https://github.com/bebraven/platform/pull/336/files#diff-3b19af8b66fab7a9e99d57507024a4b37cf0d1067becbc93ba054769c8a43151R16-R23

PR #452, opened 2020-11-12
Enhance DryCrud to not require setting dry_crud_enabled? false

sadleb:
### Description

The solution here is so simple, just check if the model class is defined
and don't register the before_actions, but i wrestled with trying to get
it to use the `model_class` method from within the `included` context so
I just ended up not being very DRY and just copy/pasting the code that looks
for the model class. Oh the irony.

Also, we include `DryCrud` on the base `ApplicationController` but it's logic needs to
apply for the subclasses. It listens for when it's subclassed and defines the 
`before_actions` on the subclass if necessary.

### Types of changes

### Checklist:


lib/dry_crud.rb
```diff
@@ -11,41 +11,36 @@ module Controllers
     # NOTE: you can alway override this behavior by defining whatever you want in the 
     # associated controller's index, show, edit, etc methods.
     included do
-      before_action :set_models_instance, only: [:index]
-      before_action :set_model_instance, only: [:show, :edit, :update, :destroy]
-      before_action :new_model_instance, only: [:new]
+      if controller_path.classify.safe_constantize
```
sadleb, line 14: This is the code to turn the controller's name into the model class constant. Same as in the `model_class` method. E.g. `controller_path` for `FakeModelsController` would be `fake_models`

lib/dry_crud.rb
```diff
@@ -11,41 +11,36 @@ module Controllers
     # NOTE: you can alway override this behavior by defining whatever you want in the
```
skaex, line 11: This seems to me like it concerns only the controllers. Is there any reason for this being here instead of the concerns folder within the controller folder?

sadleb: Good call. I'll move it. Our first `concern`!

app/controllers/concerns/dry_crud.rb
```diff
@@ -8,44 +8,51 @@ module Controllers
...
+        def self.inherited(subclass)
+          super
+          potential_model_name = subclass.controller_path&.classify&.safe_constantize
+          if potential_model_name && potential_model_name <= ActiveRecord::Base
+            subclass.send(:before_action, :set_models_instance, only: [:index])
+            subclass.send(:before_action, :set_model_instance, only: [:show, :edit, :update, :destroy])
+            subclass.send(:before_action, :new_model_instance, only: [:new])
+          end
+        end
+      end
```
sadleb, line 32: I tried to write this more cleanly by using methods for the re-usable stuff but wasted too much time. It's confusing b/c we're in the context of a module and trying to do meta-programming. Didn't seem worth the effort to clean this up.

app/controllers/concerns/dry_crud.rb
```diff
@@ -8,44 +8,51 @@ module Controllers
...
 
-    # Override this in any class that needs to turn this behavior off.
-    #
-    # Note: there must be a better way to do this, but I couldn't figure it out.
-    # This module is included in ApplicationController and all controller inherit that,
-    # so if I try to rely on respond_to?(), it applies to ApplicationController and not the subclass.
-    def dry_crud_enabled?
-      true
+      potential_model_name = controller_path.classify.safe_constantize
+      if potential_model_name && potential_model_name <= ActiveRecord::Base
```
sadleb, line 16: `potential_model_name <= ActiveRecord::Base` is checking if the class inherits from `ActiveRecord::Base`

rshipp (approved):

PR #458, opened 2020-11-13
Add ability to publish a Waivers assignment from Course Management to Canvas

sadleb:
### Description
This is the first step of [Fellows/LCs must sign NDA and liability waivers before they can access course content](https://app.asana.com/0/1174274412967132/1198301329741772). It allows us to publish and delete a Canvas assignment that will launch the to-be-implemented embedded Form Assembly waivers.

Note: the only thing we need to store in the platform database related to Waivers is whether the Fellow or LC has signed them so that we stop showing the waivers if they open that assignment up again. This will happen when we implement the `create` action that will also send a submission to Canvas in order to unlock future modules.

I could have implemented this by storing the `canvas_waivers_assignment_id` of the published Waivers assignment on the `BaseCourse` and setting the LTI launch URL to be course specific. E.g. https://platform.braven.org/base_course/1/waivers/new That would require handling when we launch a new program and adjusting the assignment ID in our database and the launch URL in canvas. Instead, the plan is to use a static endpoint that we don't have to adjust. We'll pull the course ID out of the `LtiLaunch`. The only potential need left for storing information locally would be to know whether to show the Add or Delete buttons on the Course Mgmt page. This is a low traffic page, so it seemed fine to always query the Canvas API. We'll use this same approach for pre/post-accelerator surveys, peer evaluations, attendance forms, etc.

I cutover `app/services/initialize_new_course.rb` in this PR https://github.com/bebraven/platform/pull/462 to use this new `app/services/fetch_canvas_assignments_info` service. In future PRs, we can expand it to support the rest of the assignment types we'll be supporting and need information about on the fly.

The next phase of this will be to turn the prototype of embedded a Form Assembly form [found here](https://github.com/bebraven/platform/pull/349) into a real PR.

### Types of changes

### Checklist:


public/waiver.html
```diff
@@ -0,0 +1,81 @@
+<html>
```
sadleb, line 1: Whoops. Accidental commit. I’ll remove this file

app/services/fetch_canvas_assignments_info.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def add_project_or_survey_info!(base_course_custom_content_version)
+    if base_course_custom_content_version.is_a?(BaseCourseProjectVersion)
+      @base_course_project_versions << base_course_custom_content_version
+    elsif base_course_custom_content_version.is_a?(BaseCourseSurveyVersion)
+      @base_course_survey_versions << base_course_custom_content_version
+    else
+      raise FetchCanvasAssignmentsInfoError, "BaseCourseCustomContentVersion type not recognized: #{base_course_custom_content_version.inspect}"
+    end
+  end
```
sadleb, line 68: This is unrelated to this PR, but I'll cutover the nearly identical logic in `initialize_new_course` to use this class instead in a future PR. I just wanted to get a sense for what the code would feel like

rshipp (approved):
approving based on standup discussion

PR #460, opened 2020-11-13
Handle copy/paste of inputs

rshipp:
### Description

Automatically change input names / ids when pasting, so you don't accidentally end up with duplicate names/ids.

For review: If you have feedback on where I should add comments, I'm too deep in this logic to be able to tell what's clear and what's not.

### Types of changes

### Checklist:


app/javascript/ckeditor/inputnameediting.js
```diff
@@ -0,0 +1,99 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import UpcastWriter from '@ckeditor/ckeditor5-engine/src/view/upcastwriter';
+import UniqueId from './uniqueid';
+import { ELEMENT_NAME_PREFIX, ELEMENT_ID_PREFIX } from './uniqueid';
+
+export default class InputNameEditing extends Plugin {
```
rshipp, line 6: This name is no longer accurate, and I'm drawing a blank as to what to call this plugin.

The function of this plugin is to change `name` and `id` attributes for `input` and `textarea` elements when you paste them.

jessouyangbraven: Do we need to add `<select>` to this as well?

jessouyangbraven: All the elements we've mentioned here are `form` elements [based on this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element#Forms).

Maybe `FormInputUniqueIdentifierEditing`?

I don't know if the `UniqueIdentifier` part is confusing, but it's hard coming up with something that covers both `id` and `name`. `UniqueAttributeValue`?

jessouyangbraven (approved):
Thanks for taking care of this!

I have a general architectural question, which is whether we should do this in its own `Plugin` or have each element (e.g., `input` `textarea`'s own plugins/commands) attach relevant listeners that we put in a library or something. I don't know what's possible or preferred with CKE. We can talk about this in stand-up if it's too much to type out.

app/javascript/ckeditor/inputnameediting.js
rshipp: Good catch, added `select`. 

I think we should take out the `Editing` piece, since this doesn't have any schema/converters like the other `*Editing` plugins. It's also explicitly tied to clipboard functionality - unless we merge the heading ID listener with this one too?

PR #468, opened 2020-11-18
Add peer reviews

rshipp:
### Description

Add Peer Reviews.

Task: https://app.asana.com/0/1174274412967132/1191666190028025/f

Based on SurveySubmissions and #336.

Note: Be sure to run `db:seed` after deployment to get the default PeerReviewQuestions.

Screen Shot 2020-11-20 at 1 02 07 PM

### Types of changes

### Checklist:


app/controllers/survey_submissions_controller.rb
```diff
@@ -1,62 +1,24 @@
...
-    doc = Nokogiri::HTML.parse(@survey_submission.survey_version.body)
-    doc.xpath("//input|//select|//textarea").map{ |input| input[:name] }
+  def answers_params_hash
+    # Since this controller accepts arbitrary params to #create, explicitly remove
+    # the params we know we don't want.
+    remove_params = [
+      "base_course_survey_version_id",
+      "type",
+      "controller",
+      "action",
```
rshipp, line 18: TODO: I have *no* idea if these are the right params. This is what the specs were sending. I'm not sure how to test survey submission but we should definitely manually test this before merging.

lib/lti_score.rb
```diff
@@ -38,7 +38,7 @@ def self.new_project_submission(canvas_user_id, submission_url)
 
   # Generates an LtiScore that represents a new Impact Survey submission
   # where no manual grading is needed. The score is for full credit.
-  def self.new_survey_submission(canvas_user_id, submission_url)
+  def self.new_full_credit_submission(canvas_user_id, submission_url)
```
rshipp, line 41: @jessouyangbraven I couldn't find anything that made this function specifically tied to surveys, so I renamed it to be more generic. not sure this name is accurate though.

app/controllers/concerns/submittable.rb
```diff
@@ -0,0 +1,63 @@
...
+    return redirect_to previous_submission if previous_submission
+
+    # Record in our DB first, so we have the data even if updating Canvas fails.
+    instance_variable.save_answers!(answers_params_hash)
+
+    # Update Canvas
+    lti_score = LtiScore.new_full_credit_submission(
+      @current_user.canvas_user_id,
+      # E.g. survey_submission_url(@survey_submission, protocol: 'https').
+      self.send("#{instance_variable_name}_url", instance_variable, protocol: 'https'),
```
rshipp, line 45: TODO: i didn't test this at all

app/controllers/concerns/submittable.rb
```diff
@@ -0,0 +1,63 @@
+require 'lti_advantage_api'
+require 'lti_score'
+
+module Submittable
```
sadleb, line 27: Wondering if we can code this a bit more defensively to make sure that both the `LtiHelper` and `DryCrud::Controllers::Nestable` modules have been included since we rely on them.

Not sure how but I would look into either using `responds_to?` to check if the method is defined, or using something like `Object.const_missing(name)` or `method_missing`, or trying to look at the `ancestors` or `descendents`.

sadleb, line 27: I would definitely leave a comment about what this is and how it's intended to be used.

sadleb (commented):
I'll review the rest later, but just to give feedback on the concern refactor, it's pretty cool! I left a couple of small comments.

app/controllers/concerns/submittable.rb
rshipp: I think it's possible to include a concern from a concern too, so we should think about that.

app/controllers/survey_submissions_controller.rb
```diff
@@ -1,62 +1,26 @@
...
+    ]
+    params.permit!.to_h.filter do |key|
+      not remove_params.include? key
```
rshipp, line 23: this `not` seems unrailsy but i don't know another way to do this

app/controllers/concerns/submittable.rb
```diff
@@ -0,0 +1,86 @@
+# "Submittable" behavior for controllers like SurveySubmissionsController
+# and PeerReviewController.
+#
+# This concern heavily depends on DryCrud::Controllers::Nestable, and requires
+# the including controller to be nested under one parent.
+#
+# Usage:
+#
+#   include DryCrud::Controllers::Nestable
```
sadleb, line 9: It also needs `LtiHelper` for the `lti_launch` utilities.

app/controllers/concerns/submittable.rb
```diff
@@ -0,0 +1,86 @@
...
+#   include Submittable
+#   nested_resource_of MyParentResource
+#
+# When using this concern, you must also define a method `answers_params_hash`
+# in the including controller. For example:
+#
+#   def answers_params_hash
+#     params.permit(:permitted_stuff).to_h
```
sadleb, line 17: I'm having trouble visualizing that this hash looks like without being deep in the code. Can you add an example? Or alternatively reference a spec that has an example?

app/controllers/survey_submissions_controller.rb
```diff
@@ -1,62 +1,26 @@
...
+  def answers_params_hash
+    # Since this controller accepts arbitrary params to #create, explicitly remove
+    # the params we know we don't want.
+    remove_params = [
+      "base_course_survey_version_id",
+      "type",
+      "controller",
+      "action",
+    ]
+    params.permit!.to_h.filter do |key|
```
sadleb, line 22: Would using `params.except` work here and be cleaner? https://apidock.com/rails/ActionController/Parameters/except

app/models/peer_review_submission.rb
```diff
@@ -0,0 +1,28 @@
...
+  belongs_to :course, foreign_key: :base_course_id, class_name: "Course"
+
+  has_many :peer_review_submission_answers
+  alias_attribute :answers, :peer_review_submission_answers
+
+  validates :user_id, :base_course_id, presence: true
+
+  # Takes a nested hash like:
+  #   { for_user_id => { peer_review_question_id: input_value (string) } }
+  # and adds them as PeerReviewSubmissionAnswers to this submission.
```
sadleb, line 12: Ooh, yeah, this is what I wanted to see in my previous comment when I had just seen the module. Maybe just reference this in the submittable module class comments.

app/controllers/concerns/submittable.rb
rshipp: Noted, I'll add more comments. The concern doesn't care what the hash looks like, it just has to match what the model `save_answers!` method accepts. If you look at the hashes for SurveySubmission and PeerReviewSubmission, they're completely different.

app/controllers/concerns/submittable.rb
rshipp: Whoops, that include should actually have been inside the concern, where I have the imports.

app/controllers/survey_submissions_controller.rb
rshipp: probably! i'll come back to this once I know what the correct things to exclude are.

app/controllers/concerns/submittable.rb
sadleb: Yeah, if Nestable is a requirement, then it makes sense to have it extend that and just mention in the comments here that you should include Nestable directly and not this.

app/controllers/concerns/submittable.rb
```diff
@@ -0,0 +1,90 @@
...
+# for SurveySubmission and PeerReviewSubmission look completely different.
+
+require 'lti_advantage_api'
+require 'lti_score'
+
+module Submittable
+  extend ActiveSupport::Concern
+  include LtiHelper
+
+  included do
```
sadleb, line 34: Can we check and throw a nice error if `answers_params_hash` isn't defined on the controller we're being included in? I haven't tried it, but maybe this would work: `raise "nice exception telling what's wrong" unless self.class.respond_to?(:answers_params_hash)`

app/controllers/survey_submissions_controller.rb
```diff
@@ -1,62 +1,25 @@
-require 'lti_advantage_api'
-require 'lti_score'
-require 'nokogiri'
-
 class SurveySubmissionsController < ApplicationController
-  include LtiHelper
   include DryCrud::Controllers::Nestable
+  include Submittable
```
sadleb, line 3: Can you add a comment here that `show`, `new`, and `create` are defined by Submittable? Or even add the definitions and just call `super`? I would be confused stumbling upon this for the first time after seeing some route and coming to this controller only for there not to be any associated actions.

config/routes.rb
```diff
@@ -35,7 +35,9 @@
   get 'course_management/launch', to: 'base_courses#launch_new'
   post 'course_management/launch', to: 'base_courses#launch_create'
 
-  resources :courses, controller: 'base_courses', type: 'Course'
+  resources :courses, controller: 'base_courses', type: 'Course' do
+    resources :peer_review_submissions, only: [:new, :create]
```
sadleb, line 39: We're going to need a way to test things as we setup a course template. Right now, we allow you to submit a project or a survey to a course template. Do we want to allow it for peer reviews too for the time being and figure out a general approach for testing and setting up a course template later?

db/seeds.rb
```diff
@@ -16,3 +16,10 @@ def yaml label
   bv_admin = User.create! first_name: 'Dev', last_name: 'AdminBV', email: 'admin@bebraven.org', password: "#{ENV['DEV_ENV_USER_PASSWORD']}", confirmed_at: DateTime.now
   bv_admin.add_role :admin
 end
+
+if PeerReviewQuestion.count == 0
+  PeerReviewQuestion.create!(text: "Actively contributed to team success")
+  PeerReviewQuestion.create!(text: "Met deadlines and fulfilled responsibilities in a timely manner")
+  PeerReviewQuestion.create!(text: "Gave feedback to others to help them be more successful and productiver")
+  PeerReviewQuestion.create!(text: "Embraced different perspectives on the team with openness and a sense of possibility")
+end
```
sadleb, line 25: Add to `db/dummies` too so that you don't have to blow away the DB and re-seed to get them there?

lib/lti_score.rb
sadleb: I like this approach. I'll end up refactoring the waivers PR to use this too.

jessouyangbraven (approved):
This looks reasonable to me. Let me know if you'd like me to re-review if you make any major changes based on inline comments?

app/controllers/survey_submissions_controller.rb
jessouyangbraven: I can help with this. Or if you want to leave the old `Nokogiri` stuff, I can update it once this PR is merged.

db/seeds.rb
sadleb: We talked. Seeds can just be run. It's prod data so just have it here.

app/controllers/concerns/submittable.rb
sadleb: We talked about this on standup and decided to just try and detect if the Nestable stuff isn't there from the `included` block and raise a nice error message. If it takes much time though, just queue up a backlog task b/c this is just a nice to have to devs since it'll break anyway if you mess it up.

rshipp:
The state thing is broken, I opened a task here https://app.asana.com/0/1174274412967132/1199344732354181/f

app/controllers/survey_submissions_controller.rb
```diff
@@ -1,62 +1,26 @@
...
+    # Since this controller accepts arbitrary params to #create, explicitly remove
+    # the params we know we don't want.
+    params.except(
+      :base_course_survey_version_id,
+      :type,
+      :controller,
+      :action,
+      :commit,
+      :state,
+      :authenticity_token,
```
jessouyangbraven, line 23: Just tested this manually, the things we want to blacklist here:

 - authenticity_token
- state
- commit

No harm to filtering out the rest, though.

config/routes.rb
sadleb: Also, this is going to break on new program launch. What we had talked about before was just using static endpoints that read the course from the LtiLaunch context for anything other than CustomContents. That's what I did with the waivers PR.

sadleb: The difference with Waivers though is that they don't need to use the `Submittable` module. Ugh... I guess it's better just to keep it this way and make sure `initialize_new_course#initialize_assignments` works

rshipp: > using static endpoints that read the course from the LtiLaunch context

I brought this up in standup and we decided to do the same thing impact surveys do, read from the nested parent and *not* the lti launch. We decided to to this before because having a GET request to a resource endpoint that fetches a different resource based on an lti state breaks basic http assumptions.

sadleb: GET https://platform.braven.org/peer_review_submissions/new doesn't feel like it breaks basic HTTP assumptions imo.

rshipp: Peer reviews are tied to a specific course, because a single student can be in multiple courses and do peer reviews for all of them. There will be different peers per course, and the Submission objects the form creates will be tied to a different course. If you go to the form in a course where you've already submitted, it redirects you to a "thank you" page instead. All those things being different based on something not in the url doesn't seem right. If we nest, we get /courses/10/peer_review_submissions/new and /courses/20/peer_review_submissions/new, and it's immediately obvious that you're requesting (or POSTing to) different resources.

rshipp: just had a thought, writing it down for discussion: since the problem seems to be "we have to parse URLs in course launch", what if we had a static url like `/lti/peer_review_submissions/new` that we tell Canvas about, and it reads the state and redirects to `/courses/10/peer_review_submissions/new`? That makes me feel better about the URL, and it means we don't have to update canvas. Is it overkill?

PR #469, opened 2020-11-18
Implement showing the Waivers forms

sadleb:
### Description

Task: https://app.asana.com/0/1174274412967132/1198301329741772

This PR implements [this waivers flow](https://docs.google.com/document/d/1-CzbBGFIqgXWC9HZ0W0neXrrcqoOcgWkteeXKF4OSm8/edit#heading=h.151cfk32uxbj).

I put the specs in a [separate PR](https://github.com/bebraven/platform/pull/472). Not sure if that's helpful or not.

### Motivation and Context
We want to lock down our course content, with the exception of the first module, until after Fellows have signed their NDA and liability waivers. We also want them to sign a media release waiver, but they aren't required to. To make this work, we:

* [Publish a "Waivers" assignment to Canvas from the Course Management page](https://github.com/bebraven/platform/pull/458)
* Move the assignment into a Canvas grouping module that has requirements for completion set.
<details><summary>Screenshot</summary>

</details>

* Then when Fellow opens this assignment, we provide a page asking them to launch the form. We have to do this b/c the e-Signature functionality requires access to cookies and we don't have that inside Canvas.
* When launched, we look up the FormAssembly ID of the form to launch on the Salesforce Program: 
<details><summary>Screenshot</summary>

</details>

* We use the Form Assembly API to get the form contents and render it on a platform page. We have to do this so that we can insert the `LtiLaunch.state` and tie the submission back to the Canvas assignment.
* There is a flow of several forms that the Fellow progresses through and when they get to the last one, it must be configured to POST back to our `create` action.
<details><summary>Screenshot</summary>

</details>

* When that happens, we create a submission in Canvas for this assignment. This is what allows the Pre-Requisites Canvas grouping module to be considered complete and unlock the other Canvas grouping modules (aka the rest of our course content).
* We also use whether there was a Canvas submission to determine whether to show them the button to launch the waivers.

### Types of changes

### Checklist:


app/assets/stylesheets/custom_contents.scss
```diff
@@ -37,7 +37,7 @@
 
 .bz-assignment {
     width: unset !important;
-    max-width: 672px;
+    max-width: 42rem;
```
sadleb, line 40: Unrelated to this PR. Just making this match what's in bz-assignment since I do the same thing for the form_assembly CSS below.

app/assets/stylesheets/form_assembly.scss
```diff
@@ -0,0 +1,28 @@
...
+    overflow-y: auto;
+
+    h1, h2, h3, h4, h5, h6 {
+      margin-top: 1rem;
+    }
+
+    p {
+      margin-top: 0.5rem;
+    }
+}
```
sadleb, line 28: This is just the bare minimum to make the pages in this flow look OK. We need to do a big CSS cleanup. Outside the scope of this PR.

app/controllers/waivers_controller.rb
```diff
@@ -35,14 +53,43 @@ def launch
...
+    # We just show a message about the Waivers only being loaded for a launched Course.
+    return if @base_course.is_a?(CourseTemplate)
+
+    if params[:tfa_next].present?
+      @form_head, @form_body = FormAssemblyAPI.client
+        .get_next_form_head_and_body(params[:tfa_next])
+    else
+      @form_head, @form_body = FormAssemblyAPI.client
+        .get_form_head_and_body(form_assembly_info.waivers_form_id, participantId: form_assembly_info.participant_id)
+    end
```
sadleb, line 69: The initial launch uses the `else` clause to show the form. Then as you proceed through each form before getting to the last one, the "submit" or "next" buttons keep passing the `tfa_next` param with the path to what you should show next.

app/policies/waivers_policy.rb
```diff
@@ -4,6 +4,18 @@ def launch?
...
+  end
+
+  def create?
+    !!user
+  end
+
+  def completed?
+    !!user
+  end
+
```
sadleb, line 18: It doesn't feel worth authorizing the `@base_course` and ensuring that this user is enrolled in it. Lmk if you disagree.

lib/lti_advantage_api.rb
```diff
@@ -44,6 +45,15 @@ def create_score(lti_score)
     JSON.parse(response.body)
   end
 
+  # Gets the Result object for a score created using the `create_score` method
+  # or nil if there is no score
+  #
+  # See: https://canvas.instructure.com/doc/api/result.html
+  def get_result()
+    results = get_line_item_for_user(@canvas_user_id)
```
sadleb, line 53: We'll likely want to get rid of `get_line_item_for_user` and just have this `get_result` method. I'm leaving it alone for now b/c it's used in specs even though it's not used functionally and this PR is already too big. The idea for it was that we'd be using the `LtiAdvantageAPI` to look up the submitted data more generally and we'd need the ability for a TA to view the Fellow's submission.

lib/salesforce_api.rb
```diff
@@ -135,6 +150,13 @@ def get_participants(program_id = nil, contact_id = nil, last_modified_since = n
     JSON.parse(response.body)
   end
 
+  def get_participant_id(program_id, contact_id)
+    soql_query = "SELECT Id FROM Participant__c " \
+                 "WHERE Program__r.Id = '#{program_id}' AND Contact__r.Id = '#{contact_id}'"
+    response = get("#{DATA_SERVICE_PATH}/query?q=#{CGI.escape(soql_query)}")
+    JSON.parse(response.body)['records'][0]['Id']
+  end
```
sadleb, line 158: There are already like 3 other methods in this API to get participants and participant info. One would sort of work, `find_participant`, but it wasn't fully implemented to handle the situation where a single person has multiple Participant records and I didn't want to muck with it at the moment b/c it's all the Sync To LMS stuff which is a bit complicated to follow. At some point, we may want to make the participant related part of this API a bit cleaner, but not today batman.

lib/lti_score.rb
```diff
@@ -48,6 +48,18 @@ def self.new_survey_submission(canvas_user_id, submission_url)
...
+  # The score is for full credit.
+  def self.new_waivers_submission(canvas_user_id, submission_url)
+    submission_data = {
+      new_submission: true,
+      submission_type: 'basic_lti_launch',
+      submission_data: submission_url,
+    }
+    # The number 100 is arbitrary. This is just full credit: 100/100.
+    generate(canvas_user_id, SUBMITTED, FULLY_GRADED, submission_data, 100, 100)
+  end
```
jessouyangbraven, line 61: `new_full_credit_submission `? https://github.com/bebraven/platform/pull/468#discussion_r528960304

config/routes.rb
```diff
@@ -83,9 +83,10 @@
 
   resources :lesson_contents, only: [:new, :show, :create]
 
-  resources :waivers, only: [] do
+  resources :waivers, only: [:new, :create] do
     collection do 
       get :launch
+      get :completed
```
jessouyangbraven, line 89: I don't know if it's worth doing in this PR or queueing up a task in Asana, but these new actions would be part of `WaiverSubmissionsController` and the routes would look like the ones added in #468:

```
  resources :courses, controller: 'base_courses', type: 'Course' do
    resources :waiver_submissions, only: [:new, :create]
  end

  resources :waiver_submissions, only: [:show]
```

app/views/waivers/launch.html.erb
```diff
@@ -1,6 +1,14 @@
...
+<div class="form-instructions-container">
+  <p>In order to participate in Braven, please sign and submit the waivers.</p>
+  <p><a href="<%= @new_waivers_url %>" class="btn btn-primary btn-lg" target="_parent">Click here to get started</a></p>
+</div>
+
+<% else %>
+
+<%= render partial: 'completed' %> 
+
+<% end %>
```
jessouyangbraven, line 14: Detect previous submission and `redirect` in the controller?

app/policies/waivers_policy.rb
jessouyangbraven: I think this is fine, but defer to @rshipp on the pundit policy stuff.

app/controllers/waivers_controller.rb
```diff
@@ -35,14 +56,63 @@ def launch
     authorize :waivers
   end
 
+  # Show the FormAssembly waiver form for them to sign.
+  #
+  # GET /waivers/new
+  def new
```
jessouyangbraven, line 62: I mentioned this in the `routes.rb` comment, but these new actions would go in `WaiverSubmissionsController` and the `#new` action would check if there was a previous submission and redirect to `#completed`.

app/controllers/waivers_controller.rb
```diff
@@ -35,14 +56,63 @@ def launch
...
   end
 
+  # Show the FormAssembly waiver form for them to sign.
+  #
+  # GET /waivers/new
+  def new 
+    authorize :waivers
+
+    # We just show a message about the Waivers only being loaded for a launched Course.
+    return if @base_course.is_a?(CourseTemplate)
```
jessouyangbraven, line 66: Probably not for this PR, but I think we should discuss how we want to handle different logic for Courses/Templates. I think we have a lot of options and it'd be best to be consistent:

Some things we currently do:
 - Nest under `type: Courses` in `routes.rb`
 - Some `before_action` validation in the controller
 - Conditionals in controller/views

I don't know if it makes sense to eventually have entirely separate controller/views for {waivers, peer reviews, submittable things} in {courses, templates} instead of the conditionals.

app/controllers/waivers_controller.rb
```diff
@@ -35,14 +56,63 @@ def launch
...
+      completed_waivers_url(protocol: 'https'),
+    )
+    lti_advantage_api_client.create_score(lti_score)
+  end
+
+  # Shows a "thank you for submitting" page
+  # GET /waivers/completed
+  def completed
+    authorize :waivers
+  end
```
jessouyangbraven, line 106: nit: we're using `*SubmissionsController#show` for this in impact surveys and peer reviews

jessouyangbraven (commented):
This looks reasonable to me. The FormAssembly stuff is still really foreign to me, so I'll see if @rshipp has comments on that part of the code.

My biggest question is whether we move this code to `WaiverSubmissionsController` so we're consistent with the rest of our code. I'm cool if we want to queue up a tech debt task for it, though, and get this merged so we can demo.

app/controllers/waivers_controller.rb
sadleb: I did this b/c it's a static endpoint and there is no `id` of a waiver to show. I guess I could override `show` in the routes and not require an `id`, but that seems wonky.

app/controllers/waivers_controller.rb
sadleb: Hmmm. I see what you're saying. I'm struggling here b/c there is no Waivers or WaiverSubmission model. This whole WaiversController is really just a static endpoint to facilitate connecting platform, canvas, and form assembly. There are no models to show or create.

I raised this on standup asking if it made sense to create a waiver `ActiveModel` and a `WaiverSubmission` where we just store that it happened but there is no data. Do you think it's worth doing that just to make it cognitively consistent with the other things we submit that have actual data?

app/controllers/waivers_controller.rb
sadleb: Yeah, figuring this out is rising in priority. Task here to take next steps: https://app.asana.com/0/1174274412967132/1199343471937405

app/views/waivers/launch.html.erb
sadleb: Good call. I'll do that instead.

app/controllers/waivers_controller.rb
```diff
@@ -65,11 +135,54 @@ def unpublish
...
+  # "tfa_26" depending on how many and what order you add fields. See:
+  # https://help.formassembly.com/help/prefill-through-the-url
+  #
+  # TODO: if you try to go Back in the browser after submitting the final e-signature review
+  # form, the call to the FormAssembly API with that tfa_next param returns an empty body and this 
+  # throws an exception. Do something more graceful: https://app.asana.com/0/1174274412967132/1199231117515065
+  def setup_body
+    doc = Nokogiri::HTML::DocumentFragment.parse(@form_body)
+    form_node = doc.at_css('form')
+    form_node.add_child('<input type="hidden" value="' + @lti_launch.state + '" name="state" id="state">')
```
rshipp, line 160: this is high-severity xss. Since state is user-controlled, you can inject anything you want in there and it'll show up as html on our site. A nefarious actor could link to `platform.braven.org/this_page?state="><script>evil()</script>`. Since the valid characters in `state` are pretty limited, can we maybe remove any character that's not in `\w-`? We don't need to worry about escaping things because other characters would never show up in a legit `state`.

app/policies/waivers_policy.rb
rshipp: I don't think it's merge-blocking, but just for consistency's sake it seems better to have everything that's tied to a course check enrollment. I opened a task here https://app.asana.com/0/1174274412967132/1199344732354185/f - if we had it so you could just call like `EnrollmentPolicy.new(user, course).enrolled?`, would you use it here?

app/controllers/waivers_controller.rb
jessouyangbraven: > Do you think it's worth doing that just to make it cognitively consistent with the other things we submit that have actual data?

Honestly, I don't know.

Waivers are different from all the other things that we put in Canvas as assignments, so it might be more overhead than it's worth.

I lean towards not changing what you have in this PR for now and changing it when there's a clear technical advantage or actual user story for doing it.

app/views/waivers/launch.html.erb
```diff
@@ -1,6 +1,14 @@
...
 
-<p>In order to participate in Braven, please sign and submit the waivers.</p>
-<br/>
-TODO: Not Yet Implemented
-<p><a href="#" class="btn btn-primary btn-lg" target="_parent">Click here to get started</a></p>
+<% unless @waivers_already_signed %>
+
+<div class="form-instructions-container">
+  <p>In order to participate in Braven, please sign and submit the waivers.</p>
+  <p><a href="<%= @new_waivers_url %>" class="btn btn-primary btn-lg" target="_parent">Click here to get started</a></p>
```
rshipp, line 7: I haven't seen target="_parent" before. I know target="_blank" has accessibility issues, idk if they also apply here. Does axe have any input on this if you check the browser console?

app/controllers/waivers_controller.rb
sadleb: This reads the state from the `@lti_launch`. That's not user controlled. We generate it when we insert the record and then use it to look it up again.

sadleb: I'll add a note about never switching it to read from the query param though.

app/views/waivers/launch.html.erb
sadleb: This is a place where we have no choice. We can't open this form embedded in Canvas b/c we need third party cookie access. It's similar to the LinkedIn Authorization issue we had. In that situation we popped open a new window similar to how other OAuth flows work. A new window didn't seem worth it here b/c it's a lot more than just entering a username/password and hitting submit.

app/policies/waivers_policy.rb
sadleb: Yeah, if it was that easy. I'll link the task in the code.

config/routes.rb
sadleb: I'm going to introduce a `WaiverSubmissionsController` but the routes will be slightly different. We want to use a static endpoint that doesn't need to be adjusted when we launch a new program. So they won't be embedded under a course (or base course). We read that out of the `LtiLaunch` context.

sadleb:
```
resources :waiver_submissions, only: [:new, :create] do
  collection do
    get :launch
    get :completed
  end
end
```

sadleb: Note that we don't use `show` b/c there is no model to show and hence no `id` to pass to the route. I'm using `completed` instead as a generic static endpoint.

app/controllers/waivers_controller.rb
sadleb: I ended up putting it in a method with comments so that someone won't accidentally switch to reading it from params without thinking about this.

app/views/waivers/launch.html.erb
rshipp: A new window is way worse, to be clear. Does axe tell you anything about target="_parent" though? There's usually a workaround that lets you use things in a way that is accessible.

PR #472, opened 2020-11-20
Add specs for the Waivers PR

sadleb:
### Description
Adds specs for [this waivers PR](https://github.com/bebraven/platform/pull/469).


spec/factories/users.rb
```diff
@@ -19,6 +19,7 @@
     
     factory :registered_user do
       sequence(:password) { |i| "password#{i}" }
+      sequence(:salesforce_id) { |i| "003#{i}100001iyv8IAAQ" }
```
sadleb, line 22: We never had any specs that needed this before, but it's a required value for a fully registered user.

spec/factories/salesforce_fellow_form_assembly_info.rb
```diff
@@ -0,0 +1,35 @@
...
+  # Represents the info returned from SaleforceAPI.get_fellow_form_assembly_info(...)
+
+  factory :salesforce_fellow_form_assembly_info, class: Hash do
+    skip_create # This isn't stored in the DB.
+
+    transient do
+      sequence(:program_id) { |i| "a2Y#{i}1000001HY5mEAG" }
+    end
+
+    sequence(:Id) { program_id }
```
sadleb, line 12: This is a little hacky, but Salesforce always sends back `Id` no matter what and I couldn't reference `Id` in the `attributes` key b/c it thinks I'm trying to use a class in ruby.

sadleb, line 12: This is a little hacky, but Salesforce always sends back Id no matter what and I couldn't reference Id in the attributes key b/c it thinks I'm trying to use a class in ruby.

spec/factories/lti_score.rb
```diff
@@ -8,10 +8,10 @@
     userId { 50 }
     # The Current score received in the tool for this line item and user, scaled to
     # the scoreMaximum
-    scoreGiven { 50 }
+    scoreGiven { 100.0 }
     # Maximum possible score for this result; it must be present if scoreGiven is
     # present.
-    scoreMaximum { 50 }
+    scoreMaximum { 100.0 }
```
sadleb, line 14: I figured it's better to use the realistic values we're currently using. Makes no difference.

spec/factories/lti_result.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    after :build do |the_object, evaluator|
+      # This isn't shown in the API docs, but the extension key with the submission
+      # data is sent back if it was specified when calling create_score which hits:
+      # https://canvas.instructure.com/doc/api/score.html#method.lti/ims/scores.create
+      the_object['https://canvas.instructure.com/lti/submission'] = {
+        'new_submission' => true,
+        'submission_data' => evaluator.submission_data,
+        'submission_type' => evaluator.submission_type
+      }
```
sadleb, line 33: Have to do this in `after :build` b/c the `'https://canvas.instructure.com/lti/submission'` key isn't a valid Ruby variable name.

PR #476, opened 2020-11-24
Update sync for nlu

skaex:
### Description
This updates the way sync to lms works for NLU program.

### Motivation and Context
NLU has a slightly different requirement for how fellows join our system. The fellow need to have their accounts created when running a sync to lms.

https://app.asana.com/0/1185361437008970/1199201476879535

### Types of changes

### Checklist:

### DIFF
--- a/app/services/account_creator.rb
+++ b/app/services/account_creator.rb
@@ -1,12 +1,8 @@
 # frozen_string_literal: true
 
-class AccountCreator
+class AccountCreator # rubocop:disable Style/Documentation
   def initialize(sign_up_params:, for_nlu: false)
-    @salesforce_contact_id = sign_up_params['salesforce_id']
-    @password_params = {
-      password: sign_up_params['password'],
-      password_confirmation: sign_up_params['password_confirmation']
-    }
+    @sign_up_params = sign_up_params
     @salesforce_contact = nil
     @for_nlu = for_nlu
   end
@@ -20,7 +16,9 @@ class AccountCreator
 
   private
 
-  attr_reader :salesforce_contact_id
+  def salesforce_contact_id
+    @sign_up_params['salesforce_id']
+  end
 
   def setup_portal_user!
     SetupPortalAccountJob.perform_later(salesforce_contact.id)
@@ -45,10 +43,16 @@ class AccountCreator
 
   def password_params
     if @for_nlu
+      # NLU students don't need a password on platform. They use their SSO to
+      # log into portal not the platform. This is why we just create a random
+      # password on platform just to have user on platform but not for log in
       password = Devise.friendly_token[0, 20]
       { password: password, password_confirmation: password }
     else
-      @password_params
+      {
+        password: @sign_up_params['password'],
+        password_confirmation: @sign_up_params['password_confirmation']
+      }
     end
   end
 
--- a/app/services/account_creator.rb
+++ b/app/services/account_creator.rb
@@ -1,13 +1,14 @@
 # frozen_string_literal: true
 
 class AccountCreator
-  def initialize(sign_up_params:)
+  def initialize(sign_up_params:, for_nlu: false)
     @salesforce_contact_id = sign_up_params['salesforce_id']
     @password_params = {
-      'password' => sign_up_params['password'],
-      'password_confirmation' => sign_up_params['password_confirmation']
+      password: sign_up_params['password'],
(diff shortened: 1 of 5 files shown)


app/services/account_creator.rb
```diff
@@ -1,13 +1,14 @@
...
+  def initialize(sign_up_params:, for_nlu: false)
     @salesforce_contact_id = sign_up_params['salesforce_id']
     @password_params = {
-      'password' => sign_up_params['password'],
-      'password_confirmation' => sign_up_params['password_confirmation']
+      password: sign_up_params['password'],
+      password_confirmation: sign_up_params['password_confirmation']
     }
     @salesforce_contact = nil
+    @for_nlu = for_nlu
```
sadleb, line 11: I think it would be cleaner if you stored the `sign_up_params` in an instance variable. Then in `password_params` do all the logic to convert `sign_up_params` to `password_params`.

app/services/account_creator.rb
```diff
@@ -42,6 +43,15 @@ def platform_user_params
              })
   end
 
+  def password_params
+    if @for_nlu
+      password = Devise.friendly_token[0, 20]
+      { password: password, password_confirmation: password }
```
sadleb, line 49: Can you add a comment about why this is? E.g. we don't actually use the password for NLU, they go through their own SSO server but we need some password to be there for Devise. Or is that not what this is doing?

app/services/setup_portal_account.rb
```diff
@@ -36,7 +37,11 @@ def sync_portal_enrollment!
   end
 
   def send_confirmation_notification(user)
-    user.send_confirmation_instructions
+    if sf_program.nlu?
+      user.skip_confirmation!
```
sadleb, line 41: If someone just stumbled across this code for the first time they would have no clue why we skip it. Can you add a comment about why?

app/services/setup_portal_account.rb
```diff
@@ -94,9 +99,11 @@ def docusign_template_id
   end
 
   def portal_username
-    # TODO: For now it's email but there are cases like NLU where is
-    # #{user_student_id}@nlu.edu
-    sf_participant.email
+    if sf_program.nlu?
+      "#{sf_participant.student_id}@nlu.edu"
```
sadleb, line 103: Ditto on comments. If we can't more cleanly split the NLU specific logic out from the normal logic with a big comment about how NLU is different and what the implications are, then I think you need to add comments inline where you do the special logic to help folks understand why this code is here. Aka: this is the email format that the NLU SSO server sends sends in the authentication exchange and needs to match with the email Canvas has. Actually, is that the case? Did you test this? I have a vague recollection that we had to do something special somewhere to make sure the NLU SSO server actually could successfully authenticate with Canvas and match up the user.

lib/salesforce_api.rb
```diff
@@ -98,7 +102,7 @@ def patch(path, body, headers={})
 
   def get_program_info(program_id)
     soql_query = 
-      "SELECT Id, Name, Target_Course_ID_in_LMS__c, LMS_Coach_Course_Id__c, School__c, " \
+      "SELECT Id, Name, Target_Course_ID_in_LMS__c, LMS_Coach_Course_Id__c, School__c, Program_Shortname__c, " \
```
sadleb, line 105: Why use shortname instead of school? Is school just a Salesforce ID? Seems more brittle to rely on the shortname vs the actual school name.

sadleb (changes requested):
I think this stuff is confusing enough without the full higher level context that some clarifying comments are needed. I also think we need to write a `spec` or two for this to make it clear that the intended behavior is b/c this seems like a really likely place that some future developer will want to refactor and cleanup with confidence that they aren't breaking the expected logic.

lib/salesforce_api.rb
skaex: I'll change this. However, IMO the shortname is used internally and we have more control about how it'll change. Even on SF, it looks easier to update the school name than the shortname.

sadleb: Go with whichever is less brittle. If you think that's shortcode, works for me.

sadleb (approved):
If this were for the Highlander codebase I’d want to spend a little more on the specs, but I think this is good enough to merge for the Portal code. Thanks for taking action on all the feedback!

PR #483, opened 2020-12-02
Add Modules to Canvas assignments through Course Management page

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199394675251953

### Motivation and Context

This PR moves inserting Modules into Canvas courses to our course management page. This helps centralize how we manage content hosted on Canvas Cloud.

Because a module can be in multiple courses, we need to add a join table, e.g. `course_rise360_module_versions`. This is similar to how we handle `CustomContent`, e.g. projects and impact surveys.

- Added specs
- Use **Add Module** on the course's edit page:
Screen Shot 2020-12-09 at 2 46 15 PM

- Select a module from the dropdown (you can add new modules using **Add New** on the previous page):
Screen Shot 2020-12-09 at 2 46 22 PM

- See success message:
Screen Shot 2020-12-09 at 2 46 32 PM

- View on Canvas by clicking link:
Screen Shot 2020-12-09 at 2 47 55 PM

- Click **Publish Latest**, see success message (this button does nothing because we can't update a `Rise360Module` get, which is the next PR):
Screen Shot 2020-12-09 at 2 47 18 PM

- Click **Delete**, go through confirm prompt:
Screen Shot 2020-12-09 at 2 47 23 PM

- See success message for deleting:
Screen Shot 2020-12-09 at 2 47 29 PM

- Check assignment list for course in Canvas and module has been deleted there as well:
Screen Shot 2020-12-09 at 2 47 39 PM

### Types of changes

### Checklist:


app/controllers/concerns/publishable.rb
```diff
@@ -68,6 +74,10 @@ def unpublish
   end
 
 private
+  def set_canvas_assignment_id(canvas_assignment_id)
+    # By default, do nothing. Override it if your controller cares.
+  end
```
sadleb, line 79: I don't think we should leave it up to `Publishable` to default to a NOOP. I'd rather each controller make the decision. Plus, the `method_missing` stuff below will never catch a controller meaning to set it but forgetting to with this here.

sadleb: I haven't tried it, but this may more or less just work. It uses `model_class` and `instance_variable_name` from `DryCrud`.

app/controllers/lesson_contents_controller.rb
```diff
@@ -1,36 +1,64 @@
...
+
+  before_action :create, only: [:publish]
+  before_action :destroy, only: [:unpublish]
+
+  # before_action :set_lti_launch, only: [:show]
+  # skip_before_action :verify_authenticity_token, only: [:create, :show], if: :is_sessionless_lti_launch?
 
   def new
     authorize @lesson_content
   end
```
sadleb, line 21: My head isn't fully in this, but does `new` show the actual module content inside Canvas? If so, I don't think we want to switch the `layout` to admin. We'd want to move the `new` action to something equivalent to `ProjectSubmissionsController` - perhaps`ModuleSubmissionsController` ?

app/controllers/lesson_contents_controller.rb
```diff
@@ -1,36 +1,64 @@
...
-    @deep_link_return_url, @jwt_response = helpers.lti_deep_link_response_message(@lti_launch, lesson_content_url(@lesson_content))
+  def base_course
+    @base_course
   end
 
-  def show
-    authorize @lesson_content
-    # TODO: this may be while previewing the the Lesson before inserting it through the
-    # assignment selection placement. Don't configure it to talk to the LRS in that case.
-    # https://app.asana.com/0/search/1189124318759625/1187445581799823
```
sadleb, line 24: Note that we had to use the "preview" flow when publishing from the Assignment Selection placement simply b/c we needed an extra step to be able to create the deep linking response message and submit it back. When publishing through the Course Mgmt page, I think we can get rid of the preview and just mimic what Project and Survey publishing does in terms of making the Canvas assignment LTI launchable.

spec/controllers/rise360_modules_controller_spec.rb
```diff
@@ -24,28 +24,6 @@
...
-
-        redirect_url = Addressable::URI.parse(response.location)
-        expected_url =  Addressable::URI.parse(rise360_module_with_zipfile.launch_url)
-        expect(redirect_url.path).to eq(expected_url.path)
-
-        # Specific LRS query parameters are tested in LtiHelper
-        expect(redirect_url.query_values).not_to be_empty
-      end
-    end
-  end
```
jessouyangbraven, line 47: Moved to `rise360_module_versions_controller_spec.rb` and updated.

app/controllers/rise360_modules_controller.rb
```diff
@@ -3,20 +3,7 @@
...
-  def show
-    authorize @rise360_module
-    # TODO: this may be while previewing the the Lesson before inserting it through the
-    # assignment selection placement. Don't configure it to talk to the LRS in that case.
-    # https://app.asana.com/0/search/1189124318759625/1187445581799823
-    url = Addressable::URI.parse(@rise360_module.launch_url)
-    url.query_values = helpers.launch_query
-    redirect_to url.to_s
-  end
 end
```
jessouyangbraven, line 9: Moved to `rise360_module_versions_controller.rb`'s `#lti_show` action.

This means that end-users never see `Rise360Module`s, only `Version`s of them, which is what we want. 

It we ever do want `show` back in this controller, we no longer need to attach the `launch_query` parameters, etc. so we don't communicate with the LRS.

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    @course_rise360_module_version.rise360_module_version.name
+  end
+
+  def lti_launch_url
+    authorize @course_rise360_module_version
+    lti_show_rise360_module_version_url(
+      @course_rise360_module_version.rise360_module_version,
+      protocol: 'https',
+    )
+  end
```
jessouyangbraven, line 54: There's some redirection here that's a bit confusing. I'm open to feedback on how to make this more obvious.

We use the `#lti_show` endpoint for the module version when creating the Canvas assignment. This is because we need the dynamic `params[:state]` to construct the `launch_query` to communicate with the LRS, which we won't have until the end-user actually goes to view the module version.

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    rise360_module = Rise360Module.find(params[:rise360_module_id])
+    @course_rise360_module_version = CourseRise360ModuleVersion.create!(
+      course: @course,
+      rise360_module_version: rise360_module.create_version!(@current_user),
+    )
+  end
+
+  def destroy
+    # FIXME: Dry crud?
+    @course_rise360_module_version = CourseRise360ModuleVersion.find(params[:id])
```
jessouyangbraven, line 68: For the life of me, I can't get this work, even after staring at the `dry_crud.rb` code. 

I tried adding `before_action :set_model_instance, only: [:destroy]` to see if it'd fix this, but it doesn't.

Help, @sadleb?

sadleb: This is because the `destroy` action is not being called through Rails. The action called is `unpublish`. In other words, this line isn't running in `DryCrud`:
```
before_action :set_model_instance, only: [:show, :edit, :update, :destroy]
```

sadleb: I *think* it may be ok to add `publish` and `unpublish` to that list of actions that DryCrud knows about. I feel like we had it there before at one point. @rshipp can you think of any reason not to add these two non-standard actions to `DryCrud`?

sadleb: The other option is to make `Publishable` destroy the model on `unpublish`. I don't think we'd want to blanket assume that though so I think we'd put it behind a `destroy_on_unpublish` method that defaults to false but controllers where it's safe and intended can have it return true. It's safe here b/c it's a join between the course and module version, so it doesn't make sense to have one without the other.

sadleb: E.g. something like this in `Publishable`
```
  def unpublish

    CanvasAPI.client.delete_assignment(
      course.canvas_course_id,
      params[:canvas_assignment_id],
    )
  
    if self.respond_to?(:destroy_on_unpublish) and destroy_on_unpublish
      instance_variable.destroy!
    end
      
    respond_to do |format|
      format.html { redirect_to(
        edit_course_path(course),
        notice: message % { subject: assignment_name, verb: 'deleted from' }
      ) }
      format.json { head :no_content }
    end
  end
```
and this in `CourseRise360ModuleVersionController`
```
  # For Publishable
  def destroy_on_unpublish 
    true
  end
```

app/controllers/concerns/publishable.rb
```diff
@@ -81,4 +85,8 @@ def method_missing(name, *args, &block)
   def method_missing_error_msg(name)
     "Publishable module expects method `#{name}` to be defined for #{self.class}. It's missing."
   end
+
+  def instance_variable
```
sadleb, line 89: It prob make more sense to define this in `DryCrud` right next to where `instance_variable_name` is defined and remove it from `Attachable` and `Submittable` too.

app/controllers/course_rise360_module_versions_controller.rb
sadleb: I wouldn't put more time into this at the moment. Let's tackle this holistically as part of: https://app.asana.com/0/1174274412967132/1199352155608256

app/controllers/rise360_module_versions_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  before_action :set_lti_launch, only: [:lti_show]
+
+  def lti_show
+    authorize Rise360ModuleVersion
+    @rise360_module_version = Rise360ModuleVersion.find(params[:id])
+    url = Addressable::URI.parse(@rise360_module_version.launch_url)
+    url.query_values = helpers.launch_query
+    redirect_to url.to_s
+  end
```
sadleb, line 16: It's a little confusing how the other ones do this in the `XXXSubmissionsController` but Modules do them right on ModuleVersion controller. Can you add a comment about why this is different and mention that the plan is to make this a redirect to the particular `CourseRise360ModuleVersionsController` as part of https://app.asana.com/0/1174274412967132/1199352155608256 ?

sadleb (approved):
I have some comments / suggestions, but nothing merge blocking.

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    rise360_module = @course_rise360_module_version.rise360_module
+    @course_rise360_module_version.update!(
+      rise360_module_version: rise360_module.create_version!(@current_user),
+    )
+
+    respond_to do |format|
+      format.html { redirect_to(
+        edit_course_path(course),
+        notice: message % { subject: assignment_name, verb: 'updated in' }
+      ) }
```
rshipp, line 34: i didn't know you could redirect inside a respond like this. is this railsy? if so I want to steal it for the project submissions /new redirect.

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      format.json { head :no_content }
+    end
+  end
+
+  # For Publishable
+  def course
+    @course
+  end
```
rshipp, line 42: is there a benefit to calling this as a `course` method instead of a `@course` variable? should we always be defining accessors like this?

rshipp: follow-up, is it ok to have methods in controllers that aren't actions?

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def assignment_name
+    @course_rise360_module_version.rise360_module_version.name
+  end
+
+  def lti_launch_url
+    authorize @course_rise360_module_version
```
rshipp, line 49: is this authorize here because it's missing in the action(s) that call this method?

app/controllers/course_rise360_module_versions_controller.rb
rshipp: I'm confused by the fact that `create` and `destroy` in this controller are being used as callbacks, and not actions. I'm leaning towards the last option @sadleb described above, and deleting those two methods from this controller.

app/controllers/rise360_module_versions_controller.rb
rshipp: I don't think this needs to be `lti_show`. I think i called course_resources `lti_show` only because there is no `params[:id]`, and which resource you show depends on the lti launch. which would also go away with that lti redirect controller task.

app/views/course_rise360_module_versions/_list.html.erb
```diff
@@ -0,0 +1,25 @@
...
+        </tr>
+      </thead>
+
+      <tbody>
+        <% course_rise360_module_versions.each do |crmv| %>
+          <tr>
+            <td><%= link_to crmv.rise360_module_version.name, crmv.canvas_url %></td>
+            <td class="text-right smaller">
+              <%= link_to 'Publish Latest', course_course_rise360_module_version_path(@course, crmv.id), method: :put %> |
+              <%= link_to 'Delete', unpublish_course_course_rise360_module_version_path(@course, id: crmv.id), method: :delete,
```
rshipp, line 17: why is this not either `unpublish_course_rise360...` or `destroy_course_course_rise360...`? i thought "unpublish object" meant "destroy the join record between course and object"?

app/controllers/course_rise360_module_versions_controller.rb
jessouyangbraven: Thanks @sadleb. That makes sense! I'm going to just leave fetching the model instance in the `#destroy` action for now because it's simpler and it's the only place we want that behavior so far.

app/controllers/course_rise360_module_versions_controller.rb
jessouyangbraven: >is there a benefit to calling this as a course method instead of a @course variable? should we always be defining accessors like this?

I went back and forth on which to use when I added `Publishable`. @sadleb was for `course` at the time, if I remember correctly. I think it was to reduce the dependence between the Concern and the Controller. I think all the methods currently implemented are just `@course`, though.

Worth discussing in our next stand-up?

jessouyangbraven: >follow-up, is it ok to have methods in controllers that aren't actions?

...probably not great. I think they should all be `private`. I can move them.

app/controllers/course_rise360_module_versions_controller.rb
jessouyangbraven: Good catch. That's a copy-pasta-typo and shouldn't be there.

app/controllers/course_rise360_module_versions_controller.rb
sadleb: The `course` thing was about reducing dependency and making sure you use `Publishable` and other concerns correctly. The method_missing stuff in Publishable makes it so that if you blindly include the module without realizing it’s requirements, it’ll give you a nice hint about what you didn’t do correctly. 

Another way to accomplish this would have been to put calls to `respond_to?` in all the places we use these instance variables inside `Publishable` but we still had the `lti_launch_url` method. This seemed cleaner? Happy to discuss in standup though bc I’m still on the fence.

sadleb: I’m fighting my inner “rails magic is good” vs “everything needs to be declared and configured so you see what is going on” demons.

sadleb: I want to lay out patterns that we blindly mimic for future use. I could see us moving the accessors to the concern to doing the `respond_to?` check there in order to make it easy to see what went wrong.

PR #484, opened 2020-12-02
Get rid of Course vs CourseTemplate models and routes

sadleb:
### Description

Gets rid of `Course` vs `CourseTemplate` which use Single Table Inheritance and are stored in `BaseCourse`.
Task: https://app.asana.com/0/1174274412967132/1199355052634253

**IMPORTANT**: this will break all previously published assignments that had `base_course` in their route. E.g. Surveys, Projects, Peer Reviews, etc.

Note that I wasn't sure what to call a non-launched course in the UI so I just stuck with Course Template here. If anyone has a better idea, I'm all ears. I'm thinking we can just iterate on that as we get further along with trying out the launch logistics and course template organization and management, surfacing the pain points, and introducing features to make everything easier.

### Motivation and Context
Originally, we thought we were going to treat Course Templates drastically different from launched Courses. This had begun to create problems when populating a Course Template b/c things didn't work and we needed to come up with a plan for how to manage that. Piling on, we got a good deal of pushback on the general concept of "A launched course cannot be edited." It's looking very likely that we'll have to implement certain types of edits that are allowed on launched courses that are live. We also got feedback that the general approach for iterating on and launching variations of the product (e.g. BravenX, InternPro, etc) would be to always start from the main Accelerator template b/c that's where the core edits to the content happen. Maintaining separate templates and having to iterate on them in parallel was a sub-par approach. Lastly, it made the code really non-intuitive and hard to work with.

For all these reasons, we're backing out that split in the code. Everything will be course. You can launch a new course from a course, which blows away the sections and users. You can launch a program from two courses which sets up the Salesforce connection and other standard "new launch" stuff. We added a `is_launched` column to indicate if something is launched so that we can treat it different and prevent dangerous actions, like deleting an assignment where folks have already submitted things. But this will allow a lot more flexibility. 

### Checklist:


app/controllers/cas_controller.rb
```diff
@@ -177,6 +179,8 @@ def loginpost
 
           begin
             logger.info("Redirecting authenticated user '#{@username}' at '#{@st.client_hostname}' to service '#{@service}'")
+            # Devise tends to flash a "You need to sign in or sign up before continuing." alert.
+            flash.delete(:alert)
```
sadleb, line 183: The iframed LinkedIn auth button feature spec was breaking. I think I introduced this in the CSS cleanup PR when I added `alert`'s to the root application layout but we never noticed b/c it's CI excluded. I got really into the weeds on this and it was too complicated to really figure out how to do this correctly, so this is a slight hack. It should get rid of those weird "You need to sign in or sign up before continuing." alerts that we occasionally see even though we're still logged in (from the end user standpoint).

app/controllers/courses_controller.rb
```diff
@@ -0,0 +1,115 @@
+class CoursesController < ApplicationController
```
sadleb, line 1: A lot of these actions are sort of left over from when we initially rolled this out and wanted dev to be easy but are essentially obsolete now. I queued up a task to clean them up more: https://app.asana.com/0/1174274412967132/1199503585782989

app/views/courses/index.html.erb
```diff
@@ -2,21 +2,21 @@
...
 ) %>
 
 <%= render(
   'list',
   title: 'Course Templates',
   button_label: 'Add Course Template',
-  button_url: new_course_template_path,
-  base_courses: @base_courses.course_templates,
+  button_url: new_course_path,
+  courses: @courses.course_templates,
```
sadleb, line 14: This is the main thing that still feels a little confusing. When splitting up the launched courses from the non-launched courses on the Course Management page I just stuck with the term `course_templates` (in the UI as well). Does this make sense or do you have any thoughts on how to make things more intuitive for both devs and designers?

app/views/courses/launch_new.html.erb
```diff
@@ -11,8 +11,8 @@
       </div>
 
       <div class="form-group">
-        <%= form.label :fellow_course_template_id, 'Fellow course template' %>
-        <%= form.collection_select :fellow_course_template_id, @course_templates, :id, :name, {include_blank: true}, class: 'form-control' %>
+        <%= form.label :fellow_source_course_id, 'Fellow course template' %>
+        <%= form.collection_select :fellow_source_course_id, @course_templates, :id, :name, {include_blank: true}, class: 'form-control' %>
```
sadleb, line 15: See https://github.com/bebraven/platform/pull/484/files#r534380291 for my comment on how this could be confusing. Ideas?

db/migrate/20201201033923_migrate_type_for_courses.rb
```diff
@@ -0,0 +1,23 @@
...
+    add_column :courses, :type, :string
+
+    # When rolling back, do the opposite of above.
+    Course.where(:is_launched => false).update_all(:type => 'CourseTemplate')
+    Course.where(:is_launched => true).update_all(:type => 'Course')
+
+    change_column_null :courses, :type, false
+    remove_column :courses, :is_launched
+  end
+end
```
sadleb, line 23: This is sort of the meat of the PR. It migrates `type` to `is_launched` and we use that in the handful of places where we want to treat a launch vs non-launched Course differently.

app/views/courses/index.html.erb
rshipp: i'm not sure, we might need to get input from Helen on this? I could see it being helpful to have unlaunched things at the top for easy access, but I can also see it being clearer to get rid of the word "template" entirely, since you can now launch from any course.

sadleb: Just spoke with Helen. We brainstormed a bit and she still likes Templates as the group of courses that haven't been launched (which may be test courses as well). 

However, we currently don't have a way to launch from a live course in the UI so we discussed that. The idea we agreed on is to switch the list of launched courses below to a list of launched programs. Over time, as we build tools to management execution and operations, that would make more sense if regional folks were going to navigate to their "program management" page. Then when you click on a program, a new page to launch that existing program as a new one for the next semester would be there. I'm going to queue this up as a future user story to make that change though since I *think* we would want to bring back a `Program` model rather than just hackily pulling the program information off a course to show what we need.

sadleb: https://app.asana.com/0/1174274412967132/1199503585782993

sadleb:
TODO: get rid of course mgmt path

PR #485, opened 2020-12-02
Rename LessonInteraction-> Rise360ModuleInteraction, add Rise360ModuleState

rshipp:
### Description

Part 1 of many towards removing the LRS dependency. Prep for #486.

Task: https://app.asana.com/0/1174274412967132/1197477442813577/f

### Types of changes

### Checklist:


app/models/module_state.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+class ModuleState < ApplicationRecord
```
sadleb, line 2: I think you should comment this class bc `ModuleState` is nearly the most vague name I can imagine for someone coming across this with no context.

We could name this `ModulePickUpWhereYouLeftOffNotUsingRetainedData` if we want to keep some of the pizzazz of the old code around 😂 /s

db/migrate/20201202182711_create_module_states.rb
```diff
@@ -0,0 +1,12 @@
+class CreateModuleStates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :module_states do |t|
+      t.bigint :canvas_course_id, null: false
+      t.bigint :canvas_assignment_id, null: false
```
sadleb, line 5: Should we add an index on assignment id or some combo of these columns to match the query. This is going to be hit a decent amount.

sadleb (approved):
Nothing blocking a merge. Just a couple nice-to-haves

db/migrate/20201202182711_create_module_states.rb
rshipp: whoops, I did this in the other PR. that would have been a disaster...

rshipp: backported the index to this branch. does an index with a uniqueness constraint still make search faster? i assume so but i don't actually know

app/models/module_state.rb
rshipp: is this comment enough?

sadleb: I think once this is renamed to Rise360ModuleState then this comment is enough. If it was just `Module` I think talking about what a Module is (aka not a Ruby `Module` ) or referencing the class that talks about it would be necessary.

db/migrate/20201202182711_create_module_states.rb
sadleb: Quick google on how Postgres unique constraint indexes seems to indicate yes, generally. If we're querying with all columns in the constraint in the where clause, I would assume yes, but I'm not sure what happens if you only query on a subset.

sadleb:
Lgtm after the "rename everything again".

PR #482, opened 2020-12-01
Remove LRS, use ProjectSubmissionAnswer

jessouyangbraven:
### Screenshots (if appropriate):

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #486, opened 2020-12-02
Add module state implementation, change proxy to mock

rshipp:
### Description

Part 2 of many towards removing LRS dependency. Based on #485.

Removes all mention of "LRS proxy", replacing it with an "LRS mock". Adds support for the state endpoint, alongside existing statement/interaction support. 

Note: This breaks projects until #482 is completed and merged. 

Task: https://app.asana.com/0/1174274412967132/1197477442813577/f

Testing:

1. Open a (working) Rise360 module assignment on Canvas.
2. Click through it, watch the states/statements get saved
3. Refresh the page, see that the state gets loaded and it picks up where you left off.

Screen Shot 2020-12-02 at 1 11 56 PM
Screen Shot 2020-12-02 at 12 41 09 PM
Screen Shot 2020-12-02 at 4 45 07 PM

### Types of changes

### Checklist:


app/controllers/lrs_xapi_mock_controller.rb
```diff
@@ -0,0 +1,20 @@
+require "lrs_xapi_mock"
+
+class LrsXapiMockController < ApplicationController
+  skip_before_action :verify_authenticity_token
+  wrap_parameters false # Disable putting everything inside a "lrs_xapi_mock" param. This controller doesn't represent a model.
+
+  def xAPI
+    authorize :lrs_xapi_mock, :xAPI?
```
rshipp, line 8: since we no longer have user_override, we don't need a complex policy here, so `xAPI_read?` and `xAPI_write?` are gone. the policy is now just `!!user`.

app/controllers/lrs_xapi_mock_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    # If this excepts, we let Rails handle it and return the 500.
+    if response_hash = LrsXapiMock.handle_request!(request, request.params['endpoint'], current_user)
+      # Note: have to be explicit with the content type here b/c it may "look" like json (and is) sometimes but
+      # the Rise360 package PUT/GET's it as plain data and fails if we say it's json. I'm at a loss as to why.
+      render plain: response_hash[:body], status: response_hash[:code], content_type: LrsXapiMock::OCTET_STREAM_MIME_TYPE and return
```
rshipp, line 14: since there's no GET /statements, there's never a JSON content_type.

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+#   stateId=suspend_data:       {"v":1,"d":[123,34]]}
+class LrsXapiMock
+
+  XAPI_STATE_API_ENDPOINT = 'activities/state'
+  XAPI_STATEMENTS_API_ENDPOINT = 'statements'
+  OCTET_STREAM_MIME_TYPE = 'application/octet-stream'
+  SUSPEND_DATA_STATE_ID = 'suspend_data'
+  CUMULATIVE_TIME_STATE_ID = 'cumulative_time'
+  BOOKMARK_STATE_ID = 'bookmark'
+  BOOKMARK_MAX_LENGTH = 256
```
rshipp, line 20: 256 is an arbitrary number. i just want this to be short enough that we at least inconvenience anyone trying to use this as an arbitrary file host.

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+  def self.handle_request!(request, endpoint, user)
+    Honeycomb.start_span(name: 'LrsXapiMock.handle_request!') do |span|
+      span.add_field('path', "/#{endpoint}")
+      span.add_field('user.id', user.id)
+      span.add_field('user.email', user.email)
+      span.add_field('method', request.method)
+
+      return unless request.method == 'GET' || request.method == 'PUT'
+
+      # Handle PUT for both endpoints.
```
rshipp, line 33: reactions on the if / case branching in this function? I thought about doing it other ways instead

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+            body: nil,
+          }
+        when XAPI_STATE_API_ENDPOINT
+          state_id = request.params['stateId']
+          activity_id = request.params['activityId']
+
+          # Save the state.
+          save_state!(activity_id, state_id, data, request.authorization, user)
+          return {
+            code: 204,  # No Content
```
rshipp, line 52: rails has :symbols for status codes, but I honestly find that more confusing here. I can change this if someone disagrees though.

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+      user: user,
+      state_id: state_id,
+    }
+
+    if module_state = Rise360ModuleState.find_by(attributes)
+      module_state.update!(value: data)
+    else
+      attributes[:value] = data
+      Rise360ModuleState.create!(attributes)
+    end
```
rshipp, line 133: rails 6 introduced `upsert`, which i thought would be great here, but it bypasses activerecord validations and callbacks 😞 . so instead I take advantage of the `hash` param syntax supported by `find_by` and `create!` and manually do "if exists, update; else create" logic.

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+  end
+
+  private_class_method def self.get_state(activity_id, state_id, authorization, user)
+    # Get course and assignment IDs from the auth header.
+    parsed = parse_activity_id(authorization)
+
+    Rise360ModuleState.find_by(
```
rshipp, line 140: find_by returns nil if it didn't find anything, so when we call `get_state` above we just check `if state` to see if it's a 200 or a 404.

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+      canvas_course_id: parsed[:course],
+      canvas_assignment_id: parsed[:assignment],
+      activity_id: activity_id,
+      user: user,
+      state_id: state_id,
+    )
+  end
+
+  private_class_method def self.save_interaction!(data, authorization, user)
```
rshipp, line 149: this function is unchanged except for refactoring out `parse_activity_id` so i can use it in `save_state!` too.

spec/controllers/lrs_xapi_mock_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe LrsXapiMockController, type: :controller do
```
rshipp, line 3: this was empty before too.

spec/feature/project_submissions/project_submit_spec.rb
```diff
@@ -33,13 +33,12 @@
...
   before(:each) do
     # We need to ignore these hosts because VCR can't record AJAX requests
-    lrs = Addressable::URI.parse(Rails.application.secrets.lrs_url)
     lti = Addressable::URI.parse(lti_launch.request_message.line_item_url)
 
     VCR.configure do |c|
       c.ignore_localhost = true
       # Ignore AJAX requests to platformweb, xapi_assignment, Canvas LTI
-      c.ignore_hosts Capybara.server_host, lrs.host, lti.host
+      c.ignore_hosts Capybara.server_host, lti.host
```
rshipp, line 41: lrs host is gone now.

spec/lib/lrs_xapi_mock_spec.rb
```diff
@@ -0,0 +1,334 @@
...
+
+    before(:each) do
+      allow(request).to receive(:query_parameters).and_return(query_parameters.with_indifferent_access)
+      allow(request).to receive(:raw_post).and_return(post_body) if method == 'PUT' || method == 'POST'
+      allow(request).to receive(:method).and_return(method)
+      allow(request).to receive(:authorization).and_return(authorization)
+      response # This is what makes the request for each test and sets this to the response
+    end
+
+    context 'unknown endpoint' do
```
rshipp, line 25: all the specs in this file are completely rewritten to match the new behavior.

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+      # Handle PUT for both endpoints.
+      if request.method == 'PUT'
+        data = request.raw_post
+
+        case endpoint
+        when XAPI_STATEMENTS_API_ENDPOINT
+          # Save relevant info from statements, for grading purposes.
+          save_interaction!(data, request.authorization, user)
+          return {
+            code: 204,  # No Content
```
rshipp, line 42: all this `204` behavior is based on what the LRS did previously.

app/controllers/lrs_xapi_mock_controller.rb
```diff
@@ -0,0 +1,19 @@
+require "lrs_xapi_mock"
```
jessouyangbraven, line 1: nit: `# frozen_string_literal: true`

lib/lrs_xapi_mock.rb
jessouyangbraven: nit: add this as a comment?

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+# Mocks responses to LRS/xAPI requests.
+#
+# This allows us to support Rise 360 content, without actually having an LRS.
+# Rise360 sends and receives payloads to the /activities/state xAPI endpoint that are things like
+# like the following, both JSON and just plain content, but for both it expects the Content-Type to
+# be 'application/octet-stream' (or perhaps something equivalent, but this is what the browser shows).
+#   stateId=cumulative_time:    123456 or NaN
+#   stateId=bookmark:           #/lessons/xQvXXBKjohmGHnPUlyck9VqMlQ0hgcgy
+#   stateId=suspend_data:       {"v":1,"d":[123,34]]}
+class LrsXapiMock
```
jessouyangbraven, line 12: nit: I actually am not sure, but is this supposed to be `module` instead of `class` since we're never meant to do `LrsXapiMock.new`?

lib/lrs_xapi_mock.rb
jessouyangbraven: I'm cool with what you have here.

app/controllers/lrs_xapi_mock_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class LrsXapiMockController < ApplicationController
```
jessouyangbraven, line 3: I know we agreed on having a single controller handle both state and statements, but I keep wondering if we should be using separate `Rise360Module{Interaction, State}Controller`s because the models are different.

That being said, I think we should merge this and revisit later. I'm mostly just thinking out loud now that I'm looking at the entirety of the code.

jessouyangbraven (approved):
This looks good to me!

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+    # arbitrary file host for nefarious purposes.
+    # The following are known Rise360 state IDs and their expected types.
+    # Raise an exception if any of them fail validation.
+    case state_id
+    when SUSPEND_DATA_STATE_ID
+      # E.g.: {v: 1, d: [123, 34, 112, 114, ...]}
+      JSON.parse(data)
+    when CUMULATIVE_TIME_STATE_ID
+      # E.g.: 37561
+      Integer(data)
```
sadleb, line 109: I hope someone doesn't take long enough to do a module that they overflow an int 😂

lib/lrs_xapi_mock.rb
rshipp: oh, I... actually have no idea. I'll try it and see if something breaks?

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+    when BOOKMARK_STATE_ID
+      # E.g.: #/lessons/xQvXXBKjohmGHnPUlyck9VqMlQ0hgcgy
+      raise StandardError.new("bookmark max length exceeded") unless data.length < BOOKMARK_MAX_LENGTH
```
sadleb, line 112: I generally prefer creating actual error types per class to raise so that we don't end up in a world where honeycomb is just filled with `StandardError` and our only recourse for making heads from tails is string based filtering when analyzing the data.

app/controllers/lrs_xapi_mock_controller.rb
rshipp: yeah i forgot to bring that up again but it's a valid question.

lib/lrs_xapi_mock.rb
```diff
@@ -0,0 +1,190 @@
...
+
+    attributes = {
+      canvas_course_id: parsed[:course],
+      canvas_assignment_id: parsed[:assignment],
+      activity_id: activity_id,
+      user: user,
+      state_id: state_id,
+    }
+
+    if module_state = Rise360ModuleState.find_by(attributes)
```
sadleb, line 128: Ditto on the above style nit. Do ya'll like this style? I've always avoided it b/c 1) I always have to ask myself whether the we're checking the assigned value or whether the assignment succeeded (but may have been an assignment to a falsey value 2) I've shot myself in the foot in the past with `=` vs `==` vs `===` and prefer to just avoid the mistakes.

spec/feature/lrs_xapi_proxy/project_xapi_spec.rb
```diff
@@ -70,7 +70,7 @@
     # "sends data to the LRS". 
     # It might be that the shared LRS and reusing the question_id is causing 
     # the value to get overwritten?
-    it "fetches the most recent data from the lrs - FLAKY", ci_exclude: true do
+    xit "fetches the most recent data from the lrs - FLAKY", ci_exclude: true do
```
sadleb, line 73: Add a TODO and link to the task in case we somehow lose track of needing to do this and stumble upon these later wondering why?

spec/lib/lrs_xapi_mock_spec.rb
sadleb: I'm already completely out of steam reviewing at this point. I'm just going to trust you and if there are bugs we have plenty time to find and fix before launch.

sadleb (commented):
The two comments that are potentially merge blocking are about
1. return unless authorization&.start_with? LtiAuthentication::LTI_AUTH_HEADER_PREFIX
2. use of StandardError

lib/lrs_xapi_mock.rb
rshipp: I actually tested this because I was worried about it, and Ruby Integers can get ... very large, so I don't think it's an issue.

```
2.2.3 :040 > Integer('999999999999999999999999999912901289812989899999999999999')
 => 999999999999999999999999999912901289812989899999999999999
```

lib/lrs_xapi_mock.rb
sadleb: Ruby modules are things that you include in a class. They are ruby's solution for composition vs inheritance as a way to share code. I think it makes sense for this to be a class because it IS an `LrsApiMock` as opposed to a mocking behavior that we include in another class.

rshipp: ok then i'm leaving this alone for now

PR #487, opened 2020-12-02
Fix Course routes.

sadleb:
### Description
After getting rid of the course_management paths, it exposed an issue
where `/courses/launch` was being interpreted as trying to `show` the course
with `id=launch`. The duplicate `resources :courses` routes seemed to cause
the confusion. Consolidating it makes it all work again.

Task: https://app.asana.com/0/1174274412967132/1199508029687637

*Note:* we still need to revisit whether all these routes should be defined
and/or what authorization we should put in place for them to avoid mistakes
against real, launched courses or real fully populated course templates. Task
for that: https://app.asana.com/0/1174274412967132/1199503585782989

### Test Plan:
Make sure you can go to https://platformweb/courses and use the Launch New Program button as well as edit/delete/add a course template

### Types of changes

### Checklist:


config/routes.rb
```diff
@@ -48,6 +42,8 @@
     resources :course_project_versions, controller: 'course_custom_content_versions', type: 'CourseProjectVersion', only: [:new]
     resources :course_survey_versions, controller: 'course_custom_content_versions', type: 'CourseSurveyVersion', only: [:new]
 
+    resources :peer_review_submissions, only: [:new, :create]
+
```
rshipp, line 46: this would break peer review submissions, since it relies on the parent course being set

rshipp (changes requested):

config/routes.rb
sadleb: It's nested under courses. The diff is hard to see. It's
```
resources :courses do
  ...
  resources :peer_review_submissions, only: [:new, :create]
  ...
end
```

jessouyangbraven (approved):

config/routes.rb
rshipp: ohh I see, whoops

PR #489, opened 2020-12-03
Add some CSS styling for Peer Reviews.

sadleb:
### Description
This is just a first pass, but I'm pretty happy with it. We used the
pinned screenshot in this task: https://app.asana.com/0/1174274412967132/1193980665758237
as inspiration. This was from a prototype we did in FormAssembly.

One thing to note is that this CSS is very specific to this form.
It was a challenge to get the radio list with 5 options and inline
text to lay out nice horizontally and switch to vertical at the
right media breakpoints. See the big comment in
`peer_review_submissions.scss` for why we have a weird hard-coded
non-standard media breakpoint in there. I did pretty extensively check
how the form looked at different screen sizes when shown inside Canvas.

### Before
<details><summary>Screenshot</summary>
</details>

### Types of changes

### Checklist:


app/assets/stylesheets/layouts/lti_canvas.scss
```diff
@@ -24,8 +25,9 @@
 // mobile device. If you decide to change the max-width for a certain view,
 // make sure you write your content responsively!
 .bv-lti-container {
+  height: 100vh; // Height of iframe viewport layout, not the actual screen
   max-width: $container-max-width-md;
   margin: 0 auto;
-  height: 100vh; // Height of iframe viewport layout, not the actual screen
+  padding: 0;
```
sadleb, line 31: We already show everything with left and right margins. This allows a Bootstrap container inside to be able to take advantage of the full width of the iframe when arranging content in a grid.

app/views/peer_review_submissions/new.html.erb
```diff
@@ -1,43 +1,73 @@
 <% content_for :title, 'Peer Reviews' %>
 
-<div class="peer-review">
```
sadleb, line 3: Moved this wrapper up to the parent `bv-lti-container` so that we could set the media breakpoints relative to the full iframe space we have.

app/views/peer_review_submissions/new.html.erb
```diff
@@ -1,43 +1,73 @@
...
+
+        <div class="peer-review-question-wrapper" role="radiogroup" aria-labelledby="<%= input_name %>">
+          <h5 id="<%= input_name %>" class="peer-review-question">
+            <%= question.text %>
+          </h5>
 
-          <input id="radio-<%= user.id %>-<%= question.id %>-10" type="radio" value="10"
-                 name="peer_review[<%= user.id %>][<%= question.id %>]" required>
-          <label for="radio-<%= user.id %>-<%= question.id %>-10">10 - Always</label>
+          <div class="container-fluid p-0">
```
sadleb, line 31: Full width, no-gutter, container to handle laying out the radiolist horizontally vs vertically. Note that I tried using Boostrap's `form-check` and `form-check-inline` stuff but it was just making it worse and giving me a headache. We really just need 5 columns that have nice margins/padding that switch to vertical at the point where they can no longer fit.

app/assets/stylesheets/peer_review_submissions.scss
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  .peer-review-question-wrapper input {
+    margin-left: 0.5rem;
+    margin-right: 0.25rem;
+  }
+
+  .peer-review-full-name {
+    color: $bv-skyblue-800;
+    text-transform: capitalize;
```
rshipp, line 48: let's remove this, we don't want to mess up capitalization in peoples names. i'd rather show a lower-case name for someone who entered it that way than broken capitalization for someone who entered their name correctly.

app/assets/stylesheets/peer_review_submissions.scss
```diff
@@ -0,0 +1,60 @@
...
+    margin: 0;
+  }
+
+  .peer-review-question {
+    color: $bv-black;
+    font-size: 1.25rem;
+    font-weight: normal;
+    text-transform: capitalize;
```
rshipp, line 56: This Makes The Questions Read Very Stilted To Me, can we get rid of it? these are legends, not headings, so there's no consistency issue here.

app/views/peer_review_submissions/new.html.erb
```diff
@@ -1,43 +1,73 @@
...
+minutes right now reflecting on your teammates and evaluating them. Your evaluation will be averaged with all the
+others, resulting in a final Teamwork score for each Fellow in your cohort.</p>
+
+<%= form_with(url: course_peer_review_submissions_path, method: "post", local: true) do |form| %>
+
+  <% @peer_users.each do |user| %>
+
+    <fieldset class="single-peer-review-wrapper">
+      <legend>
+        <h3 class="peer-review-full-name">
```
rshipp, line 17: have to remove this h3, that's [not valid here](https://stackoverflow.com/a/8005953/12432170). we can add a class to the legend if needed.

app/views/peer_review_submissions/new.html.erb
```diff
@@ -1,43 +1,73 @@
...
+          <%= user.full_name %>
+        </h3>
+      </legend>
 
       <% @questions.each do |question| %>
-        <fieldset>
-          <legend><%= question.text %></legend>
```
rshipp, line 16: need to put this fieldset and legend back, this is an accessibility requirement. it's ok to have a fieldset within a fieldset, if you want to have it grouped by both user and question.

app/views/peer_review_submissions/new.html.erb
```diff
@@ -1,43 +1,73 @@
...
+        <% input_name = "peer_review[#{user.id}][#{question.id}]" %>
+        <% input_id_prefix = "radio-#{user.id}-#{question.id}" %>
+
+        <div class="peer-review-question-wrapper" role="radiogroup" aria-labelledby="<%= input_name %>">
+          <h5 id="<%= input_name %>" class="peer-review-question">
```
rshipp, line 27: be sure to remove this heading once this is back to a legend

rshipp (changes requested):
see comments

app/views/peer_review_submissions/new.html.erb
rshipp: this aria-labelledby seems wrong too, I don't think that should be on a div? we already have a label for the input

sadleb: Hmm, I was just going off what I found in the FormAssembly CSS. It seems like the`role="radiogroup" aria-labelledby=<the_id>` where the text for the group has the ID is an alternate approach to accessibility: https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria -- but I didn't research it much.

Anyway, I like being consistent with the fact that we use fieldset/legend for the grouping of the user so I'll change this.

app/views/peer_review_submissions/new.html.erb
sadleb: See below for more context. I'll put this back.

app/views/peer_review_submissions/new.html.erb
rshipp: as a general rule, semantic elements are preferred over the aria alternatives. things like `<a aria-role="button">` are workarounds for when you're constrained on what you can do with the HTML for some reason (e.g. in a legacy codebase where it's easier to add an attribute than change an element and have to fix all your css/js)

app/assets/stylesheets/layouts/lti_canvas.scss
```diff
@@ -24,8 +25,9 @@
 // mobile device. If you decide to change the max-width for a certain view,
 // make sure you write your content responsively!
 .bv-lti-container {
+  height: 100vh; // Height of iframe viewport layout, not the actual screen
```
jessouyangbraven, line 28: Should this be a `$contaniner-...` variable in `shared/variables`?

jessouyangbraven (commented):
Just the question on `shared/variables` inline. Otherwise, @rshipp's review's really thorough and I'll defer to him on this PR.

app/assets/stylesheets/layouts/lti_canvas.scss
sadleb: I don't think so. It's not a "style" that we would use in multiple places and want them all changed in one go if we decide to change the value. This is just explicitly setting the height of this container to 100% of the screen dimensions available to it inside the iframe. This is specific to this layout where we show everything in an iframe. When we're not in an iframe, the `html, body` are already set to %100 which effectively does the same thing. See here for more detail: https://stackoverflow.com/a/27804632

app/assets/stylesheets/peer_review_submissions.scss
sadleb: Yup, I agree. I'll get new screenshots up for you to react to once I finish the changes. I am still using DIN font for the questions which I'm curious to get your thoughts on.

sadleb:
I've addressed the feedback and updated the screenshots. Re-requested your review @rshipp

rshipp (approved):

app/assets/stylesheets/peer_review_submissions.scss
rshipp: Yeah, it might be better with our regular paragraph font. IIRC one of the style guide things was "don't over-use" the headings font. This is a design question and not a a11y one though, so I'm fine with changing it later.

PR #499, opened 2020-12-08
Ability to manage Canvas enrollments from User's Admin dashboard

sadleb:
### Description
We each have a test user or two that we've created in Canvas and configured our platform app to use. When testing things out, we need to make sure that the user is enrolled in the correct Canvas course/section/role that matches the local platform `Section` and `Role` for things to work. This PR makes it easy to set that up from Admin Users page.

### Motivation and Context
This is intended to be used only on unlaunched courses for Designers and Developers to manage Test and Demo accounts. Originally, I had planned to allow creating a new user in Canvas from the `Add User` page but decided not to because:
1. We don't want to encourage creating tons of test accounts that will litter Canvas and that we'll have to pay for.
1. Instead, we should have each dev / designer have a couple accounts for different purposes and write tools to manage their state (like blowing away submissions for example to get them in a fresh state).
1. We don't want to open the door for this being used by product support and regional folks as a crutch to hack around issues with the core Salesforce and user provisioning flow. We should fix the core issues.
1. It doesn't really work that well as a solution in dev b/c the user has to be in the production platform for SSO to work. 

Guidance on how to create new test/demo accounts [can be found here](https://github.com/bebraven/platform/wiki/Test-&-Demo-Accounts).

#### Select the `Enrollment Type` and fill out the settings

<small>*See the notes at the bottom about how it works*</small>

### Types of changes

### Checklist:


app/policies/users_role_policy.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def destroy?
+    super # admin
+
+    # Admin UsersRoles controller shouldn't be able to delete enrollments on live courses
+    # Salesforce should control that so that everything is in sync and consistent (and we
+    # dont make mistakes).
+    record.is_a?(Role) and record.resource.course.is_launched == false
+  end
```
sadleb, line 17: Note that I'm already hiding the launched courses from the `new` UI's dropdown. It didn't seem worth trying to figure out how to authorize that in the policy as well. Lmk if you disagree.

config/initializers/constants.rb
```diff
@@ -9,7 +9,7 @@ class RoleConstants
   # Add all section-level roles to this list.
   # I.e. anything that can be used like `user.add_role XXX_ENROLLMENT, section`.
   SECTION_ROLES = [
-    TA_ENROLLMENT,
     STUDENT_ENROLLMENT,
+    TA_ENROLLMENT,
   ]
```
sadleb, line 14: This made it show up as Student then TA in the dropdown.

db/migrate/20201207185957_drop_user_sections.rb
```diff
@@ -0,0 +1,5 @@
+class DropUserSections < ActiveRecord::Migration[6.0]
+  def change
+    drop_table :user_sections
+  end
+end
```
sadleb, line 5: This was another one of the models that we created before Highlander really kicked off when we were trying to data model "the Braven universe." It's not used and `UsersRoles` is the equivalent way to get at the same info with our current setup.

spec/controllers/users_roles_controller_spec.rb
```diff
@@ -0,0 +1,132 @@
+require 'rails_helper'
+
+RSpec.describe UsersRolesController, type: :controller do
```
sadleb, line 3: Very basic specs. I just want them to detect if someone refactors `SyncPortalEnrollmentForAccount` and breaks this but didn't think it was worth spending much time here for a hacky, admin only feature.

app/policies/users_role_policy.rb
```diff
@@ -0,0 +1,18 @@
...
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    raise Pundit::NotAuthorizedError, "must be logged in" unless user
+    @user = user
+    @record = record
+  end
+
+  def destroy?
+    super # admin
```
rshipp, line 11: I think you need some type of return/branching here, otherwise this is just calling the parent's `destroy?` and immediately discarding its results

app/policies/users_role_policy.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    # Admin UsersRoles controller shouldn't be able to delete enrollments on live courses
+    # Salesforce should control that so that everything is in sync and consistent (and we
+    # dont make mistakes).
+    record.is_a?(Role) and record.resource.course.is_launched == false
```
rshipp, line 16: we don't generally do `record.is_a?`, is there a reason that's important here? would you ever call this method with something else?

app/views/users/_form.html.erb
```diff
@@ -13,39 +15,58 @@
...
-      <% end %>
+  <%# 
+    Note: We don't currently have a way to add the user to Canvas using the admin page.
+    a) We want to discourage abuse b) you can't do it from dev and be able to login.
+    The user must be in the prod platform for SSO to work (unless we added new SSO auth
+    options pointed at each dev box). The current guidance is to manually create test Canvas
+    users, then go to the prod User admin page and create the same user, then set the
+    Canvas User ID to match.
+  %>
+  <% unless is_new_user %>
```
rshipp, line 51: not creating a canvas user from platform is fine, but why is it important to hide the canvas user id field when creating a new user? My usual flow would be create user on canvas -> get the canvas user id -> create new user on platform, setting the canvas user id. this adds an extra step where you have to go back and edit the user you just created.

app/views/users_roles/_list.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    <table class="table mt-2 mb-0">
+      <thead>
+        <tr>
+          <th>Canvas Course</th><th>Section</th><th>Role</th><th></th>
+        </tr>
+      </thead>
+      <tbody>
+      <% section_roles = user.section_roles %>
+      <% unless section_roles.empty? %>
+        <% user.section_roles.each do |role| %>
```
rshipp, line 11: looks like this got only partially refactored to use the variable

rshipp (changes requested):
minor changes requested

app/policies/users_role_policy.rb
sadleb: I just thought it was easier to read knowing what it is.

app/views/users/_form.html.erb
sadleb: oh, good point. i'll add that part back.

app/policies/users_role_policy.rb
rshipp: since you're already redefining the `initialize` method, what about calling it `role` instead of `record`?

PR #502, opened 2020-12-10
Module Uploader on main page, can create/edit/delete

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199394675251953

This PR adds **Module Uploader** to the main Braven page, where you can add/edit/delete individual `Rise360Module`s.

### Screenshots (if appropriate):

- Click **Module Uploader** on main page:
Screen Shot 2020-12-11 at 3 13 34 PM

- Here's the `Rise360Modules` `#index` page:
Screen Shot 2020-12-11 at 3 13 38 PM

- **Add Module**:
Screen Shot 2020-12-11 at 3 13 52 PM

- After **Save**:
Screen Shot 2020-12-11 at 3 14 57 PM

- **Preview** iff there's an attachment, goes to the proxied URL without any LRS query params, so there aren't any xAPI statements:
Screen Shot 2020-12-11 at 3 15 04 PM

- Update the name and zipfile:
Screen Shot 2020-12-11 at 3 15 37 PM

- After **Save**:
Screen Shot 2020-12-11 at 3 15 18 PM

- You can add the module to a course going to a Course's edit page:
Screen Shot 2020-12-11 at 3 19 42 PM

- Then update the module name:
Screen Shot 2020-12-11 at 3 20 02 PM

- **Publish Latest** reflects that change:
Screen Shot 2020-12-11 at 3 20 45 PM

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #503, opened 2020-12-10
Allow updates to Rise360Module

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199394675251953

This PR facilitates #502, which allows us to update `Rise360Module`s (e.g, its name, attached zipfile) through the UI.

### Motivation and Context

`Rise360Module`'s `rise360_zipfile` used to be immutable after creation. As a result, we only had to run run the callbacks `publish` and `update_metadata!` when the model was created. The metadata pertaining to the xAPI statements generated by the `rise360_zipfile`, `activity_id` and `quiz_questions`, allow us to look up relevant `Rise360ModuleInteraction`s in order to grade Fellows on module progress and mastery questions.

Now that we're going to allow updating a `Rise360Module`, specifically its attached zipfile, we need to:

1. Run `publish` and `update_metadata!` after `save` (regardless of whether the save comes from `create` or `update`) on `Rise360Module`
2. `publish` the `rise360_zipfile` on the `Rise360ModuleVersion` so it exists on S3 separate of the `Rise360Module`
3. Save `activity_id` and `quiz_questions` to `Rise360ModuleVersion` so it remains in-sync with its `rise360_zipfile`
4. Enforce that a `Rise360ModuleVersion` is immutable after creation

### Test

- specs

### Types of changes

### Checklist:


app/models/rise360_module.rb
```diff
@@ -24,21 +30,42 @@ def new_version(user)
...
+
+      # Note: This must be set **before** we call update_metadata! below.
+      # Otherwise, the update! call in update_metadata! will trigger an
+      # infinite loops on publish.
+      self.rise360_zipfile_changed = false
+      Rise360Util.update_metadata!(self)
+    else
+      self.rise360_zipfile_changed = false
+    end
+  end
```
jessouyangbraven, line 70: I consolidated both `Rise360Util` calls to the single `publish` callback because I had to set `rise360_zipfile_changed = false` after `publish` or else I'd trigger an infinite loop from the `update` and `update_metadata()`. But as soon as I unset the flag, `update_metadata, if: rise360_zipfile_changed` wouldn't run. I could have added another `attr_accessor` for `publish` to flag `update_metadata`, but I thought that was more confusing.

app/models/rise360_module_version.rb
```diff
@@ -1,13 +1,29 @@
 require 'rise360_util'
 
 class Rise360ModuleVersion < ApplicationRecord
+  # Versions are immutable after they're created. You can delete, though.
+  before_update { raise ActiveRecord::ReadOnlyRecord }
```
jessouyangbraven, line 5: I didn't do anything at the DB level for now. Let me know if you think that should be part of this PR.

lib/rise360_util.rb
```diff
@@ -64,10 +64,10 @@ def self.update_metadata!(rise360_module, bucket = nil)
...
     # See https://stackoverflow.com/questions/4690737/nokogiri-xpath-namespace-query.
     tincan_data.remove_namespaces!
-    rise360_module.quiz_questions = tincan_data.xpath(QUIZ_QUESTION_XPATH).count
-    rise360_module.activity_id = tincan_data.xpath(COURSE_XPATH).first['id']
-
-    rise360_module.save!
+    rise360_module.update!(
+      quiz_questions: tincan_data.xpath(QUIZ_QUESTION_XPATH).count,
+      activity_id: tincan_data.xpath(COURSE_XPATH).first['id'],
+    )
```
jessouyangbraven, line 70: Kind of unrelated tidying. I was staring at this code because of the infinite loop I mentioned above.

spec/models/rise360_module_spec.rb
```diff
@@ -8,6 +8,52 @@
...
+      end
+
+      it 're-publishes to S3' do
+        subject
+        expect(Rise360Util).to have_received(:publish)
+      end
+
+      it 'updates metadata' do
+        subject
+        expect(Rise360Util).to have_received(:update_metadata!)
```
sadleb, line 52: Should you add a `once` test here to make sure we don't get in that infinite loop situation if someone refactors incorrectly?

sadleb (approved):
lgtm. One small nit.

PR #507, opened 2020-12-12
Use Publishable in Course{Rise360Module, Project, Survey}VersionsControllers

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199186575082133

### Motivation and Context

This PR began with removing network calls while holding transaction locks in `CourseCustomContentVersion`. 

Then I realized that the logic I was removing from the model was similar to what we were already doing in `CourseRise360ModuleVersionsController`, specifically, using `Publishable` to update Canvas Assignments API and then using the actions `#create`, `#update`, `#destroy` to handle Model-related state in our DB.

After removing the network calls, I refactored the common code into `Publishable#publish_latest` from `CourseRise360ModuleVersionsController` and started using the concern in `CourseCustomContentVersionsController`, which now has the actions: `#publish`, `#publish_latest`, and `#unpublish`.

I also split up the controller into `CourseProjectVersionsController` and `CourseSurveyVersionsController`, because only the former supports adding rubrics. The shared UI for these two controllers is in partials in `app/views/custom_contents/_form.html.erb`. 

### Test

- rspec

### Types of changes

### Checklist:


app/controllers/concerns/publishable.rb
```diff
@@ -1,82 +1,202 @@
...
+# If you want to have versioning history for your content, you will need to
+# have a Model that implements Versionable.
 
 require 'canvas_api'
 
 module Publishable
   extend ActiveSupport::Concern
 
+  included do
+    before_action :verify_can_edit!
```
jessouyangbraven, line 70: We weren't checking this before in `{PeerReviews,Waivers}Controller`, but I moved it here because I think we want to check that the course isn't launched for everything that modified assignments. If that's not the intended behavior, I can put this back into `Course{Rise360Module, Project, Survey}VersionsController`, where it originally was.

app/controllers/concerns/publishable.rb
```diff
@@ -1,82 +1,202 @@
...
+    after_action :destroy, only: [:unpublish]
+  end
+
   def publish
     # We can't use `model_class` here because we don't have a PeerReview model
     authorize controller_path.classify.to_sym
 
     assignment = CanvasAPI.client.create_lti_assignment(
-      course.canvas_course_id,
+      @course.canvas_course_id,
```
jessouyangbraven, line 81: I took out `def course` because it was always the same. I added a comment on how your controller needs to be nested under `Course` to use this concern.

app/controllers/concerns/publishable.rb
```diff
@@ -1,82 +1,202 @@
...
 
     instance_variable.update!(
       canvas_assignment_id: assignment['id'],
     ) if model_class && instance_variable.respond_to?(:canvas_assignment_id)
 
+    CanvasAPI.client.update_assignment_lti_launch_url(
+      @course.canvas_course_id,
+      assignment['id'],
+      lti_launch_url,
+    )
```
jessouyangbraven, line 93: This needs to happen as a separate update because of projects and surveys, which require a `canvas_assignment_id` before you can save the record. But you need a saved record in order to get the `new_submission_url`. This was just the most straightforward way to break that dependency.

app/controllers/course_project_versions_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def versionable_instance
+    params[:custom_content_id] ?
+      Project.find(params[:custom_content_id]) : # publish
+      @course_project_version.project_version.project # publish_latest
+  end
+
+  def version_name
+    'project_version'
+  end
```
jessouyangbraven, line 46: I tried to meta-program this out in the concern, but the param name was tripping me up. It was either `:custom_content_id` (for projects and surveys that share the same UI) or `:rise360_module_id`. I couldn't figure out a clean way to programmatically generate it. It was have been possible if it was `:project_id` and `:survey_id`, but I wanted to share the `_form.html.erb.`

app/controllers/course_survey_versions_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  prepend_before_action :set_model_instance, only: [:publish_latest, :unpublish]
+
+  layout 'admin'
+
+  def new
+    authorize CourseSurveyVersion
+    @custom_contents = Survey.all - @course.survey
+    @course_custom_content = CourseSurveyVersion.new(course: @course)
+  end
```
jessouyangbraven, line 18: Ditto trying to put `#new` in the concern. It's @custom_contents for projects/surveys and something else for the Rise360 modules. I didn't really want to generalize the UI for all three cases or split up projects and surveys.

app/models/course_custom_content_version.rb
```diff
@@ -17,94 +18,6 @@ class CourseCustomContentVersion < ApplicationRecord
...
-          course.canvas_course_id,
-          canvas_assignment_id,
-        )
-      end
-    rescue RestClient::NotFound
-      # This gets thrown when the assignment doesn't exist in Canvas.
-      # It's fine to delete the record from our DB in this case.
-      destroy!
-    end
-  end
```
jessouyangbraven, line 102: This was the code I originally wanted to delete.

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -20,45 +16,8 @@ def new
...
-    )
-  end
-
-  def destroy
-    @course_rise360_module_version = CourseRise360ModuleVersion.find(params[:id])
-    authorize @course_rise360_module_version
-    # For Publishable
-    params[:canvas_assignment_id] = @course_rise360_module_version.canvas_assignment_id
-    @course_rise360_module_version.destroy!
-  end
```
jessouyangbraven, line 55: These are all logically the same for the `Course*Version` models, so I put it in the concern.

app/controllers/concerns/publishable.rb
```diff
@@ -1,82 +1,202 @@
...
 require 'canvas_api'
 
 module Publishable
   extend ActiveSupport::Concern
 
+  included do
+    before_action :verify_can_edit!
+    before_action :create, only: [:publish]
+    before_action :update, only: [:publish_latest]
+    after_action :destroy, only: [:unpublish]
```
jessouyangbraven, line 73: TODO: Rename so it's obvious they're not the normal actions.

lib/canvas_api.rb
```diff
@@ -254,6 +254,12 @@ def create_lti_assignment(course_id, name, launch_url = nil)
     JSON.parse(response.body)
   end
 
+  def update_assignment_name(course_id, assignment_id, name)
+    body = { :assignment => { :name => name } }
+    response = put("/courses/#{course_id}/assignments/#{assignment_id}", body)
+    JSON.parse(response.body)
+  end
```
jessouyangbraven, line 261: This fixes the bug we saw in the sprint planning demo where the `Publish Latest` didn't work for modules.

spec/controllers/course_survey_versions_controller_spec.rb
```diff
@@ -0,0 +1,250 @@
+require 'rails_helper'
+require 'canvas_api'
+
+RSpec.describe CourseSurveyVersionsController, type: :controller do
```
jessouyangbraven, line 4: Tragically, this is the same as `CourseProjectVersionsController`.

Is there an existing Asana task I should link to for the spec clean up? Or should I make a new one for a `TODO` here?

spec/controllers/course_project_versions_controller_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+          .to receive(:create_lti_assignment)
+          .and_raise(RestClient::BadRequest)
+      end
+
+      it 'does not create the join table entry' do
+        expect {
+          subject rescue nil
+        }.not_to change(CourseProjectVersion, :count)
+      end
+    end
```
jessouyangbraven, line 114: This is the test that, even without a transaction, we don't end up with an invalid join table entry when there's no corresponding Canvas assignment.

spec/controllers/course_project_versions_controller_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+        allow(CanvasAPI).to receive(:client).and_return(canvas_client)
+        allow(canvas_client)
+          .to receive(:delete_assignment)
+          .and_raise(RestClient::NotFound)
+      end
+
+      it 'deletes the join table entry' do
+        expect { subject rescue nil }.to change(CourseProjectVersion, :count).by(-1)
+      end
+    end
```
jessouyangbraven, line 143: This is so we can delete the join table entry if someone accidentally deletes the Canvas assignment in Canvas and not the course management page. `RestClient::NotFound` is the specific error we get from Canvas when it can't find `(course_id, assignment_id)`.

spec/controllers/course_project_versions_controller_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+        allow(CanvasAPI).to receive(:client).and_return(canvas_client)
+        allow(canvas_client)
+          .to receive(:delete_assignment)
+          .and_raise(RestClient::BadRequest)
+      end
+
+      it 'does not delete the join table entry' do
+        expect { subject rescue nil }.not_to change(CourseProjectVersion, :count)
+      end
+    end
```
jessouyangbraven, line 156: This is for errors other than `RestClient::NotFound`. It tests that, even without transactions, we don't delete our join table entry if we can't delete the Canvas assignment.

spec/controllers/course_project_versions_controller_spec.rb
jessouyangbraven: Side note: I don't test whether or not the `ProjectVersion` gets created because I think it'll just change/break a bunch if we change the underlying behavior. I think it's okay to have unused `Versions` of content in our DB, and it's only critical that the join table accurately reflects what's in Canvas assignments.

spec/models/course_custom_content_version_spec.rb
```diff
@@ -14,300 +14,4 @@
       it { is_expected.to be_valid }
     end
   end
-
-  describe '#publish!' do
```
jessouyangbraven, line 18: All of these moved into the controller specs.

app/controllers/concerns/publishable.rb
sadleb: I think this is good for now. Once we get to the future user stories about what we make editable it will be nice to have all the logic in one place so we can approach it holistically

spec/controllers/course_survey_versions_controller_spec.rb
sadleb: Make a new one. If we happen to have another we’ll merge when we find it but I don’t think we do.

app/controllers/concerns/publishable.rb
```diff
@@ -1,82 +1,204 @@
...
+      @course.canvas_course_id,
+      assignment['id'],
+      lti_launch_url,
+    )
+
+    CanvasAPI.client.add_rubric_to_assignment(
+      @course.canvas_course_id,
+      assignment['id'],
+      params[:rubric_id]
+    ) if params[:rubric_id].present?
```
sadleb, line 96: Since this is only for Projects and not a general Publishable behavior, does it make more sense to have an `after_update_canvas_assignment` method that’s called here which is a NOOP here but the projects controller overrides with this rubric stuff?

app/controllers/concerns/publishable.rb
```diff
@@ -1,82 +1,204 @@
...
-      course.canvas_course_id,
-      params[:canvas_assignment_id],
-    )
+    begin
+      CanvasAPI.client.delete_assignment(
+        @course.canvas_course_id,
+        canvas_assignment_id,
+      )
+    rescue RestClient::NotFound
+      # This gets thrown when the assignment doesn't exist in Canvas.
```
sadleb, line 142: Can we add a field to Honeycomb indicating that this happened and sending the assignment ID as well in case we run into issues and need to troubleshoot in production and at least know if this code path is being executed?

app/controllers/course_project_versions_controller.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class CourseProjectVersionsController < ApplicationController
+  include DryCrud::Controllers::Nestable
+  nested_resource_of Course
+
+  # Adds the #publish, #publish_latest, #unpublish actions
+  include Publishable
+
+  prepend_before_action :set_model_instance, only: [:publish_latest, :unpublish]
```
sadleb, line 10: Interesting

lib/canvas_api.rb
sadleb: This and the launch URL could probably be refactored into a single API call but it’s not worth the effort at the moment. If we add a third assignment update attribute that we need to handle, we should deal with it then

sadleb (approved):
I think you should add the Honeycomb field in the swallowed exception, but no other comments are merge blocking. 

Note to self: be careful about declaring a “tech debt sprint” bc Jess will take it to the extreme and get fancy.

app/views/course_rise360_module_versions/_list.html.erb
```diff
@@ -11,7 +11,7 @@
       <tbody>
         <% course_rise360_module_versions.each do |crmv| %>
           <tr>
-            <td><%= link_to crmv.rise360_module_version.name, edit_rise360_module_path(crmv.rise360_module_version) %></td>
+            <td><%= link_to crmv.rise360_module_version.name, edit_rise360_module_path(crmv.rise360_module_version.rise360_module) %></td>
```
sadleb, line 14: 🙏

app/models/course_project_version.rb
```diff
@@ -1,4 +1,15 @@
 # Represents a model joining a Course to a ProjectVersion
 class CourseProjectVersion < CourseCustomContentVersion
   belongs_to :project_version, foreign_key: "custom_content_version_id"
+
+  def project
+    project_version.project
+  end
```
rshipp, line 7: nit: i think this could be a `has_one through:`

app/models/course_survey_version.rb
```diff
@@ -1,4 +1,15 @@
 # Represents a model joining a Course to a SurveyVersion
 class CourseSurveyVersion < CourseCustomContentVersion
   belongs_to :survey_version, foreign_key: "custom_content_version_id"
+
+  def survey
+    survey_version.survey
+  end
```
rshipp, line 7: same here

spec/controllers/course_project_versions_controller_spec.rb
```diff
@@ -0,0 +1,273 @@
...
+      it 'does not delete the version' do
+        expect { subject }.not_to change { ProjectVersion.count }
+      end
+
+      it 'does not delete the course' do
+        expect { subject }.not_to change { Course.count }
+      end
+
+      it 'redirects to course edit page' do
+        subject
```
rshipp, line 186: do you have to explicitly call `subject` like this?

jessouyangbraven: Yes. The `subject` is just a lambda for `delete :unpublish...` because I didn't want to have the whole thing in every test case because I always have to update the params across all of them. It doesn't do the fancy auto-magic calling of `subject` that we can usually get in the model specs. Should I call the lambda something else so it's less confusing?

app/controllers/course_project_versions_controller.rb
```diff
@@ -9,6 +9,7 @@ class CourseProjectVersionsController < ApplicationController
 
   prepend_before_action :set_model_instance, only: [:publish_latest, :unpublish]
   before_action :set_rubrics, only: [:new]
+  after_action :add_rubric, only: [:publish]
```
sadleb, line 12: Even better solution than what I proposed 🙂

PR #509, opened 2020-12-15
Add a New Course Template initialization process

sadleb:
### Description

Before this, we had no way to create a Course Template. The way it was done was to create a brand new Course in Canvas and manually create the database entries to make it a template. The template had to be setup from scratch with all projects and modules added manually. However, for the InternPro launch (or shortly there-after) we'll want to have two templates, one for Fall Accelerator and one for InternPro. This PR implements a feature allowing us to take an existing Course or Course Template and initialize a new Template using it. Task: https://app.asana.com/0/1174274412967132/1198935151567022

### Motivation and Context
The previous approach of just having one master Template for each product (Accelerator, BravenX, InternPro, etc) was unwieldy and inflexible. Designers iterate on the content in one main place (the Accelerator). When launching a variation of the Accelerator, they need the ability to customize it before launching. Further, it can be tough to align the big content changes to the demands of upcoming program launches.

This ability to create a Template from anything gives them the flexibility to manage launches and content iteration in a more decoupled fashun. E.g. if there is a big content refactor they are working on that isn't quite ready for the upcoming semester, then can choose to simply re-launch last semester's course again by turning it into a template (there are plans to streamline that where you can just choose a launched Program and choose to re-launch it). 

This also gives developers (and designers) the ability to setup Playground/Sandbox templates to play around with.

### Types of changes

### Checklist:


app/controllers/courses_controller.rb
```diff
@@ -15,8 +15,7 @@ def show
 
   # GET /courses/new
   def new
-    @course = Course.new(new_params)
-    authorize @course
+    authorize Course
```
sadleb, line 18: We don't actually use the `@course` instance. This was left-over from the initial implementation before we implemented New Program Launch and now New Course Template.

app/services/clone_course.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      if (Time.now - start_time) > 3.minutes # Fail safe so this doesn't poll forever.
+        raise CloneCourseError, 'Canvas course copy timed out after 3 minutes'
+      end
+
+      sleep 5.seconds
+
+      @canvas_clone_response = CanvasAPI.client.get_copy_course_status(progress_url)
+    end
+    block.call
```
sadleb, line 69: Direct cutover from `LaunchProgram` service

app/services/initialize_new_course.rb
```diff
@@ -31,19 +27,22 @@ def initialize(new_course, section_names)
...
-    # Add an AssignmentOverride date object to each assignment for each section so that 
-    # the Edit Assignment Dates page shows the sections.
-    CanvasAPI.client.create_assignment_overrides(@new_course.canvas_course_id, canvas_assignment_ids, canvas_section_ids)
+    if @section_names.present?
+      canvas_section_ids = create_sections(canvas_assignment_ids)
+
+      # Add an AssignmentOverride date object to each assignment for each section so that
+      # the Edit Assignment Dates page shows the sections.
+      CanvasAPI.client.create_assignment_overrides(@new_course.canvas_course_id, canvas_assignment_ids, canvas_section_ids)
+    end
```
sadleb, line 37: `LaunchProgram` always went to Salesforce to get the CohortSchedules to use as Canvas section names but this New Course Template flow clones a course without setting up sections so we need to support both options.

app/services/launch_program.rb
```diff
@@ -8,59 +8,30 @@ def initialize(salesforce_program_id, fellow_source_course_id, fellow_course_nam
...
-      course_resource: @fellow_source_course.course_resource
-    )
-
+    @fellow_course_name = fellow_course_name
     @lc_source_course = Course.find(lc_source_course_id)
-    @lc_destination_course = Course.create!(
-      name: lc_course_name,
-    )
+    @lc_course_name = lc_course_name
   end
```
sadleb, line 14: The `Course.create!` now happens in `CloneCourse.run`

app/services/launch_program.rb
```diff
@@ -8,59 +8,30 @@ def initialize(salesforce_program_id, fellow_source_course_id, fellow_course_nam
...
-    start_time = Time.now
-    while (response['workflow_state'] != 'completed' && response['workflow_state'] != 'failed')
-      raise LaunchProgramError, "Canvas course copy failed. See: #{response['migration_issues_url']}" if response['workflow_state'] == 'failed'
-      raise LaunchProgramError, 'Canvas course copy timed out after 3 minutes' if (Time.now - start_time) > 3.minutes # Fail safe so this doesn't poll forever.
-
-      sleep 5.seconds
-
-      response = CanvasAPI.client.get_copy_course_status(progress_url)
-    end
-    block.call
```
sadleb, line 63: Moved to `CloneCourse`

app/controllers/courses_controller.rb
```diff
@@ -25,27 +31,25 @@ def edit
...
-      if @course.save
-        format.html { redirect_to courses_path, notice: "Course was successfully created." }
-        format.json { render :show, status: :created, location: @course }
-      else
-        format.html { render :new }
-        format.json { render json: @course.errors, status: :unprocessable_entity }
-      end
-    end
+    @source_course = Course.find(create_params[:create_from_course_id])
+    authorize @source_course
```
rshipp, line 40: nit: this will check `can current_user create @source_course`, but that's not what this action is doing. i think `authorize Course` would be more clear here. there's currently no functional difference, since CoursePolicy is just a simple admin-only policy.

app/controllers/courses_controller.rb
```diff
@@ -104,12 +108,12 @@ def fetch_canvas_assignment_info
...
 
-  def course_params
+  def update_params
     params.require('course').permit(:name, :is_launched, :course_resource_id, :canvas_course_id)
   end
 
-  def new_params
-    params.permit()
+  def create_params
+    params.require('course').permit(:name, :create_from_course_id)
```
rshipp, line 116: nit: could you break out `:course_from_course_id` into a `*_params` method, so that it doesn't have to be nested under `course`? it's a little confusing to have this non-column field inside `course`, especially since it's not nested in `new`'s params in the action above.

app/views/courses/_list.html.erb
```diff
@@ -20,14 +20,14 @@
...
-            <%= link_to course.name, course.canvas_url %>
+            <%= link_to course.name, course_path(course) %>
             <!-- Add the badge / tags back if we make a way to categorize the templates, like test vs official ones:
                  <span class="badge badge-primary">Course</span>-->
           </td>
           <td class="text-right align-middle p-0">
             <%# Eventually we may want to support editing or deleting a launched course, but we don't currently %>
             <% unless course.is_launched %>
-              <%= link_to 'Edit', edit_course_path(course), class: 'btn btn-secondary m-0' %> |
+              <%= link_to 'View On Canvas', course.canvas_url, target: :_blank %> |
```
rshipp, line 30: can you remove `target: :_blank`, (or add the text `(Opens in new tab)` to the link)? about:blank is [bad practice](https://css-tricks.com/use-target_blank/) in general, besides being an accessibility issue.

sadleb: I thought about that. This is just for Devs and Designers whose main workflow when clicking "View On Canvas" will be to preview things (I think). I guess I could add the `new tab` text. It's just going to be long.

rshipp: but there's not a reason to force everyone to open it in a new tab, that's just the workflow you prefer. it's easy to command-click links if you like opening things in new tabs, but it's not possible to just use it as a normal link if it has target:_blank.

sadleb: kk, i'll remove it (and put a hint about using command-click somewhere on the page perhaps). You're going to be incredulous (or not) but I didn't know about command-click. I always right-click -> open in new tab which is annoying.

app/controllers/courses_controller.rb
sadleb: Welp, that turned into more than a 1 minute nit fix. Turns out the `form.hidden_field_tag` just automatically nests the param if it's a `form_for(model)`. I did what other views did and just hand wrote the `<input>` element. There is probably a better way.

PR #510, opened 2020-12-15
Sync TAs on Salesforce Program to LMS

Rowantatai:
### Background:
**User Story:**  We need a further automated process for TA onboarding to the Portal/Zoom/Slack
**Context:** We've added TAs into Salesforce to automate the onboarding experience and manage this group of program participants similar to our LCs, CPs, & Fellows.
**Related Asana Tasks:** https://app.asana.com/0/1185361437008970/1199529205874053/f

### Types of changes

### Checklist:

- [? ] My PR title follows the convention in the guide
- [?] All new and existing tests passed.


lib/sync_to_lms.rb
```diff
@@ -112,6 +112,10 @@ def enroll_user(canvas_user_id, program, role, fellow_course_section_name)
     elsif role == :'Leadership Coach'
```
skaex, line 112: You actually don't need to change this. We don't use this file anymore.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -142,4 +152,12 @@ def canvas_client
   def logger
     Rails.logger
   end
+
+  def user
+    @user ||= User.find_by(email: sf_participant.email)
+  end
+
+  def can_edit_role?
+    section.exist? && user.exist?
```
Rowantatai, line 161: This is the start of where the tests are failing

rshipp: `find_by` returns `nil` if it can't find anything, so i think `section.exist?` will fail. https://apidock.com/rails/ActiveRecord/FinderMethods/find_by

I think you want something like `section && user` instead.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -26,7 +26,7 @@ def run
 
   private
 
-  attr_reader :portal_user, :sf_participant, :sf_program, :user
+  attr_reader :portal_user, :sf_participant, :sf_program
```
sadleb, line 29: We had a bigger discussion about these attribute readers before and decided not to use them instead of the instances directly generally. Not worth refactoring this class at the moment though, so since you are using the `user` attribute in the code, you should put this back as an `attr_reader`

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -79,9 +90,7 @@ def drop_course_enrollment(canvas_course_id)
     # https://app.asana.com/0/1174274412967132/1197893935338145/f
     # At that point this should find_by! instead so it doesn't cover up issues.
     section = Section.find_by(canvas_section_id: enrollment.section_id)
-    if section
-      user.remove_role enrollment.type, section
-    end
+    user.remove_role enrollment.type, section if can_edit_role?
```
sadleb, line 93: What does this do? If there is no section, now this will fail? Should it be 
```
user.remove_role enrollment.type, section if section and can_edit_role?
```
?

sadleb: Also, is this going to break the nightly SyncToLMS? What user does the rake task run as? If they aren't an admin I think it will fail.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -100,7 +109,8 @@ def sync_enrollment(canvas_course_id, role, section_name)
 
   # Pass in a local db section.
   def enroll_user(canvas_course_id, role, section)
-    user.add_role role, section
+    user.add_role role, section if can_edit_role?
```
sadleb, line 112: Ditto on being able to run a nightly Sync To LMS rake task. Will this work as whatever use the task runs as?

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -142,4 +152,12 @@ def canvas_client
...
     Rails.logger
   end
+
+  def user
+    @user ||= User.find_by(email: sf_participant.email)
+  end
+
+  def can_edit_role?
+    section && user
+  end
```
sadleb, line 162: Oh, now that I got to this code my comment above needs to change slightly. This won't work. `section` isn't defined here. Also, `can_edit_role?` makes me think you're checking Pundit's authorization policies but you're not. You're just making sure these two variables are defined. Even if this was going to work, it should be named something more like `user_has_section?`

sadleb (changes requested):

app/services/sync_portal_enrollment_for_account.rb
Rowantatai: A user method is already defined though

app/services/sync_portal_enrollment_for_account.rb
Rowantatai: I know nothing about a SyncToLMS rake task nor do I see it in the Heroku Scheduler

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -79,9 +90,7 @@ def drop_course_enrollment(canvas_course_id)
     # https://app.asana.com/0/1174274412967132/1197893935338145/f
     # At that point this should find_by! instead so it doesn't cover up issues.
     section = Section.find_by(canvas_section_id: enrollment.section_id)
-    if section
-      user.remove_role enrollment.type, section
-    end
+    user.remove_role enrollment.type, section if user_can_edit_role?(section)
```
sadleb, line 93: What happens if you just leave the old code in place? I still think `user_can_edit_role?` is very confusing and I don't understand what this is doing.

app/services/sync_portal_enrollment_for_account.rb
sadleb: gotcha

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -100,7 +109,8 @@ def sync_enrollment(canvas_course_id, role, section_name)
 
   # Pass in a local db section.
   def enroll_user(canvas_course_id, role, section)
-    user.add_role role, section
+    user.add_role role, section if user_can_edit_role?(section)
```
sadleb, line 112: The section was already passed in. Why do we need to check if it exists? What happens if you just leave the old code in place?

app/services/sync_portal_enrollment_for_account.rb
rshipp: @sadleb can_edit_role already checks the section. and to my knowledge rake tasks don't run as a user.

app/services/sync_portal_enrollment_for_account.rb
sadleb: Cc @rshipp

sadleb: From Slack for prosperity:
-----
Rowan Tatai Yeah, I don’t think it’ll break existing logic. But to your question around the user_can_edit_role? method, it is important bc there’s a chance the user or section will not exist so its reasonable to check first. Does that answer suffice?
11:55 AM
Brian The section not existing was already handled by the old code.
Are you saying that with this change, the user may not exist where before we only ran this for users that did exist? How will the TA user be created and how will this be called? If the user doesn't exist, does the Sync do anything or should it just short-circuit out in the `run` with a warning log about the user not existing? (edited)

rshipp: i'm not following. the old code only checked section, the new code checks section and user. if you look in `initialize` above, the user is no longer guaranteed to exist, so it makes sense to check both here. if there's no user, removing its role doesn't make any sense, so this code seems fine to me.

app/services/sync_portal_enrollment_for_account.rb
rshipp: the section being passed in doesn't mean it exists, it could be `nil`

sadleb: It's created in `sync_enrollment` and passed into `enroll_user`.

app/services/sync_portal_enrollment_for_account.rb
sadleb: That's my main question. Why are we running this at all for a user that doesn't exist. I don't understand the bigger picture here that makes us want to sync enrollments for a non-existing user @Rowantatai

Rowantatai: So, here's some additional background context to explain the core of this change:
- Regions needed TAs to have periscope dashboards set in advance. 
- TA dashboards are created using mycohortinfo data.
- Mycohortinfo breaks if the TA isn't on canvas. So, I was asking about a way we can have TAs in canvas while still allowing them to create a platform account. 

NEW information discovered: Mycohortinfo doesn't appear to break if we just upload the CSV with the TAs name exactly matching that on salesforce (and match on the name on the Periscope side.)

sadleb: I would add this context to `user_can_edit_role?` as a comment then @Rowantatai - after that, I'm good with the rest.

app/services/sync_portal_enrollment_for_account.rb
Rowantatai: Ditto on [Ryan's comment](https://github.com/bebraven/platform/pull/510#discussion_r544378607)

sadleb (approved):

PR #514, opened 2020-12-17
Implement pre- and post-accelerator surveys

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1189491188178912

### Motivation and Context

For every Canvas course, there is one pre-accelerator survey and one post-accelerator survey. We assume they are unique to a course, like the waivers and peer reviews assignments. Like waivers and peer reviews, we only allow you to submit the accelerator surveys once. Once you've submitted, viewing to the assignment shows you the same confirmation page. Like waivers, pre- and post-accelerator surveys are built using FormAssembly and we do not store anything in `platformdb` about the submissions or responses. In order to determine whether these surveys have been submitted, we use the `LTIAdvantageAPI` to send a full-credit score to Canvas when the survey is submitted, which we query when we determine whether to render the form or show the confirmation screen.

There's a fair amount of set-up you need to do in order to get this to work end-to-end.

**Canvas enrollment**
- You have to be enrolled as a Student in the Canvas Course to submit the Accelerator Surveys and have the `canvas_user_id` set on the `User` in `platformdb`. This is enforced by the `LTIAdvantageAPI`. The `create_score` fails if the `user.canvas_user_id` isn't a student in the course. (This is the same for anything we do an `LTIAdvantageAPI` call for, for example, project submissions.)

**Salesforce program fields**
- You need to configure a Salesforce program so its `Highlander Accelerator Course ID` is your test course's Canvas ID. (For example, I used https://bebraven--highndrdev.lightning.force.com/lightning/r/Program__c/a2Y11000001Hg7hEAC/view and my playground course with ID `43`).
- You also need to configure the `FA ID Fellow PreSurvey` and `FA ID Fellow PostSurvey` Salesforce fields with FormAssembly IDs. I used `4810855` and `4810858`. 
Screen Shot 2020-12-17 at 6 54 56 PM

**Form Assembly connectors**
- For those forms, you will need to set up Connectors in FormAssembly, go to https://braven.tfaforms.net/forms/edit/summary/<your-form-assembly-form-id>
- Configure a **PostRedirect** connector that points to `AcceleratorSurveySubmissionsController#create`:
Screen Shot 2020-12-17 at 6 41 05 PM
Screen Shot 2020-12-17 at 6 41 16 PM
- The endpoint for pre-accelerator surveys is: https://platformweb/preaccelerator_survey_submissions
- The endpoint for post-accelerator surveys is: https://platformweb/postaccelerator_survey_submissions

Once you've done the above, you can manually test the functionality.

- On the Course's edit page, you can **Add** pre- and post-accelerator surveys:
Screen Shot 2020-12-17 at 6 32 44 PM

- Logged in as a student in Canvas, you can go through the survey like this:
Screen Shot 2020-12-17 at 6 48 02 PM
Screen Shot 2020-12-17 at 6 48 11 PM
Screen Shot 2020-12-17 at 6 48 22 PM

- The only difference between pre- and post-accelerator surveys (aside from the FormAssembly form) is the text in the UI:
Screen Shot 2020-12-17 at 6 46 26 PM

### Types of changes

### Checklist:


app/controllers/accelerator_survey_submissions_controller.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+require 'form_assembly_api'
+
+class AcceleratorSurveySubmissionsController < FormAssemblyController
```
sadleb, line 5: Let's add a comment to this class talking about what an `AcceleratorSurvey` is. We're deep in this stuff and intuitively know it's talking about Pre and Post Accelerator surveys but a future dev might think this is more general (which it may eventually be if we add more FormAssembly driven surveys which is why I like the name).

app/controllers/accelerator_survey_submissions_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+  # Because AcceleratorSurveySubmission doesn't exist in the DB, we always set
+  # it using .new before any of the actions so we have an instance.
+  before_action :set_new_model_instance
+  before_action :set_form_id!, only: [:new]
+  before_action :setup_formassembly, only: [:new]
+
+private
+  # For Submittable
+  def answers_params_hash
+    {} # Nothing to save in our DB for these surveys
```
sadleb, line 23: Can you mention how FormAssembly uses the SalesforceConnector to submit the data directly back to Salesforce?

app/controllers/accelerator_survey_submissions_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def parent_variable_instance
+    course
+  end
+
+  # We fake the #show endpoint for the submission since there aren't real IDs
+  def instance_path(_instance)
+    return preaccelerator_survey_submission_path(id: 0) if params[:type] == 'Pre'
+    return postaccelerator_survey_submission_path(id: 0) if params[:type] == 'Post'
+    throw NotImplementedError
```
sadleb, line 48: This bit is a little confusing. I assume the `show` route isn't used anywhere and this is just to make submittable work?

app/models/accelerator_survey_submission.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# This is not an ApplicationModel because there's no corresponding table, e.g.
+# `accelerator_survey_submissions`, backing it.
+# This behaves like an ActiveModel so we can use `Submittable` in the
+# AcceleratorSurveySubmissionsController and with our AcceleratorySurveyPolicy.
+class AcceleratorSurveySubmission
```
jessouyangbraven, line 7: TODO: Get rid of this

app/controllers/accelerator_survey_submissions_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def submission_url(_instance)
+    return preaccelerator_survey_submission_url(id: 0) if params[:type] == 'Pre'
+    return postaccelerator_survey_submission_url(id: 0) if params[:type] == 'Post'
+    throw NotImplementedError
+  end
```
jessouyangbraven, line 55: Use #completed endpoint instead

app/models/accelerator_survey_submission.rb
jessouyangbraven: So this ended up staying because of the `AcceleratorSurveySubmissionPolicy`. I didn't now how else to neatly pass in both `user` and `course` as part of the `record` otherwise. I looked at the `WaiverSubmissionPolicy`, but it doesn't check for course enrollment yet. If there's a better way to write the policy that also lets me get rid of this model, I'm all ears. Otherwise, I might have to go ahead and add a `WaiverSubmission` model to make things consistent?

app/policies/accelerator_survey_submission_policy.rb
```diff
@@ -0,0 +1,53 @@
...
+    enrolled = false
+    record.course.sections.each do |section|
+        RoleConstants::SECTION_ROLES.each do |role|
+          enrolled = true if user.has_role? role, section
+          break if enrolled
+        end
+    end
+
+    # Fail early if you aren't enrolled in the course
+    return false unless enrolled
```
jessouyangbraven, line 43: I don't know if this is overkill, since you can't really get the link in Canvas unless you're in the course already, but the specs were pointing out to me that without this check that a TA could theoretically view their Fellow's submissions for another course.

config/routes.rb
```diff
@@ -93,22 +107,19 @@
...
+  resources :preaccelerator_survey_submissions, controller: 'accelerator_survey_submissions', type: 'Pre', only: [:new, :create] do
+    collection do
+      get :completed
+    end
+  end
+
+  resources :postaccelerator_survey_submissions, controller: 'accelerator_survey_submissions', type: 'Post', only: [:new, :create] do
+    collection do
+      get :completed
+    end
```
jessouyangbraven, line 122: I don't think we need to change the LTI launch URLs for the pre- and post-accelerator surveys in Canvas when we clone a course. Ditto for `waiver_submissions`. I think that leaves peer reviews to migrate for the assignments we only have one of per course.

app/controllers/accelerator_survey_submissions_controller.rb
```diff
@@ -0,0 +1,132 @@
...
+    form_id = form_assembly_info.send("#{@type}_accelerator_survey_form_id")
+    if params[:tfa_next].present?
+      @form_head, @form_body = FormAssemblyAPI.client
+        .get_next_form_head_and_body(params[:tfa_next])
+    else
+      @form_head, @form_body = FormAssemblyAPI.client
+        .get_form_head_and_body(form_id, participantId: form_assembly_info.participant_id)
+    end
+    setup_head()
+    setup_body()
```
rshipp, line 73: can we rename all the `setup` stuff to a verb, for clarity? maybe just `set_up_*`?

app/policies/accelerator_survey_submission_policy.rb
rshipp: Oh, that's a good point. I wonder if I've missed that in other places...

If i'm following correctly, I think this solution still has an edge case:

* You're enrolled in this course (as any role, including as a student), and as a TA for this student in a different course
  * You shouldn't be able to see this submission, but you can

app/views/courses/edit.html.erb
```diff
@@ -62,7 +62,8 @@
 
       <div class="form-group">
         <div class="row">
-          <span class="col-md-4"><p>Peer Reviews</p></span>
+          <span class="col-md-3"><p>Peer Reviews</p></span>
+          <span class="col-md-9 align-right"><small><em>Note: peers in cohort are controlled by Salesforce</em></small></span>
```
rshipp, line 66: this is probably outside the scope of this PR, but as a note - `<span>` is an "inline element" and shouldn't have `<p>`s inside it

spec/policies/accelerator_survey_submission_policy_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      expect(subject).to permit user, accelerator_survey_submission
+    end
+
+    it "disallows non-admin users not enrolled in a course attached to the project" do
+      expect(subject).not_to permit user, accelerator_survey_submission
+    end
+  end
+
+  permissions :create? do
+    it "allow a fellow enrolled in course to submit" do
```
rshipp, line 67: add a check for non-enrolled user? or is that overkill?

app/models/accelerator_survey_submission.rb
rshipp: I don't remember all the things we talked about regarding this, but having an empty or mostly-empty class just to keep things consistent seems fine to me.

sadleb: Ditto. Seems fine.

app/controllers/accelerator_survey_submissions_controller.rb
```diff
@@ -0,0 +1,132 @@
...
+    {}
+  end
+
+  # Overrides Submittable
+  # We give a full-credit score to the Fellow for submitting their surveys,
+  # so we can check Canvas to see whether this exists in order to determine
+  # whether there's already been a submission.
+  def previous_submission
+    @line_item ||= LtiAdvantageAPI.new(@lti_launch).get_result().present?
+    return @line_item ? @accelerator_survey_submission : nil
```
sadleb, line 92: It doesn't seem like `@line_item` is used anywhere. Is it just to ensure that subsequent calls to `previous_submission` don't hit the `LtiAdvantageAPI` again? 

Also, can you mention why we have to return a fake `@accelerator_survey_submission` here. Is it only for Pundit authorization purposes? Or do we also need it for the `instance_path` which we also override since there isn't an actual instance path.

app/controllers/accelerator_surveys_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  include Publishable
+
+  layout 'admin'
+
+  prepend_before_action :set_type!
+
+private
+  # For Publishable
+  def assignment_name
+    "TODO: Complete %{type}-Accelerator Survey" % { type: @type }
```
sadleb, line 23: Is there a reason for this syntax vs `"TODO: Complete #{@type}-Accelerator Survey"`?

app/controllers/concerns/publishable.rb
```diff
@@ -192,6 +192,7 @@ def canvas_assignment_id
     if model_class && instance_variable.respond_to?(:canvas_assignment_id)
       return instance_variable.canvas_assignment_id
     end
+    params.require(:canvas_assignment_id)
     params[:canvas_assignment_id]
```
sadleb, line 196: I think both `params.require(:canvas_assignment_id)` and `params[:canvas_assignment_id]` return the same thing so you can just return `params.require(:canvas_assignment_id)` from this?

app/models/accelerator_survey_submission.rb
```diff
@@ -0,0 +1,15 @@
...
+  include ActiveModel::Model
+
+  attr_accessor :user, :course
+
+  # For Submittable
+  def save_answers!(_input_values_by_user_and_question)
+  end
```
sadleb, line 14: Maybe mention to see the comments in `accelerator_survey_submissions_controller.rb` for why this is a NOOP?

app/policies/accelerator_survey_submission_policy.rb
```diff
@@ -0,0 +1,53 @@
...
+  def new?
+    # Admins can see the form
+    return true if user.admin?
+
+    # Anyone enrolled in the course can see the form
+    record.course.sections.each do |section|
+        RoleConstants::SECTION_ROLES.each do |role|
+          return true if user.has_role? role, section
+        end
+    end
```
sadleb, line 15: Worth putting this in a private `is_enrolled?` method and calling it from `completed` as well?

config/routes.rb
sadleb: That seems right.

spec/controllers/accelerator_survey_submissions_controller_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+      end
+    end
+
+    context 'invalid parameters' do
+      it 'raises an error' do
+        [
+          { type: 'Pre' }, # missing state
+          { state: lti_launch.state }, # missing type
+          { state: lti_launch.state, type: 'Foo' }, # invalid type
+        ].each do | invalid_params |
```
sadleb, line 129: Whoa, gettin' fancy. I've never thought of doing it this way.

sadleb (approved):
None of my comments are merge blocking. Just readability related when I hit something that was confusing and I had to think about it and look at the other code to understand what was going on.

app/controllers/accelerator_survey_submissions_controller.rb
jessouyangbraven: `setup_head()` and `..._body()` are actually defined in `FormAssemblyController`. I'll do this in a follow-up PR so I don't blow up this PR's changeset anymore than I already have. I opened a task so I'd remember: https://app.asana.com/0/1168520191527013/1199512664295639

app/controllers/accelerator_survey_submissions_controller.rb
jessouyangbraven: I'll add comments in the code! But to answer your questions:

>It doesn't seem like @line_item is used anywhere. Is it just to ensure that subsequent calls to previous_submission don't hit the LtiAdvantageAPI again?

Yes.

>Also, can you mention why we have to return a fake @accelerator_survey_submission here. Is it only for Pundit authorization purposes? Or do we also need it for the instance_path which we also override since there isn't an actual instance path.

It's for the `instance_path` called in `Submittable`.

app/controllers/accelerator_surveys_controller.rb
jessouyangbraven: I think it's copy-pasta from somewhere else. I'll change to our usual syntax.

app/policies/accelerator_survey_submission_policy.rb
jessouyangbraven: You're right about the edge case here. I was relying on `user.ta_for?` to handle that, but I just stared at the code and realized it's section-agnostic.

I tried adding `ta_for?(target_user, section=nil)` to fix this, but realized we don't never determine a specific section to use in the policy or controller.

I think the fix is going to require a bit of thought, so I opened a task: https://app.asana.com/0/1168520191527013/1199512664295642

I'll add a `TODO` here with that URL, too.

app/views/courses/edit.html.erb
jessouyangbraven: I think I've copy-pasted this all over our views by now. :(

Task to figure out how widespread this is: https://app.asana.com/0/1168520191527013/1199512664295643

PR #519, opened 2020-12-18
Copy assignment overrides from cohort schedule once cohort mapping has taken place

rshipp:
### Description

Task: https://app.asana.com/0/1174274412967132/1198893636951321/f

General flow for new participants is:

1. Participants are enrolled in the course, and given a Cohort Schedule in Salesforce, e.g. "Tuesdays at 6pm" (representing the time/day of the Learning Lab), which acts as a placeholder for their cohort.
2. Sync to LMS runs, and creates a Canvas section for the Cohort Schedule
3. Someone manually edits all assignment due dates in Canvas, assigning the section "Tuesdays at 6pm" a due date of e.g. "Tuesdays at 11:59 PM".
4. At some point early on in the course, possibly after Learning Lab 1, "Cohort Mapping" takes place. This is a manual process in which Fellows are assigned to Cohorts, each Cohort with its own LC. These Cohorts are added to the Participant records in Salesforce.
5. Sync to LMS runs again, sees the participants have a cohort field, and moves them from their existing section in Canvas (in this case, the Cohort Schedule section), to a new section specific to their Cohort.
6. When Sync to LMS creates a new Cohort-specific section in Canvas, it copies all due dates from the Cohort Schedule section to the new Cohort section. This means Braven staff only have to set due dates once per cohort-schedule (usually 1 or 2, depending on the size of the program), instead of once per cohort (a much larger number).

Note some other edge cases are also handled in this PR; see https://docs.google.com/document/d/1PRwY74-vgIJwD5ow093hGmKaf8ZNqxijjNtJuVISHOg/edit?usp=sharing for more details.

### Types of changes

### Checklist:


sadleb:
@rshipp -- looks like Slack is down, so copy/pasta from there:

"over the break I didn't get a chance to revisit https://github.com/bebraven/platform/pull/519 -- I think the best way to wrap this up is just to re-write the specs more robustly so that it ensures the requirements are met. We'll test it manually as part of the Mock Launch coming up if that's too much effort in order to get it merged. I tried to [update the technical design](https://docs.google.com/document/d/1PRwY74-vgIJwD5ow093hGmKaf8ZNqxijjNtJuVISHOg/edit#heading=h.iql75orj74cx) to match where we left off by writing out the "rules" instead of trying to map out the SF fields -> Canvas/Platform sections permutations."

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -127,27 +127,26 @@ def find_or_create_section(canvas_course_id, section_name)
       if sf_participant.cohort
         # The user has been added to a cohort without an existing section.
         # Copy the due dates from the cohort-schedule section to the new section.
-        cohort_schedule_section = find_canvas_section(canvas_course_id, sf_participant.cohort_schedule)
+        cohort_schedule_section = find_canvas_course_section(canvas_course_id, sf_participant.cohort_schedule)
         # TODO: get assignment_overrides from cohort_schedule_section and copy them to canvas_section.
         cohort_schedule_overrides = []
-        assignment_ids = get_assignments(canvas_course_id).map { |a| a['id'] }
+        assignment_ids = canvas_client.get_assignments(canvas_course_id).map { |a| a['id'] }
```
sadleb, line 133: Whoops

app/controllers/users_roles_controller.rb
```diff
@@ -116,7 +117,7 @@ class Mocks
 
     CanvasUser = Struct.new(:id)
 
-    SalesforceParticipant = Struct.new(:email, :platform_role, :cohort, :status) do
+    SalesforceParticipant = Struct.new(:email, :platform_role, :cohort, :cohort_schedule, :status) do
```
rshipp, line 120: @sadleb I'm not sure of the changes in this file, because I don't have full context around when the users-roles controller is used. Is it ok to have a "nil" cohort-schedule here?

app/jobs/sync_salesforce_program_to_lms_job.rb
```diff
@@ -10,6 +10,7 @@ def perform(program_id, email)
   end
 
   rescue_from(StandardError) do |_exception|
+    Rails.logger.error(_exception)
```
rshipp, line 13: don't hide exceptions that happen during the job

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -94,12 +94,15 @@ def drop_course_enrollment(canvas_course_id)
...
   # Enroll or update their enrollment in the proper course and section
   def sync_enrollment(canvas_course_id, role, section_name)
+
     section_name = section_name.blank? ? DEFAULT_SECTION : section_name
     section = find_or_create_section(canvas_course_id, section_name)
     enrollment = find_user_enrollment(canvas_course_id)
     if enrollment.nil?
       enroll_user(canvas_course_id, role, section)
-    elsif !enrollment.section_id.eql?(section.id) || !enrollment.type.eql?(role)
+    elsif !enrollment.section_id.eql?(section.canvas_section_id) || !enrollment.type.eql?(role)
```
rshipp, line 103: this is an unrelated change, but an important bugfix. this check previously compared a canvas section ID with a local db section ID, which would have always failed. this change is also reflected in the specs at https://github.com/bebraven/platform/pull/519/files#diff-3727a5051affc065fc37d2228632a96f80b0ee1a91e81375b72f197c78d63ddeL69

lib/canvas_api.rb
```diff
@@ -283,6 +283,20 @@ def create_assignment_overrides(course_id, assignment_ids, section_ids)
...
   end
 
+  def get_assignment_overrides(course_id, assignment_id)
+    response = get("/courses/#{course_id}/assignments/#{assignment_id}/overrides")
+    get_all_from_pagination(response)
+  end
+
+  # This function name makes absolutely zero sense but so does every other name
+  # and ruby doesn't have method overloading.
+  def create_copy_assignment_overrides(course_id, overrides)
```
rshipp, line 293: Question: is there a better way to name this method, so that it's easy to differentiate from `create_assignment_overrides`? I thought about `create_assignment_overrides_from` but that also seems unclear. "Copy" is how we use this in Sync to LMS, but nothing about this function is actually specific to copying. What it really does is create based on the overrides you pass in, as opposed to https://github.com/bebraven/platform/pull/519/files#diff-40a0ff3e21920d81a7633cc6e4965caa7c945c815a80e33e89e38f75288fb292R273 which creates from the assignment/section IDs you pass in.

TODO: update this comment appropriately.

spec/services/sync_portal_enrollment_for_account_spec.rb
```diff
@@ -66,14 +86,14 @@
         expect(lms_client).not_to have_received(:create_lms_section)
       end
 
-      it 'enrolls the user to course if not enrolled' do
+      it 'does not re-enroll the user' do
```
rshipp, line 89: this test used to be "when the participant is enrolled, it enrolls them if they are not enrolled", which doesn't make any sense. this test was previously passing because of the broken section ID check mentioned above.

app/jobs/sync_salesforce_program_to_lms_job.rb
sadleb: Send to honeycomb too? Honestly, I think we should send the email and then re-raise letting it bubble up normally

app/controllers/users_roles_controller.rb
sadleb: Yeah, I think that’s fine. It’s just for devs and designers to more easily create test users. There is a field to type the section into which is sent as the cohort. So as long as the code can create a canvas and platform section from the cohort without then already haven been in a cohort schedule, this is good

lib/canvas_api.rb
sadleb: Maybe rename the other one to `create_assignment_override_placeholders` and then make this one just plain `create_assignment_overrides` since that is what it’s doing?

app/jobs/sync_salesforce_program_to_lms_job.rb
rshipp: I added a honeycomb line, but raising inside a rescue_from doesn't seem to be supported behavior. see https://simonecarletti.com/blog/2009/11/re-raise-a-ruby-exception-in-a-rails-rescue_from-statement/

app/services/sync_portal_enrollment_for_account.rb
jessouyangbraven: Is this fix something we need to add to the non-highlander platform branch?

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -116,12 +119,47 @@ def enroll_user(canvas_course_id, role, section)
   end
 
   # Creates a local DB Section and a Canvas section, but only returns the local one.
+  # Also handles copying Canvas assignment overrides when appropriate.
   def find_or_create_section(canvas_course_id, section_name)
     canvas_section = find_canvas_course_section(canvas_course_id, section_name)
     if canvas_section.nil?
+      # If the Canvas section didn't already exist, create it now.
```
jessouyangbraven, line 126: I keep getting confused between `section_name`/`section` vs. `cohort_schedule_section`. Does `section` refer to the cohort section?

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -116,12 +119,47 @@ def enroll_user(canvas_course_id, role, section)
...
       canvas_section = canvas_client.create_lms_section(course_id: canvas_course_id, name: section_name)
+
+      if sf_participant.cohort
```
jessouyangbraven, line 129: Why do we only run this when the `canvas_section` is first created?

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -116,12 +119,47 @@ def enroll_user(canvas_course_id, role, section)
...
+        # The user has been added to a cohort without an existing section.
+        # Copy the due dates from the cohort-schedule section to the new section.
+        cohort_schedule_section = find_canvas_course_section(canvas_course_id, sf_participant.cohort_schedule)
+        # In the edge case where the cohort-schedule section doesn't already
+        # exist in Canvas, we just skip all this assignment-override stuff.
```
jessouyangbraven, line 134: Shouldn't we also create the `cohort_schedule_section` here like the `section` above? Or does that get created by something else?

jessouyangbraven (commented):
I'm pretty confused with the `sync_portal_enrollment_for_account` code. I left comments inline, but I think it might be easier to clarify if I just ask them in stand up today.

app/services/sync_portal_enrollment_for_account.rb
rshipp: Because we only want to copy assignment overrides from the cohort-schedule section to the cohort section when the cohort section is first created. On subsequent runs, when more fellows are moved to the existing cohort section, the overrides have already been copied.

app/services/sync_portal_enrollment_for_account.rb
rshipp: Only sometimes 🙃. See `cohort_section_name` [above](https://github.com/bebraven/platform/pull/519/files#diff-f7c530ffb06b845aac9a89e1c327913e1a578a735c4a63cb8401841a1d1d6a5fR62-R65). It defaults to cohort, and falls back to cohort-schedule if the fellow has not yet been mapped to a cohort. That's what makes all of the rest of this work.

app/services/sync_portal_enrollment_for_account.rb
rshipp: No, because the only reason we care about the cohort-schedule section is if we're copying assignment overrides from it. If the cohort-schedule section doesn't already exist, it can't have any assignment overrides, and so we can safely skip this step.

app/services/sync_portal_enrollment_for_account.rb
rshipp: i think the impact for this is "we un-enroll and re-enroll people a lot for no reason". the fact that no one has noticed makes me think this just comes down to "we do a lot of canvas API calls for no reason". i'll leave it to @sadleb whether it's worth backporting.

sadleb: Spring Accelerator is launching soon, so if there is an issue, it prob won’t be noticed until launch time. We should ask Rowan to verify that the folks being synced actually end up in the correct sections when moved between cohorts. If so, let’s not touch the Accelerator code

PR #526, opened 2021-01-05
Small change to password reset to redirect to Portal afterwards

hcli-work:
https://app.asana.com/0/1179330853755533/1199540555646309/f 

After resetting password on platform, previous functionality redirected via `service` param to platform's `users/services` endpoint. This caused lots of confusion among users. Instead, we're going to redirect to Portal by default after a password reset. This should make it slightly more intuitive for users and a little more confusing for platform admins.

Test Plan:
click on forget password
check logs to grab reset password URL
visit reset URL
reset password
check that new service URL in params points to Portal
login with new password and click to login as staff

Also checked that Accelerator/BravenX, Booster, Highlander, and dev environments all have the `CANVAS_URL` environment variable


a-willis (approved):

sadleb:
Note @hcli-work - here was the task in the Highlander project for this: https://app.asana.com/0/1174274412967132/1199599522871561 -- imma link to this PR so we can merge this to the `highlander` branch and sanity check that it fixes it there.

PR #529, opened 2021-01-06
Re-enable duplicate section name 

Rowantatai:
**Issue:** Section related logic intended for highlander is causing issues for current programs. Here's an example 

`ActiveRecord::RecordNotUnique (PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_sections_on_name_and_base_course_id"
Jan 06 07:57:20 braven-platform app/web.1: DETAIL:  Key (name, base_course_id)=(Default Section, 2) already exists.`

"there is a bug that is going to break more than TAs if we have the same section name in multiple courses. since we don't actually use this BaseCourse and UserSection stuff in the non-Highlander logic, it's fine to do whatever we want with that dummy data” - Bsadler


sadleb (approved):

hcli-work:
in the future, can you make sure to include a test plan section at the bottom of PRs?

**Test Plan:** bundle exec rake db:migrate. # no errors popped up

PR #531, opened 2021-01-07
Fix Peer Review for non-fellows and only fellow in section

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199508237905859

### Motivation and Context

There are a couple of cases we want to handle better for showing `PeerReviewSubmission#new`:
- User who is the only Fellow in a section, e.g. no peers to review (yet)
- User who is not a Fellow, e.g. no "peers" ever and you can't actually submit to `LTIAdvantageAPI`, but we want to support previewing the page

- **After**: this is what both cases described above see now:
Screen Shot 2021-01-07 at 11 11 43 AM

- **Before** for Fellow without any peers:
Screen Shot 2021-01-06 at 7 33 48 PM

- **Before** for Admin:
Screen Shot 2021-01-06 at 8 42 03 PM

- **Before** for TA:
I didn't test manually, but the new spec I added is failing on the policy for `new?`

### Types of changes

### Checklist:


app/views/peer_review_submissions/new.html.erb
```diff
@@ -8,64 +8,68 @@ Leadership Coach for your contributions to your team throughout the Challenge. T
 minutes right now reflecting on your teammates and evaluating them. Your evaluation will be averaged with all the
 others, resulting in a final Teamwork score for each Fellow in your cohort.</p>
 
-<%= form_with(url: course_peer_review_submissions_path, method: "post", local: true) do |form| %>
+<% unless @peer_users.present? %>
+  You have no peers to review.
```
jessouyangbraven, line 12: RFC: Any feedback on how this looks?

jessouyangbraven: RFC: Also, should the messaging be different for fellows and non-fellows?

I'm guessing that in production we never want the situation where a user is the only fellow in their section, so this isn't as a big of a deal.

But LCs/TAs/admins technically never have any peers to review on this assignment page, so I'm guessing we're just letting them preview here. Is this even a useful preview without any people in the form? But if we do let them somehow submit, the `LTIAdvantageAPI` errors out because they're not a student.  Should this be a separate placeholder view that says "Hey this isn't relevant to you?".

spec/controllers/peer_review_submissions_controller_spec.rb
```diff
@@ -5,70 +5,125 @@
...
+    end
+
+    context 'as non-fellow (TA) user' do
+      let!(:user) { create :ta_user, section: section }
+      let!(:fellow_user) { create :fellow_user, section: section }
+
+      # FIXME: Failing on PeerReviewSubmissionPolicy.new?
+      # Should we show "No Peers to Review" and omit the form like we do admins?
+      # it_behaves_like 'valid request'
+      # it_behaves_like 'no peers to review'
```
jessouyangbraven, line 88: RFC: See comment above. Not sure if we should handle this like admins for now, or do something else.

jessouyangbraven (commented):
I figured out what was breaking, which wasn't actually "I'm the only fellow in this section" (that'd show an empty form and the Save button). It was actually non-fellows, like admins and TAs that were returning `nil` for `section`. I left some RFCs inline to get feedback on how to handle these scenarios.

app/controllers/peer_review_submissions_controller.rb
```diff
@@ -11,10 +11,16 @@ class PeerReviewSubmissionsController < ApplicationController
   before_action :set_questions, only: [:new, :create]
 
 private
-
   def set_peer_users
     section = current_user.student_section_by_course(@course)
-    @peer_users = section.students.where.not(id: current_user.id)
+    if section&.students
+      # You're enrolled as a student and there are other students in this course
```
rshipp, line 17: is there somewhere else we're doing a "you're enrolled as a student" check? this doesnt seem to do that

app/views/peer_review_submissions/new.html.erb
rshipp: I would just do whatever's the simplest way to show a "you have no peers" message to anyone who's not a student with peers.

sadleb: Agree with Ryan. I think it’s fine to show the same message for everyone. I would go with: “<p>You have no peers to review. If you think this is wrong, please reach out to support@bebraven.org.</p>”

spec/controllers/peer_review_submissions_controller_spec.rb
sadleb: Yeah, let’s show the “no peers to review” msg

app/views/peer_review_submissions/new.html.erb
sadleb: Huh, git removed the < p > tags. I think it should be in a new paragraph to give it some space

app/controllers/peer_review_submissions_controller.rb
jessouyangbraven: It's not super obvious, but the line above, 15, that does `section = current_user.student_section_by_course(@course)` does that.

It only returns the section that you're enrolled in as a student.

Should I add a comment or rename the variable to make it more obvious?

rshipp: Ahh i missed the `student_section_by_course` part, thanks for clarifying.

sadleb (approved):
Lgtm

PR #539, opened 2021-01-08
Add FellowEvaluationSubmissionAnswer (2/?)

jessouyangbraven:
**dev env breaking changes**
- In `deva`, run `bundle exec rails db:migrate`.

### Description

**Task**: https://app.asana.com/0/1174274412967132/1196915754308340

### Motivation and Context

`FellowEvaluationSubmissionAnswer` represents a response to a `FellowEvaluationQuestion` that's part of a `FellowEvaluationSubmission` that the LC submits when they're reviewing their Fellows. 

It is like `PeerReviewSubmissionAnswer`.

### Test

- specs

### Types of changes

### Checklist:


rshipp (approved):

db/migrate/20210108003729_create_fellow_evaluation_submission_answers.rb
```diff
@@ -0,0 +1,10 @@
+class CreateFellowEvaluationSubmissionAnswers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :fellow_evaluation_submission_answers do |t|
+      t.references :fellow_evaluation_submission, null: false, foreign_key: true, index: { name: :index_fellow_evaluation_submission_answers_on_submission_id }
+      t.references :for_user, null: false, foreign_key: { to_table: :users }
+      t.string :input_name
```
jessouyangbraven, line 6: This used to be a reference to `fellow_evaluation_question`, which it now just an `input_name`. (This is also how we store `ProjectSubmissionAnswer` and `SurveySubmissionAnswer`.)

The `input_name` strings are taken from `data-bz-retained` in the HTML view of this Portal Fellow Assessment:
https://portal.bebraven.org/courses/1/pages/fellow-assessment/edit?

Examples are: `would-hire-entry-level-role`, `how-ready-to-be-professional`.

I think it makes sense to keep them the same/similar the data analysis doesn't have to change as much.

db/schema.rb
```diff
@@ -103,6 +103,15 @@
     t.string "type"
   end
 
+  create_table "fellow_evaluation_submission_answers", force: :cascade do |t|
+    t.bigint "fellow_evaluation_submission_id", null: false
+    t.bigint "for_user_id", null: false
```
sadleb, line 108: Is there a reason why this is on the answer and not the submission model? A single submission will have multiple answers, but they are all for the same user, right?

jessouyangbraven: >A single submission will have multiple answers, but they are all for the same user, right?

A single submission has answers for all the users you're reviewing. We do the same thing for `PeerReviewSubmission` and `Answers`.

I think we had a reason for doing it this way, rather than having a submission per `for_user`. I think it was because we only allow you to submit once and only ever show the form for everyone you need to review?

I'm going to leave it as-is for now to keep it consistent with peer reviews, but if we want to change the data model, we can do it for both of them with a few migrations. It should be pretty straightforward?

jessouyangbraven: Task to keep track: https://app.asana.com/0/1174274412967132/1199544270990456

sadleb: Oh, this makes sense. I'm going to delete the follow-up task.

PR #540, opened 2021-01-08
FellowEvaluationSubmission controller, policy, views (3/?)

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1196915754308340

### Motivation and Context

This PR adds the controller, policy, and views for `FellowEvaluationSubmission`. 

### Test

- This is what the Fellow Evaluation form looks like:
Screen Shot 2021-01-11 at 1 32 26 PM
Screen Shot 2021-01-11 at 1 32 33 PM

- This is what the confirmation looks like:
Screen Shot 2021-01-11 at 1 31 00 PM

- specs

### Types of changes

### Checklist:


app/views/fellow_evaluation_submissions/new.html.erb
```diff
@@ -0,0 +1,73 @@
...
+          <% input_name = "fellow_evaluation[#{user.id}][#{question.id}]" %>
+          <% input_id_prefix = "radio-#{user.id}-#{question.id}" %>
+
+          <fieldset class="peer-review-question-wrapper">
+            <legend class="peer-review-question">
+              <%= question.text %>
+            </legend>
+
+            <div class="container-fluid p-0">
+              <div class="row peer-review-radios">
```
jessouyangbraven, line 32: RFC: I ran into a snag with this because the questions pretty much all have different input type/labels/values.

1. "Likely", values (10, 8, 6, 4, 2, **0**)
2. "Ready", values (10, 8, 6, 4, 2)
3. Same as no. 2
4. Checklist + text input: 
Screen Shot 2021-01-07 at 9 52 59 PM
5. Same as  no. 4
6. Text input

Sources: https://portal.bebraven.org/courses/148/pages/fellow-assessment?module_item_id=7679, and https://portal.bebraven.org/courses/1/pages/fellow-assessment/edit (in HTML mode is how I'm getting the values).

jessouyangbraven: I'm not sure if we want `FellowEvaluationQuestion` anymore. Maybe we just hard-code all of this in views using partials to cut down on the duplication between 2/3 and 4/5?

app/controllers/fellow_evaluation_submissions_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  before_action :set_fellow_users, only: [:new, :create]
+  before_action :set_questions, only: [:new, :create]
+
+private
+  def set_fellow_users
+    # Get all sections where this user is a TA
+    sections_as_ta = current_user.sections_with_role(RoleConstants::TA_ENROLLMENT)
+
+    # FIXME: What if you're a TA in multiple courses over multiple runs?
```
jessouyangbraven, line 24: RFC: I know we said not to worry about this and just add all the fellows in one giant list, but I think it'd be weird to see fellows from previous runs of courses. I'm not sure how to fix it though.

sadleb: Good point re: previous runs. Hmmmm, I can't think of a way to handle that without knowing which Accelerator course the LC Playbook is associated with.

sadleb: @rshipp we've been able to avoid doing this up to now, but is it time to add some way to associate the two to our data model? I think introducing a Program model would be the best way? We're going to need that for the future enhancement to show a list of programs and allow staff to re-launch a program from the Course Mgmt page.

If we introduce that, then you would just look up the current `Program` using the current canvas course id and grab the associated Accelerator course so you could filter the sections down to just that.

app/views/fellow_evaluation_submissions/new.html.erb
sadleb: Woof. Yeah, I suppose hard-coding is the better route for now rather than try and data-model out how to store these in the db.

rshipp: Hardcoding seems fine to me, unless we can get a quick decision from product that it's OK to normalize 1/2/3 to the same format, and then we could put those in the DB and just hardcode 4.

app/controllers/fellow_evaluation_submissions_controller.rb
rshipp: That does seem like the correct route :/ It'll complicate the salesforce accessors we have set up in the Course model.

jessouyangbraven: Salesforce route: Course -> Program -> Look up in SF -> Find Accelerator Course (issues if SF is down)
Data model route: Program data model for future

app/views/fellow_evaluation_submissions/_checkbox_question.html.erb
```diff
@@ -0,0 +1,44 @@
...
+               name="<%= "fellow_evaluation[#{user.id}][#{question_name}-#{value}]" %>">
+            <label for="<%= checkbox_id %>"><%= label %></label>
+          </div>
+      <% end %>
+    </div>
+    <div class="col-sm pr-0">
+      <label for="<%= "checkbox-#{user.id}-other" %>">Other</label>
+      <input id="<%= "checkbox-#{user.id}-other" %>" type="text" name="<%= "fellow_evaluation[#{user.id}][#{question_name}-other]" %>">
+    </div>
+  </div>
```
jessouyangbraven, line 42: The `name` attribute on the `<input>s` was super fiddly, but the data is nicely formatted by fellow ID when we `POST`:

Screen Shot 2021-01-11 at 1 21 45 PM

rshipp (approved):

app/views/fellow_evaluation_submissions/new.html.erb
```diff
@@ -0,0 +1,80 @@
+<% content_for :title, 'Fellow Evaluations' %>
+
+<% content_for :additional_container_css, 'peer-review-container' %>
+
+<h2 id="cohort-teamwork-evaluation">Accelerator Fellow Evaluation</h2>
+<p>Thank you for being a Leadership Coach this semester! We really appreciate all the work and time you’ve dedicated to helping and teaching our Fellows. As we continue to grow as an organization, we've found that feedback from Coaches (who know their Fellows better than we do), has been crucial in helping us tailor post-course programming to meet Fellows' needs. So, we'd like to get some feedback on the Fellows in your cohort. Your responses will be used by Braven staff only (to better understand strengths and growth areas for Fellows and our program overall) and will not be reported back to your cohort members.</p>
+<p>Please reach out to ayodele@bebraven.org with any questions. Thanks for your responses!</p>
```
sadleb, line 7: @jessouyangbraven I queued up a future task to un-hardcode the point-of-contact: https://app.asana.com/0/1174274412967132/1199539421634461

PR #547, opened 2021-01-13
Ignore Visual Studio Code files

sadleb:
This is the local VSCODE config stuff, like launch.json
for launching a debugging session (still haven't gotten that
working btw).


sadleb:
No need to code review this.

PR #550, opened 2021-01-13
Fix dropdown accessibility, copy/paste issues

rshipp:
### Description

Dropdowns were being created without labels. While fixing that, I noticed clipboad functionality was broken in several places, so I fixed that too. The changes are hard to break apart, so it's in 1 PR.

Task: 
* https://app.asana.com/0/1174274412967132/1199733407122996/f
* https://app.asana.com/0/1174274412967132/1199733407122994/f

Note the new visible labels for dropdowns mean that the wording has changed. (See screenshots below.)

Also note there is **no automigration** for the new structure. This means all old surveys, and anywhere else that used dropdowns, **will break** when this PR is merged. They must be migrated by hand. Task: https://app.asana.com/0/1174274412967132/1199565750950565/f.

### Motivation and Context

Accessibility best practices say to prefer visible text as the label. This means `<label for=` is better than `<select aria-label=`. I removed the `aria-label` attribute from dropdowns, and replaced it with an `id` attribute, and an associated `label` element. For CKE code simplicity, that meant I needed a wrapper div to place both those elements inside (`selectWrapper`).

I then had to handle copy/paste for `selectWrapper` and its children, since we want to be able to copy/paste selectWrappers and have the pasted elements use new, unique IDs and names. 

I noticed radio copy/paste was broken, so I fixed that at the same time.

I updated `InsertNumericalSelectorCommand` and `InsertIndustrySelectorCommand` to use the new `selectWrapper` and appropriate children. While doing that, I remembered the `fieldset data-radio-group` issue, which was also breaking copy/paste for Industry Selectors, and fixed that too.

### Testing

(From the editor, with `?debug` CKE Inspector enabled.

1-10 Dropdown:

1. Add a new 1-10 dropdown
2. Note it has a label and a selection handle on the wrapper
2. Copy the dropdown
3. Paste the dropdown
4. Note the ID and name of the new dropdown has changed, and the label's `for` attribute points to the new dropdown ID.

Industry selector:

1. Add a new industry selector
2. Note it has new, editable labels 
3. Cut the entire industry selector block
4. Paste it
5. Note the ID and name of both dropdowns have *not changed*. Only selectWrappers directly inside the root of the document will be updated.
6. Note the industry selector fieldset does *not* have a `data-radio-group` attribute.

Radio:

1. Add a radio list
2. Optionally press Enter inside the radio label, to create more radio buttons
3. Copy the radio list
4. Paste
5. Note the fieldset's data-radio-group has changed
6. Note *all* radio inputs inside the fieldset have a `name` that matches the fieldset's data-radio-group
7. Note all radio inputs have a new, unique name
8. Note each radio input has the same value as it did before copy/paste

Before:

Screen Shot 2021-01-13 at 1 39 27 PM

After:

Screen Shot 2021-01-13 at 1 38 53 PM

### Types of changes

### Checklist:


app/assets/stylesheets/custom_contents.scss
```diff
@@ -55,6 +55,10 @@
     height: 43px; // Height of button plus 2px for the 1px border.
     border-width: 0;
   }
+
+  div.select-wrapper > label.select-label {
+    padding-right: 0.5rem;
```
rshipp, line 60: adds some space between the label and the dropdown

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -57,17 +57,40 @@ export default class ContentCommonEditing extends Plugin {
...
+         *     <selectOption value="value-1">$text</selectOption>
+         *     <selectOption value="value-2">$text</selectOption>
+         *   </select>
+         * </selectWrapper>
+         */
+        schema.register( 'selectWrapper',  {
             isObject: true,
-            allowAttributes: [ 'aria-label', 'name' ],
-            allowIn: [ '$root' ],
+            allowIn: [ '$root', '$fieldset' ],
```
rshipp, line 73: note `fieldset` is required because `selectWrapper` is used inside the industry selector fieldset.

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -82,25 +105,34 @@ export default class ContentCommonEditing extends Plugin {
...
             model: ( viewElement, { writer } ) => {
-                return writer.createElement( 'fieldset', {
-                    'data-radio-group': viewElement.getAttribute( 'data-radio-group' ),
-                } );
+                let attributes = {};
+                const radioGroup = viewElement.getAttribute( 'data-radio-group' );
+                if ( radioGroup ) {
+                    attributes['data-radio-group'] = radioGroup;
+                }
+                return writer.createElement( 'fieldset', attributes );
```
rshipp, line 113: only add `data-radio-group` if there is one. this was also breaking copy/paste for industry selector blocks.

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -217,37 +249,101 @@ export default class ContentCommonEditing extends Plugin {
...
+                name: 'select',
+            },
+            model: ( viewElement, { writer } ) => {
+                return writer.createElement( 'select', {
+                    'id': viewElement.getAttribute('id'),
+                    'name': viewElement.getAttribute('name'),
+                } );
+            }
+        } );
+        conversion.for( 'downcast' ).elementToElement( {
```
rshipp, line 339: since `select` is no longer a widget, we can simplify into a single downcast

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -217,37 +249,101 @@ export default class ContentCommonEditing extends Plugin {
...
+                name: 'label',
+                classes: [ 'select-label' ],
             },
             model: ( viewElement, { writer } ) => {
-                return writer.createElement( 'select', {
-                    'name': viewElement.getAttribute('name'),
-                    'aria-label': viewElement.getAttribute('aria-label') || '',
+                return writer.createElement( 'selectLabel', {
+                    // HACK: Get the id of the select this label corresponds to.
+                    'for': viewElement.parent.getChild(1).getAttribute('id')
```
rshipp, line 293: we use this for radio input/labels too. the `getChild(1)` is hacky, but I don't think handling other cases here is worth the effort.

app/javascript/ckeditor/inputuniqueattributeediting.js
```diff
@@ -25,11 +25,14 @@ export default class InputUniqueAttributeEditing extends Plugin {
             let newContentElements = [];
 
             for ( const element of data.content.getChildren() ) {
-                if ( [ 'textarea', 'input', 'select' ].includes( element.name ) ) {
+                if ( [ 'textarea', 'input' ].includes( element.name ) ) {
```
rshipp, line 28: remove `select` from here, since it can no longer exist at the $root level, only inside selectWrapper

app/javascript/ckeditor/inputuniqueattributeediting.js
```diff
@@ -38,6 +38,54 @@ export default class InputUniqueAttributeEditing extends Plugin {
                             element.getChildren(),
                         )
                     );
+                } else if ( element.name == 'div' && element.hasClass( 'select-wrapper' ) ) {
```
rshipp, line 41: new handling for `selectWrapper` and its children.

app/javascript/ckeditor/inputuniqueattributeediting.js
```diff
@@ -58,10 +106,11 @@ export default class InputUniqueAttributeEditing extends Plugin {
                                 // Radio button.
                                 const newRadioAttributes = new Map( child.getAttributes() );
                                 if ( newRadioAttributes.has( 'name' ) && newRadioAttributes.get( 'name' ).startsWith( ELEMENT_NAME_PREFIX ) ) {
-                                    newAttributes.set( 'name', this._getNewName() );
+                                    // Always set the radio name to the fieldset's data-radio-group value.
+                                    newRadioAttributes.set( 'name', newAttributes.get( 'data-radio-group' ) );
                                 }
                                 if ( newRadioAttributes.has( 'id' ) && newRadioAttributes.get( 'id' ).startsWith( ELEMENT_ID_PREFIX ) ) {
-                                    newAttributes.set( 'id', this._getNewId() );
+                                    newRadioAttributes.set( 'id', this._getNewId() );
```
rshipp, line 113: this was the radio copy/paste bug. this functionality is totally broken in prod.

app/javascript/ckeditor/insertnumericalselectorcommand.js
```diff
@@ -6,8 +6,10 @@ const NUM_OPTIONS = 10;
 export default class InsertNumericalSelectorCommand extends Command {
   execute( ) {
     this.editor.model.change( writer => {
-      const inputName = this.editor.plugins.get('UniqueId').getNewName();
-      const selector = createSelector( writer, inputName );
+      const uniqueId = new UniqueId();
+      const inputName = uniqueId.getNewName();
```
rshipp, line 10: since the UniqueId plugin is stateless, it's simpler to make a new instance rather than finding the one from `editor.plugins`. this is the pattern we use in other plugins.

app/javascript/ckeditor/insertnumericalselectorcommand.js
```diff
@@ -16,22 +18,30 @@ export default class InsertNumericalSelectorCommand extends Command {
...
-function createSelector( writer, inputName ) {
+function createSelector( writer, inputName, inputId ) {
+  const selectWrapper = writer.createElement( 'selectWrapper' );
+
+  const selectLabel = writer.createElement( 'selectLabel' );
+  writer.insertText( 'Choose a number:', selectLabel );
+
   const selector = writer.createElement( 'select', {
     'name': inputName,
-    'data-bz-retained': inputName,
```
rshipp, line 28: removed old data-bz-retained reference, since we stopped relying on that

sadleb (commented):
I'll defer to Jess on the code review here. Nothing stuck out to me worth discussing, but I don't understand the CKEditor stuff.

jessouyangbraven (approved):
I'm pretty sure I messed all of this up with industry selector. :( Thanks for fixing!

app/javascript/ckeditor/contentcommonediting.js
rshipp: yeah good call

app/javascript/ckeditor/contentcommonediting.js
rshipp: no because 1-10 dropdowns (which are also selectWrappers now) still live in the $root, only industry selector dropdowns are in a fieldset.

but i just read this again and realized it says `$fieldset` instead of `fieldset` and should never have worked, so i'll figure out what that's about.

rshipp: fieldset is `allowContentOf: '$root',`, so i'm resetting this to just `allowIn: [ '$root' ]`

PR #554, opened 2021-01-14
Add AttendanceEvents (1/?)

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199526687431917

### Motivation and Context

This is the first step of adding attendance to Canvas as assignments.

The `AttendanceEvent`, is like a `CustomContent`, in that it's Braven-specific content stored in our database.

In this case, the only content stored is `title`, the title that will eventually be used for the Canvas assignment. Currently, we think these titles will be in the format of: **CLASS: Learning Lab 2: Lead Authentically**.

We store this in our database (1) for analytics, so we can associate the same logical attendance event across multiple runs of a course, (2) for designers, so they can add a pre-existing attendance event to a course (again, this is because it's conceptually the same event/assignment across runs of the course).

This is the DB diagram to show how `AttendanceEvents` fits into the rest of the data model that will be implemented for assignments (https://dbdiagram.io/d/5ffdeae380d742080a3609eb):
Screen Shot 2021-01-13 at 5 12 35 PM

- `https://platformweb/attendance_events`:
Screen Shot 2021-01-13 at 5 16 27 PM

- **Add Attendance Event** and **Save**:
Screen Shot 2021-01-13 at 5 16 42 PM

- It appears on #index:
Screen Shot 2021-01-13 at 5 16 48 PM

- Delete:
Screen Shot 2021-01-13 at 5 17 02 PM

### Types of changes

### Checklist:


app/controllers/attendance_events_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+        notice: 'Attendance event was successfully created.'
+      ) }
+      format.json { head :no_content }
+    end
+  end
+
+  # Note: We do **not** support #edit and #update currently.
+  # If we do, we also need to support the corresponding #publish_latest
+  # action for the CourseAttendanceEventsController so there's a way to update
+  # the Canvas assignment title reflect the change.
```
rshipp, line 35: why is this? i thought publish_latest was there because we're actually linking versions, not the thing itself?

rshipp (approved):

app/controllers/attendance_events_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    authorize @attendance_event
+    @attendance_event.destroy!
+    respond_to do |format|
+      format.html { redirect_to(
+        attendance_events_path,
+        notice: 'Attendance event was successfully deleted.',
+      ) }
+      format.json { head :no_content }
+    end
+  end
```
sadleb, line 47: What happens if it's already published? Would things in Canvas start to fail? Maybe for now just make sure we can't delete it on a launched course and we'll cross that bridge when figuring out what to support re: editing live courses?

sadleb (approved):
One comment, but not merge blocking.

app/controllers/attendance_events_controller.rb
jessouyangbraven: This is because `AttendanceEvent` is weird. Because the only content here is `:title`, it is sort of its own version of the Canvas assignment.

So even if we've published the `AttendanceEvent` to a Canvas assignment, we can update on our end without breaking the existing assignments because Canvas still have `title` saved on their side.

This wasn't true of the other content because there was additional content, like the module package or content editor HTML, that would change on our side that Canvas didn't keep track of.

All of this to say, I think we could just put in `#edit` and `#update` here and then add `#publish_latest` to the `course#edit` page for attendance events and it'd all magically work without us having to add a `AttendanceEventVersion`.

Do you think I should add it now with these PRs?

I should add that by not having `AttendanceEventVersion`, we can't look at prior `:title`s, or support reverting. But we don't do that for the other assignment/content types now, so it might not be a big deal?

app/controllers/attendance_events_controller.rb
jessouyangbraven: >What happens if it's already published? Would things in Canvas start to fail?

Yeah, things would fail.

But I only show a `Delete` link on the `attendance_events#index` page, not the `course#edit` one that we use for course management. There's not a direct link to the `attendance_events#index` page anywhere in the UI, so I'm not that worried about someone getting there and deleting on accident.

We do the same thing for `CustomContent`, where `custom_contents#index` will let you delete, but there's not a link to that page in the UI.

I mainly have this here so there's some way to delete these things from a UI. I can also rip this out if you think it's not worth having/maintaining at this point.

app/controllers/attendance_events_controller.rb
rshipp: Ohh that makes sense, I forgot the fact that the title is also stored on the Canvas side. Thanks for the explanation. I don't think this needs any changes, I was just curious about the comment.

app/controllers/attendance_events_controller.rb
sadleb: I think it’s fine to leave it for now as long as designers don’t have a way to get to it and accidentally delete

PR #555, opened 2021-01-14
Add existing, create new attendence events in course#edit (2/?)

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199526687431917

### Motivation and Context

This PR builds on #554 and allows you to add an existing attendance event to a Canvas course as an assignment through the course management page. It also adds a button so you can go to the `attendance_event#new` page directly from `courses#edit` if you want to create a new attendance event and then add it.

This PR adds `course_attendance_events` shown in this DB diagram:
Screen Shot 2021-01-13 at 5 12 35 PM

Like the `course_{custom_content_version, rise360_module_version}` types, attendance events and courses have a many-to-many relationship. Unlike the other join-table types, we don't version `course_attendance_events`, since the only content is the title of the Canvas assignment.

- Go to a `course#edit` page and see `Edit Attendance Events` section at the bottom:
Screen Shot 2021-01-13 at 7 35 21 PM

- **New Attendance Event** takes you to the `attendance_events#new` page:
Screen Shot 2021-01-13 at 7 35 30 PM

- Adding a new event redirects you back to the `course#edit` page you were on (instead of the `attendence_events#index` default redirect)
Screen Shot 2021-01-13 at 7 35 36 PM

- **Add Attendance Event** lets you select an existing `AttendanceEvent` from a drop-down (it also filters out ones you've already added to the course)
Screen Shot 2021-01-13 at 7 35 47 PM

- **Add** updates the list on the `course#edit` page:
Screen Shot 2021-01-13 at 7 35 55 PM

- **View on Canvas** shows this in the Accelerator course:
Screen Shot 2021-01-13 at 7 36 03 PM

### Types of changes

### Checklist:


app/controllers/course_attendance_events_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  # For Publishable
+  def assignment_name
+    @course_attendance_event.attendance_event.title
+  end
+
+  def lti_launch_url
+    launch_attendance_event_url(
+      @course_attendance_event.attendance_event,
+      protocol: 'https',
+    )
```
jessouyangbraven, line 28: There are two separate commits in this PR.

The second commit is where I attempted a static endpoint conversion because I realized it was relatively straightforward to do in this case.

This is a course-agnostic URL (aka "static endpoint")  to the `AttendanceEvent`, so when we clone this, the URL still works.

jessouyangbraven:
Making this a draft again. I just realized that even with a static endpoint, I need to create a new `CourseAttendanceEvent` when we do `initialize_new_course`. Will push up another commit for that.

app/controllers/attendance_events_controller.rb
```diff
@@ -22,7 +27,7 @@ def create
 
     respond_to do |format|
       format.html { redirect_to(
-        attendance_events_path,
+        create_params[:redirect_to] || attendance_events_path,
```
jessouyangbraven, line 30: This was added so we can do a redirect back to `course#edit` if we go click the `New Attendance Event` button.

app/controllers/course_attendance_events_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  # Note: The versioning here is faked by AttendanceEvent
+  def versionable_instance
+    AttendanceEvent.find(params[:attendance_event_id])
+  end
+
+  def version_name
+    'attendance_event'
+  end
```
jessouyangbraven, line 38: This is to make `Publishable` work as-is. Everything else that uses that concern (e.g., modules, custom contents), has versioned content.

app/models/attendance_event.rb
```diff
@@ -3,4 +3,14 @@
...
+
+  # For use with Publishable in CourseAttendanceEventsController
+  include Versionable
+
+  # For Versionable
+  # We don't version attendance events because the only content is the title,
+  # which is stored in Canvas independent of our DB.
+  def create_version!(user)
+    self
+  end
```
jessouyangbraven, line 15: Ditto. This is also coercing `Publishable` to work. I started refactoring it to handle the "I have a model but not a version for that model" case, but this turned out to be cleaner. I'm open to feedback on this, though.

app/controllers/attendance_events_controller.rb
```diff
@@ -46,8 +51,30 @@ def destroy
...
   end
 
+  def launch
+    # Note: we authorize with the **Course**AttendanceEventPolicy here because
+    # we need the course to check for enrollment.
+    authorize @course_attendance_event
+
+    # Override the layout since we're in a Canvas LTI context
+    render layout: 'lti_canvas'
+  end
```
jessouyangbraven, line 61: This is the static endpoint used for the Canvas assignment's LTI launch URL. I called it `#launch` because that's what we tend to use for LTI things. I also wanted to distinguish it from `attendance_event#show`, which isn't implemented yet, but if we did, I think that'd be a more admin-y view like the rest of this controller.

app/views/attendance_events/launch.html.erb
```diff
@@ -0,0 +1,3 @@
+<% content_for :title, @attendance_event.title %>
+
+<p>Your LC will take attendance for this event.</p>
```
sadleb, line 3: I would spell out Leadership Coach

app/views/course_attendance_events/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<div>
+  <div class="row">
+    <div class="col-sm-4">
+      <h4>Select an attendance event to a Canvas course</h4>
```
sadleb, line 4: Grammar. “... to publish to...”

app/views/courses/edit.html.erb
```diff
@@ -152,6 +152,12 @@
   <h3>Edit Impact Surveys</h3>
   <%= link_to 'Add Impact Survey', new_course_course_survey_version_path(@course), class: 'btn btn-sm btn-secondary' %>
   <%= render :partial => 'course_custom_content_versions/list', locals: {course_custom_content_versions: @course.course_survey_versions} %>
+
+  <h3>Edit Attendance Events</h3>
+  <%= link_to 'Add Attendance Event', new_course_course_attendance_event_path(@course), class: 'btn btn-sm btn-secondary' %>
+  <%= link_to 'New Attendance Event', new_attendance_event_path(:redirect_to => edit_course_path(@course)), class: 'btn btn-sm btn-secondary' %>
```
sadleb, line 158: Is “Create New” more clear?

db/migrate/20210114012202_create_course_attendance_events.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCourseAttendanceEvents < ActiveRecord::Migration[6.0]
+  def change
+    create_table :course_attendance_events do |t|
+      t.references :attendance_event, null: false, foreign_key: true
+      t.references :course, null: false, foreign_key: true
+      t.bigint :canvas_assignment_id
```
sadleb, line 6: Do we save it first and then update with the assignment id? Is that why it can’t be non-null?

app/controllers/attendance_events_controller.rb
sadleb: Should viewing either an unsubmitted or submitted attendance be in the AttendanceEventSubmission controller and that’s the launch URL we use for both Fellows and LCs to see or submit attendance?

app/controllers/course_attendance_events_controller.rb
sadleb: See above, but I think AttendanceEvent should just be for publishing and other course mgmt related things. For end users, we should use a submission, right?

app/controllers/attendance_events_controller.rb
jessouyangbraven: Note from stand-up: move this to `AttendanceEventSubmissionsController`.

db/migrate/20210114012202_create_course_attendance_events.rb
jessouyangbraven: I didn't actually think about this when writing the migration. Looking at `db/schema`, it looks like we're enforcing non-null canvas assignment IDs on similar things. I'll change and see if I can get that to work.

app/views/attendance_event_submissions/launch.html.erb
```diff
@@ -0,0 +1,3 @@
+<% content_for :title, @attendance_event.title %>
+
+<p>Your Leadership Coach will take attendance for this event.</p>
```
sadleb, line 3: We talked about changing the language here so that it made sense both before and after attendance had been taken. Here is my stab at that copy, but feel free to tweak - I'm not thrilled with it: "When your Leadership Coach takes attendance the gradebook for this event will reflect that."

sadleb (approved):
lgtm. Just reminding you about the language change we discussed in standup.

PR #556, opened 2021-01-14
RFC: Fix UX from Account Creation Process

Rowantatai:
Enhance UX when people use the account creation link
- Support 18 character IDs in SF APIs so that all platform users are notified to login instead AND direct them to a path that will ultimately get them back to canvas
- Inform end user signing up with any validation failures (ex: password mismatch etc)

### Motivation and Context

Reduce the numbers of "Something went wrong" or "The change you wanted was rejected" experiences

Screen Shot 2021-01-13 at 10 51 27 AM

### Types of changes

### Checklist:

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -25,6 +25,9 @@ class Users::RegistrationsController < Devise::RegistrationsController
     AccountCreator.new(sign_up_params: sign_up_params).run
 
     redirect_to action: :show
+  rescue ActiveRecord::RecordInvalid => e
+    flash[:notice] = e.message
+    redirect_to new_user_registration_path(u: sign_up_params['salesforce_id'])
   end
 
   # GET /resource/edit
@@ -58,7 +61,7 @@ class Users::RegistrationsController < Devise::RegistrationsController
   end
 
   def find_user_by_salesforce_id
-    User.find_by(salesforce_id: salesforce_id)
+    User.where('salesforce_id LIKE ?', "#{salesforce_id}%").first
   end
 
   protected
--- a/app/views/devise/registrations/already_exists.html.erb
+++ b/app/views/devise/registrations/already_exists.html.erb
@@ -1,6 +1,6 @@
 <%= render "devise/shared/error_messages", resource: resource %>
 <h2>You've already signed up!</h1>
 <div>
-  <p>Please <%= link_to "log-in here.", new_user_session_path %></p>
+ <p>Please <%= link_to "log-in here.", canvas_url %></p>
   <p>If believe you don't have an account, please <a href="mailto:support@bebraven.org">contact us here</a></p>
 </div>
--- a/app/views/devise/registrations/already_exists.html.erb
+++ b/app/views/devise/registrations/already_exists.html.erb
@@ -1,6 +1,6 @@
 <%= render "devise/shared/error_messages", resource: resource %>
 <h2>You've already signed up!</h1>
 <div>
- <p>Please <%= link_to "log-in here.", canvas_url %></p>
+ <p>Please <%= link_to "log-in here.", cas_login_url %></p>
   <p>If believe you don't have an account, please <a href="mailto:support@bebraven.org">contact us here</a></p>
 </div>
--- a/app/views/devise/shared/_error_messages.html.erb
+++ b/app/views/devise/shared/_error_messages.html.erb
@@ -9,3 +9,11 @@
     </ul>
   </div>
 <% end %>
+
+<% if flash[:notice] %>
+  <div id="error_explanation">
+    <ul>
+      <li><%= flash[:notice] %></li>
+    </ul>
+  </div>
+<% end %>


app/controllers/users/registrations_controller.rb
```diff
@@ -58,7 +61,7 @@ def salesforce_id
   end
 
   def find_user_by_salesforce_id
-    User.find_by(salesforce_id: salesforce_id)
+    User.where('salesforce_id LIKE ?', "#{salesforce_id}%").first
```
Rowantatai, line 64: @hcli-work "The API tools will accept the 15 character IDs as input but it will always provide you the 18 character ID when you export data. **It is a Salesforce best practice to use 18 character IDs in APIs, formulas or as reference due to its case insensitive nature.**"

app/controllers/users/registrations_controller.rb
```diff
@@ -25,6 +25,9 @@ def create
     AccountCreator.new(sign_up_params: sign_up_params).run
 
     redirect_to action: :show
+  rescue ActiveRecord::RecordInvalid => e
+    flash[:notice] = e.message
```
Rowantatai, line 29: TODO: Find a nicer way to output this information

app/controllers/users/registrations_controller.rb
hcli-work: this makes me uneasy because the solution is wed to this 15/18 character trait of Salesforce but the query is much more generic than that.  how about in the definition of `salesforce_id` you stripped it of leading/trailing whitespace and ensured it was either 15 or 18 characters. if it's not, then raise an error. this feels important because it's a user provided value.

next, can we implement a User validation check that the salesforce_id we save in the database is always 18 characters after stripping of whitespace? moreover, how about also implementing a uniqueness constraint on that column?

app/controllers/users/registrations_controller.rb
hcli-work: putting a todo in the comments isn't going to remind anyone to remember to do it. maybe make an Asana task in our backlog instead?

hcli-work (changes requested):

app/controllers/users/registrations_controller.rb
sadleb: I need more info on how to reproduce the issue you're trying to solve before I can weigh in here. The Salesforce ID here comes from API so it should always be the same. How are we getting different formats?

Once I understand that, I would recommend that you look for the specific error we know about and flash a human readable message about the issue. E.g. "You already have an account, please login." if we know that the reason is b/c they are trying to sign-up a second time.

app/controllers/users/registrations_controller.rb
```diff
@@ -25,6 +25,9 @@ def create
...
+    redirect_to new_user_registration_path(u: sign_up_params['salesforce_id'])
```
sadleb, line 30: This doesn't make sense to me. The `create` action is called when the `new_user_registration_path` form is submitted. The only thing they type in that form is their password. Is this to handle the situation of a password mismatch? Or to handle some Salesforce ID issue? If we're just trying to handle password mismatches, we should explicitly look for that and only redirect in that situation. Otherwise, we're going to mask any number of issues that could go wrong and never get the exceptions sent to Honeycomb/Sentry/logs to be able to debug.

app/controllers/users/registrations_controller.rb
Rowantatai: @sadleb It seems like this is happening when we're manually concatenating it (at the minimum). Here's a sample:
15 chars (i.e., account creation link on salesforce)
- Link: https://platform.bebraven.org/users/sign_up?u=0031J00001dGfqK
- Current Behaviour: "YOU'VE ALREADY SIGNED UP!"

18 chars (i.e., manual concatenation)
- Link: https://platform.bebraven.org/users/sign_up?u=0031J00001dGfqKQAS
- Current Behaviour: "CREATE ACCOUNT" then "The change you wanted was rejected" Error (bc email is already in use)

app/controllers/users/registrations_controller.rb
Rowantatai: Handle other situations. lol the thesis of this PR is to enhance the UX account creation process...but now I see why it may be confusing. Thanks for flagging the confusion. 

This is intended to handle the situation of a password mismatch, password already created, Password is too short, and email already signed up (this should direct to "you've already signed up" though). All of the `ActiveRecord::RecordInvalid` I've encountered were around the ones mentioned above. What are some other issues you suspect will get masked from this?

app/controllers/users/registrations_controller.rb
rshipp: Oh interesting, this is something I ran into when I was doing salesforce stuff too, but I didn't have a clue what was going on. @sadleb per @Rowantatai's comment above about the Salesforce best practices, can we change the autogenerated sign_up link to use the full, 18 character IDs? That will simplify validation/querying a lot on our end.

sadleb: If there is some Salesforce API call we can make to take a 15 char ID and get the 18 char ID back, I think we should put a database validation in place and only store the
18 char IDs. It would require us to fix the places where a 15 char ID is currently coming through though. If we can change the auto-generated links, that would be the first place. The second that I can think of is the admin dashboard that allows you to manually change an ID.

app/controllers/users/registrations_controller.rb
sadleb: Currently, I bet those are the only situations that would cause this. But someone could refactor the code and mess things up and forget to set their first name or something in certain situations. That would lead to an error happening of when they were manually testing and happened to hit that situation but be masked in prod of manual testing didn’t catch it.

sadleb: Btw, we’ve already put in some fixes in a slightly different way for some of this stuff (not the SF ID character length issue tho) in Highlander. So we shouldn’t spend much time on this and instead figure out what is left of the things that cause all this noise which still need to be fixed in Highlander.

rshipp:
All existing (Highlander Platform) users are migrated to use the 18-char IDs, and 18-char IDs are enforced (in Highlander) as of #685. Leaving this PR open just in case until Highlander is merged to `staging`.

sadleb:
I think most of this is implemented elsewhere at this point. We should re-open narrower user stories / PRs as we run into pain points

PR #558, opened 2021-01-15
Only send one confirmation email.

sadleb:
When I refactored the services into RegisterUserAccount, I introduced
a bug where the confirmation email was being sent twice. Once when the
new User model was created and once by explicitly calling the method to
send. This just relies on the User model creation to have it go out.

Note: accidentally using the sign-up form twice is untested. I'll
queue up a future task to sanity test that.

Test Plan:
- bundle exec rspec spec/services/register_user_account_spec.rb


app/services/register_user_account.rb
```diff
@@ -13,25 +13,21 @@ def initialize(sign_up_params)
     # they gave us when they signed up / applied.
     @salesforce_participant = sf_client.find_participant(contact_id: sign_up_params[:salesforce_id])
     @salesforce_program = sf_client.find_program(id: @salesforce_participant.program_id)
-
-    @new_user = User.find_or_create_by(
-      salesforce_id: sign_up_params[:salesforce_id]
-    )
-    @new_user.update(sign_up_params.merge(salesforce_contact_params))
+    @create_user_params = sign_up_params.merge(salesforce_contact_params)
```
sadleb, line 16: Just moving this from `initialize` to `run` since you wouldn't expect anything to happen until you actually run the service. Unrelated to this fix.

app/services/register_user_account.rb
```diff
@@ -13,25 +13,21 @@ def initialize(sign_up_params)
...
+      @new_user = User.find_or_create_by(
+        salesforce_id: @create_user_params[:salesforce_id]
+      )
+      @new_user.update(@create_user_params)
       span.add_field('user.id.created', @new_user.id)
 
-      # Do this before syncing their enrollment just in case it fails.
-      @new_user.send_confirmation_instructions
-      span.add_field('confirmation.instructions.sent', true)
-
```
sadleb, line 34: This is the fix. Don't explicitly call send b/c it happens when `User.find_or_create_by` creates the record. We didn't used to use `find_or_create_by` which led to the ability to create duplicate accounts.

sadleb:
Approved over slack

PR #560, opened 2021-01-20
Make sure and grade the proper Rise360ModuleVersion

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1199733407123005

* When publishing multiple `Rise360ModuleVersion`s to the same course or across different courses, they can have a different number of mastery questions so it's important that we choose the proper version when grading.
* Also handle the situation with 0 mastery questions in a Module more gracefully. Only count the engagement portion towards the total points available.

### Motivation and Context

There are a few things about modules that we can't control that make grading difficult:
1. Articulate Rise360 determines the `activity_id` for the exported `.zip` package. It's not something we can change. It is also the same when you edit an existing module and export it again. This means that we can have multiple `Rise360ModuleVersion`s in our databases with the same `activity_id`.
2. Articulate Rise360 determines the xAPI statements and data sent. This includes the `activity_id`, but we can't introduce additional fields ourselves to add, for example, the `Rise360ModuleVersion` ID instead.

These things make it difficult to grade the mastery questions, in particular, for a Fellow going through a module. The number of quiz questions is the denominator for grading, can change with different versions, but can't be distinguished from the `activity_id` alone.

This isn't an issue for progress through the module, as that's always sent as a percentage, so we don't need to know anything about the module itself to compute the participation portion of the grade.

### Test
1. Publish a Module to a course. In Canvas, set the points to 100.
2. Go in the database and set the `quiz_questions` to 0 for your module version. E.g. `update rise360_module_versions set quiz_questions = 0 where id = <your_id>;`
2. Using a valid Fellow enrolled in the course as a student, do at least one section of the module so that a row is inserted into `rise360_module_interactions` with a `progress` value. 
4. Run `bundle exec rake grade:modules`
5. Notice that the grade is set to the `progress` value. This is b/c there are NO quiz questions and engagement counts towards 100% of the grade
2. Use `Publish Latest` to upload a new version of the same exact module. Make sure it actually has a non-zero value for `quiz_questions`.
6. Do more of the module so a new value of `progress` is recorded in `rise360_module_interactions`
7. Run `bundle exec rake grade:modules` again and check your grade. It should be half of the `progress` value since there are now quiz questions and you only did some of the engagement portion of the grade which is weighted to half the total points.
8. Publish the same Module to a different course (same activity_id). Update the quiz_questions in the DB to 0 for that module version on a different course.
9. Do more of the module so a new value of `progress` is recorded and run the rake task again. The grade in Canvas should be half the `progress` value b/c it used the correct module version with non-zero number of a questions and not the new one published to a different course with 0 questions.

### Types of changes

### Checklist:


lib/module_grade_calculator.rb
```diff
@@ -47,6 +51,8 @@ def self.grade_module_engagement(interactions)
   # Returns [0, 100] for number of questions user got right out of the
   # questions in the module.
   def self.grade_mastery_quiz(interactions, total_questions)
+    return 0 if total_questions.nil? || total_questions.zero?
```
sadleb, line 54: Actually, I think this should be 100 if there are no mastery questions. The entire grade should be composed of the engagement portion and not be impact by the mastery portion. Ideally, the algorithm would switch to be 100% engagement and 0% mastery, but that's outside the scope of this so I'll queue up a task. I don't think there are plans to have modules with no mastery questions, so it's an edge case but one that could bite us one day.

app/models/rise360_module_interaction.rb
```diff
@@ -6,14 +6,6 @@ class Rise360ModuleInteraction < ApplicationRecord
...
   validates :user, :activity_id, :verb, :canvas_course_id, :canvas_assignment_id, presence: true
 
-  scope :for_user_and_activity, -> (user_id, activity_id) do
-    where(
-      "activity_id LIKE ? AND user_id = ?",
-      "#{activity_id}%",
-      user_id,
-    )
-  end
-
```
sadleb, line 16: Removing this since this PR adds a new `canvas_assignment_id` condition and `Rise360ModuleInteraction.for_canvas_assignment_and_user_and_activity` didn't feel more readable than just directly writing the `where` clause in the place where the code uses it. This is only used in one spot, so it's not here for reusability, just readability.

spec/lib/module_grade_calculator_spec.rb
```diff
@@ -86,12 +90,45 @@
...
+          .once
+
+        expect(grade).to eq(100)
+      end
     end
 
     context "module engagement grade" do
       it "returns 0 for no interactions" do
         interactions = Rise360ModuleInteraction
-          .for_user_and_activity(user.id, activity_id)
```
sadleb, line 94: Removed all these lines here and below b/c it felt like this was testing the implementation a bit more than the behavior so it wasn't worth refactoring with the new `where` clause that filters down to the `canvas_assignment_id` as well?

jessouyangbraven:
@sadleb , this looks good to merge to me. Going to let @rshipp review and approve, since I can't as the author of the PR.

rshipp (approved):

PR #561, opened 2021-01-20
Update LTI launch URL for Fellows viewing attendance events

jessouyangbraven:
### Description

**Task**: https://app.asana.com/0/1174274412967132/1197571375872706

This PR changes the LTI launch URL for fellows viewing attendance assignments in the Accelerator course. It's prep work for the actual code that lets LCs take attendance for learning labs and 1-1s, etc.

### Motivation and Context

The LTI launch URL for attendance event assignments that Fellows see used to be `AttendanceEventSubmission#launch`.

This PR changes it to `AttendanceEventSubmissionAnswer#launch`. There are two reasons for this:
1. Our data model has an `Answer` per `for_user`, where the `for_user` is a fellow the LC is taking attendance for. This lets us be as restrictive as possible with our policy, because we can check whether `current_user == answer.for_user`.
2. This frees up `AttendanceEventSubmission#launch` to use for the **Take Attendance** link in the LC Playbook course.

The second part will be implemented in follow-up PRs.

### Test

- rspec
- Add an attendance event from `course#edit` page, verify it still renders in Canvas:
Screen Shot 2021-01-20 at 1 02 49 PM
- Check the assignment settings for updated LTI launch URL:
Screen Shot 2021-01-20 at 1 03 27 PM

### Types of changes

### Checklist:


app/controllers/attendance_event_submission_answers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  
+  # For the #launch action
+  include LtiHelper
+  before_action :set_lti_launch, only: [:launch]
+  before_action :set_course_attendance_event, only: [:launch]
+  before_action :set_attendance_event_submission_answer, only: [:luanch]
+
+  layout 'lti_canvas'
+
+  def launch
```
jessouyangbraven, line 14: This is the entry-point for the attendance assignments in the Accelerator course.

app/controllers/attendance_event_submission_answers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    if @attendance_event_submission_answer
+      authorize @attendance_event_submission_answer
+    else
+      # If the LC hasn't taken attendance yet, use the CourseAttendanceEvent
+      authorize @course_attendance_event, :launch?
+    end
```
jessouyangbraven, line 20: I wasn't sure how best to do this. This seemed okay since it was what I had in `AttendanceEventSubmissionsController`? I think we stop having as many issues with this once we add an `IsEnrolledPolicy` that looks at the current_user and a course.

app/controllers/attendance_event_submission_answers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    )
+  end
+
+  def set_attendance_event_submission_answer
+    attendance_event_submission = AttendanceEventSubmission.find_by(
+      course_attendance_event: @course_attendance_event,
+    )
+    @attendance_event_submission_answer = attendance_event_submission&.find_by(
+      for_user: current_user,
+    )
```
jessouyangbraven, line 39: This doesn't actually get created anywhere yet because we haven't hooked up/written most of `AttendanceEventSubmissionController`.

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -2,26 +2,63 @@
 
 class AttendanceEventSubmissionsController < ApplicationController
   include DryCrud::Controllers
-  
+
```
jessouyangbraven, line 5: I went back and forth about all the changes in this file. I'm cool if we want to rip all of this out and do it in a separate PR.

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -2,26 +2,63 @@
...
   end
 
 private
   # For #launch
+  def set_course
+    lc_playbook_course = Course.find_by(
+      canvas_course_id: @lti_launch.request_message.canvas_course_id,
+    )
+    # TODO: get the Accelerator course from Salesforce
+    @course = nil
```
jessouyangbraven, line 41: All the specs are commented out because nothing works for real until this logic goes in.

jessouyangbraven (commented):
I want to be able to merge the LTI launch URL changes independent of the LC grading stuff Ryan and I are going to work on, so let me know what makes sense to pull out of the change set.

app/controllers/attendance_event_submission_answers_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class AttendanceEventSubmissionAnswersController < ApplicationController
+  include DryCrud::Controllers
+  
+  # For the #launch action
+  include LtiHelper
+  before_action :set_lti_launch, only: [:launch]
+  before_action :set_course_attendance_event, only: [:launch]
+  before_action :set_attendance_event_submission_answer, only: [:luanch]
```
rshipp, line 10: typo: launch

app/controllers/attendance_event_submission_answers_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    @attendance_event = @course_attendance_event.attendance_event
+  end
+
+private
+  # For #launch
+  def set_course_attendance_event
+    @course_attendance_event = CourseAttendanceEvent.find_by(
+      canvas_assignment_id: @lti_launch.request_message.custom['assignment_id'],
```
rshipp, line 29: instead of pulling this directly out of the lti launch, what do you think about having a separate lti-specific controller that pulls out of the launch and redirects to a standard url-based rails route like `/course_attendance_events/1/etc`? i think this lti stuff is causing complexity in other parts of this controller, like the `&.find_by` below and the `if`/`authorize` tree in the `launch` action.

app/controllers/attendance_event_submission_answers_controller.rb
sadleb: I don't know either. This policy stuff is still confusing to me. It seems weird that you have to explicitly pass `launch?` in though.

app/controllers/attendance_event_submission_answers_controller.rb
sadleb: That's more inline with that "static endpoints" backlog task we have. Should we do it as part of that b/c it will blow up the scope of this? Or maybe give it a shot and if it's not too bad then go add this PR as an example to that task to follow when we go do it for all of the endpoints?

app/controllers/attendance_event_submission_answers_controller.rb
sadleb: I think you can `find_by` multiple attributes in one call? Something like
```
AttendanceEventSubmission.find_by(
      course_attendance_event: @course_attendance_event,
      for_user: current_user,
    )
```

app/controllers/attendance_event_submissions_controller.rb
sadleb: I lean towards having this PR be as focused as possible on the Answer stuff only and leave the Submission.launch (the one the LC uses to actually submit) to a separate PR where we implement it for real.

sadleb (approved):
lgtm

rshipp (approved):

PR #565, opened 2021-01-21
Redirect non admin users to the portal

Rowantatai:
### Description

While authenticated, our platform.bebraven.org serves admins only. When a non-admin authenticated user stumbles upon platform, they will be redirected to canvas. 

### Motivation and Context

Currently, they're being directed to "We're sorry, but something went wrong" error page. In the logs, it translates to `Pundit::NotAuthorizedError (not allowed to index? this Symbol):`

### Test Plan
_Case 1 - non-admin user_
1. Login to platform
2. Visit platform.bebraven.org
Outcome: redirects to canvas

_Case 2 - admin user_
1. Login to platform
2. Visit platform.bebraven.org
Outcome: stays on platform

### Checklist

- [?] All new and existing tests passed.


spec/controllers/custom_content_versions_controller_spec.rb
```diff
@@ -57,7 +57,7 @@
     context "non-admin attempting to view all versions" do
       let(:user) { create :fellow_user }
 
-      it 'raises a not-authorized error' do
+      xit 'raises a not-authorized error' do
```
Rowantatai, line 60: I tried to get redirection working instead of raising the error but haven't been successful yet

app/controllers/application_controller.rb
```diff
@@ -17,6 +17,8 @@ class ApplicationController < ActionController::Base
   # reason to call authorize in devise or CAS controller acitons.
   after_action :verify_authorized, unless: :devise_controller?
 
+  rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
```
sadleb, line 20: I'm a bit worried about this. It might be fine in non-Highlander since there is nothing other than the CAS SSO stuff on platform for non-admins. But in Highlander, this is likely going to get us in trouble when the authorization failed for some other reason and we end up in an infinite loop of sorts. Also, it will mask the actual errors from our logs and Honeycomb to be able to troubleshoot.

How about we just do this sort of thing in the `home_controller`? I think we've probably already mostly solved the issue by switching the `new_user_session_path` to `cas_login_url`, but having the redirect only on the home controller feels like a safer fallback for whatever root cause we may have missed. E.g. does password reset end up causing this and we need to switch a URL there as well?

When you implement the redirect in the `home_controller` I would explicitly check if the user has an admin role and redirect to Canvas if not instead of relying on an exception.

sadleb (changes requested):

app/controllers/application_controller.rb
hcli-work: +1

spec/controllers/custom_content_versions_controller_spec.rb
hcli-work: commenting out a test that doesn't work feels like not the right step.

hcli-work (changes requested):

app/controllers/application_controller.rb
sadleb: One more potential root cause I just remembered. I think log-out can cause this b/c it log's you out to the platform log-in page but loses the `service` URL that tells login to bring you to Canvas. The above solution with the `home_controller` should fix that scenario, but in a hacky way. The proper fix is probably to preserve the `service` URL. This wasn't as much of a problem before b/c the Join server would * mostly * be able to figure out where you should go and redirect you but now with it largely out of the picture everyone who logs out and back in is perhaps ending up at the "something went wrong screen"

spec/controllers/custom_content_versions_controller_spec.rb
Rowantatai: Agreed. The decision to update the test depended on alignment of the changes. Not necessary anymore.

app/controllers/application_controller.rb
Rowantatai: I moved the rescue from the application controller and instead ensured the user is an admin in the home controller, as requested. I created a backlog task to preserve the service url [here].(https://app.asana.com/0/1179330853755533/1199679486201904/f)

sadleb (approved):
I don’t know where `:canvas` is defined in the redirect, but assuming it is and it’s the correct URL, lgtm!

sadleb:
> I don’t know where `:canvas` is defined in the redirect, but assuming it is and it’s the correct URL, lgtm!

Wait, actually, do you need to do redirect `and return` ? Generally, I think the rest of the code executes after calling redirect so does that mean the Not Authorized exceptions are being thrown but it still redirects? I don’t know how it works in a before action, but assuming you tested this and it works, still lgtm

hcli-work (approved):
when this goes live, make sure to test for both Booster and Accel.

PR #566, opened 2021-01-21
Completed Waivers page was showing login form in Chrome incognito.

sadleb:
Task: https://app.asana.com/0/1174274412967132/1199627462312679

We need to pass the state param through when redirecting to pages shown
in the context of an LtiLaunch in case folks are using a browser that
doesn't support third party cookies since we can't access session in
that scenario and have to use the state to authenticate.

Related follow-up task: https://app.asana.com/0/1174274412967132/1199627462312687

### Test Plan:
- Comment out the "if" in the waiver_submissions_controller#launch
  so that it always redirects.
- Open the Waivers assignment in Chrome incognito with third party cookies
  disabled and see that it shows the login form.
- Now apply this fix and see that it shows the Completed page.


app/controllers/waiver_submissions_controller.rb
```diff
@@ -28,7 +28,7 @@ class WaiverSubmissionsController < FormAssemblyController
   def launch
     authorize :waiver_submission
 
-    redirect_to completed_waiver_submissions_path if waivers_already_signed?
+    redirect_to completed_waiver_submissions_path(state: params[:state]) if waivers_already_signed?
```
jessouyangbraven, line 31: Do we care that this leaks `state` in the URL?

rshipp: no, we have to do that everywhere. URLs are hidden inside https everywhere except [in our own logs](https://app.asana.com/0/1174274412967132/1199623626113021/f), where we have to strip them out

sadleb: Yes and no. We care, but we decided to "guard" ourselves is by using a really long state that can't be brute forced and expiring them after 2 weeks. It's exposed all over the place.

rshipp (approved):

PR #570, opened 2021-01-25
Add "Rate This Module"

rshipp:
### Description

Add "Rate This Module" page, iframeable within Rise360 content, which itself is iframed within Canvas.

* New LTI helper callback `set_lti_launch_from_referrer`
* New authentication support for LTI state in referrer header (lowest priority).
* Static URL `/rate_this_module_submissions/launch` will ALWAYS find/create/redirect to the proper Submission, so Content Designers can always use the same URL in Rise360 iframes.
* Only save the latest response for each input field. There are always 0-1 Submission per user/module_version and 0-2 Answers per user/module_version.
* Only admins can view Submissions from other users; no one can update submissions from other users. 
* No new CSP/security exceptions, so this page will *only* work when Rise360 content is hosted on the same domain as the page itself. I.E. everything must be on prod, or everything must be on your local dev environment.

Note: When adding an iframe in Rise360, use the following code:

    <iframe src="https://platform.braven.org/rate_this_module_submissions/launch" width="100%" height="750px">

Adjust `height` as desired, but note it is not respected on mobile, and the height of the content will change depending on the window width, so there is no perfect height. You *must* specify a width, or Rise360 will not treat `height` correctly. You *must* use double-quotes, not single-quotes.

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1199165501769796/f

Screen Shot of rate this module iframe

### Types of changes

### Checklist:


app/controllers/rate_this_module_submissions_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  layout 'lti_canvas'
+
+  def launch
+    # Note: we immediately discard this `new` submission after using it
+    # to check the policy, and use find_or_create_by below instead.
+    # It's just easier to always pass in a submission object to the policy.
+    authorize RateThisModuleSubmission.new(
+      user: current_user,
+      course_rise360_module_version: @course_rise360_module_version
+    )
```
sadleb, line 18: Why don't we pass the one gotten using `find_or_create_by` below in to authorize?

app/helpers/lti_helper.rb
```diff
@@ -22,6 +22,26 @@ def set_lti_launch
...
+
+    @lti_launch = LtiLaunch.current(state)
+  end
+
+  # We call this class method from the LTI middleware too.
+  def self.get_lti_state_from_referrer(request)
+    begin
+      referrer = Addressable::URI.parse(request.referrer)
+      referrer_query_params = referrer.query_values
+      referrer_query_params['auth'][/LtiState (.*)$/, 1]
```
sadleb, line 39: Is there a clean way to share the `LTI_AUTH_HEADER_PREFIX` from `lti_authentication` ? Maybe move it to our `constants` file or reference it directly?

app/helpers/lti_helper.rb
```diff
@@ -22,6 +22,26 @@ def set_lti_launch
...
+    rescue NoMethodError
```
sadleb, line 40: What causes this? If the `referrer_query_params['auth']` is nil? Should we just proactively check instead of relying on an exception?

app/policies/rate_this_module_submission_policy.rb
```diff
@@ -0,0 +1,39 @@
...
+    # No one can update submissions that aren't their own.
+    # Only admins can read submissions that aren't their own.
+    return false unless user == record.user
+
+    # Users can rate modules attached to a course they are enrolled in.
+    record.course.sections.each do |section|
+      RoleConstants::SECTION_ROLES.each do |role|
+        return true if user.has_role? role, section
+      end
+    end
```
sadleb, line 35: Outside the scope of this PR, but we do this same thing in like 10 places now and it should prob be a centralized method somewhere.

sadleb (approved):
I left some comments about things that were confusing or may be worth cleaning up a bit more, but nothing merge blocking. I am surprised that we don't have to do anything to ensure that the `referer` sends the full URL similar to what we do for FormAssembly forms using the `no-referrer-when-downgrade` meta element (or doing it in the headers).

app/controllers/rate_this_module_submissions_controller.rb
rshipp: Because it might `create`, and doing that before checking authorization wouldn't make sense.

app/policies/rate_this_module_submission_policy.rb
rshipp: That was the idea behind https://app.asana.com/0/1174274412967132/1199344732354185/f

app/helpers/lti_helper.rb
rshipp: at least `request.referrer` and `referrer_query_params[]`. Handling the exception is easier than checking every variation of what could be missing here. I'm not sure which way is more Ruby-idiomatic; since I'm coming from Python, where exceptions are preferred, I'm biased towards using them.

rshipp:
The reason we don't have to add any exceptions for the referrer is that this request is same-domain. All involved content is served from platform.

app/helpers/lti_helper.rb
rshipp: This helper is used by `./config/initializers/lti_authentication.rb`, so we can't do circular dependency. I'm open to reorganizing this if you think a different way is better.

sadleb:
> The reason we don't have to add any exceptions for the referrer is that this request is same-domain. All involved content is served from platform.

Oh yeah, I just remembered that it was a form assembly processing endpoint on their servers that was the problem. We host the FA forms on our domain as well

app/helpers/lti_helper.rb
sadleb: Can we just include the `LtiAuthentication` module in the controller (maybe it already is on the base app controller) and just add this stuff to `fetch_state` which we have access to if the module is included?

app/helpers/lti_helper.rb
sadleb: I don’t know what is more Ruby idiomatic but I get scared anytime I see exception handling that assumes one thing is the cause but doesn’t actually check for it bc it can end up masking other problems.

rshipp: The key here is that we don't actually care why it failed, because the "failure" states are more common than the "good" state, and the expected behavior is to continue happily. I'll add a comment to make that clear.

app/helpers/lti_helper.rb
rshipp: todo: put the prefix in a constant and refactor both places

PR #571, opened 2021-01-26
Select `CourseAttendanceEvent` automatically by date

jessouyangbraven:
### Description
Task: https://app.asana.com/0/1174274412967132/1199155164193684

### Motivation and Context
When taking attendance, we show a dropdown of AttendanceEvents
that the user can choose from. This PR makes it so that the dropdown
defaults to select the most likely event that they'll want to take attendance
for given the current date that they are opening up the attendance taking
view.

The algorithm basically boils down to "on the day of the event
(aka midnight) start selecting that attendance event instead of the previous one"

Note that if there are multiple attendance events on the same day (e.g. Mock Interviews at night),
select the one that is coming up soonest. This isn't ideal and we could probably do something
smarter but it doesn't seem worth the complexity and likelihood of bugs. 

Note:
- Canvas assignment's `due_at` time is a string in `iso8601` format, set to UTC, and not local time
- Using [this API](https://apidock.com/rails/v6.0.0/Time/beginning_of_day) to simplify date computations and comparisons

### Checklist:


app/services/select_attendance_event_by_date.rb
```diff
@@ -0,0 +1,96 @@
...
+    return true if same_day_as?(other_date)
+    return @date.after?(other_date)
+  end
+
+  def in_between?(date1, date2)
+    return (@date.after?(date1) && @date.before?(date2))
+  end
+
+  def same_day_as?(other_date)
+    return @date.beginning_of_day == other_date.beginning_of_day
```
rshipp, line 94: I think this check is wrong. This code is TZ-aware, which means "beginning of day" is different depending on the timezone. Is there a way to just compare the date instead?

```
irb(main):005:0> d1 = Time.now.in_time_zone(-6)
=> Fri, 05 Feb 2021 11:56:11 CST -06:00
irb(main):006:0> d1.beginning_of_day
=> Fri, 05 Feb 2021 00:00:00 CST -06:00
irb(main):009:0> d2 = Time.parse('2021-02-05T23:59:00-07:00')
=> 2021-02-05 23:59:00 -0700
irb(main):010:0> d1.beginning_of_day == d2.beginning_of_day
=> false
```

spec/services/select_attendance_event_by_date_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+  let(:course) { create :course }
+  let(:section) { create :section, course: course }
+  let(:ta_user) { create :ta_user, section: section }
+  let(:fellow_user) { create :fellow_user, section: section }
+  let(:date) { Time.now }
+  let(:assignment_overrides) { [] }
+  let(:canvas_client) { double(CanvasAPI) }
+
+  describe '#run' do
+
```
rshipp, line 15: i think we'd benefit from some timezone tests here too. According to the docs, the canvas api returns things with a local timezone, presumably the one that's tied to the course? The platform server could be in local(?) time, or UTC.

app/services/select_attendance_event_by_date.rb
rshipp: More info https://stackoverflow.com/questions/5419148/rails-get-beginning-of-day-in-time-zone

sadleb: Pushed a fix for this by just doing everything in UTC. We now use `Time.now.utc` to do comparisons to the UTC timestamps the Canvas API sends.

spec/services/select_attendance_event_by_date_spec.rb
sadleb: The examples in the Attendance API docs are wrong. Canvas returns the timestamps in UTC. See here: https://community.canvaslms.com/t5/Question-Forum/Canvas-API-and-time-discrepancies/td-p/164211

rshipp (approved):

PR #575, opened 2021-01-27
Fix errors seen by Canvas users (non-admin) when ending up on Platform.

sadleb:
We have two main entry points for the Braven apps, either Canvas or the Platform. The Platform is only meant for staff (aka admin role) and Canvas is meant for everyone else. This PR makes it so that users are sent to the proper `cas/login` endpoint since that's what we use and not Devise's default `users/sign_in` and it also makes it so that once they login, they are sent to the proper start page depending on if they are a Fellow/LC or staff.

Note: some of this is really hard to test in dev b/c it requires coordination b/n the prod Canvas instance and the SSO server (platform). I'm going to do some full end-to-end testing once we push this live to prod and either rollback or fix things that are still broken.

### Test Plan:
- Change your password, the Log In button should bring you to /cas/login endpoint
- Try to sign-up an account that is already signed up, the Log In link should be /cas/login
- Logout of Canvas (in both Chrome Incognito and normal), the Log In link should be /cas/login
- At the /cas/login page, logging in as a non-admin should bring you to Canvas regardless
  of the `service` parameter (which will be to platform)
- Logging in as an admin should bring you to the Platform admin dashboard

Note: in Chrome incognito, Canvas was sending a `url` parameter as the `continue_url`
which lead to a non-existant `logout` view. Logout should always render the `login`
page.

### Types of changes

### Checklist:


sadleb:
We did a code review synchronously in our standup meeting.

PR #577, opened 2021-01-29
Make Attendance handle Sections properly.

sadleb:
### Description

When an LC submits the attendance form, they will create a submission
that is unique to them, the course_attendance_event and the section
they have a TA role for.

Originally I thought we'd have to change the data model to account for
the section, but by restricting them to only be allowed in one section,
we just need to create or find the submission per LC per course attendance
event.

If they are in multiple sections, this just shows an error page letting
them know we don't support this and to reach out to us if they're seeing it.

### Motivation and Context

1. Before this, each LC that submitted attendance for an event would overwrite the other LC's attendance.
2. We need to use the `Section` that the LC is a TA for in order to implement [automatic attendance event selection](https://github.com/bebraven/platform/pull/571/files) -> this provides access to that b/c we needed to prove we could without altering the data model.

Note: the proper data model is likely just a mapping of Fellows to their Attendance status/answer and we use a React component to load and submit that. This data model is a bit funky in order to support the Rails routes to make the UI work.

image

### Types of changes

### Checklist:


app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -33,16 +33,23 @@ def launch
...
+
     # This is a find_or_new_by() so we have an object to authorize against
-    attendance_event_submission = AttendanceEventSubmission.find_by(
+    @attendance_event_submission = AttendanceEventSubmission.find_by(
+      user: current_user,
       course_attendance_event: @course_attendance_event,
     ) || AttendanceEventSubmission.new(
+      user: current_user,
       course_attendance_event: @course_attendance_event,
     )
```
sadleb, line 48: This is the meat of the change. Everything else is just to ensure that they are only a TA for one section so that we have confidence we don't need to change the data model in order to access the section so that we can magically choose the most likely event they are taking attendance for.

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -82,19 +89,22 @@ def update
...
-      lc_playbook_course.canvas_course_id,
-    )
+      # Figure out the Accelerator course using Salesforce
+      accelerator_canvas_course_id = SalesforceAPI.client.get_accelerator_course_id_from_lc_playbook_course_id(
+        lc_playbook_course.canvas_course_id,
+      )
 
-    @accelerator_course = Course.find_by(canvas_course_id: accelerator_canvas_course_id)
+      @accelerator_course = Course.find_by(canvas_course_id: accelerator_canvas_course_id)
+    end
```
sadleb, line 107: Not related to this PR. I just thought it was safer to protect against calling it twice since it makes API calls.

app/views/attendance_event_submissions/multiple_sections.html.erb
```diff
@@ -0,0 +1,20 @@
+<% content_for :title, 'Take Attendance' %>
+
+<div class="row h-50">
+  <div class="col-md align-self-center">
+
+    <h1>Take Attendance</h1>
+    <p>
+      <%= "It appears that you are an LC for multiple Cohorts. We don't currently handle that." %>
```
sadleb, line 8: TODO

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -30,19 +30,31 @@ class AttendanceEventSubmissionsController < ApplicationController
...
+      user: current_user,
       course_attendance_event: @course_attendance_event,
     ) || AttendanceEventSubmission.new(
+      user: current_user,
       course_attendance_event: @course_attendance_event,
     )
-    authorize attendance_event_submission
-    attendance_event_submission.update!(user: current_user)
+    authorize @attendance_event_submission
+    @attendance_event_submission.update!(user: current_user)
```
rshipp, line 55: safe to remove this update?

app/controllers/attendance_event_submissions_controller.rb
rshipp: nit: could rewrite as `return if @accelerator_course`, `...rest of function...` to remove the `begin` block and be more in line with other functions.

rshipp (approved):
non-blocking comments

app/controllers/attendance_event_submissions_controller.rb
sadleb: If it happens to be the `new` part above, this won't be in the database. I think it's there to handle that and I don't see an easy way to make it cleaner.

rshipp: Oh I missed that it was `find_by || new`, whoops!

sadleb: TODO: change `update` to `save!`

PR #579, opened 2021-02-04
Missed null constrain in db/schema.

sadleb:
This is from PR: https://github.com/bebraven/platform/pull/570


sadleb:
Approved over Slack.

PR #581, opened 2021-02-08
Add controller that creates a new page in platform that'll allow stud…

a-willis:
Project Task:
https://app.asana.com/0/1179330853755533/1199391832901943/f

Test Plan:
http://platformweb:3020/users/portal_password/5823/edit

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "liveServer.settings.port": 3021
+}
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -3,6 +3,7 @@
 class Users::PasswordsController < Devise::PasswordsController
   layout 'accounts'
 
+  skip_before_action :require_no_authentication, :only => [:edit, :update]
   # GET /resource/password/new
   # def new
   #   super
@@ -13,10 +14,10 @@ class Users::PasswordsController < Devise::PasswordsController
   #   super
   # end
 
-  # GET /resource/password/edit?reset_password_token=abcdef
-  # def edit
-  #  super
-  # end
+   # GET /resource/password/edit?reset_password_token=abcdef
+   # def edit
+   #  super 
+   # end
 
   # PUT /resource/password
   # def update
--- a/app/controllers/users/portal_password_controller.rb
+++ b/app/controllers/users/portal_password_controller.rb
@@ -1,0 +1,25 @@
+class Users::PortalPasswordController < UsersController
+  
+  before_action :find_user, only: %i[show edit update]
+
+  # GET	/users/portal_password/5823/edit 
+  def edit
+    authorize @user
+  end 
+
+  # POST	/users/portal_password/:id
+  def update 
+    authorize @user
+  end 
+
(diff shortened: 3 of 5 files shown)


hcli-work:
Lexi, I'm a little confused here. I thought you were creating a whole new controller instead of mucking around with Devise anymore. Did I misunderstand that?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
hcli-work, line 1: you'll need to add `.vscode` to your `.gitignore` file so that git ignores these config files for your editor

app/controllers/users/passwords_controller.rb
```diff
@@ -3,6 +3,7 @@
 class Users::PasswordsController < Devise::PasswordsController
   layout 'accounts'
 
+  skip_before_action :require_no_authentication, :only => [:edit, :update]
```
hcli-work, line 6: should this be here? it feels like if we are adding a new controller, we shouldn't have to touch this existing one.

app/controllers/users/passwords_controller.rb
```diff
@@ -13,10 +14,10 @@ class Users::PasswordsController < Devise::PasswordsController
   #   super
   # end
 
-  # GET /resource/password/edit?reset_password_token=abcdef
-  # def edit
-  #  super
-  # end
+   # GET /resource/password/edit?reset_password_token=abcdef
```
hcli-work, line 17: similarly, let's revert this so it doesn't conflate whitespace issues with the meat of this PR

app/controllers/users/portal_password_controller.rb
```diff
@@ -0,0 +1,25 @@
+class Users::PortalPasswordController < UsersController
```
hcli-work, line 1: can you explain why there's a need to inherit from `UsersController` instead of `ApplicationController`?

app/controllers/users/portal_password_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  
+  before_action :find_user, only: %i[show edit update]
+
+  # GET	/users/portal_password/5823/edit
```
hcli-work, line 5: perhaps we shouldn't allow passing in of ids at all. it feels like it would leave opportunity for someone to do something bad like alter another's password. can we just grab the id based on who is logged in?

typically, the `ApplicationController` will have a method on it called `current_user`. we can just use `current_user.id` which would close up the security vulnerability and make the `find_user` method and instance variable `@user` unnecessary.

hcli-work (changes requested):

sadleb:
For Highlander, this is related to this task: https://app.asana.com/0/1174274412967132/1199950461423843 -> if we implement a page that allows folks to change their email, we should also allow them to change their password from there (while logged in, which is different from password reset used when you can’t login in the first place).

sadleb:
Closing bc this got deprioritized in favor of folks using the password reset flow after logging out in order to reset their password. The more common user sorry is that someone can’t login and needs to reset their password. We should implement a way for them to change it while they’re logged in at some point, but that could be months or years from now.

PR #583, opened 2021-02-08
fixing refresh

hcli-work:
these small changes make the database refresh go a little smoother. fewer errors and consistent with canvas-lms repo


sadleb (approved):
If it works better, I don't see any issues or things to consider with the change.

PR #585, opened 2021-02-10
Add rails-erb-loader Webpacker package/loader

sadleb:
### Description
This is PR 1 of several aimed at injecting our own custom CSS into the Rise360 content. It adds the ability to write Javascript files using ERB so that we can access Rails helpers, such as asset_path('some_file'). We need to inject the CSS stylesheet using Javascript b/c CSS doesn't apply to the inside of iframes even if it's on the same domain, BUT you can use Javascript to access the inside of an iframe on the same domain.

This is the result of running: `bundle exec rails webpacker:install:erb` and then
changing the loader's config to use `bundle exec bin/runner` instead of plain
`bin/runner` so that it works in our environment.
    
Example **app/javascript/packs/some_file.js.erb**:
```
<% helpers = ActionController::Base.helpers %>
console.log('CSS path: <%=  helpers.asset_path('some_file.css') %>')
// This will print out something like /assets/some_file-1978412c120978716cf71d516733.css
// assuming that some_file.css is compiled as part of the Sprockets pipeline
```

### Types of changes

### Checklist:


rshipp (approved):
i don't understand the webpack/er stuff and it was an enormous pain to get running, but if this works, i don't see a problem with it. I think our pipeline works differently in dev than prod, so hopefully this works when deployed too.

sadleb:
> i don't understand the webpack/er stuff and it was an enormous pain to get running, but if this works, i don't see a problem with it. I think our pipeline works differently in dev than prod, so hopefully this works when deployed too.

I had it in the back of my head that I needed to think about prod differences. That’s a good push though. I should probably try it out on a prod build locally (or fix Review apps) before merging.

sadleb:
> > i don't understand the webpack/er stuff and it was an enormous pain to get running, but if this works, i don't see a problem with it. I think our pipeline works differently in dev than prod, so hopefully this works when deployed too.
> 
> I had it in the back of my head that I needed to think about prod differences. That’s a good push though. I should probably try it out on a prod build locally (or fix Review apps) before merging.

I tested with a local prod build and it worked. Fingers crossed that it's fine with the Heroku build.

PR #586, opened 2021-02-10
Fix Waivers and Pre-Accelerator Survey "Completed" CSS

sadleb:
### Description
Tasks:
1. [Waivers styling](https://app.asana.com/0/1174274412967132/1199922520026063)
2. [Pre-Accelerator styling](https://app.asana.com/0/1174274412967132/1199922442556434)

The "Completed" template was written to look nice in the `lti_canvas`
layout. We introduced a `form_assembly` layout and changed pre/post-accelerator
survey flow to be outside of Canvas. In doing both of these, we missed
a couple spots where we should use `lti_canvas` as the layout instead
of the `form_assembly` one. This makes the styling look how it was originally.

Note that it's a bit hacky that it uses the lti_canvas layout when not
shown inside Canvas for the `create` actions, but it didn't seem worth
trying to make cleaner.

**Waivers before this change**
image

**Waivers after this change**
image

**Pre-Accelerator before this change**

**Pre-Accelerator after this change** (Note: the "after" was hard to actually test so I did some hacky dev env stuff to gain confidence that the change works, but it will look like the Waivers one here:
image

### Types of changes

### Checklist:


app/controllers/accelerator_survey_submissions_controller.rb
```diff
@@ -81,6 +81,11 @@ def redirect_after_create?
     false
   end
 
+  # Override default Submittable behavior.
+  def create_layout_to_render
+    'lti_canvas'
+  end
```
sadleb, line 87: Just doing the same slight hack as above instead of investing time in this.

app/controllers/waiver_submissions_controller.rb
```diff
@@ -81,6 +81,7 @@ def create
   # GET /waiver_submissions/completed
   def completed
     authorize :waiver_submission
+    render layout: 'lti_canvas'
```
sadleb, line 84: The default layout is the form assembly one with none of our styling applied. I just missed this when introducing that layout to deal with conflicts bn our CSS and theirs

rshipp (approved):

PR #588, opened 2021-02-12
Inject CSS into the Rise360 content

sadleb:
### Description

[Task](https://app.asana.com/0/1174274412967132/1199552136290617)

Put Rise360 Modules in an iFrame and inject our own stylesheet into it. This builds on the [add rails-erb-loader](https://github.com/bebraven/platform/pull/585) PR to allow us to apply custom styles to Rise360 content. One (or more) PRs will follow this to actually write the custom CSS we want to apply.

We do it this way b/c CSS in a parent frame doesn't apply to a child iframe, but you can use Javascript from a parent frame to access the inside of the child iframe IF you are on the same domain.

Notice how "Resume Lesson" has green text. That is the result of uncommenting the example in `rise360_content.scss ` 

image

### Types of changes

### Checklist:


rshipp (approved):
very relieved we're not html parsing, thank you

app/controllers/rise360_module_versions_controller.rb
```diff
@@ -21,6 +21,6 @@ def show
     @rise360_module_version = Rise360ModuleVersion.find(params[:id])
     url = Addressable::URI.parse(@rise360_module_version.launch_url)
     url.query_values = helpers.launch_query
-    redirect_to url.to_s
+    @launch_path = "#{url.path}?#{url.query}"
```
sadleb, line 24: I construct the relative path here instead of using the full URL b/c this will only work if we're on the same domain so it seems more likely to prevent a future dev from just pointing at an endpoint on another domain.

PR #594, opened 2021-02-12
Copy AttendanceEvents as part of "Launch Program"

rshipp:
### Description

Add support for attendance events to Launch Program flow.

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1199922442556423/f

Attendance events successfully copied from my "Playground" course in local dev:

Screen Shot 2021-02-12 at 3 43 36 PM

### Types of changes

### Checklist:


app/services/initialize_new_course.rb
```diff
@@ -85,7 +91,7 @@ def initialize_new_course_custom_content_version(canvas_assignment_id, old_cours
   end
 
   def create_new_course_custom_content_version!(canvas_assignment_id, old_course_custom_content_version)
-    if old_course_custom_content_version.course_id == @new_course.canvas_course_id
+    if old_course_custom_content_version.course.canvas_course_id == @new_course.canvas_course_id
```
rshipp, line 94: unrelated bug fix, comparing the wrong IDs. I fixed it in `create_new_course_attendance_event!`, so thought I should fix it here too.

app/services/fetch_canvas_assignments_info.rb
```diff
@@ -78,6 +84,13 @@ def parse_assignment_info!(lti_launch_url, canvas_assignment)
     cccv = CourseCustomContentVersion.find_by_lti_launch_url(lti_launch_url) 
     add_project_or_survey_info!(cccv, canvas_assignment) and return if cccv
 
+    # Doesn't matter which Course the CourseAttendanceEvent is attached to, because we'll be
+    # replacing the course. Just get one with the right AttendanceEvent.
+    attendance_event = AttendanceEvent.find_by(title: canvas_assignment['name'])
+    course_attendance_event = CourseAttendanceEvent.find_by(attendance_event_id: attendance_event&.id)
```
rshipp, line 90: this bit took me a while to figure out, but this seems to be the behavior the cccv stuff above does with find_by_lti_launch_url. lmk if anyone finds a problem here though- the fact that this "fetch canvas stuff for this course" service actually fetches stuff from arbitrary other courses seems strange to me.

app/services/fetch_canvas_assignments_info.rb
```diff
@@ -16,7 +16,8 @@ class FetchCanvasAssignmentsInfo
               :canvas_peer_reviews_url, :canvas_peer_reviews_assignment_id,
               :canvas_fellow_evaluation_url, :canvas_fellow_evaluation_assignment_id,
               :course_project_versions, :course_survey_versions,
-              :course_custom_content_versions_mapping # Maps the fetched canvas assignment ID to the cccv.
+              :course_custom_content_versions_mapping,  # Maps the fetched canvas assignment ID to the cccv.
+              :course_attendance_events, :course_attendance_events_mapping
```
rshipp, line 20: note: at first glance `course_custom_content_versions_mapping` vs `course_attendance_events, course_attendance_events_mapping` seems strange, but it's because what would be `course_custom_content_versions` is broken up into `course_project_versions` and `course_survey_versions`.

app/services/fetch_canvas_assignments_info.rb
jessouyangbraven: I think this works as long as we don't allow updating `AttendanceEvent` titles. (Or we'll have to support versioning like CCCV).

I was trying to think of another way to do this, but all we have are the URL and title to go off of.

And the launch URL is a static URL without any IDs in it, so that doesn't help us figure out which attendance event.

jessouyangbraven, line 90: Nit: we could do this in `add_attendance_event_info` after we've checked the URL and then we wouldn't have to pass the `course_attendance_event`.

app/services/fetch_canvas_assignments_info.rb
```diff
@@ -109,6 +123,12 @@ def add_project_or_survey_info!(course_custom_content_version, canvas_assignment
     @course_custom_content_versions_mapping[canvas_assignment['id']] = course_custom_content_version
   end
 
+  def add_attendance_event_info(course_attendance_event, canvas_assignment)
+    @course_attendance_events << course_attendance_event
```
jessouyangbraven, line 127: I don't think this is being read anywhere? Do we need this array at all?

app/services/initialize_new_course.rb
```diff
@@ -98,6 +104,22 @@ def create_new_course_custom_content_version!(canvas_assignment_id, old_course_c
...
   end
 
+  def initialize_new_course_attendance_event(canvas_assignment_id, old_course_attendance_event)
+    create_new_course_attendance_event!(canvas_assignment_id, old_course_attendance_event)
+  end
+
+  def create_new_course_attendance_event!(canvas_assignment_id, old_course_attendance_event)
+    if old_course_attendance_event.course.canvas_course_id == @new_course.canvas_course_id
+      raise InitializeNewCourseError, "Canvas Assignment[#{canvas_assignment_id}] is already associated with #{@new_course}"
+    end
```
jessouyangbraven, line 114: I know we do this for CCCV, but I don't really understand this check.

I guess we're making sure we don't have a duplicate join table entry for the same attendance_event in the course, but I can't figure out how that could happen.

If we didn't need the check, we could just pass `AttendanceEvent` and not `old_course_attendance_event`, which would make more sense to me.

jessouyangbraven (approved):
I had to stare at this for a while and I'm pretty sure it works. That being said, the way these two services interact is really confusing, but that's not for this PR to solve.

app/services/fetch_canvas_assignments_info.rb
sadleb: Should we start off by trying to determine if we're processing an attendance event by running the `lti_launch_url =~ /#{attendance_event_submission_answer_path}/` stuff below and then if we know that it's an attendance event trying to look it by name and then look up the `course_attendance_event` IFF we find it by name?

sadleb: But otherwise, yeah, I can't think of another way to do this with the current way we have things setup and I think this works.

app/services/initialize_new_course.rb
sadleb: I think this was just a failsafe in case we messed up something in the code, it would be easier to have it fail with a clear message so that we could fix instead of ending up with duplicates and not realizing until after it was painful to fix b/c the course had already started and folks started working on things.

sadleb (approved):
Non-merge blocking comments, but I think it could avoid some stuff if you change the order of how you're looking up things.

app/services/fetch_canvas_assignments_info.rb
rshipp: Doesn't look like it. I took it out. We also don't appear to be using course_project_versions / course_survey_versions, but I don't want to mess with them and risk breaking something.

app/services/fetch_canvas_assignments_info.rb
rshipp: moved the check above, but left these queries here since it matches what we do for custom content above

rshipp: moved the check above

PR #602, opened 2021-02-18
Better error handling for user creation

rshipp:
### Description

Let Devise handle model validation errors (like "password too short"), instead of raising exceptions.

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1199951305978275/f

The error we were experiencing came from here: https://github.com/bebraven/platform/blob/0db524a1618f202da479d3aa8d0a4899a6853b97/app/services/register_user_account.rb#L81

In general, Devise expects a flow like this:

```ruby
def create
  user = User.new(sign_up_params)
  if user.save
    # It worked, do success stuff
    redirect_to user_path(user)
  else
    # Model validation errors, re-render
    # The errors should be in resource.errors
    self.resource = user
    render :new
  end
end
```

But in our case, `sign_up_params` only contains `salesforce_id`, because that's all we have access to at this point. Inside the `RegisterUserAccount` service, we use the salesforce ID to look up user information like email and name, and form `@create_user_params`: https://github.com/bebraven/platform/blob/0db524a1618f202da479d3aa8d0a4899a6853b97/app/services/register_user_account.rb#L16

which we then use to create the user record: https://github.com/bebraven/platform/blob/0db524a1618f202da479d3aa8d0a4899a6853b97/app/services/register_user_account.rb#L28

Note we use `update`, not `update!`, so it won't raise an exception if the model validations fail here. Later on though, we do have to use `update!`, because we look up this record in the `SyncPortalEnrollmentForAccount` service: https://github.com/bebraven/platform/blob/0db524a1618f202da479d3aa8d0a4899a6853b97/app/services/register_user_account.rb#L58-L65

with a `find_by!`, which will also raise an exception if the record hasn't been persisted to the db:

https://github.com/bebraven/platform/blob/0db524a1618f202da479d3aa8d0a4899a6853b97/app/services/sync_portal_enrollment_for_account.rb#L12

We can't use a `find_by` instead here, because that account needs to exist for any of the rest of `SyncPortalEnrollmentForAccount` to make sense, and that service is an integral part of the signup process - we _want_ to see an exception at signup time if it fails.

Since we can't remove the root cause - the possibility of exceptions in the signup flow - I tried a different workaround, adding a `yield` directly below the first `save` (or in this case, `update`) attempt. It will only yield if a block is passed in, so we can continue using this service as usual if we don't need the Devise error-handling. In the signup's `#create` action though, we can detect model validation errors with `persisted?`, short-circuit the rest of the service with an early return, and render the `:new` view with the `resource` set to the still unsaved `@new_user`.

Screen Shot 2021-02-18 at 3 21 56 PM

### Types of changes

### Checklist:


sadleb:
I’ve gotta think about this some more, but it’s cool to see the use of blocks and yield. I’ve used it a handful of times but always forget and this is a good reminder of the power.

app/controllers/users/registrations_controller.rb
```diff
@@ -22,10 +22,14 @@ def new
...
   # POST /resource
   def create
-
     # Registers the new user in all of our systems and creates the local User record
     # to send the confirmation email on success
-    RegisterUserAccount.new(sign_up_params).run
+    RegisterUserAccount.new(sign_up_params).run do |user|
+      unless user.persisted?
+        self.resource = user
+        return render :new
```
sadleb, line 30: The thing that confuses me here is how the new action successfully renders without passing the u=<sf_id>. Any insight into that? Otherwise, this makes sense to me.

sadleb: Is it because the create never actually completed and rendered so we’re still on new just with the user in a state that devise knows to add the error message when we re-render. That’s probably it?

sadleb:
I think it’s worth trying to style the “please try again” message to be a smaller heading.

app/controllers/users/registrations_controller.rb
rshipp: It doesn't need `u`, since it has `resource.salesforce_id` which it renders as a hidden input. 
Screen Shot 2021-02-19 at 9 39 59 AM

rshipp:
Does it look better with `<strong>`? 

Screen Shot 2021-02-19 at 9 44 46 AM

Our heading styles are not great. Since it's under an `h1`, this should be an `h2`, if anything, but our `h2`s look like this

Screen Shot 2021-02-19 at 9 46 25 AM

Even then, that puts the password inputs contextually beneath the heading "please try again", which doesn't make a lot of sense. Headings should describe the document and allow people to navigate it, not be used just to change font styles.

sadleb:
> Does it look better with `<strong>`?
> 
> Screen Shot 2021-02-19 at 9 44 46 AM
> 
> Our heading styles are not great. Since it's under an `h1`, this should be an `h2`, if anything, but our `h2`s look like this
> 
> Screen Shot 2021-02-19 at 9 46 25 AM
> 
> Even then, that puts the password inputs contextually beneath the heading "please try again", which doesn't make a lot of sense. Headings should describe the document and allow people to navigate it, not be used just to change font styles. Also, since this is inside a partial that's reused all over our app, it's risky to use any heading at all unless we know this will always be rendered under an `h1` and nothing else.

Paragraph text with string seems good. I think there are other bootstrap styles you can apply if you want to tinker, but strong is good enough imo

sadleb:
> > Does it look better with `<strong>`?
> > Screen Shot 2021-02-19 at 9 44 46 AM
> > Our heading styles are not great. Since it's under an `h1`, this should be an `h2`, if anything, but our `h2`s look like this
> > Screen Shot 2021-02-19 at 9 46 25 AM
> > Even then, that puts the password inputs contextually beneath the heading "please try again", which doesn't make a lot of sense. Headings should describe the document and allow people to navigate it, not be used just to change font styles. Also, since this is inside a partial that's reused all over our app, it's risky to use any heading at all unless we know this will always be rendered under an `h1` and nothing else.
> 
> Paragraph text with strong seems good. I think there are other bootstrap styles you can apply if you want to tinker, but strong is good enough imo

You could try additionally the `lead` class along with strong

rshipp:
The `lead` class looks like it increases the font size and decreases the font weight, so using it with `<strong>` wouldn't make sense. It's not easy for me to get back to this page, so I'd like to keep iterations to a minimum. Do you want me to use `lead` instead of `strong`?

sadleb:
> The `lead` class looks like it increases the font size and decreases the font weight, so using it with `<strong>` wouldn't make sense. It's not easy for me to get back to this page, so I'd like to keep iterations to a minimum. Do you want me to use `lead` instead of `strong`?

strong

jessouyangbraven (approved):
This looks reasonable to me, but I'll defer to @sadleb since he's familiar with devise. Also, thanks for the detailed description! Otherwise, I wouldn't have been able to make heads or tails of this.

sadleb:
Yeah, the detailed write-up is SUPER helpful!

sadleb (approved):

PR #603, opened 2021-02-19
Make "Rate This Module" look good on mobile.

sadleb:
### Description
This does a pretty ok job on various screen sizes making the Rate This Module content look the same as the Rise360 content.  Notably, the font's now match, there is no border, and the iframe doesn't scroll.

See the novel sized comment about how I punted on trying to get the content to line up properly on various in-between widths.

### Motivation and Context

Rise360 respects a height set on an embedded iframe for Desktop's but not on mobile. This is a pretty hacky solution to manually set the iframe height to various sizes that we know will fit the content and not scroll, but it looks much better than it did before on various screen sizes.

Mobile 1:

Mobile 2:

Desktop:
image

Example of how it doesn't line up on certain screen-sizes that overlap b/c no-sidebar vs sidebar views:
image

### Types of changes

### Checklist:


app/views/rate_this_module_submissions/_form.html.erb
```diff
@@ -1,39 +1,52 @@
...
 
-  <input type="hidden" name="state" value="<%= @lti_launch.state %>">
+      <div class="form-group">
+        <label for="module-other-feedback" class="mt-4">
+          Do you have any other feedback on this module to share?
+        </label>
+        <textarea id="module-other-feedback"
+                  name="rate_this_module_submission[module_feedback]"
+                  rows="3"
+                  aria-label="Other feedback"
```
rshipp, line 30: can't have both a label and an aria-label. label needs to be a `<p>`, or if we really want that text to be the label, we should remove the aria-label.

rshipp:
for the last screenshot in the pr body, is there still a scrollbar on the iframe?

rshipp (approved):
idk, this seems good enough. fingers crossed that we just never have to touch it again.

sadleb:
> for the last screenshot in the pr body, is there still a scrollbar on the iframe?

No, for all screensize's there is never a scrollbar. Here's how the bottom of the iframe looks in the one of the funky sizes. The only scrollbar is for the whole Module, but there is still quite a bit of space between the frame and the next content, which is how it was before too. This is b/c of the buffer I have there to make sure things will be unlikely to have a scrollbar even if your base font is a good deal bigger.
image

sadleb:
> idk, this seems good enough. fingers crossed that we just never have to touch it again.

seriously

rshipp:
ah ok, just making sure the bottom of the content was still visible somehow, since the scrollbars are hidden in the screenshot

PR #604, opened 2021-02-19
WIP: Check if a user exists with a 15 or 18 SF character ID 

Rowantatai:
### Description

If the Salesforce Contact ID parameter in the signup link is either 18-characters or 15-characters AND the user already created an account, redirect user to the “YOU'VE ALREADY SIGNED UP!” page

### Motivation and Context

We have a “YOU'VE ALREADY SIGNED UP!” page but it only works when the id matches that in platform. While majority of the IDs are 18-characters in platform (which should be what we get from the API), there’s currently 744 platform users saved with 15-character SFID, which majority, if not all, appear to be coming from the summer/fall 2020. Still investigating why.

When a  signup link with the opposing id (i.e., we have 18-chars saved on platform but the person is using a link with 15) is used by a program participant, it raises “Email looks like you've already signed up. Please log-in instead.”…and there’s no link to login instead, making it confusing for an end user.

### Test Plan / Screenshots (if appropriate):

TBD

### Types of changes

### Checklist:

### DIFF
--- a/lib/salesforce_id_formatter.rb
+++ b/lib/salesforce_id_formatter.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+# Salesforce id converter. Credits: https://github.com/heroku/salesforce_id_formatter
+
+module SalesforceIdFormatter
+  extend self
+
+  InvalidId = Class.new(StandardError)
+
+  def to_18(id)
+    raise InvalidId unless valid_id?(id)
+
+    id.size == 15 ? convert_15_to_18(id) : id
+  end
+
+  def to_15(id)
+    raise InvalidId unless valid_id?(id)
+
+    id.size == 18 ? convert_18_to_15(id) : id
+  end
+
+  def valid_id?(id)
+    [15, 18].include?(id.size) && id.match(/[^a-zA-Z0-9]/).nil?
+  end
+
+  private
+
+  def sfdc_base32_to_decimal(char)
+    sfdc_base32[char]
+  end
+
+  def decimal_to_sfdc_base32(dec)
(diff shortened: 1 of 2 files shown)


lib/salesforce_id_formatter.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+# Salesforce id converter. Credits: https://github.com/heroku/salesforce_id_formatter
```
Rowantatai, line 3: I [read](https://www.toptal.com/ruby-on-rails/top-10-mistakes-that-rails-programmers-make#:~:text=Common%20Mistake%20%235%3A%20Using%20too%20many%20gems&text=Excessive%20use%20of%20gems%20makes,configurations%20and%20increased%20operating%20costs.) it's bad practice to use too many gems, so copied what we needed

sadleb:
My initial hunch on where 15 char IDs came from is folks who were altered using the admin page bc copying the ID from the URL is the 15 char one, BUT 744 is a lot to have been fixed manually. I suppose it’s possible, but damn yo if so. Gotta be something else, gotta (maybe).

rshipp:
All existing (Highlander Platform) users are migrated to use the 18-char IDs, and 18-char IDs are enforced (in Highlander) as of #685. Leaving this PR open just in case until Highlander is merged to `staging`.

sadleb:
We’re very unlikely to merge / implement this at this point considering the current code doesn’t let you get into this state in the first place. Closing for now. Hopefully we’ll remember this exists if we ever need to be able to migrate 15 char IDs to 18 char IDs

PR #605, opened 2021-02-23
move rowan_bot into platform repository

hcli-work:
https://app.asana.com/0/1109150244034467/1199129462241589

This is mostly just a straight copy from the rowan_bot source into here. The point is to have a single repository to troubleshoot issues rather than having to juggle and keep in sync two repos even while developing.

next steps will be to clean this up and implement tests. we'll likely need to rip out Abdul's Salesforce API code in favor of the cleaner Highlander approach.

**Test Plan:** ran this locally with Booster prod env variables. I attempted doing the Booster Slack sync and also the `rake sync:docusign_to_salesforce[1]` task. No errors on either.


Gemfile
```diff
@@ -126,3 +120,10 @@ gem 'rails_same_site_cookie'
 # authorization
 gem "pundit"
 gem "rolify"
+
+# needed for rowan_bot code
+gem 'docusign_esign'
+gem 'faraday'
+gem 'restforce', '~> 4.2.2'
+gem 'selenium-webdriver'
+gem 'slack-ruby-client'
```
rshipp, line 129: can we make these all `require: false`? i can never remember what that means, just that we're supposed to do it bc it makes load faster somehow

hcli-work: I'm a little hesitant as I don't fully understand what that does and want to minimize introducing bugs since this code is super annoying to have to develop with right now due to the wacky gem ref stuff going on. I'm up for trying that out when it's in a single code base though.

sadleb: Require false means that it won’t automatically be loaded when the server restarts. For things that aren’t needed everywhere, it’s better to set require false and explicitly require the gem/class from where it’s needed to improve restart times and memory pressure (which can quickly spiral out of control). To do this, you would need to actually execute (or most) code paths. We should definitely lean toward all new gems being require false unless they are used widely. For cutting over, this feels like a good thing to switch as you introduce basic specs for all the features / functionality.

sadleb: Is selenium-webdriver needed for production code? I assume so and if yes, maybe remove it from the test group so it’s not duplicated?

sadleb:
I’m going to hold back any actual comments / suggestions until you’ve taken the next steps and we try and merge some of this to Highlander. I will note that in addition to the SalesforceAPI stuff being a next step, I think you should add consolidating the selenium-webdriver related config with the same stuff we do in the specs so that future bug fixes / iterations impact both.

For the old apps this is a step in the right direction.

sadleb (approved):

Gemfile
hcli-work: removed. it's needed to auto add folks to Slack. the API for adding is only available on the Enterprise plan

PR #601, opened 2021-02-18
Inject CSS into the Rise360 content

sadleb:
### Description

[Task](https://app.asana.com/0/1174274412967132/1199552136290617)

Put Rise360 Modules in an iFrame and inject our own stylesheet into it. This builds on the [add rails-erb-loader](https://github.com/bebraven/platform/pull/585) PR to allow us to apply custom styles to Rise360 content. One (or more) PRs will follow this to actually write the custom CSS we want to apply.

We do it this way b/c CSS in a parent frame doesn't apply to a child iframe, but you can use Javascript from a parent frame to access the inside of the child iframe IF you are on the same domain.

### Screenshots (if appropriate):
Notice how "Resume Lesson" has green text. That is the result of uncommenting the example in `rise360_content.scss ` 

image

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #606, opened 2021-02-24
More Rise360 CSS customization (matching, fill in blank, inline feedback + hide X Incorrect sometimes)

sadleb:
### Description

[Task](https://app.asana.com/0/1174274412967132/1199552136290617)

Follow-up on https://github.com/bebraven/platform/pull/601 , https://github.com/bebraven/platform/pull/589 , and https://github.com/bebraven/platform/pull/603 to add a bit more custom styling to Rise360 content.

* Fixes the inline feedback, making it green instead of red, for some places I missed in the first pass (matching, fill-in-the-blank, and checklists with different settings from before)
* Hides the "X Incorrect" overall feedback for checklists in the lesson content (not in a mastery quiz) b/c we don't expect them to get them all correct 

### Test Plan
For lesson content:
- Checkboxes correct shows "<check> Correct"
- Checkboxes incorrect hides "X Incorrect"
- Radio list shows both correct and incorrect
- Fill in the blank shows both correct and incorrect
- Matching interaction shows both correct and incorrect

For quiz content (aka mastery quizzes):
- All of the above shows both Correct and Incorrect

### Motivation and Context
Grades and feedback on things that feed into grades are one of the things that generates the most noise. It's worth making the experience of everything related to grades as streamlined as possible to make it less confusing and reduce noise.

Matching inline feedback:
image

Fill in the blank inline feedback:
image

Hide "X Incorrect" for checkbox lists in lesson content only:
image

### Types of changes

### Checklist:


app/assets/stylesheets/rise360_content.scss
```diff
@@ -50,23 +50,45 @@
...
+// (not a Mastery Quiz). We use these to have them think about the question and attempt to
+// answer so we can give more context, but we don't expect them to get it all correct.
+// This is VERY brittle b/c there is no way to tell if it's a radio list of checkbox list
+// without javascript except looking at the aria-label
+.page__content .block-knowledge div.block-knowledge__wrapper[aria-label*="Multiple Response"] {
+  div.quiz-card__feedback-icon:not(.quiz-card__feedback-icon--correct),
+  div.quiz-card__feedback-icon:not(.quiz-card__feedback-icon--correct) + .quiz-card__feedback-label {
+    display: none;
+  }
+}
```
sadleb, line 91: If we don't use the `aria-label` to target this, which is much more likely to change than the CSS classes, we'd have to use Javascript to start at the div we're targeting, go up a couple levels, go over to a sibling, and go down a couple levels to determine that it's a list of checkboxes vs a radiolist. When this does eventually break, maybe we invest the time to use JS to make it more robust, but for now being able to test the JS in dev is so hard that it feels harder to maintain by going that route.

Eventually I would like to come up with a solution to have `ngrok`'d iframes in Rise360 content load without having to manually edit the HTML in the Chrome Inspector.

rshipp: doesn't the "hard to test in dev" part go away now that the main PR is merged? you should be able to edit css/js live in-browser in prod now, right?

app/assets/stylesheets/rise360_content.scss
```diff
@@ -50,23 +50,45 @@
...
 
+// Hide overall "X Incorrect" icon and label for multiple choice checkboxes when in a Lesson
+// (not a Mastery Quiz). We use these to have them think about the question and attempt to
+// answer so we can give more context, but we don't expect them to get it all correct.
+// This is VERY brittle b/c there is no way to tell if it's a radio list of checkbox list
+// without javascript except looking at the aria-label
+.page__content .block-knowledge div.block-knowledge__wrapper[aria-label*="Multiple Response"] {
+  div.quiz-card__feedback-icon:not(.quiz-card__feedback-icon--correct),
+  div.quiz-card__feedback-icon:not(.quiz-card__feedback-icon--correct) + .quiz-card__feedback-label {
+    display: none;
```
rshipp, line 89: just a heads up in case we run into it, there can be some weird edge cases where `display:none` [is still read by screenreaders](https://juicystudio.com/displayarticle.php/screen-readers-display-none.php?page=screen-readers-display-none.php), even though it's technically not supposed to be. shouldn't be a huge issue here even if it messes up though.

rshipp (approved):

app/assets/stylesheets/rise360_content.scss
sadleb: Yeah, I suppose it's easier since I don't need to change anything with the logic to find the the iframe in the first place. I can just directly run the JS and see if it does the right thing having confidence that it will run against the real thing on-load.

Linking here from this task to revisit at some point: https://app.asana.com/0/1174274412967132/1199978731682878

PR #609, opened 2021-03-02
IN CASE WE NEED IT: Migrating SSO Discovery page over

hcli-work:
We think that NLU will be off the NLU SSO login feature (for Highlander) by the time a new NLU launch happens, which is the only reason we need this page, but just in case we’re keeping this PR in our back pocket for the scenario we have to rush out support for NLU SSO login after the Join server is deleted.
 ——-
https://app.asana.com/0/1109150244034467/1199981838773832

This pull request migrate the SSO Discovery page that is currently on Join over to Platform. After this PR goes live, we will have to do the following for staging and prod platform:
1. add CANVAS_SERVER and BOOSTER_PORTAL_SERVER environment variables
2. update the Discovery URL here http://canvasweb:3000/accounts/1/authentication_providers

Note that the fonts are a little off on this implementation versus the original. I was wary of adding fonts in though as I didn't know if it would screw up assets elsewhere.

Test Plan: logged out locally and logged back into canvasweb:3000. presented with this page on platform
image

clicked on one of them (not NLU or Booster) and logged in


hcli-work:
I am pretty hazy on this pull request. Help would be appreciated especially around the assets stuff.

config/secrets.yml
```diff
@@ -4,6 +4,8 @@ development:
   aws_access_key: <%= ENV['AWS_ACCESS_KEY_ID'] %>
   aws_secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
   aws_region: <%= ENV.fetch('AWS_REGION') { 'us-east-1' } %>
+  booster_portal_server: <%= ENV['BOOSTER_PORTAL_SERVER'] %>
+  canvas_server: <%= ENV.fetch('CANVAS_SERVER') %>
```
rshipp, line 8: one of these formats, i forget which, will crash if the variable isn't set. is this the behavior you want/expect for these variables?

app/views/layouts/minimal.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Braven - Choose Your University</title>
+    <meta name="description" content="Braven">
+    
+    <!-- Until we add a landing page and login, keep our site out of search engines -->
+    <meta name="robots" content="noindex">
+    <link rel="favorite icon" href="<%= image_path("favicon.ico") %>" />
+    <link rel="apple-touch-icon" href="<%= image_path("favicon-152.png") %>" />
+    <%= stylesheet_link_tag 'minimal', media: "all", "data-turbolinks-track" => true %>
+    <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,300,700,400italic' rel='stylesheet' type='text/css'>
```
rshipp, line 13: can we get rid of this? it's an extra network request for a fallback font

app/views/layouts/minimal.html.erb
```diff
@@ -0,0 +1,36 @@
...
+<html lang="en">
+  <head>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Braven - Choose Your University</title>
+    <meta name="description" content="Braven">
+    
+    <!-- Until we add a landing page and login, keep our site out of search engines -->
+    <meta name="robots" content="noindex">
+    <link rel="favorite icon" href="<%= image_path("favicon.ico") %>" />
+    <link rel="apple-touch-icon" href="<%= image_path("favicon-152.png") %>" />
```
rshipp, line 11: do we actually have any of these files?

app/views/cas/sso_discovery.erb
```diff
@@ -0,0 +1,72 @@
...
+    </div>
+
+    <br />
+    <br />
+
+    Braven Staff Member? <a class="login-link" href="//<%= Rails.application.secrets.canvas_server %>/login/cas/1"> Login to the Braven Portal</a>
+  </div>
+
+  <div style="display: none;" id="nlu_university_choice">
+    <h2>National Louis University <a onclick="return showAll();" href="./">Change</a></h2>
```
rshipp, line 44: inline scripts are disabled most places by our CSP. might need to add an exception for this page if we need this code.

app/views/cas/sso_discovery.erb
```diff
@@ -0,0 +1,72 @@
...
+      <a href="//<%= Rails.application.secrets.canvas_server %>/login/cas/1">
+        <div>I am an...</div>
+        <span>Instructor/LC<br />&#9654;</span>
+      </a>
+    </div>
+
+  </div>
+
+  <script>
+    function showAll() {
```
rshipp, line 61: see inline scripts comment

app/views/layouts/minimal.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    <%= stylesheet_link_tag 'minimal', media: "all", "data-turbolinks-track" => true %>
+    <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,300,700,400italic' rel='stylesheet' type='text/css'>
+    <%= csrf_meta_tags %>
+    <script>
```
rshipp, line 15: see inline scripts comment

app/views/cas/sso_discovery.erb
```diff
@@ -0,0 +1,72 @@
+<div id="sso_discovery">
+  <script>document.body.className = 'sso_discovery';</script>
```
rshipp, line 2: see inline scripts comment

app/assets/stylesheets/minimal.css.scss
```diff
@@ -0,0 +1,137 @@
...
+      padding: 1em;
+      line-height: 4em; /* (height - vertical padding) / 2 */
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      background-color: white;
+    }
+
+    span {
+      font-family: "TradeGothicNo.20-CondBold", "Oswald", "Open Sans", sans-serif;
```
rshipp, line 95: we don't use tradegothic anymore, can we replace this with our standard fonts?

app/assets/stylesheets/minimal.css.scss
```diff
@@ -0,0 +1,137 @@
...
+    text-decoration: underline;
+    font-weight: bold;
+  }
+
+  .choose-your-university {
+    /*display: flex;*/
+  }
+
+  .choose-your-role a {
+    font-family: "TradeGothicNo.20-CondBold", "Oswald", "Open Sans", sans-serif;
```
rshipp, line 28: see fonts comment

app/views/cas/sso_discovery.erb
rshipp: note local dev behavior is different than prod behavior, so even if it works locally, it may still fail in prod.

app/assets/stylesheets/minimal.css.scss
rshipp: actually it looks like that change is only on highlander, not staging, so TradeGothic is fine here as long as we don't expect to ever merge this code to highlander.

rshipp:
I don't know why the fonts aren't working, I'd guess it's not including the application css correctly somehow. this is where they're defined https://github.com/bebraven/platform/blob/staging/app/assets/stylesheets/application.scss

config/secrets.yml
hcli-work: good catch! I just copied these straight from Join. :/

app/views/layouts/minimal.html.erb
hcli-work: I'm fine with that and using whatever default font Platform uses

app/views/layouts/minimal.html.erb
hcli-work: these two are part of the PR

app/views/layouts/minimal.html.erb
hcli-work: let me try to rip these out and see what happens...

app/views/layouts/minimal.html.erb
rshipp: whoops, thought i looked at the files

app/views/cas/sso_discovery.erb
```diff
@@ -0,0 +1,72 @@
...
+      </a>
+    </div>
+
+    <br />
+    <br />
+
+    Braven Staff Member? <a class="login-link" href="//<%= Rails.application.secrets.canvas_server %>/login/cas/1"> Login to the Braven Portal</a>
+  </div>
+
+  <div style="display: none;" id="nlu_university_choice">
```
rshipp, line 43: i missed this in my first pass, but inline CSS is also disabled in our CSP

hcli-work: that's cool. I'm going to rip this stuff out and just make it two web pages instead of this silly inline JS nonsense

PR #610, opened 2021-03-02
Grade the Module in real-time when they complete it.

sadleb:
### Description

Task: https://app.asana.com/0/1174274412967132/1199952642130847

### Motivation and Context
It's confusing to finish a module and not see your grade. Only doing it if they complete the module should
be fine computationally, since it's a background job, and will reduce the nightly load of the rake task as well.

### Test Plan
Do half a Module and go to the grades page as a Fellow. There shouldn't be a grade and it should show a message about you having ungraded work. Now go complete the Module and go back. The message should go away and you should see a grade.


app/jobs/grade_module_for_user_job.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def perform(user, canvas_course_id, canvas_assignment_id, activity_id)
+
+    # TODO: only grade Modules and not things in the LC Playbook b/c it generates
+    # email notifications and it's weird.  We probably still want to grade LCs and staff
+    # in Modules just so they can test it out though:
+    # https://app.asana.com/0/1174274412967132/1199946751486950
+
+    Honeycomb.start_span(name: 'GradeModuleForUserJob.perform') do |span|
+      span.add_field('grade.module.for.user.job.user_id', user.id)
```
rshipp, line 17: this is a lot of dots. any specific reason to name the field that way?

sadleb: Whoops, I was trying to follow the convention of the built-in rails fields but went overboard. I think it should be `grade_module_for_user_job.user_id` instead.

rshipp (approved):

PR #612, opened 2021-03-03
Module grading improvements

rshipp:
### Description

Complete overhaul of automatic module grading code, with comprehensive tests.

New grading formula:

When the module has mastery questions:
* 40% of total grade is engagement (% progress through the module)
* 40% of total grade is mastery score (# of correct mastery question answers / # of total mastery questions in module)
* 20% of total grade is timeliness (did you reach 100% progress before due date?)

When the module does not have mastery questions:
* 80% of total grade is engagement (% progress through the module)
* 20% of total grade is timeliness (did you reach 100% progress before due date?)

The contents of the module grading rake task were moved to a service, and new functionality was added to both that service and the grading lib.

Testing manually:

Set up a course, user, module, etc. Interact with the module as a user. Then run the rake task, e.g.:

    docker-compose exec platformweb bundle exec rake grade:modules

Here's an example Honeycomb trace from the rake task: https://ui.honeycomb.io/braven/datasets/development/result/GySTANVPAiZ/trace/iD661yGo88R

### Motivation and Context

Module grading:

* did not take "lateness" into account
* did not give automatic zero grades to people who did not interact with the module at all
* did not have enough automated tests
* autograded LC Playbook modules, sending annoying emails

Task: https://app.asana.com/0/1174274412967132/1199946751486971/f

### Types of changes

### Checklist:


lib/salesforce_api.rb
```diff
@@ -95,6 +95,19 @@ def patch(path, body, headers={})
...
+  def get_current_and_future_accelerator_programs()
+    soql_query = 
+      "SELECT Id, Name, Highlander_Accelerator_Course_ID__c FROM Program__c " \
+      "WHERE RecordType.Name = 'Course' AND Highlander_Accelerator_Course_ID__c <> NULL AND Status__c IN ('Current', 'Future')"
+
+    response = get("#{DATA_SERVICE_PATH}/query?q=#{CGI.escape(soql_query)}")
+    response_json = JSON.parse(response.body)
+    raise SalesforceDataError, "Got paginated response which isn't implemented" if response_json['nextRecordsUrl'].present?
+    response_json
+  end
```
sadleb, line 109: Example response:
```
{
  "totalSize": 2,
  "done": true,
  "records": [
    {
      "attributes": {
        "type": "Program__c",
        "url": "/services/data/v49.0/sobjects/Program__c/a2Y15555555555UAY"
      },
      "Id": "a2Y15555555555UAY",
      "Name": "CHI: NLU Summer 2020",
      "Highlander_Accelerator_Course_ID__c": "555"
    },
    {
      "attributes": {
        "type": "Program__c",
        "url": "/services/data/v49.0/sobjects/Program__c/a2Y155555555UAI"
      },
      "Id": "a2Y155555555UAI",
      "Name": "Braven Tech (TEST) Highlander Spring 2021",
      "Highlander_Accelerator_Course_ID__c": "35555"
    }
  ]
}

```

lib/tasks/grade_modules.rake
```diff
@@ -3,6 +3,7 @@
 # at some point we decide we need to calculate grades more frequently, we may need to optimize this
 # task to be more memory- and/or time-efficient.
 
+require 'time'
```
rshipp, line 6: not sure if this is already included by Rails somewhere, but since i'm calling Time directly, i figured better safe than sorry

lib/tasks/grade_modules.rake
```diff
@@ -3,65 +3,102 @@
...
+                  "no interactions and assignment isn't due yet"
+              next
+            end
+
+            # ASSUMPTION: Root activity ID will be the same for all canvas_assignment_id,
+            # so just select the first one.
+            root_activity_id = Rise360ModuleInteraction
+              .where(canvas_assignment_id: canvas_assignment_id)
+              .first
+              .root_activity_id
```
rshipp, line 85: @sadleb does this sound correct?

rshipp: note to self: try to get rid of activity id and use canvas_assignment_id instead, this might be a leftover from old code

lib/tasks/grade_modules.rake
```diff
@@ -32,8 +32,8 @@ namespace :grade do
       Honeycomb.start_span(name: 'rake:grade:modules:course') do |span|
         # We're doing some less-readable queries here because they're drastically
         # more efficient than using the more-readable model associations would be.
-        sections = Section.where(course: course)
-        roles = Role.where(resource: sections)
+        sections = Section.select(:id).where(course: course)
+        roles = Role.select(:id).where(resource: sections)
```
sadleb, line 44: This might be the same thing, but FYI, you can also use `pluck` at the end of an active record query chain to limit the query to only the column(s) you’re interested in and it returns an array instead of active records.

sadleb: Also, TIL, you can tack an `explain` onto the end and it will return the query plan. Whoa.

rshipp: oh that helps for the places i was doing `.select.where.map`, thanks. these two queries i need to be activerecord relations, because i'm relying on the fact that activerecord combines these two with the next UserRole one into one long query with inner queries, which is more efficient than `select id`-ing a bunch of intermediate things into Rails memory and then back into new queries.

lib/module_grade_calculator.rb
```diff
@@ -85,6 +126,21 @@ def self.grade_mastery_quiz(interactions, total_questions)
     100 * (correct_answers.to_f / total_questions.to_f)
   end
 
+  # Returns 0 *or* 100, representing whether the user reached 100%
+  # progression before the due date.
+  def self.grade_completed_on_time(interactions, due_date)
+    on_time_progress = interactions.where('created_at <= ?', due_date)
+      .order(:created_at)
+      .last
+      .progress
```
sadleb, line 135: Is it safer to use `max(progress)`? I suppose it doesn't matter. I don't think it's possible.

lib/module_grade_calculator.rb
```diff
@@ -85,6 +126,21 @@ def self.grade_mastery_quiz(interactions, total_questions)
...
+
+    if progress < 100
+      0
+    else
+      100
+    end
+  end
+
   # Returns SQL that strips off _<timestamp> from the activity_id.
   def self.activity_id_without_timestamp
```
sadleb, line 145: Is this still used?

app/services/grade_modules.rb
```diff
@@ -0,0 +1,153 @@
...
+  def run
+    Honeycomb.start_span(name: 'GradeModules.run') do |span|
+      # From the list of "running" "programs" in Salesforce, fetch a list of "accelerator"
+      # (non-LC) courses.
+      programs = SalesforceAPI.client.get_current_and_future_accelerator_programs
+      if programs.empty?
+        puts "Exit early: no current/future accelerator programs"
+        return
+      end
+      # TODO: Remember to swap this hardcoded Highlander stuff out when we switch to prod.
```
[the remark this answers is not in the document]
rshipp: not sure what you mean, it's a note about the hardcoded `Highlander_Accelerator_Course_ID__c` below

app/services/grade_modules.rb
```diff
@@ -0,0 +1,153 @@
...
+    Honeycomb.start_span(name: 'GradeModules.grade_course') do |span|
+      # We're doing some less-readable queries here because they're drastically
+      # more efficient than using the more-readable model associations would be.
+      sections = Section.select(:id).where(course: course)
+      roles = Role.select(:id).where(resource: sections)
+      # We're loading all the User IDs into memory right now, so keep an eye out
+      # if this needs to be batched or something.
+      # NOTE: Don't copy this UserRole code anywhere else unless you *really* need the performance.
+      # TODO: Can this be .where.group.pluck?
+      user_ids = UserRole.select(:user_id).where(role: roles).group(:user_id).map { |ur| ur.user_id }
```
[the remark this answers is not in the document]
rshipp: see https://github.com/bebraven/platform/pull/612#discussion_r588484467

lib/module_grade_calculator.rb
rshipp: i'm relying on your testing where you mentioned it never gives a lower progression.

lib/module_grade_calculator.rb
rshipp: it is, see `grade_mastery_quiz` function

app/services/grade_modules.rb
```diff
@@ -0,0 +1,153 @@
...
+        return
+      end
+      # TODO: Remember to swap this hardcoded Highlander stuff out when we switch to prod.
+      canvas_course_ids = programs['records'].map { |r| r['Highlander_Accelerator_Course_ID__c'] }
+      span.add_field('canvas_course_ids', canvas_course_ids)
+
+      # Eliminate courses with no module interactions, and exit early if that
+      # leaves us with an empty list.
+      courses = Course.where(canvas_course_id: canvas_course_ids)
+        .filter { |c| Rise360ModuleInteraction.where(canvas_course_id: c.canvas_course_id).exists? }
```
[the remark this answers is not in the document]
rshipp: probably wouldn't hurt, it doesn't look like there's one for course.

app/services/grade_modules.rb
```diff
@@ -0,0 +1,153 @@
...
+
+  def grade_assignment(canvas_assignment_id, user_ids)
+    Honeycomb.start_span(name: 'GradeModules.grade_assignment') do |span|
+      span.add_field('canvas_assignment_id', canvas_assignment_id)
+
+      # Initialize map of grades[canvas_user_id] = 'X%'.
+      grades = Hash.new
+
+      # Skip assignments with zero interactions; they probably haven't opened yet.
+      unless Rise360ModuleInteraction.where(canvas_assignment_id: canvas_assignment_id).exists?
```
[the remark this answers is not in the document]
rshipp: there's one for `user_id, canvas_assignment_id` already, but i added another for just assignment id. i think we're better off having too many rather than too few, but I don't actually know what the cost is for adding more indices.

app/services/grade_modules.rb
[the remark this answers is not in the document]
rshipp: updated the code to remove my TODO, and add a comment clarifying what this query looks like

app/services/grade_modules.rb
```diff
@@ -0,0 +1,159 @@
...
+      # NOTE: the `new` column should only be considered an estimate with +/- 1 day resolution.
+      max_id = Rise360ModuleInteraction.maximum(:id)
+      span.add_field('interactions.max_id', max_id)
+      span.add_field('users.count', user_ids.count)
+
+      # All users in the course, even if they haven't interacted with this assignment.
+      user_ids.each do |user_id|
+        # If we're before the due date, and there are no *new* interactions, skip this user.
+        # This might crash if there are no due dates at all, but that shouldn't ever
+        # happen in a real, running course.
```
rshipp, line 115: todo: probably need to just make this not crash 😞

app/services/grade_modules.rb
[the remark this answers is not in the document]
sadleb: That's the name of the Salesforce field. It shouldn't change. If we do decide to one day rename it, hopefully we'll grep the code for where it's used and not just assume it's only used in SalesforceAPI. I think you can delete this TODO.

app/services/grade_modules.rb
```diff
@@ -0,0 +1,169 @@
...
+      # From the list of "running" "programs" in Salesforce, fetch a list of "accelerator"
+      # (non-LC) courses.
+      programs = SalesforceAPI.client.get_current_and_future_accelerator_programs
+      if programs['records'].empty?
+        puts "Exit early: no current/future accelerator programs"
+        return
+      end
+      # TODO: Remember to swap this hardcoded Highlander stuff out when we switch to prod.
+      canvas_course_ids = programs['records'].map { |r| r['Highlander_Accelerator_Course_ID__c'] }
+      span.add_field('canvas_course_ids', canvas_course_ids)
```
sadleb, line 28: Generally, I've been trying to following a convention of sending all fields specific to particular class prefixed with the class. E.g. `span.add_field('grade_modules.canvas_course_ids', canvas_course_ids)`. Should that be a convention?

app/services/grade_modules.rb
```diff
@@ -0,0 +1,169 @@
...
+
+      # Eliminate courses with no module interactions, and exit early if that
+      # leaves us with an empty list.
+      courses = Course.where(canvas_course_id: canvas_course_ids)
+        .filter { |c| Rise360ModuleInteraction.where(canvas_course_id: c.canvas_course_id).exists? }
+      span.add_field('courses.count', courses.count)
+      if courses.empty?
+        puts "Exit early: no accelerator courses with interactions"
```
sadleb, line 36: The logger doesn't seem to output to STDOUT from Rake tasks, which is why I was using `puts` in the Rake task. Generally, we should always use the logger instead of `puts` so that in the future we can do fancy stuff under the hood. I feel like we should switch to using the logger in this service and if it's not being output when the Rake task is run, then queue up a separate task to somehow enable logging when run in a Rake context?

rshipp: I wasn't sure if logger worked from services, but I can switch that out.

app/services/grade_modules.rb
```diff
@@ -0,0 +1,169 @@
...
+      #   SELECT "users_roles"."user_id" FROM "users_roles" WHERE "users_roles"."role_id" IN (
+      #     SELECT "roles"."id" FROM "roles" WHERE <...> AND "roles"."resource_id" IN (
+      #       SELECT "sections"."id" FROM "sections" WHERE "sections"."course_id" = <course.id>
+      #     )
+      #   ) GROUP BY "users_roles"."user_id"
+      sections = Section.where(course: course)
+      roles = Role.where(resource: sections, name: RoleConstants::STUDENT_ENROLLMENT)
+      # We're loading all the User IDs into memory right now, so keep an eye out
+      # if this needs to be batched or something.
+      # NOTE: Don't copy this UserRole code anywhere else unless you *really* need the performance.
```
sadleb, line 62: I'm curious as to why we should avoid it? What is the downside of using it? I was having trouble following the linked github issue where this solution came from in order to understand why it's bad, other than the fact that it's not the built-in Rollify API.

app/services/grade_modules.rb
rshipp: I don't think so, what's the reason you were thinking for that? The fields are inside spans, so that wouldn't add any context that wasn't there already. I can't think of any reason where a field from somewhere else would conflict, if you were trying to query across a bunch of traces. If the field was something that existed with a completely different meaning somewhere else, like "action", I could see namespacing it, but here I think it'd just make the field names longer.

app/jobs/grade_module_for_user_job.rb
```diff
@@ -6,32 +6,47 @@
...
     # TODO: only grade Modules and not things in the LC Playbook b/c it generates
     # email notifications and it's weird.  We probably still want to grade LCs and staff
     # in Modules just so they can test it out though:
     # https://app.asana.com/0/1174274412967132/1199946751486950
 
     Honeycomb.start_span(name: 'GradeModuleForUserJob.perform') do |span|
+      # Note a lot of this code is duplicated (but simplified) from app/services/grade_modules.rb.
       span.add_field('grade_module_for_user_job.user_id', user.id)
       span.add_field('grade_module_for_user_job.canvas_course_id', canvas_course_id)
       span.add_field('grade_module_for_user_job.canvas_assignment_id', canvas_assignment_id)
```
sadleb, line 20: Should we use the same field names for things that are sent in multiple places? E.g. you're introducing the following in `services/grade_modules`
```
span.add_field('canvas_assignment_id', canvas_assignment_id)
span.add_field('course.id', course.id)
span.add_field('course.canvas_course_id', course.canvas_course_id)
```
Maybe change these to match? Generally I think if a field is specific to a class, it should be prefixed with the class but if it's something that would be sent from all over, then we should try and consistently send the same field name.

app/services/grade_modules.rb
rshipp: Because I chose a sub-par solution that works here, instead of a more complete solution that would've taken more work. If we ever want to use this model for other things, we should take the time to set up the associations in User and Role models, rename the table to follow pluralization convention, closer to this https://github.com/RolifyCommunity/rolify/issues/318#issuecomment-276480921. Otherwise there might be weird behavior with some of the model associations.

app/services/grade_modules.rb
```diff
@@ -0,0 +1,169 @@
...
+        # into grade_assignment because grabbing the list of users is slow.
+        grade_assignment(canvas_assignment_id, user_ids)
+      end
+    end
+  end
+
+  def grade_assignment(canvas_assignment_id, user_ids)
+    Honeycomb.start_span(name: 'GradeModules.grade_assignment') do |span|
+      # Select course again, because it's not that expensive and saves us passing it in.
+      course = CourseRise360ModuleVersion.find_by(canvas_assignment_id: canvas_assignment_id).course
```
sadleb, line 85: I don't get the reason to avoid passing it in

rshipp: Passing a bunch of things around in a class is gross. I don't feel strongly on this one, it's a small change if you'd rather pass in a `course` object to this function.

app/jobs/grade_module_for_user_job.rb
rshipp: I don't think it matters a lot, since you're unlikely to be writing a single query that acts on these fields across both Job spans and Service spans. Do you want to remove the namespace on the Job fields?

app/views/rise360_module_grades/show.html.erb
```diff
@@ -42,27 +47,51 @@
...
+      Your grade is recalculated as:
+    <p/>
+
+    <p><code>(Engagement grade) + (Mastery grade) + (On time) = (Total grade)</code></p>
+    <p><code>(100% of 40%) + (90% of 40%) + (0) = (X% of total)</code></p>
+    <p><code>(1.0 * 0.4) + (0.9 * 0.4) = 0.76</code></p>
+    <p>so... </p>
+    <p><code>76% * 10 points = 7.6 points</code></p>
+    <p>You'll see <strong>GRADE: 7.6 / 10</strong> at the top in this example.</p>
+
```
sadleb, line 95: Whoa, thank you for remembering to update this page!

lib/module_grade_calculator.rb
```diff
@@ -1,61 +1,104 @@
...
+    section_overrides = assignment_overrides.filter { |o| canvas_section_ids.include? o['course_section_id'] }
+
+    # Then, look for user-specific overrides.
+    user_overrides = assignment_overrides.filter { |o| o['student_ids']&.include? user.canvas_user_id }
+
+    # Return the latest due date.
+    [
+      section_overrides,
+      user_overrides,
+    ].flatten.map { |o| o['due_at'] }.max
```
sadleb, line 99: Do we need to instantiate `Time` objects from the value of `o['due_at']` so that `max` works? Or is the string in milliseconds which Ruby/Rails magic treats as an `int` so that's why this works? Or some other reason why this works?

rshipp: No, it works because iso8601 time strings are already sortable alphabetically

lib/module_grade_calculator.rb
```diff
@@ -85,6 +128,24 @@ def self.grade_mastery_quiz(interactions, total_questions)
     100 * (correct_answers.to_f / total_questions.to_f)
   end
 
+  # Returns 0 *or* 100, representing whether the user reached 100%
+  # progression before the due date.
+  def self.grade_completed_on_time(interactions, due_date)
+    # Use due_date unless it's nil, then fall back to current date.
```
sadleb, line 134: Do we fallback to the current date b/c `due_date` is only `nil` if ... actually, how can `due_date` be nil? From what I can tell that's only if the assignment literally doesn't have a due date either globally, for their section, or for that user. Should we just short-circuit out in that case with full on time grade, but log it and send to Honeycomb so we'll see the fact that we missed setting up a due date if one was supposed to be (assuming we put monitoring for that particular situation)?

rshipp: Right, `due_date` is nil when the assignment doesn't have any overrides that apply to this user. It's safe to use `now` because we can safely expect to not have interactions with future `created_at` dates. Originally, this would just crash if it couldn't find a date, but assignments in my local test env didn't have dates, so I decided it was better to just handle it. 

> Should we just short-circuit out in that case with full on time grade

This code will give full credit if the user has interactions, or zero credit if there are no interactions, which I think is the behavior we want here. Short-circuiting would always give full credit. Let me know if that's wrong though. In practice, I don't expect this condition to happen except in test courses / courses that haven't been set up with due dates yet.

> but log it and send to Honeycomb

I didn't think it was worth logging, because this is expected behavior in test courses, and if we didn't have due dates in a real course, I don't think a log would be the first place we'd notice that.

spec/factories/rise360_module_interactions.rb
```diff
@@ -24,6 +24,17 @@
     factory :answered_module_interaction do
       verb { Rise360ModuleInteraction::ANSWERED }
       success { true }
+
+      factory :ungraded_answered_module_interaction do
+        # new: true is the default
+      end
+
+      factory :graded_answered_module_interaction do
+       # add_attributoe(:new) { false } # I think "new" is a reserved word, but this still wasn't working.
```
sadleb, line 33: Oh wow, I really hope the `add_attributoe` typo was after I tried it and not **the reason** that it wasn't working leading to me using the after create.

app/services/grade_modules.rb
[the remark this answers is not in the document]
rshipp: Ah, ok. will do.

app/jobs/grade_module_for_user_job.rb
sadleb: Yeah, I think we should remove the namespace on the Job fields. The reason I think it's worth it is b/c it's nice to have some really common fields, like the course id and assignment id, where you don't have to go look in the code when just tinkering around with queries doing some initial exploration. Though maybe my approach to writing queries is just woefully inadequate, in which case you need to teach me the magic of your ways.

rshipp: I definitely don't look at the code, I generally lean on the autocomplete in the query builder, or if I just want to browse, I'll look at the field list on the Raw Data tab. I can remove the namespace here though.

app/services/grade_modules.rb
sadleb: I started doing it b/c that's what the built-in Rails integrations seemed to do. I never thought about why though. The only reason I can think of is that it's nice to be querying for something in a particular controller / class and be able to see a list of what is available to choose from by typing the class name instead of looking at the code. I'm good with not namespacing them, but I think I really need more insight into how you go about writing Honeycomb queries b/c it always takes me 10 times longer than it should just to get started b/c I don't know which fields are available for the thing I'm looking for.

app/services/grade_modules.rb
sadleb: Logger should work from services when the services are called from the normal Rails application.

app/services/grade_modules.rb
sadleb: I don't feel strongly either.

lib/module_grade_calculator.rb
sadleb: I think the current logic is fine to not give full credit until you've finished it, even if there is no due date. 

Not now, but I imagine there may be things that don't have due dates in the future if they serve more as a resource. Speaking of which, is the Resources thingie graded? No, right? B/c this only runs off CourseModuleVersions... 

Side note: LC Playbook already have module with no due dates, so until we fully fix that bug, this logic will apply there (but may just work since its only still a bug for when we immediately grade at 100% progress).

sadleb (approved):

lib/module_grade_calculator.rb
rshipp: Ugh I hope it doesn't mess with Resources. I'll look into that.

app/services/grade_modules.rb
rshipp: Looking into field names now, I think I'm going to queue up a task to change a bunch of ours, and set up some consistent guidelines for us. We have a lot going on that I didn't realize. (`Honeycomb.add_field` auto-namespaces things in `app.*`, but `span.add_field` doesn't, and libhoney definitely doesnt. 🤯 )

rshipp: https://app.asana.com/0/1174274412967132/1200055379602516/f

lib/module_grade_calculator.rb
rshipp: Yeah it doesn't touch resources, those are in `CourseResource`

PR #614, opened 2021-03-03
Don't return a 404 for missing routes.

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1197923258675345

Instead, return a 401 for JSON requests and redirect to login for
normal web-browser requests.

Note: in dev when you're logged in it will still show the page that
lists all the routes which is convenient.

### Motivation and Context
This is so that bots can't enumerate paths and objects on our server.
Force an actual person to come here on github to find that out!

### Test Plan
During all of the following, look at the Network tab in Chrome inspector and make sure no information about the `missing_route` route is there.

#### Logged out
Hit https://platformweb/asdf -> get a 302 redirect to login
Hit https://platformweb/asdf.json -> get a 401 unauthorized `{ "error" : "" }`

#### Logged in
In dev hit https://platformweb/asdf -> get the "Routing Error" page with a list of routes
Not in dev hit https://platformweb/asdf -> redirect to home page


config/routes.rb
```diff
@@ -222,4 +222,10 @@
 
   # Honeycomb Instrumentation Routes
   post '/honeycomb_js/send_span', to: 'honeycomb_js#send_span'
+
+  # Catch-all route for missing ones to return an unauthorized response rather than a 404
+  get '*other', to: 'home#missing_route'
+  post '*other', to: 'home#missing_route'
+  put '*other', to: 'home#missing_route'
+  delete '*other', to: 'home#missing_route'
```
rshipp, line 230: do we need `patch` too? i think update requests use that

sadleb: fixed @rshipp

rshipp (approved):

PR #615, opened 2021-03-04
Make Rise360 Modules require authentication

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1199668138554725

### Motivation and Context
Since we are just proxying the endpoint through to an Amazon S3 bucket in middleware, this wasn't subject
to the normal ApplicationController authentication we have setup. Warden is the authentication middleware
module underpinning Devise and CAS authentication. We had to drop down to that layer.

### Test Plan
Grab a direct Rise360 package proxy URL like:
      https://platformweb/rise360_proxy/lessons/sxj0v0dsxa8qykfv36n1citzh6jz/index.html#/

Try it in the following situations:
    - normal Chrome logged out
    - normal Chrome logged in
    - Chrome Incognito logged out
    - Chrome Incognito logged in
    
Also try actually loading a Module inside Canvas in the above 4 scenarios.

image

### Types of changes


config/initializers/lti_authentication.rb
```diff
@@ -76,7 +76,10 @@ def finish_authenticate(span, status, message, url = nil, canvas_id = nil, user
...
     # request.referrer contains a state token when an iframe is loaded *inside* CustomContent (Projects).
     def fetch_state
-      lti_state = params[:state] || params[:auth]
+      lti_state = params[:state]
+      unless lti_state
+        lti_state = params[:auth][/#{LtiConstants::AUTH_HEADER_PREFIX} (.*)$/, 1] if params[:auth]
+      end
       unless lti_state
         lti_state = request.headers[:authorization][/#{LtiConstants::AUTH_HEADER_PREFIX} (.*)$/, 1] if request.headers[:authorization]
       end
```
sadleb, line 85: This was a bug thats been around forever but we never noticed b/c the endpoint was wide open. The `auth` param value is the format `LtiState the_state_value` but we weren't handling the `LtiState` prefix.

lib/lti_rise360_proxy.rb
```diff
@@ -27,7 +27,9 @@ def self.proxy_url
 
   def perform_request(env)
     request = Rack::Request.new(env)
+
     if request.path =~ PROXY_REGEX
+      return [401, {}, ['Unauthorized']] unless authenticate(request)
```
sadleb, line 32: If you return an  array of `[status, headers, body]` from any middleware, it stops the chain and returns that to the client.

lib/lti_rise360_proxy.rb
```diff
@@ -36,4 +38,22 @@ def perform_request(env)
...
+  # ourselves
+  def authenticate(request)
+    warden = request.env['warden']
+    return false unless warden
+    return true if warden.authenticated? # short circuit if already authenticated using session
+
+    # Authenticate but don't fail out b/c it would try redirecting to login.
+    # Just return if it fails so we respond with a 401
+    warden.authenticate 
+    return true if warden.user
```
sadleb, line 56: Taken from here: https://github.com/wardencommunity/warden/wiki/Overview#the-how

rshipp (approved):

PR #617, opened 2021-03-10
Feedback Requested: Migrate Braven Network search page over

hcli-work:
This is the first of a series of PRs to try and move Braven Network over to Platform from Join.
The overarching Asana task is here: https://app.asana.com/0/1179330853755533/1200040556288370
This particular PR is for the search page only: https://app.asana.com/0/1179330853755533/1200040556288354

I ended up adding a new scss file, but wanted it to look similar to the styling of accounts, so I pulled all of that stuff into a common `standalone.scss` file that gets imported into `accounts` and `network`.

I also added a partial so that there would be a top red header that would show who's logged in and give folks the ability to log out.

Currently, there is no actual search functionality as there's also no data. :)  That'll be in the next PR.


hcli-work:
image

app/models/user.rb
```diff
@@ -33,7 +33,7 @@ class User < ApplicationRecord
   validates :email, presence: true
 
   def full_name
-    [first_name, last_name].join(' ')
+    [first_name, last_name].compact.join(' ')
```
rshipp, line 36: we have presence validations on both fields, where was this `compact` needed?

app/policies/network_policy.rb
```diff
@@ -0,0 +1,5 @@
+class NetworkPolicy < ApplicationPolicy
+  def connect?
+    !!user
```
rshipp, line 3: just to make sure, this route requires a signed-in user, right? or is it anonymous access? i wasn't thinking when I suggested this policy

rshipp:
This work was merged into https://github.com/bebraven/platform/pull/655

PR #618, opened 2021-03-11
Filter sensitive data out of our logs (inc. Honeycomb and Sentry)

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1199623626113021

Note that it is very unlikely I got *everything* in this pass. We should setup alerts in these services looking for leaked data. Followup task for that: https://app.asana.com/0/1174274412967132/1200066294473216

### Motivation and Context
If there is a hack of a third party like Sentry/Honeycomb/Papertrail, then if we have been sending them sensitive data that is essentially a hack of us. Plus, I think it would be a breach of our LinkedIn API contract to have the access tokens sent to these services.

A few example screenshots:

Logs
image

image

Sentry

image

Honeycomb

image

image

image

### Types of changes

### Checklist:


lib/restclient_instrumentation.rb
```diff
@@ -8,33 +8,31 @@ module RestClientInstrumentation
...
     # I'm trying to mimic what normal Rails request's send:
     # https://guides.rubyonrails.org/active_support_instrumentation.html#process-action-action-controller
-    Honeycomb.start_span(name: "RestClient.#{method}") do |span|
-      span.add_field('class_name', calling_class)
-      span.add_field('method', method)
-      span.add_field('url', redacted_url) # Strips out password if it's in there.
-      span.add_field('timestamp', DateTime.now)
-      redacted_headers = processed_headers.dup
-      redacted_headers['Authorization'] = '[REDACTED]' if redacted_headers['Authorization']
-      span.add_field('headers', redacted_headers)
```
sadleb, line 25: This is the actual change. I'm moving this to FilterLogging class. Everything else is just renaming the fields to match our convention.

lib/core_ext/postgresql_adapter.rb
```diff
@@ -0,0 +1,31 @@
...
+    return super unless filtered_type_casted_binds
+    return {
+      sql: sql,
+      binds: binds,
+      type_casted_binds: filtered_type_casted_binds,
+      name: name,
+      connection: self,
+      cached: true
+    }
+  end
```
sadleb, line 20: The original implementation sends the `type_casted_binds` as a lambda so that it's evaluated lazily considering this is the "cached" route of reading from the DB. Sticking with that by keeping everything a lambda unless we actually need to filter it. Only then do we call `type_casted_binds(binds)` and do the filtering. The below method isn't as optimized, but to share code it also passes in a lambda.

lib/filter_logging.rb
```diff
@@ -0,0 +1,207 @@
...
+    'LoginTicket' => ['ticket'],
+    'ProxyGrantingTicket' => ['ticket'],
+    'ServiceTicket' => ['ticket'],
+    'TicketGrantingTicket' => ['ticket'],
+    'AccessToken' => ['key'],
+  }
+
+  def self.is_enabled?
+    (Rails.env.development? ? false : true)
+  end
```
sadleb, line 36: Thoughts on this? I like having it off in dev so I can just grab stuff from the log instead of the database or browser console, but maybe it would be better to have filtering turned on in dev as well so that it will be more likely that a dev would notice something sensitive sneaking through?

lib/filter_logging.rb
```diff
@@ -0,0 +1,207 @@
...
+        value.clear()
+        value.insert(0, FILTERED)
+
+      elsif param_name == 'url' || param_name == 'u' || param_name == 'pgu'
+        value.gsub!(/state\=([^&]+)/, FILTERED_STATE_QUERY_PARAM)
+        value.gsub!(/auth\=([^&]+)/, FILTERED_AUTH_QUERY_PARAM)
+
+      elsif param_name == 'restiming'
+        value.gsub!(/state\=([^"&]+)/, FILTERED_STATE_QUERY_PARAM)
+        value.gsub!(/auth\=([^"&]+)/, FILTERED_AUTH_QUERY_PARAM)
```
sadleb, line 60: This regex just adds a double quote to the possible terminating character since this is serialized json.

lib/filter_logging.rb
```diff
@@ -0,0 +1,207 @@
...
+    'ticket',
+  ]
+
+  MODELS_TO_FILTER = {
+    'LtiLaunch' => ['state'],
+    'User' => ['encrypted_password', 'reset_password_token', 'linked_in_access_token', 'confirmation_token'],
+    'LoginTicket' => ['ticket'],
+    'ProxyGrantingTicket' => ['ticket'],
+    'ServiceTicket' => ['ticket'],
+    'TicketGrantingTicket' => ['ticket'],
```
rshipp, line 30: We log a lot of these manually from Cas Controller, do those need removed too?

rshipp: example: 

```
platformweb_1      | D, [2021-03-16T17:00:47.930617 #1744] DEBUG -- : Login ticket '<ticket> has been created for '<ip>'
platformweb_1      | Rendering login form with lt: <ticket>, service: https://platformweb/users/service, renew: , gateway: false
```

lib/filter_logging.rb
rshipp: I don't remember ever needing to use a secret from the Rails log, and being able to catch things getting through filters in dev is a benefit. No strong opinion.

lib/core_ext/postgresql_adapter.rb
```diff
@@ -0,0 +1,34 @@
...
+require 'filter_logging'
+
+class ActiveRecord::ConnectionAdapters::PostgreSQLAdapter 
+
+private
+
+  # Overrides the "info" sent in an ActiveRecord::Notification for cached values.
+  def cache_notification_info(sql, name, binds)
+    filtered_type_casted_binds = FilterLogging.filter_sql(name, binds, -> { type_casted_binds(binds) })
+    return super unless filtered_type_casted_binds
```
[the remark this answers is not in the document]
sadleb: Nope. Super with no arguments passes whatever the method was called with as arguments

config/initializers/sentry.rb
```diff
@@ -0,0 +1,71 @@
...
+    #'ActionDispatch::Http::Parameters::ParseError',
+    #'ActiveRecord::RecordNotFound'
+
+    ################
+    # Taken from sentry-ruby IGNORE_DEFAULT array here:
+    # https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/configuration.rb#L151
+    'Rack::QueryParser::InvalidParameterError',
+    'Rack::QueryParser::ParameterTypeError',
+    'Mongoid::Errors::DocumentNotFound',
+    'Sinatra::NotFound',
```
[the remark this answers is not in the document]
sadleb: Yeah, you’re right. I should start with these commented out and only add them to the ignore list if they generate a bunch of red herring exceptions in the wild

lib/filter_logging.rb
```diff
@@ -0,0 +1,207 @@
...
+    'password_confirmation',
+    'encrypted_password',
+    'reset_password_token',
+    'linked_in_access_token',
+    'confirmation_token',
+    'ticket',
+  ]
+
+  MODELS_TO_FILTER = {
+    'LtiLaunch' => ['state'],
```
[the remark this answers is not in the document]
sadleb: I didn’t add most of those bc they are single use things as part of the handshake. The exception is client ID and ID token payload. I don’t think client ID is actually sensitive? I need to think about that a bit more. I will add the ID token payload though. I doubt it could be decrypted without one of our private keys, but there was a no reason to have it in there and it does contain PII.

lib/filter_logging.rb
sadleb: Good catch. I’ll go through the manual CAS logs and double check.

lib/filter_logging.rb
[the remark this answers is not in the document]
sadleb: Canvas initiates the LtiLaunch handshake with the client ID so I think we're fine. If knowing it meant a nefarious actor could initiate the handshake making it look like they were us, I'd filter it.

sadleb:
@rshipp the commits starting with "Filter sensitive data in all environments" are what I did to address your code review feedback.

rshipp (approved):

PR #623, opened 2021-03-15
Upgrade Rails to 6.1

rshipp:
### Description

Upgrade Rails from 6.0 to 6.1. 

Supporting changes:

* Ran `bin/rails app:update`, applied some changes
* Upgraded `rspec-rails` to 4.x to fix ```WrongScopeError: `name` is not available from within an example``` on all tests
* Removed `source:` on `has_many` associations without `through:` to fix `ArgumentError: Unknown key: :source` on Course model
* Applied new 6.1 defaults.

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1184591252130409/f


config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -7,61 +7,61 @@
...
 # `before_enqueue`/`before_perform` respectively halts with `throw :abort`.
-# Rails.application.config.active_job.skip_after_callbacks_if_terminated = true
+Rails.application.config.active_job.skip_after_callbacks_if_terminated = true
 
 # Specify cookies SameSite protection level: either :none, :lax, or :strict.
 #
 # This change is not backwards compatible with earlier Rails versions.
 # It's best enabled when your entire app is migrated and stable on 6.1.
-# Rails.application.config.action_dispatch.cookies_same_site_protection = :lax
+Rails.application.config.action_dispatch.cookies_same_site_protection = :lax
```
rshipp: This is the only one I think we'd care about.

sadleb (approved):
I don’t know what to look for from a review perspective. But the new settings make sense to me. It was good to see the backtrace silencer too. I want to add some of our middleware to that since it’s confusing and a red herring.

PR #625, opened 2021-03-17
Authenticate AWS S3 bucket requests for Rise360 content

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1198943778694978

We already have AWS access keys configured that are used when publishing Rise360 content using the SDK gem. This just piggybacks off the SDK to make the proxied requests for showing the Rise360 content use authenticated presigned URLs. After deploying this, turn the bucket permissions for both dev and prod to only allow the IAM accounts we want to be able to access the buckets.

### Motivation and Context
Prevents our content from being available to the world.

### Test Plan
- Make sure you can hit a URL like: https://platformweb/rise360_module_versions/9?state=some_state before this change and it works
- Merge this change and make sure it still works
- Make sure a URL like https://our-s3-bucket.s3.amazonaws.com/lessons/adsjlhbvi3njl1mau14uea2bdhre/index.html works with the bucket still not locked down
- Turn on permissions for the bucket and make sure the above URL doesn't work (no access error)
- Finally make sure the URL in step 1 still works after locking the bucket down

### Types of changes

### Checklist:


lib/lti_rise360_proxy.rb
```diff
@@ -30,8 +32,14 @@ def perform_request(env)
 
     if request.path =~ PROXY_REGEX
       return [401, {}, ['Unauthorized']] unless authenticate(request)
-      env["HTTP_HOST"] = @backend.host # e.g. some-bucket.s3.amazonaws.com
-      env['PATH_INFO'] = $1            # The match in the regex above
+
+      path = $1 # from regex match above
+      uri = URI(Rise360Util.presigned_url(path))
```
rshipp, line 37: i'll always remember the URI pain, but have forgotten what caused it - is this call safe, or should we use Addressable::URI instead?

spec/lib/lti_rise360_proxy_spec.rb
```diff
@@ -12,7 +12,8 @@
 
   let(:dummy_host) { 'platform.host' }
   let(:dummy_path) { '/some/path' }
-  let(:dummy_env) { { 'HTTP_HOST' => dummy_host, 'PATH_INFO' => dummy_path} }
+  let(:dummy_query_stinrg) { '' }
```
rshipp, line 15: typo?

rshipp (approved):

lib/lti_rise360_proxy.rb
sadleb: `URI` parsing is fine. `URI` joining and constructing is what causes pain. I just need to parse the host/path/query components, so this should be safe.

PR #631, opened 2021-03-18
Honeycomb field normalization

rshipp:
### Description

Remember!

    Honeycomb.add_field('my_field_name')
    span.add_field('app.my_field_name')

#### Notes / areas for discussion: 

Things we want to get notifications about
* `app.alert.blah`
  * `app.alert.attendance.no_events`
  * `app.alert.attendance.multiple_sections`

Things we don't want to get notifications about, but might help with debugging
* `app.<where to look in the code>.blah`
    * `app.unpublish.canvas_assignment_not_found`
    * `app.lti_launch_cleanup.current_server_time`
    * `js.app.rate_this_module_submissions.iframe.search.time`

Things that represent a count of objects, or length of a list:
* `app.<where>.blah.count` (postfix with `.count`)
    * `app.grade_modules.courses.count`

Exceptions we catch and log manually (note these are discrete from `error` and `error_detail` at the root level; we can't `Honeycomb.add_field` without the `app` prefix):
* `app.error`: e.class.name
* `app.error_detail`: e.message

Canvas:
* Always `app.canvas.user.id`, never `app.user.canvas_user_id`
* Always `app.canvas.assignment.id`, never `app.course_project_version.canvas_assignment_id`

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200055379602516/f

### Types of changes

### Checklist:


app/controllers/honeycomb_js_controller.rb
```diff
@@ -45,29 +48,32 @@ def send_span
     existing_trace_to_add_to = request.headers[X_HONEYCOMB_TRACE_HEADER]
     raise ActionController::BadRequest, "Missing '#{X_HONEYCOMB_TRACE_HEADER}' header" unless existing_trace_to_add_to 
 
-    span_name = params[:name] || 'javascript.event'
+    span_name = params[:name] || 'js.event'
```
rshipp, line 51: I changed everything `javascript` to `js`, mostly because it makes this much nicer:

Screen Shot 2021-03-18 at 4 18 36 PM

Becomes:


Screen Shot 2021-03-18 at 4 18 15 PM

app/controllers/honeycomb_js_controller.rb
```diff
@@ -45,29 +48,32 @@ def send_span
...
+      span.add_field('app.user.first_name', current_user.first_name)
+      span.add_field('app.user.last_name', current_user.last_name)
 
       translate_boomerang_fields(span)
 
-      # Add the Boomerang sent data
-      params.each { |key, value| span.add_field(key, value) }
+      # Add the data sent by the JS. Includes both Boomerang and manual fields.
+      params.each do |key, value|
+        key = "#{BOOMERANG_FIELD_PREFIX}.#{key}" unless key.start_with? CUSTOM_FIELD_PREFIX
```
rshipp, line 74: Anything without "js.app" becomes "js.boomerang". Note this includes fields like `error`, `error_detail`, which were previously coming through in the root. This behavior is consistent with our manually-added `app.error` when we catch exceptions and log instead of re-raising, so I think this is ok?

app/controllers/honeycomb_js_controller.rb
```diff
@@ -88,7 +94,7 @@ def translate_boomerang_fields(span)
         span.add_field('response.status_code', status)
       end
 
-      span.add_field('javascript.timestamp', beacon_timestamp) if params['rt.end']
+      span.add_field('js.timestamp', beacon_timestamp) if params['rt.end']
```
rshipp, line 97: This is the only field that's not in `js.app` or `js.boomerang`. I'm assuming this is just to help with time correlation between frontend/server? If so, the name change shouldn't matter.

app/controllers/linked_in_authorization_controller.rb
```diff
@@ -31,7 +31,6 @@ def launch
     redirect_url = linked_in_auth_redirect_url
 
     Honeycomb.add_field('user.linked_in_state', state)
-    Honeycomb.add_field('redirect_url', redirect_url)
```
rshipp, line 34: `redirect_url` is always the same, so I deleted this

app/controllers/linked_in_authorization_controller.rb
```diff
@@ -47,9 +46,6 @@ def launch
   def oauth_redirect
     authorize :LinkedInAuthorization
 
-    Honeycomb.add_field('params.code', params[:code]) if params[:code]
-    Honeycomb.add_field('params.error', params[:error]) if params[:error]
-    Honeycomb.add_field('params.state', params[:state])
```
rshipp, line 52: these are duplicated in the request.query_string field.

app/javascript/packs/honeycomb.js
```diff
@@ -16,15 +18,15 @@ class HoneycombSpanBase {
      *                                   extra logic before (and after) executing it. All public methods 
      *                                   will be wrapped in this if it's specified.
      */
-    constructor(controller, action, fields = {}, prefixFieldsWithControllerName = true, functionWrapper = null) {
+    constructor(controller, action, fields = {}, prefixFields = true, functionWrapper = null) {
```
rshipp, line 21: no longer prefixing things with controller name, but I thought leaving this constructor param in might (??) come in handy at some point. The new behavior says whether to prefix with `js.app` or not.

app/javascript/packs/honeycomb.js
```diff
@@ -43,7 +45,7 @@ class HoneycombSpanBase {
         this.runAfterBoomerangLoaded(() => {
             // Standard fields that should be added to every span.
             // Don't prefix these. They're common to all "controllers" 
-            window.BOOMR.addVar('javascript.controller', controller, true);
+            window.BOOMR.addVar('controller', controller, true);
```
rshipp, line 48: Removed the prefix here since it would otherwise become `js.boomerang.javascript.controller`.

app/javascript/packs/honeycomb.js
```diff
@@ -43,7 +45,7 @@ class HoneycombSpanBase {
...
             window.BOOMR.addVar('name', this.name, true);
```
rshipp, line 49: note this `name` field becomes `js.boomerang.name` - but still makes it into the span name, since we handle that before prefixing in HoneycombJsController

app/javascript/packs/project_answers.js
```diff
@@ -2,10 +2,9 @@
 // For an example, see the view used for project_submissions_controller#edit
 import Rails from '@rails/ujs';
 import { HoneycombXhrSpan } from './honeycomb'
-import { HoneycombSpan } from './honeycomb'
 
 const AUTH_HEADER = 'LtiState '+ document.querySelector('meta[name="state"]').content;
-const HONEYCOMB_CONTROLLER_NAME = 'javascript.project.answer';
+const HONEYCOMB_CONTROLLER_NAME = 'project_answers';
```
rshipp, line 7: otherwise it becomes `js.javascript.project.answer`. the dot -> underscore change is just for consistency with our Rails span names.

app/jobs/cleanup_lti_launch_table_job.rb
```diff
@@ -21,18 +21,22 @@ def perform
...
+      records_to_delete = LtiLaunch.where("updated_at < ?", cutoff_time) 
       total_count = LtiLaunch.count
       count = records_to_delete.count
       records_to_delete.delete_all if count > 0
+
       finish_msg = "#{count} out #{total_count} LtiLaunch records deleted. Done cleaning up old LtiLaunches"
-      span.add_field('lti_launch_cleanup.finish_message', finish_msg)
       Rails.logger.info(finish_msg)
+      span.add_field('app.lti_launch_cleanup.records_to_delete.count', count)
+      span.add_field('app.lti_launch_cleanup.total_records.count', total_count)
```
rshipp, line 39: I rearranged these, added some whitespace, and changed the field names. The logic is unchanged.

app/jobs/grade_module_for_user_job.rb
```diff
@@ -32,7 +32,7 @@ def perform(user, canvas_course_id, canvas_assignment_id)
 
       grade = "#{ModuleGradeCalculator.compute_grade(user.id, canvas_assignment_id, assignment_overrides)}%"
 
-      span.add_field('grade', grade)
+      span.add_field('app.grade_module_for_user.grade', grade)
```
rshipp, line 35: This is an example of the "Things we don't want to get notifications about, but might help with debugging" I mention in the PR description. Since "app.grade" in one part of the code is likely totally different from "app.grade" in another part of the code, I'm using a local prefix.

app/jobs/sync_salesforce_program_to_lms_job.rb
```diff
@@ -11,7 +11,8 @@ def perform(program_id, email)
 
   rescue_from(StandardError) do |exception|
     Rails.logger.error(exception)
-    Honeycomb.add_field('error', exception)
+    Honeycomb.add_field('error', exception.class.name)
+    Honeycomb.add_field('error_detail', exception.message)
```
rshipp, line 15: Note these show up as `app.error`, NOT `error`. There's no way around this without patching the beeline or using libhoney, and I don't think it's worth that.

app/services/grade_modules.rb
```diff
@@ -24,13 +24,13 @@ def run
         return
       end
       canvas_course_ids = programs['records'].map { |r| r['Highlander_Accelerator_Course_ID__c'] }
-      span.add_field('canvas_course_ids', canvas_course_ids)
+      span.add_field('app.grade_modules.canvas_course_ids', canvas_course_ids)
```
rshipp, line 27: Another distinction with fields that only make sense in local context - "app.canvas.course.ids" would be confusing (what is `ids`?), so I'm treating this as "Things we don't want to get notifications about, but might help with debugging".

app/services/grade_modules.rb
```diff
@@ -24,13 +24,13 @@ def run
...
 
       # Eliminate courses with no module interactions, and exit early if that
       # leaves us with an empty list.
       courses = Course.where(canvas_course_id: canvas_course_ids)
         .filter { |c| Rise360ModuleInteraction.where(canvas_course_id: c.canvas_course_id).exists? }
-      span.add_field('courses.count', courses.count)
+      span.add_field('app.grade_modules.courses.count', courses.count)
```
rshipp, line 33: Again, `courses.count` means something totally different here than `courses.count` might mean in a different part of the code. It's not a field we'd want to query against if multiple controllers were sending it, it's meant only to add context to the current span for debugging. So a localized field name makes more sense.

app/services/register_user_account.rb
```diff
@@ -18,7 +18,7 @@ def initialize(sign_up_params)
 
   def run
     Honeycomb.start_span(name: 'RegisterUserAccount.run') do |span|
-      span.add_field('salesforce.contact.id', @salesforce_participant.contact_id)
+      span.add_field('app.salesforce.contact.id', @salesforce_participant.contact_id)
```
rshipp, line 21: This is not from `SalesforceAPI`, so let's call the field `app.salesforce` to be consistent with `app.canvas`. If we ever instrumented SalesforceAPI/CanvasAPI, we'd want to use just `salesforce.` and `canvas.` there.

app/services/register_user_account.rb
```diff
@@ -28,21 +28,21 @@ def run
       @new_user.update(@create_user_params)
       # Allow error handling on model validation when called from a controller.
       yield @new_user if block_given?
-      span.add_field('user.id.created', @new_user.id)
+      span.add_field('app.user.id', @new_user.id)
```
rshipp, line 31: The `.created` postfix didn't seem to add context here (every user has been created) and made the field contents less clear (it's not a boolean, it's an ID), so I renamed for consistency.

app/services/register_user_account.rb
```diff
@@ -28,21 +28,21 @@ def run
...
       # Create a user in Canvas.
       create_canvas_user!
-      span.add_field('canvas.user.id.created', @new_user.canvas_user_id)
+      span.add_field('app.canvas.user.id', @new_user.canvas_user_id)
 
       canvas_user_id_set = sf_client.set_canvas_user_id(
                                      @salesforce_participant.contact_id,
                                      @new_user.canvas_user_id)
-      span.add_field('salesforce.canvas.user.id.set', true)
+      span.add_field('app.register_user_account.salesforce_canvas_user_id_set', true)
```
rshipp, line 40: This one *is* a boolean though, and is debug info, so using localized field name and underscores instead of dots.

config/initializers/lti_authentication.rb
```diff
@@ -17,7 +17,6 @@ def valid?
 
     def authenticate!
       Honeycomb.start_span(name: 'LtiAuthentication.authenticate!') do |span|
-        span.add_field('lti_auth.state', @lti_state)
```
rshipp, line 20: 🔏

config/initializers/lti_authentication.rb
```diff
@@ -53,11 +52,12 @@ def authenticate!
...
-      span.add_field('status', status)
-      span.add_field('lti_auth.message', message)
-      span.add_field('url', url) if url 
-      span.add_field('canvas_id', canvas_id) if canvas_id
-      span.add_field('user_id', user.id) if user
+      span.add_field('app.lti_authentication.status', status&.to_s)
+      span.add_field('app.lti_authentication.message', message)
+      span.add_field('app.lti_authentication.url', url)
+      span.add_field('app.canvas.user.id', canvas_id)
+      span.add_field('app.user.id', user&.id)
```
rshipp, line 59: note localized fields (`app.lti_authentication.url`, because `app.url` here doesn't mean the same thing as `app.url` somewhere else) vs global fields (`app.user.id`).

lib/filter_logging.rb
```diff
@@ -117,17 +125,11 @@ def self.filter_honeycomb_data(fields)
...
-      end
-      if fields.has_key?('restiming')
-        fields['restiming'] = parameter_filter.filter_param('restiming', fields['restiming'])
-      end
-      if fields.has_key?('state')
-        fields['state'] = parameter_filter.filter_param('state', fields['state'])
+      FILTER_BOOMERANG_FIELDS.each do |field_name|
+        if fields.has_key?("#{BOOMERANG_FIELD_PREFIX}.#{field_name}")
+          fields["#{BOOMERANG_FIELD_PREFIX}.#{field_name}"] =
+            parameter_filter.filter_param("#{field_name}", fields["#{BOOMERANG_FIELD_PREFIX}.#{field_name}"])
```
rshipp, line 131: still passing in the field without a prefix as the first param to `filter_param`... I didn't look up wherever that function is, but this seems correct?

lib/lrs_xapi_mock.rb
```diff
@@ -28,10 +28,9 @@ class LrsXapiMockError < StandardError; end
   # Returns a hash with keys {:code, :body}, OR nil.
   def self.handle_request!(request, endpoint, user)
     Honeycomb.start_span(name: 'LrsXapiMock.handle_request!') do |span|
-      span.add_field('path', "/#{endpoint}")
-      span.add_field('user.id', user.id)
-      span.add_field('user.email', user.email)
-      span.add_field('method', request.method)
```
rshipp, line 34: method is already duplicated in `request.method`

lib/restclient_instrumentation.rb
```diff
@@ -17,22 +17,19 @@ def execute &block
...
         Rails.logger.error("{\"Error\":\"#{e.message}\"}")
         error_response = e.http_body
         Rails.logger.error(error_response)
-        span.add_field('restclient.error', e.message)
-        span.add_field('restclient.status', e.http_code)
-        # Note: can't use error_detail field name. It's overwritten somewhere with the exception
-        # message detail, not the actual response.
-        span.add_field('restclient.error_response', error_response)
+        span.add_field('restclient.response.status_code', e.http_code)
+        span.add_field('restclient.response.body', error_response)
```
rshipp, line 32: changed everything related to the request to be in `restclient.request` and everything related to the response to be in `restclient.response` - this is more consistent with autoinstrumented `request` and `response`. The note about "can't use error_detail" is no longer true, since we're not trying to put it in the root level anymore - but also, the `error_response` is really just the `response.body`, not an exception message.

lib/rise360_util.rb
```diff
@@ -40,8 +40,8 @@ def self.publish(zipfile)
 
       zipfile.open do |file|
         Honeycomb.start_span(name: 'Rise360Util.publish.zipfile') do |span|
-          span.add_field('file.path', file.path)
-          span.add_field('file.size', file.size)
+          span.add_field('app.zipfile.path', file.path)
+          span.add_field('app.zipfile.size', file.size)
```
rshipp, line 44: This one's a little strange, but whatever.

lib/rise360_util.rb
```diff
@@ -78,8 +78,9 @@ def self.update_metadata!(rise360_module, bucket = nil)
   rescue Exception => e
     Rails.logger.error("Could not update Rise360Module metadata: #{rise360_module}")
     Rails.logger.error("  Error: #{e}")
-    Honeycomb.add_field('error', e)
-    Honeycomb.add_field('rise360_module', rise360_module)
+    Honeycomb.add_field('error', e.class.name)
+    Honeycomb.add_field('error_detail', e.message)
+    Honeycomb.add_field('rise360_module.id', rise360_module.id)
```
rshipp, line 83: The beeline didn't know what to do with this object, so it wasn't showing up in HC as anything useful. Sending just the ID is nicer.

app/controllers/honeycomb_js_controller.rb
sadleb: I would make this `js.boomerang.timestamp`. It's just converting a boomerang field to something more readable.

app/controllers/honeycomb_js_controller.rb
sadleb: I think it's fine. `error` and `error_detail` are essentially a translation of the hard to read boomerang stuff into more readable errors. But ultimately, it's just a boomerang error and not something we're manually adding. Like you said, it's consistent with the automatic `app.error` stuff.

app/controllers/concerns/publishable.rb
```diff
@@ -127,6 +127,9 @@ def unpublish
     # We can't use `model_class` here because we don't have a PeerReview model
     authorize controller_path.classify.to_sym
 
+    Honeycomb.add_field('canvas.course.id', @course.canvas_course_id)
+    Honeycomb.add_field('canvas.assignment.id', canvas_assignment_id)
```
sadleb, line 131: Outside the scope of this PR, but it would be nice if this stuff was centralized so that we automatically have the canvas related info added to controllers everywhere it's available.

app/services/register_user_account.rb
sadleb: Makes sense.

lib/filter_logging.rb
sadleb: `filter_param` calls the above `@filter_parameters_lambda` -> if the params like `pgu` are coming through from JS with no prefix and we're adding the prefix on the Rails side, then I think this works properly. That's what it looks like you're doing.

sadleb (approved):
One small change request. but otherwise looks good. Can you drop the guidance in the PR description into a Wiki page too?

rshipp:
I was going to put it in the README, is the wiki better?

lib/filter_logging.rb
rshipp: At this point, the keys in `fields` have the prefix. does that mean the first param should have the prefix too?

sadleb: Yeah, in that case it needs the prefix

sadleb: I still think this is correct. You add the prefix when looking in fields, so that part will match, but then you call the lambda with no prefix which is what that logic is written to look for to determine if it should be filtered. I think you want to leave it like that bc that lambda is also responsible for filtering the non-prefixed value went sent from JS to the controller.

sadleb:
> I was going to put it in the README, is the wiki better?

I'll leave that up to you. The README is getting pretty long and unwieldy, Currently we have some high-level examples in the README but all the details of the setup are in the Wiki. You could just edit the high-level examples in the README to be up to date if you want. I'm leaning towards starting to refactor things out of the README into Wiki's so that the README serves as a Table Of Contents of sorts. If you go that route, then remove the obsolete README examples and move them to the Wiki. Totally up to you.

PR #636, opened 2021-03-22
Handle missing assignment ID in xAPI mock

rshipp:
### Description

* When an xAPI request doesn't have a valid assignment ID (from the LTI Launch payload), exit early with a 403 ("the request was understood but the server is refusing to process it") instead of raising an exception and returning a 500.
* Add more information the the Honeycomb span, to make future xAPI events more useful.
* Refactor out a confusing method from the LtiLaunch model.

Example new HC events: https://ui.honeycomb.io/braven/datasets/development/result/7Q6cq357MCL?tab=raw

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200076185491906/f

Since "Course Resources" attempts to send xAPI requests, but is not an assignment, there is no way to save Rise360ModuleInteractions. Instead of crashing and sending weird errors to Sentry/Honeycomb, we should just handle this gracefully with an appropriate error message.

Screen Shot 2021-03-22 at 12 50 08 PM

### Types of changes

### Checklist:


sadleb (approved):
Lgtm. I like that we have course and assignment ID easily accessible on LtiLaunch now. In a future iteration we should cleanup all the references that use the request_message to use these helpers instead

PR #641, opened 2021-03-23
Remove obsolete login code (Join server CAS, etc)

sadleb:
### Description

Task: https://app.asana.com/0/1174274412967132/1173451812413601

This is in preparation for a few enhancements to login:
* [Handle trying to login before confirming email](https://app.asana.com/0/1174274412967132/1199689504093092)
* [Handle trying to sign-up a second time or if you lost your confirmation email](https://app.asana.com/0/search/1200099946072945/1199575667690414)

**TODO**: when deploying, remove the `PUBLIC_SITE_DOMAIN` ENV var from prod config.

### Motivation and Context
This stuff originally came from the Join server and was meant to work with our customized Canvas code which we hosted on our own servers. We had some custom stuff in our old Canvas code where we added some code to tell the Join server to log itself out instead of using the built-in CAS "sign-out everywhere functionality" (which may or may not work). 

Also, in that world, the login form would be embedded inside an iframe sometimes where we didn't want the header and footer to show. 

Lastly, there was a config var that let you switch b/n auth modes, local auth vs auth against Join, so we could run the same code on different servers in different modes (e.g. we enabled it for Booster before we enabled it for Accelerator).

All that stuff is obsolete now so this rips it out.

### Types of changes

### Checklist:


app/controllers/cas_controller.rb
```diff
@@ -96,29 +96,7 @@ def login
...
-
-      @form_action = params['submitToURI'] || guessed_login_uri
-
-      if @form_action
-        return render :_login_form
-      else
-        return render :json => {:response => 'Could not guess the CAS login URI. Please supply a submitToURI parameter with your request.'}, status: :internal_server_error
-      end
-    else
-      render :login
```
sadleb, line 120: I think this was for logging in from the iframed Braven Network or Resume Bank. We use LTI extension to iframe stuff now.

app/views/cas/_login_form.erb
```diff
@@ -1,41 +1,24 @@
-<% if @log_out_of_services %>
```
sadleb, line 1: The diff is hard to see. Changes in this file are:
* rip out the hidden `iframe` 
* remove `@form_action` which was set by `onlyLoginForm` param above
* remove `ENV['BZ_AUTH_SERVER']` controlling alternative login links

config/rubycas.yml
```diff
@@ -4,12 +4,10 @@ default: &default
   maximum_unused_service_ticket_lifetime: 300
   maximum_session_lifetime: 172800
   maximum_remember_me_lifetime: 604800
-  infoline: Powered by <a href="http://rubycas.github.com">RubyCAS-Server-Core</a>
```
sadleb, line 7: This is not actually currently working and it's not required by the license. Just remove it. https://github.com/rubycas/rubycas-server/blob/master/LICENSE

rshipp (approved):

PR #643, opened 2021-03-23
Rate This Module now shows an indication that it was submitted.

sadleb:
### Description

This is a bit hacky. Flash notices don't work if we don't have session, which we often don't when shown inside an iframe inside
Canvas cross domain (esp in Safari or Chrome incognito). Instead, just pass the info about the flash notice through in query params and also switch the button text.

The reason I had to use two query params is that that the Close of the flash actually redirects back to edit again and we want to keep the Re-Submit text.

I didn't think it was worth dealing with showing the Re-Submit text on subsequent launches of the Module. It will pre-load the old response, but "Re-Submit" only shows when you're working on it inline.

# Test Plan:
Load a Module with an embedded "Rate This Module" iframe in Chrome Incognito. Make sure it shows the "Feedback submitted." alert when you submit, that the button changes to Re-Submit, that you can close the alert (and it still says Re-Submit), and that re-loading the Module puts it back in a fresh state, but populated with your previous submission

Viewing it on first launch:
image

When you hit submit:
image

When you close the alert:
image

### Types of changes

### Checklist:


rshipp (approved):
seems fine. this whole page is a hacky solution.

PR #649, opened 2021-03-25
More Honeycomb field normalization

sadleb:
### Description
When trying to use the new [Honeycomb field naming guidelines](https://github.com/bebraven/platform/wiki/Monitoring-&-Troubleshooting:-Honeycomb.io) I ran into some issues that still made it hard to query for things. This PR aims to solve for that. 

1. Before, we alway overwrote the `name` field for page load and unload beacons so we would lose the information we constructed the `HoneycombSpan` with. E.g. where in the code it came from. Since multiple HoneycombSpans can be in a single beacon and ultimately end up in a single span sent to Honeycomb, we need to stick with generalizing the `name` to the beacon type. We weren't doing that for all beacon types though, so it was inconsistent where to look for stuff. I expanded the names to include the two missing beacon types below and always normalize the `name` to one of these 4:
```
name = js.boomerang.page_load
name = js.boomerang.page_unload
name = js.boomerang.xhr
name = js.boomerang.manual
```

2. This lead to the question of where to put the information we use to construct the `HoneycombSpan`. This PR gets rid of the `name` concept from the JS and replaces it with a `context` concept. By default, all fields added are prefixed with the `context` so that it's easier to call `addField()` and have the fields come through consistently tied back to the `context`. We do this on a field by field basis instead of trying to add a separate field with the context b/c async stuff can come across in separate beacons so you have to go to the trace to get from one to the other. It's much easier just to query for the field you're looking for directly. See the example below.

3. I did a bunch of renaming to clear up the confusion above about how to use these HoneycombSpans, including renaming that one to `HoneycombAddToSpan`.

4. Some Rails magic somewhere was sending in an `action` and `controller` field. These were set to `honeycomb_js` and `send_span` respectively. They are not part of the JS payload, so filter that out, as well as `state` which is sent in the query param b/c we would filter it anyway so it's worthless to add an extra field to the Honeycomb schema.

5. This also adds the `app.user.id` to the `process_action.action_controller` span which has a `process_action.action_controller.exception` if Rails threw an exception. This makes querying for exceptions tied to users easier. See the screenshot below.

[Related change to Boomerang here](https://github.com/bebraven/boomerang/pull/3) so that uncaught exceptions in async functions get instrumented.

### Example
Given this JS
```
    const honeySpan = new HoneycombAddToSpan('form_assembly', 'changeESignatureMessage');
    try {
      let esigWarning = document.querySelector('fieldset#tfa_esignature-email-warning');
      if (esigWarning) {
        honeySpan.addField('esignature_message_found', 'true');
... do more stuff
      } else {
        honeySpan.addField('esignature_message_found', 'false');
      }
    } catch (err) {
      const errorMsg = 'Some error message.';
      honeySpan.addErrorDetails(error_msg, err);
    }
  }
```
Before this change what would end up in Honeycomb is the following:
```
name = js.page.load
js.boomerang.controller = honeycomb_js
js.boomerang.name = send_span
js.app.esignature_message_found = true
```
the `form_assembly` and `changeESignatureMessage` info used to construct the `HoneycombSpan` was lost.

After this change:
```
name = js.boomerang.page_load
js.form_assembly.changeESignatureMessage.esignature_message_found = true
... the following are NOT sent anymore
js.boomerang.controller = honeycomb_js
js.boomerang.action = send_span
js.boomerang.state = the_state
```
If an error is thrown, the following fields would come through
```
js.boomerang.error = "TheExceptionName"
js.boomerang.error_detail = "Some error message."
js.boomerang.error_filename = "form_assembly"
js.boomerang.error_function = "changeESignatureMessage"
```

If more `HoneycombAddToSpan`s are used as part of the page load, those additional prefixed fields will be added to the same span that is sent to Honeycomb.

Name field:
image

Querying for JS info:
image

Querying for a particular "context":
image

Querying for an exception for a particular user:
image

### Types of changes

### Checklist:


app/controllers/honeycomb_js_controller.rb
```diff
@@ -46,16 +46,11 @@ def send_span
     authorize :honeycomb_js
 
     existing_trace_to_add_to = request.headers[X_HONEYCOMB_TRACE_HEADER]
-    raise ActionController::BadRequest, "Missing '#{X_HONEYCOMB_TRACE_HEADER}' header" unless existing_trace_to_add_to 
+    raise ActionController::BadRequest, "Missing '#{X_HONEYCOMB_TRACE_HEADER}' header" unless existing_trace_to_add_to
 
-    span_name = params[:name] || 'js.event'
+    span_name = params[:name] || BOOMERANG_FIELD_PREFIX
```
sadleb, line 51: If no manual instrumentation has happened, there will be no name, so just name it `js.boomerang` since that's the prefix for all boomerang stuff

app/controllers/honeycomb_js_controller.rb
```diff
@@ -82,30 +80,41 @@ def send_span
...
+      span.add_field("#{BOOMERANG_FIELD_PREFIX}.beacon_type", 'page_load')
+    elsif is_page_unload_beacon?
+      span.add_field("#{BOOMERANG_FIELD_PREFIX}.beacon_type", 'page_unload')
+    elsif is_xhr_beacon?
+      span.add_field("#{BOOMERANG_FIELD_PREFIX}.beacon_type", 'xhr')
+      method = params['http.method'] || 'GET' # Annoyingly, 'http.method' may not be set for GET. If missing on XHR beacon, it's GET.
+      span.add_field("#{BOOMERANG_FIELD_PREFIX}.request.method", method)
+      status = params['http.errno'] || '200'
+      span.add_field("#{BOOMERANG_FIELD_PREFIX}.response.status_code", status)
+    end
```
sadleb, line 99: The change here other than moving stuff around is just to add a `js.boomerang.beacon_type` field.

sadleb:
TODO: crap. I just remembered why I originally set the name to page load or unload and not what the `HoneycombSpan` name is. These aren’t really spans. They are a way to use the “span” pattern that we’re used to in the Rails side, but if you have multiple of these as part of loading the page, the name will end up being set to whatever the last one run is. The actual honeycomb span is the page load / unload beacon. 

The XHR one is different bc each of those does correspond to a honeycomb span. Need to think about this a bit more. I’m thinking I should rename these to `HoneycombInstrumentation` or something, put back the logic to always name the page load / unload spans consistently, and make the “name” here be used to set a `js.app.[the_filename].[the_function_name]` field with a value of true so that all of the Instrumentation objects we create just end up adding a field indicating that there are other fields related to that particular category of manual instrumentation. 

Turning them into proper spans will blow up the scope significantly.

app/javascript/packs/honeycomb.js
```diff
@@ -43,33 +42,34 @@ class HoneycombSpanBase {
...
+            // Fields that the caller passed into the constructor
             this.addFields(fields);
 
             this.startTime = window.BOOMR.now();
         });
     }
 
     /*
-     * Adds a field to the current span.
+     * Adds a field to the current span. Note that field names are prefixed by the "fileName"
```
sadleb, line 53: TODO update comment. Prefixed with context now not filename.

app/javascript/packs/honeycomb.js
```diff
@@ -96,49 +96,53 @@ class HoneycombSpanBase {
                  error: e.name,
                  error_log: log_msg,
                  error_detail: e.toString(),
-                 error_action: this.action,
+                 error_filename: this.fileName,
+                 error_function: this.functionName,
```
sadleb, line 100: I used file and function name bc those are greppable whereas context is not

app/javascript/packs/honeycomb.js
```diff
@@ -43,33 +42,34 @@ class HoneycombSpanBase {
         }
 
         this.runAfterBoomerangLoaded(() => {
-            // Standard fields that should be added to every span.
-            // Don't prefix these. They're common to all "controllers" 
-            window.BOOMR.addVar('controller', controller, true);
-            window.BOOMR.addVar('name', this.name, true);
```
sadleb, line 49: This is replaced by prefixing field names with context.

app/controllers/honeycomb_js_controller.rb
```diff
@@ -82,30 +94,41 @@ def send_span
...
 
+  # See https://developer.akamai.com/tools/boomerang/docs/BOOMR.plugins.AutoXHR.html
+  # These are Fetch/XHR calls for some AJAX'y thing.
+  def is_xhr_beacon?
+    params['http.initiator'] == 'xhr'
+  end
+
+  # See: https://akamai.github.io/boomerang/tutorial-howto-measure-arbitrary-events.html
+  def is_manual_beacon?
+    params.key?('h.pg') || params.key?('xhr.pg')
```
sadleb, line 129: I didn’t see this in the documentation, but in practice `xhr.pg` comes through. I looked in the src code and it’s part of calling `responseEnd` which is what causes the manual timer beacon to go out

app/controllers/honeycomb_js_controller.rb
```diff
@@ -46,17 +46,26 @@ def send_span
...
+    elsif is_page_load_beacon?
+      span_name = "#{BOOMERANG_FIELD_PREFIX}.page_load"
     elsif is_page_unload_beacon?
-      span_name = 'js.page.unload'
+      span_name = "#{BOOMERANG_FIELD_PREFIX}.page_unload"
       duration = 0 # the duration passed is actually for the page.load, so 0 it out.
+    elsif is_manual_beacon?
+      span_name = "#{BOOMERANG_FIELD_PREFIX}.manual"
+    else
+      span_name = BOOMERANG_FIELD_PREFIX
```
sadleb, line 68: This shouldn’t happen, but just in case.

app/javascript/packs/honeycomb.js
```diff
@@ -96,49 +96,53 @@ class HoneycombSpanBase {
...
 
     /*
      * When measuring arbitrary events, call this to force a beacon to go out with
-     * any fields you added and the h.pg field set to the name of this span with the
+     * any fields you added and the h.pg field set to the context of this span with the
      * time it took to run as the value.
      * See here for more detail: https://akamai.github.io/boomerang/tutorial-howto-measure-arbitrary-events.html
      */
-    sendTimerBeacon() {
+    sendNow() {
```
sadleb, line 114: Renamed bc you need deeper knowledge of boomerang to know that sending a timer beacon is how you send manual beacons where you just want to instrument and measure something.

sadleb:
This is hard to review bc of all the renaming, so I called out the places you should look at in comments.

sadleb:
> TODO: crap. I just remembered why I originally set the name to page load or unload and not what the `HoneycombSpan` name is. These aren’t really spans. They are a way to use the “span” pattern that we’re used to in the Rails side, but if you have multiple of these as part of loading the page, the name will end up being set to whatever the last one run is. The actual honeycomb span is the page load / unload beacon.
> 
> The XHR one is different bc each of those does correspond to a honeycomb span. Need to think about this a bit more. I’m thinking I should rename these to `HoneycombInstrumentation` or something, put back the logic to always name the page load / unload spans consistently, and make the “name” here be used to set a `js.app.[the_filename].[the_function_name]` field with a value of true so that all of the Instrumentation objects we create just end up adding a field indicating that there are other fields related to that particular category of manual instrumentation.
> 
> Turning them into proper spans will blow up the scope significantly.

This is resolved with the follow-up commits.

rshipp (approved):

app/javascript/packs/honeycomb.js
sadleb: Done

PR #650, opened 2021-03-27
Redirect users to Canvas when they login from the users/sign_up flow

sadleb:
### Description

Bug found when testing this task: https://app.asana.com/0/1174274412967132/1200028118308870

We added a new flow where if they already have an account, we ask them
to login. It was just saying "User has successfully logged in" and staying
on the login page though. This makes that login redirect to Canvas like
we normally do.

### Test Plan:
- Signup a fresh new user using /users/sign_up?u=blah and make sure it still works.
- Fully logout /cas/logout and then try hitting the same /users/sign_up?u=blah endpoint
- It should say you already have an account, login instead. Do that.
- It should redirect to Canvas instead of staying on the platform login page.


sadleb:
Foregoing a code review. There is nothing to review here.

PR #652, opened 2021-03-29
Update LinkedIn API authorization consent language.

sadleb:
### Description

Tasks:
- https://app.asana.com/0/1184939531452067/1199981673602534
- https://app.asana.com/0/1174274412967132/1200040143264135

There was one final iteration of the langauge in the contract that we missed.
This fixes that. It now says "...profile data for three years."

**Release Task** Go into the Course Template and update the language there. This PR only impacts adding the LinkedIn button in the Content Editor, but we need to fix up any already out there (just the latest version in the Course Template should be enough).


sadleb:
Foregoing a code review. This is just contract language stuff.

sadleb:
After deploying, I updated the language in the template: 
image

PR #658, opened 2021-04-01
Cleanup Honeycomb instrumentation and Sentry monitoring for our ActiveJobs

sadleb:
### Description
- Configure auto-instrumentation for all `ActiveSupport::Notification`'s
  that `ActiveJob` emits
- Automatically wrap all of our jobs in a span with the name of the job.
  This is particularly nice b/c uncaught exceptions end up in the root
  'error' and 'error_detail' field this way, magically.
- Get rid of the `rescue_from` usage. This was suppressing exceptions from
  going to Sentry even if you re-raise

### Motivation and Context
Trying to make [troubleshooting Sync To LMS](https://app.asana.com/0/1174274412967132/1200093973021815) issues easier.

These are the relevant fields for [a single trace ](https://ui.honeycomb.io/braven/datasets/stagingplatform.bebraven.org/result/63deBsEro4i/trace/DuKB7j81zNP) for a job that resulted in an uncaught "Test Exception 5"
image

The above trace in waterfall view:
image

The above exception flowing through to Sentry:
image
image


app/jobs/cleanup_lti_launch_table_job.rb
```diff
@@ -17,27 +17,25 @@ class CleanupLtiLaunchTableJob < ApplicationJob
   queue_as :low_priority
 
   def perform
-    Honeycomb.start_span(name: 'CleanupLtiLaunchTableJob.perform') do |span|
```
sadleb, line 20: Turn on "ignore whitespace" when reviewing. Removed the outer spans since they are done automatically now which led to a ton of whitespace changes.

app/jobs/cleanup_lti_launch_table_job.rb
```diff
@@ -17,27 +17,25 @@ class CleanupLtiLaunchTableJob < ApplicationJob
...
-      # The value of 'lti_launch_remember_for' can be a Rails Time helper like 2.weeks or the 
-      # number of seconds, like 1209600 (this is also 2 weeks)
-      cutoff_time = Time.at(Time.now - eval(Rails.application.secrets.lti_launch_remember_for))
-      span.add_field('app.lti_launch_cleanup.current_server_time', Time.now)
-      span.add_field('app.lti_launch_cleanup.cutoff_time', cutoff_time)
+    # The value of 'lti_launch_remember_for' can be a Rails Time helper like 2.weeks or the
+    # number of seconds, like 1209600 (this is also 2 weeks)
+    cutoff_time = Time.at(Time.now - eval(Rails.application.secrets.lti_launch_remember_for))
+    Honeycomb.add_field('lti_launch_cleanup.current_server_time', Time.now)
```
sadleb, line 23: Note that the rename from `app.blah` to just `blah` is b/c we switched from `span.add_field()` to `Honeycomb.add_field()` which adds the app prefex.

app/jobs/sync_salesforce_program_to_lms_job.rb
```diff
@@ -7,12 +7,10 @@ class SyncSalesforceProgramToLmsJob < ApplicationJob
   def perform(program_id, email)
     SyncPortalEnrollmentsForProgram.new(salesforce_program_id: program_id).run
     SyncSalesforceToLmsMailer.with(email: email).success_email.deliver_now
-  end
-
-  rescue_from(StandardError) do |exception|
+  rescue => exception
     Rails.logger.error(exception)
-    Honeycomb.add_field('error', exception.class.name)
-    Honeycomb.add_field('error_detail', exception.message)
```
sadleb, line 15: Don't need these anymore b/c the automatic span that wraps all active jobs now sets them when uncaught exceptions bubble up.

rshipp (approved):
lgtm

sadleb:
AH HA!! @rshipp so I confirmed it elsewhere that if you let an exception bubble up through a span, then it automatically captures and sets the `error` and `error_detail` on the span. We might want to revisit our guidance and have folks do this:
```
begin
  Honeycomb.start_span(...) do |span|
    ... code that fails
  end
rescue => e
  .. handle exception
end
```
instead of 
```
Honeycomb.start_span(...) do |span|
  begin
    ... code that fails
  rescue => e
    .. handle exception
    end
end
```

PR #664, opened 2021-04-05
Modify account creation flow

rshipp:
### Description

1. Create local User records during the SyncPortalEnrollmentsForProgram, OR when the user hits `sign_up`.
2. When the user hits `sign_up`, use the existing record OR create one. Still create Canvas accounts at this time.
3. If a user hits Password Reset with a valid reset token, BEFORE hitting sign_up, they are redirected to sign_up instead.
4. If a registered user hits Password Reset, it acts the same as it did before.
4. The sign_up flow is the **only** place that sends confirmation emails, and users are unable to sign in until they have confirmed.

Supporting changes:

* "Unregistered" users - those who have not completed the sign_up flow - will not be able to sign in.
* If an unregistered user (who somehow magically has a password set) attempts to log in with the correct credentials, the rejection will act the same as if they did not have the correct credentials. This prevents people from simply typing in an email and being redirected to the sign_up page.
* If a registered user attempts to hit the sign_up page, they will be sent to the login page instead. This prevents people from hitting sign_up with a valid salesforce ID and being able to set the password.

### Motivation and Context

When a user tries to log in but hasn't hit `sign_up`, they should still be able to use either the `sign_up` flow OR the reset-password flow to initialize their account.

Task: https://app.asana.com/0/1174274412967132/1200095789801975/f

Also: 
* https://app.asana.com/0/1174274412967132/1200153244154412/f
* https://app.asana.com/0/1174274412967132/1200153244154414/f

### Types of changes

### Checklist:


app/controllers/users/passwords_controller.rb
```diff
@@ -19,9 +19,13 @@ class Users::PasswordsController < Devise::PasswordsController
   # end
 
   # PUT /resource/password
-  # def update
-  #   super
-  # end
+  def update
+    super do |user|
+      if user.errors.empty? && !user.registered?
+        user.update(registered_at: DateTime.now.utc)
```
rshipp, line 25: need to have L25 call RegisterUserAccount instead, so it can create the canvas account.

app/services/register_user_account.rb
```diff
@@ -20,15 +20,17 @@ def run
...
       span.add_field('app.salesforce.contact.id', @salesforce_participant.contact_id)
 
-      # Need to have the User record saved before the Canvas sync runs since it relies on it.
-      # Essentially do an upsert
-      @new_user = User.find_or_create_by(
+      # Local user record was already created during Sync from Salesforce.
+      # If we can't find it, something's wrong; raise immediately.
+      @new_user = User.find_by!(
         salesforce_id: @create_user_params[:salesforce_id]
       )
```
sadleb, line 27: Sync to LMS runs nightly. A lot of times in the mad-dash during launch, folks are being recruited last minute (like they show up at the first event b/c they heard about Braven). I think we need to still create their local user record if it doesn't exist. Is there any downside to that with this new flow (which I'm still wrapping my head around)?

sadleb: We used to rely on Sync To LMS and staff would run it over and over again manually during this mad dash to get everyone access to the Portal

sadleb: L48 in the registrations controller would have to change. I think it's fine to allow enumeration of the sign-up ID if we use the more secure (longer, random) token we were talking about instead of the SF ID.

rshipp: Hmm, we'd have to think it through and make sure that wouldn't hide any errors. I think it'd be ok?

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -9,7 +9,10 @@ def initialize(portal_user:, salesforce_participant:, salesforce_program:)
     @portal_user = portal_user
     @sf_participant = salesforce_participant
     @sf_program = salesforce_program
-    @user = User.find_by!(email: sf_participant.email)
+    # NOTE: This can fail if there are duplicate Contacts with the same
+    # email on Salesforce. TODO: Do we care?
+    @user = User.create_with(salesforce_contact_params)
+      .find_or_create_by!(salesforce_id: @sf_participant.contact_id)
```
sadleb, line 15: We don't care if it's actually caused by duplicate email/Contact's in SF b/c that should never happen and if it does we should fix it with a SF validation. But, if a contact enrolls in one semester and then comes back to enroll in a future semester (e.g. they dropped the course the first time or they are a returning LC), will this fail too? In other words, is there anything in `salesforce_contact_params` that is different for different participants? Or does it all come from the contact and should be the same across semesters?

sadleb: Whoops, just saw the params below. Looks like it should be the same across semesters but leaving the above comment as something to consider overall with this new flow.

rshipp: If someone enrolls in multiple semesters, will they have the same Contact ID across all of them? If so, we should be good. This only fails if there are 2+ contact IDs tied to a single email.

app/controllers/users/passwords_controller.rb
sadleb: I'm still having trouble understanding how the confirmation email is generated. It doesn't look like anything in `RegisterUserAccount` does that. I assume it's something about the `registrations_controller` and the super class taking care of it on create. We'll need to make sure and generate the confirmation email ourselves if the creation happens during this password reset flow which is really a first time sign-up/registration flow.

rshipp: The confirmation email is generated by Devise when you create a new `User` record. (One of the modified tests in this PR checks that too.) We should definitely test that as part of QA though.

spec/services/register_user_account_spec.rb
```diff
@@ -44,31 +45,34 @@
...
+      expect(Devise.mailer.deliveries.count).to eq 1
+      expect(user.confirmation_sent_at).not_to be(nil)
 
-    it 'sends confirmation instructions to user' do
       Devise.mailer.deliveries.clear()
       RegisterUserAccount.new(sign_up_params).run
-      expect(Devise.mailer.deliveries.count).to eq 1
+      expect(Devise.mailer.deliveries.count).to eq 0
       user = User.find_by_canvas_user_id!(canvas_user['id'])
-      expect(user.confirmation_sent_at).not_to be(nil)
```
rshipp, line 71: asserts that the confirmation email is sent at appropriate times

app/controllers/users/passwords_controller.rb
sadleb: Huh. That doesn’t happen from the Rails console, right? Or maybe it does. We probably want to delay that for this new Sync To LMS flow until the time that they register though. Otherwise we’ll be sending random confirmation emails to folks who aren’t taking any action on their own

app/models/user.rb
```diff
@@ -110,4 +115,16 @@ def self.search(query)
...
+
+  # Allow empty passowrds for non-registered users.
+  # This lets us create accounts with empty passwords in the
+  # Sync flow, when login is still impossible, but prevents
+  # empty passwords once the user goes through the sign_up
+  # flow and sets their password for the first time.
+  def password_required?
+    return false unless registered?
+    super
+  end
```
sadleb, line 128: To gain more confidence that empty passwords won't cause folks to be allows to login with them, we may want to add checks for `registered?` in the `cas_controller` where it checks `if unconfirmed_user`

rshipp: Can't do that, it opens up some reset-password bypasses. I'd like to get rid of the existing unconfirmed_user check, actually. We do this instead: https://github.com/bebraven/platform/pull/664/files#diff-a00c1f30078cf84b4726f23eb925078de64732490b22b1377380ee05822e32f0

app/services/sync_portal_enrollment_for_account.rb
sadleb: The contact id should stay the same

app/models/user.rb
```diff
@@ -50,11 +50,16 @@ def student_section_by_course(course)
     sections_with_role(RoleConstants::STUDENT_ENROLLMENT).find_by(course: course)
   end
 
-  # True if the user has confirmed their account and can login.  
+  # True if the user has confirmed their account and can log in.
```
sadleb, line 53: 🤣

spec/services/register_user_account_spec.rb
sadleb: I think we need to add a similar test in `sync_portal_enrollment_for_account.rb ` where we're adding a new `User.create_with` call.

app/controllers/users/passwords_controller.rb
rshipp: both issues from this thread are now addressed.

app/services/register_user_account.rb
rshipp: this is now addressed.

spec/services/register_user_account_spec.rb
rshipp: this is now addressed

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -9,7 +9,7 @@ def initialize(portal_user:, salesforce_participant:, salesforce_program:)
     @portal_user = portal_user
     @sf_participant = salesforce_participant
     @sf_program = salesforce_program
-    @user = User.find_by!(email: sf_participant.email)
+    @user = User.find_by!(salesforce_id: sf_participant.contact_id)
```
sadleb, line 12: This is unrelated to the rest of this PR, right? You're just doing it b/c we have more confidence in `salesforce_id` being accurate than we do for `email` b/c it can change? If so, I'm wondering if we should actually use `canvas_user_id` instead b/c of the whole 15 vs 18 character mess with the SF IDs (which we need to fix, ideally with a database validation). If there is a reason why we want to use SF ID instead of Canvas User ID, I think we should add a database index for it b/c currently we expect you to look up users by ID or Canvas ID

app/controllers/users/passwords_controller.rb
```diff
@@ -14,9 +14,26 @@ class Users::PasswordsController < Devise::PasswordsController
...
+
+    # From https://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Recoverable/ClassMethods#reset_password_by_token-instance_method
+    original_token = params[:reset_password_token]
+    reset_password_token = Devise.token_generator.digest(User, :reset_password_token, original_token)
+
+    user = User.find_by(reset_password_token: reset_password_token)
+    if user.present? && !user.registered?
+      # Since we've verified the token, and we know the user has not done
+      # the sign_up flow yet, redirect them there to set their initial
+      # password, create their Canvas account, etc instead.
```
sadleb, line 29: Should we add a Honeycomb field here so we can analyze how often this is happening?

app/services/register_user_account.rb
```diff
@@ -20,16 +20,34 @@ def run
...
+        @new_user = User.new(
+          salesforce_id: @create_user_params[:salesforce_id]
+        )
+        # Don't send confirmation email yet; we do it explicitly below.
+        @new_user.skip_confirmation_notification!
+      end
+
+      @new_user.update(@create_user_params.merge({
+        registered_at: DateTime.now.utc,
+      }))
```
sadleb, line 41: The line below sounds like it's doing some handling related to not successfully setting your password. Are we waiting until after they've set a valid password to mark them registered?

app/services/register_user_account.rb
```diff
@@ -20,16 +20,34 @@ def run
...
       # Allow error handling on model validation when called from a controller.
       yield @new_user if block_given?
+
+      # Add field after user record is created.
       span.add_field('app.user.id', @new_user.id)
 
+      # Send confirmation email.
+      @new_user.send_confirmation_instructions
```
sadleb, line 49: I spent a few minutes thinking about whether this should come before marking them as registered, so just for future readers `registered` just means that you've successfully created a password for the first time and your `User` record is now saved. This concept was introduced b/c we actually create the user record in advance so we needed a way to gate keep the concept of setting the password and thus being officially "registered" with an account. In the normal Devise flow, that happens before confirmation. Confirmation is just the final step, to let you login as a security measure, but other than that your account is fully valid.

app/controllers/users/passwords_controller.rb
```diff
@@ -14,9 +14,26 @@ class Users::PasswordsController < Devise::PasswordsController
...
+      # TODO: Redirect by a token instead of SF ID.
+      # https://app.asana.com/0/1174274412967132/1200147504835146/f
+      return redirect_to new_user_registration_path(u: user.salesforce_id)
```
sadleb, line 32: Maybe I'm blind, but are we missing a spec for this? Seems worth writing one if so. This is pretty much the core behavior this PR is meant to solve for.

app/services/register_user_account.rb
rshipp: I think the root question you're asking is "Will we ever fail to set a password, but mark someone registered?", to which the answer is "no". If you want to understand the error handling here, it's the block in registrations controller `#create`.

app/services/sync_portal_enrollment_for_account.rb
rshipp: I think this change is standalone, yes. There are cases where we have a local account but no Canvas account, but there will never be a local account without a Salesforce Contact.

app/controllers/users/passwords_controller.rb
rshipp: That information is already available in the response status code and response location header.

app/controllers/users/passwords_controller.rb
rshipp: There are no specs for the Devise controllers, or most of the Sync / Register logic. If you want them, let's merge and QA this first, and then spend time writing tests once we're sure the behavior is correct. There are already so many open tasks for bugs I found in this flow.

sadleb (approved):

rshipp:
Merged with the understanding that we'll immediately prioritize tests after this has been QA'd.

PR #674, opened 2021-04-08
Preserve user-specific assignment overrides when updating enrollment

rshipp:
### Description

During Sync to LMS, if a user enrollment changes:
* Fetch all existing assignment overrides for the course
* [Continue with enrollment change]
  * [Canvas will have deleted any user-specific overrides for this user]
* Filter overrides to only those that match this user
* If there are matching overrides, re-create them on Canvas

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200061986057145/f

### Checklist:


app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -99,17 +99,36 @@ def drop_course_enrollment(canvas_course_id)
...
     section_name = section_name.blank? ? SectionConstants::DEFAULT_SECTION : section_name
     section = find_or_create_section(canvas_course_id, section_name)
+
     enrollment = find_user_enrollment(canvas_course_id)
     if enrollment.nil?
       enroll_user(canvas_course_id, role, section, limit_privileges_to_course_section)
     elsif !enrollment.section_id.eql?(section.canvas_section_id) || !enrollment.type.eql?(role)
       # Section or role has changed.
+      # Fetch assignment overrides, so we can replace ones that get deleted in the next step.
+      assignment_overrides = canvas_client.get_assignment_overrides_for_course(
```
sadleb, line 111: I wish there was a way to do this more efficiently. I think we should wrap this in a honeycomb span (and anywhere else we call this API endpoint), or take a moment to add fields that make it easy to segment our canvas API calls by method since all rest client calls are already wrapped in spans. Maybe we can already do that, but it feels worth double checking since we’re now calling this on every user that moves cohorts.

PR #675, opened 2021-04-09
Add Project autosave indicators

rshipp:
### Description

Add indicators for the autosave functionality in Projects.

 
### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1199612014163177/f

**Note:** the final version of "All progress saved." is shown in the "...fades away over a 5 second duration" screenshot. The padding on top and bottom is the same as the failed ones so it's more prominent since it fades away.

#### Status message when your answer saves (desktop)
image

#### Fades away over a 5 second duration
This shows the status after it's started fading away (note that the style in this screenshot isn't the final style above, but you get the idea).

image

#### Failed to save (desktop)
image

#### Some answers still unsaved
Happens if you have multiple and instead of clicking Retry you go in an input and change the value causing a retry through the normal path. The instructions about what to look for is b/c the project is long and it can be hard to find if you scroll away.

image

#### Submit button disabled if in unsaved situation
image

### Types of changes

### Checklist:


app/assets/stylesheets/custom_contents.scss
```diff
@@ -9,6 +9,8 @@
 .bv-custom-content-container {
 
   padding: 1.4rem;
+
+  // Leave space so the autosave status bar doesn't overlap the submit button.
   padding-bottom: 3rem;
```
sadleb, line 14: This will effect impact surveys too. Prob fine may may want to double check and narrow the target to only projects?

rshipp: We can if we need to, but this is pretty minor. Only increases bottom padding by 1.6rem.

sadleb: Oh, padding bottom on the last thing.. duh, that's fine. I misread and thought this would add padding between the button and what was above it.

app/javascript/packs/project_answers.js
```diff
@@ -87,13 +92,43 @@ export async function main() {
...
+            statusBar.classList.add('autosave-status-success');
+        } else {
+            statusBar.classList.remove('autosave-status-success');
+            statusBar.classList.add('autosave-status-error');
+        }
     }
 
     function sendAnswer(e) {
         const input = e.target;
+        if (input.value == input.initialValue) return;
```
sadleb, line 125: See the [comments on this commit](https://github.com/bebraven/platform/pull/675/commits/7398d2771b85e0aa17e71d32e9fcec101d684a1e) for why I took the time to do this.

app/javascript/packs/project_answers.js
```diff
@@ -122,9 +157,34 @@ export async function main() {
...
+            // User feedback.
+            input.classList.remove('autosave-input-error');
+            inputAlert.innerHTML = "";
+            setTimeout(function(){
+                if (getUnsavedInputs().length === 0) {
+                    changeAutoSaveStatus('All progress saved.', true);
+                } else {
+                    changeAutoSaveStatus('Some answers are still unsaved!', false);
+                }
+            }, 200); // Wait a split second to change status back so the 'Saving...' message is actually visible
```
sadleb, line 169: I tried to do this with CSS animations/transitions but it wasn't worth the effort b/c display and text properties don't seem to be animatable. Using `setTimeout` seems fine to me?

app/javascript/packs/project_answers.js
```diff
@@ -122,9 +157,34 @@ export async function main() {
...
         })
         .catch((error) => {
+            // User feedback.
+            input.classList.add('autosave-input-error');
+            inputAlert.innerHTML = "Failed to save answer. <a href='#'>Retry?</a>";
+            inputAlert.querySelector('a').onclick = () => { 
+              // clicking the Retry on any of the failed answers retries them all
+              let allUnsavedInputs = getUnsavedInputs();
+              allUnsavedInputs.forEach((input) => {
+                  sendAnswer({ target: input });
```
sadleb, line 182: This felt a little weird faking an event object by just passing a plain ol' javascript object with a target attribute. Does this seem like the right "javascripty" way to do this? I didn't think it was worth adding code inside the `sendAnswer` function to check the type of what was passed in.

app/javascript/packs/project_answers.js
```diff
@@ -87,13 +92,37 @@ export async function main() {
...
-    function sendAnswer(e) {
+    function changeAutoSaveStatus(text, cssClass) {
+        const statusBar = document.getElementById('autosave-status-bar');
+        statusBar.textContent = text;
+        statusBar.classList = cssClass;
+    }
+
+    function sendAnswer(e, forceSend = false) {
         const input = e.target;
+        if (!forceSend && input.value == input.initialValue) return;
```
sadleb, line 119: See the [comments on this commit](https://github.com/bebraven/platform/pull/675/commits/7398d2771b85e0aa17e71d32e9fcec101d684a1e) for why I took the time to do this.

app/javascript/packs/project_answers.js
```diff
@@ -122,9 +151,34 @@ export async function main() {
...
+            // User feedback.
+            input.classList.remove('autosave-input-error');
+            inputAlert.innerHTML = "";
+            setTimeout(function(){
+                if (getUnsavedInputs().length === 0) {
+                    changeAutoSaveStatus('All progress saved.', 'autosave-status-success');
+                } else {
+                    changeAutoSaveStatus('Some answers are still unsaved!', 'autosave-status-error');
+                }
+            }, 400); // Wait a split second to change status back so the 'Saving...' message is actually visible
```
sadleb, line 163: I tried to do this with CSS animations/transitions but it wasn't worth the effort b/c display and text properties don't seem to be animatable. Using setTimeout seems fine to me?

app/javascript/packs/project_answers.js
```diff
@@ -122,9 +151,34 @@ export async function main() {
...
         })
         .catch((error) => {
+            // User feedback.
+            input.classList.add('autosave-input-error');
+            inputAlert.innerHTML = "Failed to save answer. <a href='#'>Retry?</a>";
+            inputAlert.querySelector('a').onclick = () => { 
+                // clicking the Retry on any of the failed answers retries them all
+                let allUnsavedInputs = getUnsavedInputs();
+                allUnsavedInputs.forEach((input) => {
+                    sendAnswer({ target: input }, true);
```
sadleb, line 176: This felt a little weird faking an event object by just passing a plain ol' javascript object with a target attribute. Does this seem like the right "javascripty" way to do this? I didn't think it was worth adding code inside the sendAnswer function to check the type of what was passed in.

app/javascript/packs/project_answers.js
rshipp: Doesn't this just make it say "Saving..." for longer? I thought you wanted to fade out the "All progress saved." box? The only reason we're getting the super-quick saving box is because we're testing locally, I'm not sure we want it to act even slower when we're using a real internet connection.

app/javascript/packs/project_answers.js
rshipp: Since we already have a reference to `input`, I think this would be better as 
```suggestion
                    input.dispatchEvent(new Event('blur'));
```

app/javascript/packs/project_answers.js
rshipp: Instead of this parameter, I think it'd be cleaner to set a `data-last-saved` (or some other name) attribute on the input after a successful save. Then when this function is called, compare `input.value` with `input.dataset.lastSaved`, and if they match, exit early. That way you don't have to pass in the extra `true` directly to this function in the Retry handler below, and you can rely on the JS event system.

app/javascript/packs/project_answers.js
```diff
@@ -122,9 +151,34 @@ export async function main() {
...
             honeySpan.addField('response.status', response.status, false);
         })
         .catch((error) => {
+            // User feedback.
+            input.classList.add('autosave-input-error');
+            inputAlert.innerHTML = "Failed to save answer. <a href='#'>Retry?</a>";
+            inputAlert.querySelector('a').onclick = () => { 
+                // clicking the Retry on any of the failed answers retries them all
+                let allUnsavedInputs = getUnsavedInputs();
+                allUnsavedInputs.forEach((input) => {
```
rshipp, line 175: Can simplify to
```suggestion
                getUnsavedInputs().forEach((input) => {
```
since we're not re-using this variable anywhere.

app/assets/stylesheets/custom_contents.scss
```diff
@@ -66,7 +69,59 @@
...
+      padding: 0.25rem;
+  }
+
+  div#autosave-status-bar.autosave-status-saving {
+      display: block;
+      background-color: $bv-silver-300;
+      border: 1px solid $bv-silver-500;
+      color: $bv-oceanblue-800;
+      font-size: 1.2rem;
+      font-weight: 700;
```
rshipp, line 111: this larger font threw me off, I don't think we need to draw extra attention to the saving indicator.

app/assets/stylesheets/custom_contents.scss
```diff
@@ -66,7 +69,59 @@
...
+      padding: 0.5rem;
+  }
+
+  div#autosave-status-bar.autosave-status-error {
+      display: block;
+      background-color: $bv-gold-200;
+      border: 1px solid $bv-red;
+      color: $bv-red-800;
+      font-weight: 700;
```
rshipp, line 120: can we use `bold` instead of `700`? it's easier to read, and I'm not sure numeric weights even work with the fonts we're using here, I think it might just fall back to whichever is closer `regular` or `bold`.

app/javascript/packs/project_answers.js
rshipp: Actually, can we rename `input` too? It's technically valid, but confusing to override the outer scope with a local scope.

app/javascript/packs/project_answers.js
sadleb: If I hit the actual prod server, with my internet connection it's 150-220ms. That's still too fast of a flash to actually know what the flash is. So this sets what felt like the smallest amount of time it needed to flash for someone to be able to see what it was.

As for fading out the "All progress saved" status bar, that was the original goal. I started by trying to just make it look less intrusive and once I did that it felt fine to always have it there. But then since it's so unobtrusive if you use a biggish screen and its out of your scope of vision (or at least focus), I don't think folks would even notice it. That's why I wanted to call attention to the "Saving..." status by making it bigger and stick around long enough to see -> so your eye would notice it.

I do think leaving the "Saving..." style the same as "All progress saved" and fading it out will probably be just as good (or better?)

Maybe try this branch out locally (or we can hop on a call and I'll demo) to make a decision on whether I should try and do the latter option instead of what is implemented here?

app/assets/stylesheets/custom_contents.scss
sadleb: See my above comments about why I did this.

sadleb:
TODO: double check drop downs.

sadleb:
> TODO: double check drop downs.

They look good.

app/javascript/packs/project_answers.js
sadleb: We decided to switch the behavior to a fade out of the "All progress saved" bar. This makes it so that we don't have to purposefully make the 'Saving...' status more prominent and last longer. That's distracting for the user. Instead, just let that happen super fast (and screenreaders won't usually even have to read it) and fade out the "All progress saved" status bar which has unobtrusive styling, but now with the animation provides enough visual cues to catch your eye.

app/javascript/packs/project_answers.js
```diff
@@ -87,13 +93,46 @@ export async function main() {
     }
 
     function attachInputListeners() {
-        getAllInputs().forEach(input => { input.onblur = sendAnswer });
+        getAllInputs().forEach(input => {
+          input.onfocus = (e) => {e.target.initialValue = e.target.value};
```
sadleb, line 97: TODO: remove this line. From the old code

sadleb (approved):

PR #679, opened 2021-04-20
Change API access tokens from 16 to 255 characters.

sadleb:
### Description
16 chars is not secure, esp for a key we don't rotate frequently (or at all)

Task: https://app.asana.com/0/1174274412967132/1198943782529944

[Follow-up PR here for far more security](https://github.com/bebraven/platform/pull/680)


app/models/access_token.rb
```diff
@@ -1,26 +1,20 @@
-require 'digest/md5'
-
 class AccessToken < ApplicationRecord
-  VALID_KEY = /\A[0-9a-f]{16}\z/i
-  
+
```
sadleb, line 2: Ignore this. It was left over code from when I was trying to get this work and I remove it in a follow up PR

rshipp (approved):

PR #680, opened 2021-04-20
Tie AccessTokens to a user that owns it so we can write policies for them

sadleb:
### Description
Enhance `AccessToken` to have a user associated with each token as the "owner." When authenticating with a token set the `current_user` to the token owner in Devise. Now policies can do user / role based authorization for tokens instead of letting them have access to the entire app. Implemented this for the `SalesforceController` which now allows API access to its endpoints.

Task: https://app.asana.com/0/1174274412967132/1198943782529944

### Motivation and Context
We wanted this feature to be secure since this is the first time we are creating and using access tokens: https://app.asana.com/0/1174274412967132/1200179852436577

#### Creating a new access token
image

image

### Types of changes

### Checklist:


app/controllers/access_tokens_controller.rb
```diff
@@ -68,6 +72,11 @@ def destroy
...
     # Never trust parameters from the scary internet, only allow the white list through.
     def access_token_params
-      params.require(:access_token).permit(:name)
+      params.require(:access_token).permit(:name, :email)
+    end
+
+    def set_owner
+      owner_email = @access_token&.user&.email || access_token_params[:email]&.strip&.downcase
+      @owner = User.find_by!(email: owner_email)
     end
```
sadleb, line 81: I thought it was easier to allow creating the `AccessToken` => user association in the UI with an email instead of the `User.id` but that makes this code a bit messy. Not sure it's worth cleaning up since we don't really need a UI for this, but since we had one already I put the minimum amount of work in necessary to be able to create these tokens.

app/controllers/application_controller.rb
```diff
@@ -20,16 +20,20 @@ class ApplicationController < ActionController::Base
...
     return false unless request.format.symbol == :json
 
     key = params[:access_key] || request.headers['Access-Key']
     return false if key.nil?
     
-    !!AccessToken.find_by(key: key)
+    access_token = AccessToken.find_by(key: key)
+    return false unless access_token
+
+    sign_in(:user, access_token.user)
```
sadleb, line 35: This is what sets `current_user`

spec/controllers/salesforce_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
+require 'rails_helper'
+
+RSpec.describe SalesforceController, type: :controller do
```
sadleb, line 3: I started trying to create an `ApplicationController` spec for this access token stuff, but it seems like that's not generally the way testing common things like this is done. There are no actions to target there. As we start adding more API access, we should create some `scenario`s in a shared file that every controller can use to test API access without duplicating the below.

app/controllers/access_tokens_controller.rb
rshipp: a users dropdown seems easier to use than a text email field, but I agree it's not worth spending much time on

rshipp (approved):
nothing blocking. we should try to write policy specs for new policies, though I think I've missed some of those too.

app/controllers/access_tokens_controller.rb
sadleb: It would need to be searchable. We already have 100 users in prod and we haven't even launched yet. But yeah, not planning on doing anything else here for now.

PR #685, opened 2021-04-23
Sign up by token, not Salesforce ID

rshipp:
### Description

Changes signup flows to rely on a token instead of the Salesforce ID. Removes all references to the `u` param, replaced by one of `signup_token`, `uuid`, or nothing at all, as appropriate.

Supporting changes:
* Try to fix all the security issues where we're exposing user information unnecessarily
* Only show "re-send confirmation" prompt to users who get their password correct
* Enforce 18-character, unique salesforce IDs (but still allow nil)
* Always cross-reference users by Salesforce ID, never anything else
* Let (legitimate) users sign up through the password reset flow
* Invalidate password-reset tokens after registration, and signup tokens after password reset
* Globally replace "sign_up_email" with "signup_email" to match "signup_token" and "signup_date"
* Migrate the signup email code to use the new token
* Remove the feature that redirected already-registered users to a special login page with their email prefilled
* If signup link is invalid (expired or never existed), redirect to login, but don't expose any information
* Fully remove the ability to create local accounts via signup (all accounts must be created by Sync or the Admin views)

### Motivation and Context

Main task: https://app.asana.com/0/0/1200147504835146/f 

Other tasks: 
* https://app.asana.com/0/0/1200192173211414/f 
* https://app.asana.com/0/0/1200153244154416/f
* https://app.asana.com/0/1174274412967132/1200192173211414/f
* https://app.asana.com/0/1174274412967132/1200153244154416/f

### Types of changes

### Checklist:


app/controllers/cas_controller.rb
```diff
@@ -175,16 +166,28 @@ def loginpost
...
-      else
+      elsif auth.valid_password?(@password)
+        # The username/password is correct, but something else is making
+        # this account currently invalid for login. Be careful here.
         user = User.find_by_email(@username)
-        unconfirmed_user = user && user.confirmed? == false
-        if unconfirmed_user
-          logger.warn("Unconfirmed user tried to login: '#{@username}'")
-          redirect_to users_registration_path(u: user.salesforce_id, login_attempt: true) and return
+        if user.present? && user.registered? && !user.confirmed?
```
sadleb, line 173: See [this comment](https://github.com/bebraven/platform/pull/696/files#diff-1ecc6b6eed42222136d409895b0d6482a5e86e28199a09b63583fd43858a91b1R24-R30) on the "expire confirmation tokens" PR about a fix we need to do here too. If you're changing someone email, a new `confirmation_token` is generated that could expire. If that happens, we still want to let them log in with either their old or new email to be able to regenerate a new confirmation email. If we can't find a user by email, we need to look for them by `unconfirmed_email` too. If found, they are also an unconfirmed user.

sadleb: The scenario is that we just changed their email and sent a confirmation link. They have the `unconfirmed_email` column set to the new email but the email column is still the old one. If the confirmation token expires, they will be redirected to the login page with no info about why other than something is wrong. They were just told to use the new email to log in with, so they try that. The above tries to look up the login POST using the email field which is the old email, so it fails to log them in. Since we only redirect to the page to generate a new confirmation token / link if they successfully log in, they just get an “invalid username / password” message. We need to look them up by `unconfirmed_email` as well to determine if we can send them to the page to generate a new confirmation token / link.

app/controllers/users/registrations_controller.rb
```diff
@@ -8,19 +8,24 @@ class Users::RegistrationsController < Devise::RegistrationsController
...
+    # UUID without checking it, so we don't give away whether this
+    # UUID is tied to any user.
+    @uuid = uuid_param
   end
 
   # GET /resource/sign_up
   def new
     super do
-      return render :bad_link unless salesforce_id
+      return render :bad_link if signup_token.blank?
```
sadleb, line 21: Should this also be checking for the password reset token to determine if we should short circuit out?

app/controllers/users/registrations_controller.rb
```diff
@@ -8,19 +8,24 @@ class Users::RegistrationsController < Devise::RegistrationsController
...
 
-      user = find_user_by_salesforce_id
+      # Only select the user by one of the two token options, because this
+      # route allows setting a password for the account, and we need it
+      # to be secure.
```
sadleb, line 25: Is it worth mentioning that during this redirect, we don't want to expose anything about the account the token is tied to? Similar to the comment below about `but still don't reveal anything about which account was tied to the token.`

app/controllers/users/registrations_controller.rb
```diff
@@ -30,30 +35,46 @@ def new
...
+          return render :new
+        end
       end
+    else
+      # If the token was invalid, show the error page.
+      # This applies to already-used tokens, as well as tokens
+      # that never existed to begin with.
+      # Note that this means a bad-actor can check for valid tokens
+      # by simply POSTing to this action.
+      return render :bad_link
```
sadleb, line 72: Should this just redirect to the login page with a vague message about "either already being registered or using a bad link" similar to what we do for the confirmations controller if we encounter a bad token? This would be a nicer experience for a real user (non-bad-actor) if they try to use the sign_up link a 2nd time and reveal no information about the state of the account.

app/models/user.rb
```diff
@@ -139,6 +154,34 @@ def self.search(query)
...
+  # Used during the sign_up / registration process to securely
+  # identify users. Modeled after and uses Devise methods, but not
+  # actually a part of Devise.
+  # https://github.com/heartcombo/devise/blob/5d5636f03ac19e8188d99c044d4b5e90124313af/lib/devise/models/recoverable.rb
+  def set_signup_token
+    raw, enc = Devise.token_generator.generate(self.class, :signup_token)
+
+    self.signup_token = enc
+    self.signup_token_sent_at = DateTime.now.utc
+    save(validate: false)
```
sadleb, line 166: Are we calling `save` vs `save!` for some reason? Do we want failures to be silent or raise?

app/controllers/users/registrations_controller.rb
```diff
@@ -30,30 +35,46 @@ def new
...
-    # TODO: Use a token instead of SF ID.
-    # https://app.asana.com/0/1174274412967132/1200147504835146
-    user = User.find_by(salesforce_id: params[:user][:salesforce_id])
-
-    # Prevent using this page to reset arbitrary other users passwords.
-    # You can only do this sign_up flow once.
+    # Only select the user by one of the two token options, because this
+    # route allows setting a password for the account, and we need it
+    # to be secure.
+    user = find_user_by_signup_token || find_user_by_reset_password_token
```
sadleb, line 41: Maybe make `find_user_by_tokens` in `services/register_user_account` a class method and call that from here to share code?

rshipp: These functions are 1-line leftovers from a previous implementation where they were more complicated. I left them in because I thought they were easier to read this way. If you'd rather, I can take them out.

app/controllers/users/registrations_controller.rb
rshipp: Added a comment just in case, but the login page also no longer accepts any type of user reference.

app/controllers/users/registrations_controller.rb
rshipp: This is the `bad_link` page, for reference. It already shows a message and a link to log in. Are you suggesting we remove this page entirely? I don't love the `notice` stuff we do on the login page, letting anyone write whatever text they want directly on our login page seems like a bad idea.

Screen Shot 2021-05-03 at 11 14 33 AM

app/models/user.rb
rshipp: This function is copied from Devise https://github.com/heartcombo/devise/blob/5d5636f03ac19e8188d99c044d4b5e90124313af/lib/devise/models/recoverable.rb. I believe they use `save(validate: false)` because they want to be sure the token saves regardless of whether the rest of the record has some validation error.

I think it's technically possible for this `save` to fail if the database connection itself fails, or if we generate the same token twice (statistically unlikely). If it does fail, we'd send out signup links with an empty token, and fellows will be shown the `bad_link` view. If this raised an exception instead, it'd be handled by the Sync error handling flow and show up in the diagnostic email.

app/controllers/cas_controller.rb
rshipp: Checking both emails seems bad. Checking an unconfirmed email seems to defeat the purpose of confirmation. Where are they told to log in with the new email before it has been confirmed? That seems like the root cause of the issue you're describing. It also doesn't seem that bad if they just have to try the old email to get the re-send confirmation button.

sadleb: Devise implemented reconfirmation using the `unconfirmed_email` column. If that column is set, it just means that email is in an unconfirmed state, but the old email is still valid. Also, reconfirmation uses the same email as confirmation. So, by changing someones email with reconfirmable enabled, unless we want to dig into how to override the out-of-the-box behavior of sharing the same confirmation email for both the initial registration and for changing emails, then we wouldn't be able to let them know what email they should use to login with in the confirmation email. That was another earlier user story to make it more explicit what they should log in with.

In other words, if they try with the old email, they won't get a reconfirmation page b/c that email is confirmated. Only the unconfirmed email is in an unconfirmed state requiring reconfirmation (which clears that column and makes the email column set to the new email).

app/controllers/users/registrations_controller.rb
sadleb: I suppose this is fine for now and we can revisit if folks continue to use the signup email link more than once and reach out for support.

I agree that the notice stuff in the login page isn't great. Feels like a bigger project to change who we load the login page with a message without using a query param though.

app/controllers/users/registrations_controller.rb
sadleb: I don't mind leaving them in b/c I agree they are more readable. I was just thinking that we always want to check both so consolidating the check into a single method which we already do over in `register_user_account` seems easier to maintain in case we ever add any more fancy logic and forget to do it in all the places it's used.

app/controllers/cas_controller.rb
sadleb: In other other words, unconfirmed_email is conceptually the same as confirmed_at. It’s a column that tells whether they are confirmed.

app/controllers/users/registrations_controller.rb
rshipp: The RegisterUserAccount method is slightly different than this one, since it doesn't have to handle the nested `user[token]` case. I think I'd rather leave this as `signup || reset` rather than a single method anyway, it seems clearer to explicitly check both.

app/models/user.rb
sadleb: I think we should have it fail, show up in diagnostics, and not send the email. It should be really rare if it happens for any of the above reasons and if it turns out not to be rare, we should revisit our options.

app/controllers/users/registrations_controller.rb
sadleb: Kk, works for me

app/controllers/cas_controller.rb
rshipp: Addressed in #698

sadleb: > Addressed in #698

I don't think this was actually addressed there. See here for where I make this flow work: https://github.com/bebraven/platform/pull/727

PR #695, opened 2021-04-27
Reconcile mismatched emails across Salesforce, Platform, and Canvas.

sadleb:
### Description
Salesforce is the source of truth for what the login email should be. If it doesn't match what is in Platform and Canvas, we update it. In order for the updated email to be active and able to login, a reconfirmation email is sent out. This also serves to notify the user since staff could change their login email without their knowledge.

A big change here is that instead of updating the Canvas login email at the time of sync'ing, we wait until they actually reconfirm it by clicking the confirmation link in the email they get. Their old account will continue to work until they activate the new one. This was required so that we didn't continue to spam them with reconfirmation emails if they don't confirm the change. We just do it once and if they don't activate that email we skip trying to sync it.

* Implements [this task](https://app.asana.com/0/1174274412967132/1200197329627692)
* Refactors [this task's behavior](https://app.asana.com/0/1174274412967132/1200179852436577)
* Fixes [this bug](https://app.asana.com/0/1174274412967132/1200249851554781)
* Side effect is to implement [this user story](https://app.asana.com/0/1174274412967132/1200247977762042)

**TODO**: [follow-up PR to write specs](https://app.asana.com/0/1174274412967132/1200249851554798)

### Motivation and Context
* When updating an account's email address in Salesforce, it could fail (e.g. Canvas is temporarily down). The nightly sync should fix that if possible.
* A staff member could accidentally manually change an email in Platform or Canvas and cause them to be out of sync. This will recover from that.
* A user may not want to use the new account so this allows them to not accept the change. Ideally we'd see in the traces that someone refuses to accept the email change and reconcile things manually to use their preferred email

Trying to use a bad confirmation token (either missing or already used or expired) will always just send you here with this message.
image

Some honeycomb info about confirmation token errors:
image

### Types of changes

### Checklist:


app/controllers/users/confirmations_controller.rb
```diff
@@ -22,18 +22,32 @@ def create
   # GET /resource/confirmation?confirmation_token=abcdef
   def show
     super do
-      Rails.logger.debug("Account creation confirmed for #{resource}. Signing them in using CAS SSO.")
-      redirect_path_for_user = sign_in_and_get_redirect_path
-      set_flash_message!(:notice, :confirmed)
-      redirect_to redirect_path_for_user and return
+      if resource.errors.empty?
+        ConfirmUserAccount.new(resource).run!
```
sadleb, line 26: TODO: if this canvas stuff fails, the confirmation token is consumed, and then can’t successfully activate their account. Can we rollback the consumption of the token?

app/controllers/salesforce_controller.rb
```diff
@@ -57,61 +57,38 @@ def update_contacts
...
       Rails.logger.error(e2.message)
+      user_id = @user&.id || '<Unknown>' # we could have failed to find the user
       SyncSalesforceContactToCanvasMailer.with(
         staff_email: params[:staff_email],
         first_name: @first_name,
         last_name: @last_name,
         new_email: @new_email,
-        user_id: @user.id,
+        user_id: user_id,
         canvas_user_id: @canvas_user_id,
```
sadleb, line 92: Unrelated to this PR. Fixes one small issue with [As a Staff member, I would like to be able to change which email a Fellow uses to log in](https://app.asana.com/0/1174274412967132/1200179852436577) user story.

app/controllers/users/confirmations_controller.rb
```diff
@@ -21,19 +21,63 @@ def create
...
+        redirect_to redirect_path_for_user and return
+      else
+        # This is most likely for a bad token or a token that's already been consumed.
+        #
+        # Note that Devise does error handling if resource.errors has errors. Just instrumenting it,
+        # but not raising for that reason.
+        Rails.logger.error("Confirmation error for #{resource.inspect}. Errors = #{resource.errors.full_messages}")
+        Honeycomb.add_field('user.email', resource.email)
+        Honeycomb.add_field('error', resource.errors.class.name)
+        Honeycomb.add_field('error_detail', resource.errors.full_messages)
```
sadleb, line 53: This is the bug fix for [the confirmation link can be used to log in even after the initial confirmation is complete](https://app.asana.com/0/1174274412967132/1200249851554781/f)

app/services/sync_portal_enrollments_for_program.rb
```diff
@@ -84,16 +80,6 @@ def run
...
-  def email_inconsistent?(portal_user, participant)
-    !participant.email.casecmp(portal_user.email).zero?
-  end
-
-  def reconcile_email!(portal_user, participant)
-    ReconcileUserEmail.new(salesforce_participant: participant,
-                           portal_user: portal_user)
-                      .run
-  end
-
```
sadleb, line 96: Moved to the account level sync and implemented there.

lib/canvas_api.rb
```diff
@@ -172,22 +172,23 @@ def get_logins(user_id)
...
   # used to send Notifications. If you call this, you almost certainly want to
   # fuss with their communication channels using the methods below.
-  def change_user_login_email(user_id, new_email, account_id=DefaultAccountID)
-    logins = get_logins(user_id)
-    update_login(logins.first['id'], new_email, account_id)
+  def update_login(login_id, new_email, account_id=DefaultAccountID)
+    response = put("/accounts/#{account_id}/logins/#{login_id}", {
+      'login[unique_id]': new_email,
+    })
+    JSON.parse(response.body)
```
sadleb, line 191: We needed to check the login in order to determine if we needed to change it as part of the nightly sync / reconciliation, so this is refactored to allow consumers to do both steps (get and update).

app/controllers/users/confirmations_controller.rb
```diff
@@ -21,19 +21,67 @@ def create
...
+      # if they use valid credentials, then they'll get a message about the problem (e.g.
+      # they have to resend a confirmation email b/c the link expired).
+      Rails.logger.error("Confirmation error for #{resource.inspect}. Errors = #{resource.errors.full_messages}")
+      Honeycomb.add_field('user.email', resource.email)
+      Honeycomb.add_field('error', resource.errors.class.name)
+      Honeycomb.add_field('error_detail', resource.errors.full_messages)
+      redirect_to cas_login_path(
+          service: CanvasAPI.client.canvas_url,
+          notice: 'Your email is either already confirmed or that link was invalid. Please log in.'
+        ) and return
```
sadleb, line 59: Note that currently the `confirmation_token` doesn't expire, but we'll [implement that in a follow-up](https://app.asana.com/0/1174274412967132/1200249851554804) PR.

app/services/sync_from_salesforce_contact.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  # The Salesforce Contact is the source of truth for their email (and name). This service
+  # updates Platform to match if the email changes.
+  #
+  # Note that also update their name if we have to change the email, but leave it alone on Canvas
+  # for now. They can change it there if they want. We update locally so that our comms are
+  # correct if we made a mistake and change it later.
+  def run!
+    if @user.unconfirmed_email == @salesforce_contact.email.downcase
```
sadleb, line 21: TODO: downcast this

app/controllers/salesforce_controller.rb
```diff
@@ -57,61 +57,38 @@ def update_contacts
...
-      @user = User.find_by!(canvas_user_id: @canvas_user_id)
-
       @new_email = contact[:Email]
       # Note: we don't send these to Canvas. Rely on users do update their own names.
       @first_name = contact[:FirstName]
       @last_name = contact[:LastName]
       @salesforce_contact_id = contact[:Id]
 
+      # Raise if not found so we don't change anything in Canvas.
+      @user = User.find_by!(canvas_user_id: @canvas_user_id)
```
rshipp, line 67: I know this was old code, but do you know why we're checking by Canvas ID here and not Salesforce ID?

app/controllers/users/confirmations_controller.rb
```diff
@@ -21,19 +21,67 @@ def create
...
+    # which is what super does to look up the resource (aka user)
+    user_before_confirmation = User.find_first_by_auth_conditions(confirmation_token: params[:confirmation_token])
+    rollback_params = {
+      email: user_before_confirmation.email,
+      unconfirmed_email: user_before_confirmation.unconfirmed_email,
+      confirmation_token: user_before_confirmation.confirmation_token,
+      confirmed_at: user_before_confirmation.confirmed_at
+    } if user_before_confirmation
+
+    self.resource = resource_class.confirm_by_token(params[:confirmation_token])
```
rshipp, line 35: Since this action never renders a view, I think we should not set this `self.resource` reference. Setting this reference when we didn't need to lead to a number of information-disclosure issues in other Devise paths, that I've been having to strip out. It also makes this code easier to read if we just call the variable `user` instead of `self.resource`.

app/controllers/users/confirmations_controller.rb
```diff
@@ -21,19 +21,67 @@ def create
...
+    # staff member to trigger it. Note: don't do a database transaction to handle the
+    # rollback b/c this is a network call that can take time.
+    Honeycomb.add_field('alert.confirmations_controller.confirmation_failed', true)
+    Rails.logger.error(e.message)
+    if resource && rollback_params
+      resource.skip_confirmation_notification!
+      resource.update(rollback_params)
+      # Changing the email will regenerate the token and I couldn't figure out the Devise
+      # methods to call to skip that part too, so I'm just explicitly setting it back
+      resource.update(confirmation_token: rollback_params[:confirmation_token])
```
rshipp, line 74: i think it's `skip_reconfirmation!` https://github.com/heartcombo/devise/blob/5d5636f03ac19e8188d99c044d4b5e90124313af/lib/devise/models/confirmable.rb#L165

app/services/sync_from_salesforce_contact.rb
```diff
@@ -0,0 +1,56 @@
...
+    @user = user
+    @salesforce_contact = salesforce_contact
+  end
+
+  # The Salesforce Contact is the source of truth for their email (and name). This service
+  # updates Platform to match if the email changes.
+  #
+  # Note that also update their name if we have to change the email, but leave it alone on Canvas
+  # for now. They can change it there if they want. We update locally so that our comms are
+  # correct if we made a mistake and change it later.
```
rshipp, line 19: don't know how much it matters, but this comment doesn't match the behavior below. we only update names if the email changed, otherwise we exit early and never call the code that would update the name.

app/services/sync_from_salesforce_contact.rb
```diff
@@ -0,0 +1,56 @@
...
+      Honeycomb.add_field('sync_from_salesforce_contact.skipping', true)
+      Honeycomb.add_field('sync_from_salesforce_contact.skipping_reason', 'Email change already requested and reconfirmation email sent.')
+      return
+    elsif @user.email.downcase == @salesforce_contact.email.downcase
+      Honeycomb.add_field('sync_from_salesforce_contact.skipping', true)
+      Honeycomb.add_field('sync_from_salesforce_contact.skipping_reason', 'No email change.')
+      return
+    end
+
+    Honeycomb.add_field('sync_from_salesforce_contact.email_changed', true)
```
rshipp, line 35: could we rename this field to match `skipping` above? (or rename above to `email_changed: false`?) I think it's easier to have one field that can be true/false, rather than two fields that are only ever `true` and we have to check for `exists`.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -209,6 +212,37 @@ def find_or_create_section(canvas_course_id, section_name)
...
+           salesforce_participant.last_name)
+      ).run!
+    end
+  end
+
+  def email_inconsistent?(user_email, canvas_email, salesforce_email)
+    Honeycomb.add_field('user.email', user_email)
+    Honeycomb.add_field('canvas.user.email', canvas_email)
+    Honeycomb.add_field('salesforce.contact.email', salesforce_email)
+    emails_inconsistent = !salesforce_email.casecmp?(user_email) || !salesforce_email.casecmp?(canvas_email)
```
rshipp, line 238: any reason to use `casecmp?` here but `downcase ==` elsewhere?

app/controllers/salesforce_controller.rb
sadleb: I think I was just afraid of 15 vs 18 char mismatches with SF ID until we put in database validations and canvas id is equally unique. The SF side has a unique constraint and so does the platform side. The only issue I see using it is if we’re in code where the Cavas user doesn’t exist yet.

app/services/sync_portal_enrollment_for_account.rb
sadleb: Oh, that was from the old code. I think downcase is more readable so I’ll switch this to use that to be consistent

app/controllers/salesforce_controller.rb
rshipp: actually I can make this change in #685 too if you're fine with that, since it's outside the scope of this PR anyway.

sadleb: Yeah, let’s do this change in that PR. I agree that keeping the consistency is worthwhile!

app/controllers/users/confirmations_controller.rb
sadleb: Yup, that worked. Thanks!

PR #697, opened 2021-05-04
Don't overwrite manually entered grades

sadleb:
### Description
We need the ability for Teaching Assistants (TAs) or staff to manually give a Module grade under extenuating circumstances. This PR implements that ability. It works by detecting if a grade as been entered in Canvas by any user other than the user that API calls happen under.

Task: https://app.asana.com/0/1174274412967132/1198943782529860

### Screenshots
image

### Types of changes

### Checklist:


app/jobs/grade_module_for_user_job.rb
```diff
@@ -23,23 +23,25 @@ def perform(user, canvas_course_id, canvas_assignment_id)
     # haven't actually processed yet, causing students to get missing or incorrect grades.
     max_id = Rise360ModuleInteraction.maximum(:id)
 
-    # Fetch assignment overrides.
-    assignment_overrides = CanvasAPI.client.get_assignment_overrides(
-      canvas_course_id,
-      canvas_assignment_id
-    )
+    unless GradeModules.grading_disabled_for?(canvas_course_id, canvas_assignment_id, user)
```
sadleb, line 26: Note that I like `grading_enabled_for?` better but didn't take the time to refactor. Worth it?

I originally called this `GradeModules.grade_manually_overriden?()` but renamed it b/c I ran into the scenario where if they never opened the module then they don't have a `Rise360ModuleGrade` and we want that to also short-circuit any auto-grading. That's why it's `disabled` instead of `enabled` in the first place.

app/jobs/grade_module_for_user_job.rb
```diff
@@ -23,23 +23,25 @@ def perform(user, canvas_course_id, canvas_assignment_id)
...
 
-    grade = "#{ModuleGradeCalculator.compute_grade(user.id, canvas_assignment_id, assignment_overrides)}%"
+      assignment_overrides = CanvasAPI.client.get_assignment_overrides(
+        canvas_course_id,
+        canvas_assignment_id
+      )
 
-    Honeycomb.add_field('grade_module_for_user.grade', grade)
```
sadleb, line 34: Moved inside `compute_grade`

app/services/grade_modules.rb
```diff
@@ -114,34 +116,53 @@ def grade_assignment(canvas_assignment_id, user_ids)
...
-          new: true,
-        )
-        # Note since we only call exists?, the slow `select *` query implied above never actually runs.
-        if !interactions.exists? && !due_date.nil? && Time.parse(due_date) > Time.now.utc
-          Rails.logger.info("Skip user_id = #{user_id}, canvas_assignment_id = #{canvas_assignment_id}; " \
-              "no interactions and assignment isn't due yet")
-          next
+        Honeycomb.start_span(name: 'grade_modules.grade_user') do
+          Honeycomb.add_field('canvas.assignment.id', canvas_assignment_id)
+          Honeycomb.add_field('user.id', user_id)
```
sadleb, line 121: Doing these before calling `User.find(user_id)` in case that throws, we'll have these in the span.

app/services/grade_modules.rb
```diff
@@ -114,34 +116,53 @@ def grade_assignment(canvas_assignment_id, user_ids)
...
+          Honeycomb.add_field('canvas.user.id', user.canvas_user_id)
+
+          interactions = Rise360ModuleInteraction.where(
+            user: user,
+            canvas_assignment_id: canvas_assignment_id,
+            new: true,
+          )
+          Honeycomb.add_field('grade_modules.new_interactions.count', interactions.count)
+
+          next if GradeModules.grading_disabled_for?(course.canvas_course_id, canvas_assignment_id, user)
```
sadleb, line 132: This is the meat of this PR. It just short-circuits all auto-grading logic if they never opened the module or if we detect a user other than the current Canvas API user entered a grade.

lib/canvas_api.rb
```diff
@@ -62,6 +62,14 @@ def delete(path, body={}, headers={})
       url: "#{@api_url}#{path}", payload: body, headers: @global_headers.merge(headers))
   end
 
+  def api_user_id
+    @api_user_id ||= begin
+      response = get('/users/self')
+      api_user = JSON.parse(response.body)
+      api_user['id']
+    end
+  end
```
sadleb, line 71: Since we don't expect this to change, only make the call once and cache it (until the server restarts)

lib/module_grade_calculator.rb
```diff
@@ -55,25 +62,29 @@ def self.compute_grade(user_id, canvas_assignment_id, assignment_overrides)
...
         # If there are no mastery questions, fold the mastery weight in with the engagement
         # weight, so that engagement is just worth more.
-        (
+        grade = (
           (grade_weights[:module_engagement] + grade_weights[:mastery_quiz]) * engagement_grade +
           grade_weights[:on_time] * on_time_grade
         )
       end
+      Honeycomb.add_field('module_grade_calculator.grade', grade)
+      grade
```
sadleb, line 87: No changes in this file. Just adding Honeycomb instrumentation.

lib/module_grade_calculator.rb
```diff
@@ -31,10 +31,14 @@ def self.compute_grade(user_id, canvas_assignment_id, assignment_overrides)
         canvas_assignment_id: canvas_assignment_id,
         user_id: user_id,
       )
+      Honeycomb.add_field('module_grade_calculator.interactions.count', interactions.count)
```
sadleb, line 34: No changes in this file. Just adding Honeycomb instrumentation.

app/services/grade_modules.rb
```diff
@@ -167,4 +188,34 @@ def grade_assignment(canvas_assignment_id, user_ids)
       ).where('id <= ?', max_id).update_all(new: false)
     end
   end
+
+  def self.grading_disabled_for?(canvas_course_id, canvas_assignment_id, user)
+    rise360_module_grade = CourseRise360ModuleVersion.find_by!(canvas_assignment_id: canvas_assignment_id)
+      .rise360_module_grades.find_by(user: user)
+
+    # If they've never opened the module a Rise360ModuleGrade won't exist yet.
+    # All auto-grading logic is disabled circuited until they do.
```
rshipp, line 197:
```suggestion
    # All auto-grading logic is short-circuited until they do.
```

app/services/grade_modules.rb
```diff
@@ -167,4 +188,34 @@ def grade_assignment(canvas_assignment_id, user_ids)
       ).where('id <= ?', max_id).update_all(new: false)
     end
   end
+
+  def self.grading_disabled_for?(canvas_course_id, canvas_assignment_id, user)
+    rise360_module_grade = CourseRise360ModuleVersion.find_by!(canvas_assignment_id: canvas_assignment_id)
```
rshipp, line 193: is this exception handled somewhere or is this a "this should never happen so just die if it does"?

app/services/grade_modules.rb
```diff
@@ -167,4 +188,34 @@ def grade_assignment(canvas_assignment_id, user_ids)
...
+      .rise360_module_grades.find_by(user: user)
+
+    # If they've never opened the module a Rise360ModuleGrade won't exist yet.
+    # All auto-grading logic is disabled circuited until they do.
+    unless rise360_module_grade
+      Honeycomb.add_field('grade_modules.module_never_opened', true)
+      return true
+    end
```
rshipp, line 201: I think something like this would be easier to read?
```suggestion
    Honeycomb.add_field('grade_modules.module_opened', rise360_module_grade.present?)
    return true unless rise360_module_grade
```

app/services/grade_modules.rb
```diff
@@ -167,4 +188,34 @@ def grade_assignment(canvas_assignment_id, user_ids)
...
+
+    Honeycomb.add_field('grade_modules.grade_manually_overridden', rise360_module_grade.grade_manually_overridden)
+    return true if rise360_module_grade.grade_manually_overridden
+
+    # If a TA or staff member manually sets the grade in Canvas, now that's the grade
+    # and we turn auto-grading off going forward.
+    if CanvasAPI.client
+       .latest_submission_manually_graded?(canvas_course_id, canvas_assignment_id, user.canvas_user_id)
+
+      Honeycomb.add_field('grade_modules.grade_manually_overridden_detected_at', DateTime.now)
```
rshipp, line 211: should this timestamp field be added in other cases? also `DateTime.now.utc`

app/services/grade_modules.rb
```diff
@@ -167,4 +188,34 @@ def grade_assignment(canvas_assignment_id, user_ids)
...
+      Honeycomb.add_field('grade_modules.grade_manually_overridden', true)
+      rise360_module_grade.update!(grade_manually_overridden: true)
+      return true
+    else
+      Honeycomb.add_field('grade_modules.grade_manually_overridden', false)
```
rshipp, line 216: This is already set above, are you just repeating it for clarity?

lib/canvas_api.rb
rshipp: Since grading is run by a rake task, I think "until the server restarts" is "for this run of the task". But I think that's still fine here.

app/jobs/grade_module_for_user_job.rb
rshipp: What was the reason for this change? Adding the field inside and un-collapsing the `return 0 unless` in that function makes it harder to read for me.

app/services/grade_modules.rb
```diff
@@ -114,34 +116,53 @@ def grade_assignment(canvas_assignment_id, user_ids)
...
+          Honeycomb.add_field('user.id', user_id)
+          user = User.find(user_id)
+          Honeycomb.add_field('canvas.user.id', user.canvas_user_id)
+
+          interactions = Rise360ModuleInteraction.where(
+            user: user,
+            canvas_assignment_id: canvas_assignment_id,
+            new: true,
+          )
+          Honeycomb.add_field('grade_modules.new_interactions.count', interactions.count)
```
rshipp, line 130: This `count` is expensive, since `where` doesn't run any queries and the only place we used `interactions` was `.exists?` below, which runs the fast query `SELECT 1 ... LIMIT 1`. Is having this count in the span worth querying `SELECT COUNT` on all these interactions? Would replacing with this an `exists?` call serve the same purpose? (the `exists?` result would NOT be cached by Rails, so it would also re-run the query below, resulting in two `SELECT 1 ... LIMIT 1`s, which is still faster than `SELECT COUNT` plus `SELECT 1 ... LIMIT 1`).

app/views/rise360_module_grades/show.html.erb
```diff
@@ -19,6 +19,13 @@
       <a href="<%= canvas_url %>" target="_parent">
       Click here to go to the Module</a> and continue working on it
     </p>
+
+    <%#
+      TODO: show a warning message if auto-grading is turned off because a TA or staff
+      member manually entered a grade: https://app.asana.com/0/1174274412967132/1200275983366469
+      The Rise360ModuleGrade.grade_manually_overridden will be set to true
+     %>
```
rshipp, line 27: is this a 1-liner? if so, why are we deferring it with a TODO?

app/services/grade_modules.rb
sadleb: just die

app/services/grade_modules.rb
sadleb: voiced over thoughts here, basically this is the first time we detect it which could help in troubleshooting

sadleb: TODO: utc

app/services/grade_modules.rb
sadleb: yeah, i think `exists` serves same purpose. will change

app/jobs/grade_module_for_user_job.rb
sadleb: notes from standup: moved it b/c there are two entry points for grading a module and I didn't want to have to keep the field in sync from the consumer end. Instead, just have the grade calculator handle it under the covers.

app/views/rise360_module_grades/show.html.erb
sadleb: notes from standup: it's a couple liner but will take up to 30 min or so to write and test. I plan on doing it before releasing this to QA but the PR was already getting big.

PR #698, opened 2021-05-04
Allow re-sending confirmation for expired tokens

rshipp:
### Description

If a token is expired or invalid, redirect to a page with a re-send button. If the token was valid and expired, the button will send a new confirmation email. If the token was invalid, the button will do nothing but act like it did, so we don't expose any information about whether the token was valid or not.

### Motivation and Context

Added a new `resend` action/route because the text in the old one (which is in registrations#show instead of confirmations#new) is not clear for this usecase, and confirmations#new is already used for something else. We might want to refactor how these actions/views are set up at some point, but I don't think it's worth doing right now.

Screen Shot 2021-05-04 at 12 07 20 PM

### Types of changes

### Checklist:


app/controllers/users/confirmations_controller.rb
```diff
@@ -11,11 +11,18 @@ class Users::ConfirmationsController < Devise::ConfirmationsController
   #   super
   # end
 
+  # GET /resource/confirmation/resend
+  def resend
```
sadleb, line 15: `resend` makes it sounds like it's a POST to actually resend it. Maybe `show_invalid` or `show_failure`?

app/controllers/users/confirmations_controller.rb
```diff
@@ -113,8 +120,18 @@ def sign_in_and_get_redirect_path
...
   private
 
+  # Note this uses the nested syntax expected in the #create action.
+  def find_user_by_uuid
+    User.find_by(uuid: params[:user][:uuid])
+  end
+
+  # Note this uses the nested syntax expected in the #create action.
+  def find_user_by_confirmation_token
+    User.find_first_by_auth_conditions(confirmation_token: params.dig(:user, :confirmation_token))
```
sadleb, line 130: Any reason you use `dig` here but not above? If `:user` is guaranteed to be there I don't think we need `dig`.

app/views/layouts/accounts.html.erb
```diff
@@ -1,4 +1,4 @@
-<% content_for :title, 'Braven Login'  %>
+<% content_for :title, 'Braven Login' if content_for(:title).blank? %>
```
sadleb, line 1: Alternate syntax in case you like it better: `<% content_for :title, 'Braven Login' unless content_for?(:title) %>`. I think I like yours b/c `if` is more intuitive than `unless` for me, but I just wanted to point out the existence of `content_for?` (with a question mark) for the future.

sadleb (approved):
All my comments are nits. Defer to you on them.

app/controllers/users/confirmations_controller.rb
rshipp: Whoops, leftover from a previous iteration.

app/controllers/users/confirmations_controller.rb
rshipp: It's not a `show` action though, it's a `new`. The existing `new` is more of a `show`, just without an ID.

sadleb: Yeah, good point re: show vs new. I agree that it's not worth refactoring the built-in Devise routes to make the existing `new` action be something else since it's a bit weird. I can't come up with a good name. Maybe just `invalid` ?

sadleb: Though feel free to stick with `resend` if you don't think it's going to cause much confusion in the future.

sadleb: Decided to go with `show_resend`. We want it to be clear that this is where the button to resend is.

PR #703, opened 2021-05-06
Handle bad reset password tokens more gracefully.

sadleb:
## UPDATE- PUTTING THIS ON HOLD
We put up the two small bug fixes [here](https://github.com/bebraven/platform/pull/704). Ryan is going to take the rest of this and work on it alongside [expiring the signup_token](https://app.asana.com/0/1174274412967132/1200296467131444). The goal is to make is so that the user can easily act on the error message and be able to successfully reset their password instead of getting confused and thinking "there is an error" and end up needing support.

### Description
Task: https://app.asana.com/0/1174274412967132/1200153244154432

If a reset password token is expired or invalid, redirect to the page where you generate a password reset link by entering your email with an error message. We don't reveal any information about what is bad about the token, we just
let them try and generate a new one.

This mimics what we did for the [confirmation flow here](https://github.com/bebraven/platform/pull/698).

Also fixes: https://app.asana.com/0/1174274412967132/1200293140861534

### Motivation and Context
It's relatively common for a user to click password reset. Then do it again and get confused about this is the latest email with the working link. This just let's them go from a bad link, back to the ability to send a new one (and hopefully be more patient waiting for it to come through)

Anticipated valid user flow if they use an old reset password link behavior
image

Bad actor who edits the hidden token value or uses a script to directly post to the "Change password" endpoint
image

### Types of changes

### Checklist:


app/controllers/users/passwords_controller.rb
```diff
@@ -15,17 +15,33 @@ class Users::PasswordsController < Devise::PasswordsController
...
 
     user = User.with_reset_password_token(reset_password_token)
     if user.present? && !user.registered?
       # Since we've verified the token, and we know the user has not done
       # the sign_up flow yet, redirect to RegistrationsController#new to set
       # their initial password, create their Canvas account, etc instead.
+      #
+      # NOTE: we ignore the reset_password_within config for this path.
+      # We want to give them the same amount of time we do to register with
+      # their one-time signup token or to confirm with their confirmation token.
```
rshipp, line 31: we shouldn't do this. reset-password emails are generated by the user's own actions, and should always follow the same rules regardless of whether the account is registered or not. the TODO below is fine, but this comment is misleading.

app/controllers/users/passwords_controller.rb
```diff
@@ -15,17 +15,33 @@ class Users::PasswordsController < Devise::PasswordsController
...
+      # doesn't look like we implemented it.
+      # https://app.asana.com/0/1174274412967132/1200296467131444
       return redirect_to user_registration_path(reset_password_token: reset_password_token)
     end
 
+    # This is most likely for a bad token or a token that's already been consumed or
+    # an expired token. Don't reveal any information about the token's validity or the
+    # account tied to it for security purposes. Just send them back to the "reset password"
+    # page where they can entere their email again to generate new link.
+    redirect_to new_user_password_path(resend: true) and return unless user.present? && user.reset_password_period_valid?
```
rshipp, line 43: IDK about this. What was the original behavior here? the conditional makes me nervous, are we sure this doesn't expose information about the account/token?

rshipp: It looks like the intended behavior for these tokens is to show the form anyway. Which makes sense, because there's no reason to reveal whether the token was valid here. 

Screen Shot 2021-05-06 at 10 30 24 AM

rshipp (changes requested):
i think we only want the `reset_password_token` variable (L18-21), and not the rest of this PR

app/controllers/users/passwords_controller.rb
sadleb: The original behavior was to show the form and then when they submit the form show "Reset password token is invalid" but it was missing the "Reset password token" part of the message b/c we're using the error `message` instead of the `full_message` in app/views/devise/shared/_error_messages.html.erb, which I think was a hacky way to fix the other error messages to be more user friendly.

This just moves us exposing that it was invalid to an earlier point, when they try to load it vs when they try to post to it. It's a nicer user experience imo and doesn't expose anything more.

sadleb:
> i think we only want the `reset_password_token` variable (L18-21), and not the rest of this PR

This PR addresses 2 tasks (actually 3 now since it also expires the token before sending to new registration). Do you think it's worth splitting those out into separate PRs?

PR #706, opened 2021-05-06
Check tokens for expiry during signup

rshipp:
### Description

Check signup and reset_password tokens for expiry when used in RegisterUserAccount service. 

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200296467131444/f

### Checklist:


app/controllers/users/registrations_controller.rb
```diff
@@ -41,21 +41,17 @@ def create
...
 
-    # If the account is already registered, don't try to register again.
-    # Redirect to login, but still don't reveal anything about which account
-    # was tied to the token.
-    if user.present? && user.registered?
-      redirect_to cas_login_path(
-        service: CanvasAPI.client.canvas_url,
-        notice: 'Looks like you have already signed up. Please log in.'
-      ) and return
-    end
```
rshipp, line 52: Unrelated change, but I'm removing this because

1. This can never happen in normal signup/reset flow, since tokens are cleared after registration. The original user story about "if someone tries to click on their old signup link a second time, send them to the login page instead" will never actually be addressed by this code.
2. If a bad actor requests password resets of unregistered accounts, they can then use this endpoint to test for valid tokens.

config/initializers/devise.rb
```diff
@@ -158,6 +158,7 @@
   # prior to launch and there will be folks that don't fully take all the action needed
   # to get into Canvas. We don't want expired links adding noise if the did it partially
   # and are trying to get all the way in the day of launch
+  # Note we reuse this configuration as a "sign_up_within" too, because it's easier.
```
rshipp, line 161: I reused `confirm_within` because I spent a couple minutes looking into how to add a `sign_up_within` option from here to the User model, and it didn't look worth the effort.

app/controllers/users/registrations_controller.rb
rshipp: Currently it renders :bad_link which has a link to the login page.

sadleb: Standup notes: Brian may edit the `bad_link` page in a follow-up PR if it makes it more clear what to do when using the signup link a second time. The `bad_link` path should never be hit from the reset password flow that redirects to the sign-up flow.

app/models/user.rb
```diff
@@ -175,6 +175,12 @@ def send_signup_token(token)
     })
   end
 
+  # https://github.com/heartcombo/devise/blob/57d1a1d3816901e9f2cc26e36c3ef70547a91034/lib/devise/models/recoverable.rb#L77
```
sadleb, line 178: Helpful link.

app/services/register_user_account.rb
```diff
@@ -36,6 +39,23 @@ def run
       span.add_field('app.user.email', @user.email)
       span.add_field('app.register_user_account.salesforce_email', @register_user_params[:email])
 
+      # Mimic Devise's reset_password_by_token behavior for expired tokens.
+      # https://github.com/heartcombo/devise/blob/57d1a1d3816901e9f2cc26e36c3ef70547a91034/lib/devise/models/recoverable.rb#L145
+      # Note if someone tries to use both tokens at once (why??) and one is expired but
+      # the other isn't, we act the same as if both were expired. Because that's easier.
+      if @sign_up_params[:signup_token].present? && !@user.signup_period_valid?
+        span.add_field('app.register_user.signup_token_expired', true)
```
sadleb, line 47: Unrelated to this PR, but I had backlogged a task to talk about not using `span.add_field` anymore and instead always use `Honeycomb.add_field`. If we agree to that and decide that's the guidance, we should go cleanup everything, but just a heads up that I've started using `Honeycomb.add_field` always.

From [their docs](https://docs.honeycomb.io/getting-data-in/ruby/beeline/): 

Here is an example of adding a custom field to the current span:
```
def calculate_big_number
  # ... do work, find a big number
  Honeycomb.add_field('big_num', result)
  # ... do thing with big_num, maybe something fails
  Honeycomb.add_field('error', err)
end
```
Additional fields are added under the app. namespace. For example, the field above would appear in your event as app.big_num. The namespace groups your fields together to make them easy to find and examine.

If you prefer to avoid the app. namespace, for example in order to overwrite an automatically-populated field, you can have your block take the current span as a paramter and call add_field directly on that instance.

sadleb (approved):

app/controllers/users/registrations_controller.rb
rshipp: done here https://github.com/bebraven/platform/pull/709

PR #718, opened 2021-05-13
Specs for ability to send invitation emails to sign-up and create an account

sadleb:
### Description

https://github.com/bebraven/platform/pull/686/files
https://github.com/bebraven/platform/pull/687/files

### Screenshots (if appropriate):

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #719, opened 2021-05-13
Add signup specs

rshipp:
### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200192173211416/f

### Types of changes

### Checklist:


app/controllers/users/registrations_controller.rb
```diff
@@ -42,12 +42,21 @@ def create
...
     user = find_user_by_signup_token || find_user_by_reset_password_token
 
+    # Act the same as #new, just in case someone tried to register again
+    # with a tab that was open before or something.
+    if user.present? && user.registered?
+      redirect_to cas_login_path(
+        service: CanvasConstants::CANVAS_URL,
+        notice: 'Looks like you have already signed up. Please log in.'
+      ) and return
+    end
```
rshipp, line 52: this is the only behavior change

app/controllers/users/registrations_controller.rb
```diff
@@ -42,12 +42,21 @@ def create
...
+
     # If the token was valid, run account registration.
     if user.present?
       # Register the new user in all of our systems.
-      RegisterUserAccount.new(sign_up_params).run do |user|
-        if user.errors.any? || !user.persisted?
-          if user.errors.any? { |e| [:reset_password_token, :signup_token].include? e.attribute }
+      RegisterUserAccount.new(sign_up_params).run do |updated_user|
```
rshipp, line 57: this variable rename is just for clarity, since i realized i was overwriting the `user` in the outer scope with the `user` inside this block

app/controllers/users/registrations_controller.rb
sadleb: Word. This behavior matches what I thought we talked about doing in this case.

spec/controllers/cas/login_spec.rb
```diff
@@ -94,6 +94,18 @@
...
+      let!(:unconfirmed_user) { create(:unconfirmed_user) }
+      let(:username) { unconfirmed_user.email }
+      let(:password) { 'NotTheRightPassword' }
+
+      it "fails to log in" do
+        subject
+        # Ensure that the login was failed
+        expect(page).to have_content("Incorrect username or password")
+      end
+    end
```
sadleb, line 108: fyi: I added a PLACHOLDER task here for myself to revisit if there are other confirmation related specs we want to add to this `login_spec` file: https://app.asana.com/0/1174274412967132/1200325250500042

spec/controllers/users/passwords_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+RSpec.describe Users::PasswordsController, type: :controller do
+  render_views
+
+  describe '#edit' do
+    subject { get :edit, params: { reset_password_token: @reset_password_token } }
+
+    context 'with invalid token' do
+      before :each do
+        @reset_password_token = 'fake'
```
sadleb, line 11: I think it's fine, but I'm curious if there was a reason you're using an instance variable instead of a `let(:reset_password_token) { 'fake' }` variable? Just asking b/c it may be better to choose a style and generally stick with it across our specs. In the current test suite, it's rare to use instance variables.

spec/controllers/users/passwords_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+
+      it 'shows the form' do
```
sadleb, line 14: We've talked a ton about the reasoning behind this type of behavior, so that for security purposes it doesn't allow someone to enumerate the endpoint for valid tokens. A future dev (or me) not knowing (or forgetting) that context could easily decide to change the behavior to be more user friendly and show an error message about an invalid token. This spec gives no hints as to "why" we show the form and I think it should either as a comment or in the name of the example

spec/controllers/users/passwords_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  describe '#update' do
+    subject { post :update, params: {
+      'user[reset_password_token]': @reset_password_token,
+      'user[password]': 'password',
+      'user[password_confirmation]': 'password',
+    } }
+
+    context 'with valid token' do
```
sadleb, line 54: Should we add a spec for the invalid token behavior? It should be the same as valid token and that's important for security purposes, right?

spec/controllers/users/passwords_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      let(:user) { create(:registered_user) }
+
+      before :each do
+        @reset_password_token = user.send(:set_reset_password_token)
+      end
+
+      it 'resets signup token/sent_at' do
```
sadleb, line 61: Ditto on the why? Let's try to prevent a future dev changing this behavior without understanding the consequences.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
+require 'rails_helper'
+require 'salesforce_api'
+
+RSpec.describe Users::RegistrationsController, type: :controller do
+  render_views
+
+  describe '#show' do
+    subject { get :show, params: { uuid: uuid } }
+
+    context 'with invalid uuid param' do
```
sadleb, line 10: Ditto on above "why" comment.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+        reset_password_token: '',
+      } }
+
+      it 'renders bad_link' do
+        subject
+        expect(response.body).to include("Uh oh. Something went wrong!")
+      end
+    end
+
+    context 'with invalid signup_token' do
```
sadleb, line 64: Ditto on above "why" comment.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+
+      it 'renders form' do
+        subject
+        expect(response.body).to include("<h1>Create Account</h1>")
+        expect(response.body).to include("<form action=\"#{user_registration_path}\"")
+        expect(response.body).to include("<input value=\"#{params[:signup_token]}\" type=\"hidden\" name=\"user[signup_token]\"")
+      end
+    end
+
+    context 'with invalid reset_password_token' do
```
sadleb, line 77: Ditto on above "why" comment.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+        expect(response.body).to include("Uh oh. Something went wrong!")
+      end
+    end
+
+    context 'with invalid signup_token' do
+      let(:params) { {
+        signup_token: 'invalid token',
+      } }
+
+      it 'renders form' do
```
sadleb, line 69: Ditto on above "why" comment.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+      let(:params) { {
+        reset_password_token: 'invalid token',
+      } }
+
+      it 'renders form' do
+        subject
+        expect(response.body).to include("<h1>Create Account</h1>")
+        expect(response.body).to include("<form action=\"#{user_registration_path}\"")
+        expect(response.body).to include("<input value=\"#{params[:reset_password_token]}\" type=\"hidden\" name=\"user[reset_password_token]\"")
+      end
```
sadleb, line 87: Ditto on above "why" comment.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+      let(:params) { {
+        signup_token: 'invalid token',
+      } }
+
+      it 'renders form' do
+        subject
+        expect(response.body).to include("<h1>Create Account</h1>")
+        expect(response.body).to include("<form action=\"#{user_registration_path}\"")
+        expect(response.body).to include("<input value=\"#{params[:signup_token]}\" type=\"hidden\" name=\"user[signup_token]\"")
+      end
```
sadleb, line 74: Ditto on above "why" comment.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+  describe '#show' do
+    subject { get :show, params: { uuid: uuid } }
+
+    context 'with invalid uuid param' do
+      let(:uuid) { 'invalid uuid' }
+
+      it 'shows uuid param in hidden field' do
+        subject
+        expect(response.body).to include("<input value=\"#{uuid}\" type=\"hidden\" name=\"user[uuid]\"")
+      end
```
sadleb, line 16: Ditto on above "why" comment.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+      before :each do
+        allow(SalesforceAPI).to receive(:client).and_return(sf_client)
+        params['user[signup_token]'] = user.set_signup_token!
+        user.update!(signup_token_sent_at: 5.weeks.ago)
+      end
+
+      it 'renders bad_link' do
+        subject
+        expect(response.body).to include("Uh oh. Something went wrong!")
+      end
```
sadleb, line 240: Ditto on above "why" comment. Prevent future dev from showing a nicer error message since this was a valid token and it's easy to want to be helpful and action oriented from a UX perspective.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+      before :each do
+        allow(SalesforceAPI).to receive(:client).and_return(sf_client)
+        params['user[reset_password_token]'] = user.send(:set_reset_password_token)
+        user.update!(reset_password_sent_at: 5.days.ago)
+      end
+
+      it 'renders bad_link' do
+        subject
+        expect(response.body).to include("Uh oh. Something went wrong!")
+      end
```
sadleb, line 259: Ditto on above comment

spec/controllers/users/passwords_controller_spec.rb
rshipp: I used let everywhere else, the ones where I use an instance variable it's because I have to modify the variable inside the `before` block and use it in the `subject`. if I declare this with let, but try to set it in `before`, it actually just creates a new variable in block scope and doesn't update the one used in the `subject`.

spec/controllers/users/passwords_controller_spec.rb
rshipp: no, this is `#update`. i'm not sure why i didn't have an invalid-token context, but it'd be an error page, not the same behavior.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+      } }
+
+      before :each do
+        allow(RegisterUserAccount).to receive(:new).and_return(service)
+        params['user[signup_token]'] = user.set_signup_token!
+      end
+
+      it 'runs register service' do
+        subject
+        expect(service).to have_received(:run).once
```
sadleb, line 314: Do we want to check if the proper values were used to initialize the `service` before calling run?

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+      end
+
+      it 'redirects to #show with uuid' do
+        subject
+        expect(response).to redirect_to(users_registration_path(uuid: user.uuid))
+      end
+    end
+
+    context 'with valid signup_token, unregistered user, register fails' do
```
sadleb, line 323: Confusing example name. Almost the same as the above. Maybe `with valid signup_token, unregistered user, when missing email` or `with valid signup_token, unregistered user, on form errors` if we want to add more ways the form can error in the future.

spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+        subject
+        expect(service).to have_received(:run).once
+      end
+
+      it 'redirects to #show with uuid' do
+        subject
+        expect(response).to redirect_to(users_registration_path(uuid: user.uuid))
+      end
+    end
+  end
```
sadleb, line 374: Do we care about testing the behavior for form errors like we do above (e.g. an empty email) if they use `reset_password_token` instead of `signup_token` ? Probably not, just checking in case leaving that out was accidental.

spec/lib/salesforce_api_spec.rb
```diff
@@ -345,4 +345,35 @@
...
+      response = SalesforceAPI.client.get_contact_signup_token(contact_id)
+
+      expect(WebMock).to have_requested(:get, request_url_regex).once
+      expect(response).to eq(contact_obj[:Signup_Token__c])
+    end
+  end
+
+  describe '.update_contact' do
+    let(:contact_id) { '003170000125IpSAAU' }
+    let(:contact_obj) { { test: 1 } }
```
sadleb, line 366: Might be clearer what this is testing if this was `let(:contact_obj) { { some_field: 'some_value' } }`

spec/controllers/users/registrations_controller_spec.rb
rshipp: seems like testing a getter? there's no logic there, it just directly passes in the params permitted by Devise.

spec/controllers/users/registrations_controller_spec.rb
rshipp: would `validation fails` be clearer? `invalid user`? form errors are the result, not the context.

spec/controllers/users/registrations_controller_spec.rb
rshipp: no, that'd hit the same code paths and just be more to maintain without giving any added confidence.

spec/models/user_spec.rb
```diff
@@ -325,4 +338,158 @@
...
+        expect(sf_client).not_to have_received(:get_contact_signup_token)
+      end
+
+      it 'sends email with token url' do
+        Devise.mailer.deliveries.clear()
+        subject
+        expect(Devise.mailer.deliveries.count).to eq 1
+        # Hopefully the first part is always the text/plain one.
+        # If this starts failing, try adjusting the body part selection.
+        expect(Devise.mailer.deliveries.last.body.parts.first.to_s).to include("/users/sign_up?signup_token=3D#{token}")
```
sadleb, line 365: It's prob better to stub out the mailer used and check that it was created and the proper method to send this email with it was sent. See `spec/jobs/clone_course_job_spec.rb` for an example of where we do something like that. Then leave the actual checking of the email to the `spec/mailers/send_signup_email_mailer_spec.rb` spec I have here https://github.com/bebraven/platform/pull/718/files

Btw, I had a ton of trouble with inspecting the content of the email too and ended up using the plain text one over there too rather than the HTML one which we care about more.

spec/models/user_spec.rb
```diff
@@ -325,4 +338,158 @@
...
+        expect(sf_client).to have_received(:get_contact_signup_token)
+      end
+
+      it 'sends email with token url' do
+        Devise.mailer.deliveries.clear()
+        subject
+        expect(Devise.mailer.deliveries.count).to eq 1
+        # Hopefully the first part is always the text/plain one.
+        # If this starts failing, try adjusting the body part selection.
+        expect(Devise.mailer.deliveries.last.body.parts.first.to_s).to include("/users/sign_up?signup_token=3D#{token}")
```
sadleb, line 387: Ditto on above comment.

spec/models/user_spec.rb
```diff
@@ -325,4 +338,158 @@
...
+      allow(SalesforceAPI).to receive(:client).and_return(sf_client)
+    end
+
+    it 'calls salesforceapi' do
+      subject
+      expect(sf_client).to have_received(:update_contact)
+        .with(user.salesforce_id, {'Signup_Token__c': token})
+        .once
+    end
+  end
```
sadleb, line 436: Should we add a spec for if that API call fails? Currently it will never try again and the only way to get a valid signup link ready for the Campaign Monitor blast would be to use the Platform Admin dash to send the email (or have an engineer grab the token out of the DB).

spec/models/user_spec.rb
rshipp: i'd rather just delete this part of the spec if we're relying on mailer specs instead. i don't see any use in a spec that just asserts that a hardcoded service is run.

spec/models/user_spec.rb
```diff
@@ -325,4 +338,158 @@
...
+
+      it 'returns false' do
+        expect(subject).to eq(false)
+      end
+    end
+
+    context 'with unexpired token' do
+      before :each do
+        user.set_signup_token!
+        user.update!(signup_token_sent_at: 1.weeks.ago)
```
sadleb, line 457: Any reason to use 1 week instead of something safer like 5 minutes ago? We also have comments elsewhere about why we chose 2.5 weeks as the expiration, but if we change that later for good reason it would be annoying for this spec to break when it doesn't care about the context of how long the expiration is, just that it's expired.

spec/models/user_spec.rb
rshipp: there's no model behavior to handle api failures, that spec would be in the sync service, which already has tests for generic exceptions.

spec/services/register_user_account_spec.rb
```diff
@@ -74,5 +74,103 @@
...
+
+      it 'clears all signup/reset tokens' do
+        RegisterUserAccount.new(sign_up_params).run do |u|
+          expect(u.errors.any?).to eq(false)
+        end
+        user.reload
+        expect(user.signup_token).to eq(nil)
+        expect(user.signup_token_sent_at).to eq(nil)
+        expect(user.reset_password_token).to eq(nil)
+        expect(user.reset_password_sent_at).to eq(nil)
```
sadleb, line 97: Ditto on why comment. Not to be annoying but I really want to ingrain it into these specs all these security related decisions and work we did to make it secure in a way that prevents future devs (or PMs) from undoing our hard work without at least being forced to confront the why and understand the security implications.

spec/services/register_user_account_spec.rb
```diff
@@ -74,5 +74,103 @@
...
+      let(:user) { create(:unregistered_user) }
+
+      before :each do
+        sign_up_params[:reset_password_token] = user.send(:set_reset_password_token)
+        allow(SalesforceAPI).to receive(:client).and_return(sf_client)
+        allow(CanvasAPI).to receive(:client).and_return(canvas_client)
+        allow(SyncPortalEnrollmentForAccount).to receive(:new).and_return(enrollment_process)
+      end
+
+      it 'clears all signup/reset tokens' do
```
sadleb, line 112: Ditto on why.

fyi: in this round of specs, I finally took the time to learn and get more comfortable with `shared_examples`. I'm not sure if they would or wouldn't help make all these specs more maintainable without spending time trying to actually refactor some of this stuff into a shared example, but IF the reason you aren't using them is b/c you haven't taken the time to really tinker with them and get comfortable, I would suggest doing that sooner than later. I didn't like them much when Jess first started using them, but I'm staring to like them now that I can actually use them (and therefore read them better).

spec/models/user_spec.rb
rshipp: 5 minutes doesn't give me any confidence it's actually honoring the expiration. we could change all these hardcoded dates to pull from the devise config and add/subtract time as needed, but I decided that was 1) more work and 2) gave me less confidence, since I wouldn't be able to look at something like `devise_config.password_expire_setting - 1.day` and be able to tell if it's actually testing the dates I'd expect it to. if we change the expiration times later, I'd prefer to have to update these specs to match. I intentionally used `5.weeks.ago` and `5.days.ago` for all post-expiration test dates, and `1.days.ago` and `1.weeks.ago` for all pre-expiration test dates, to make it easy to find-and-replace.

spec/services/register_user_account_spec.rb
```diff
@@ -74,5 +74,103 @@
...
+        RegisterUserAccount.new(sign_up_params).run do |u|
+          errors = u.errors
+        end
+        user.reload
+        expect(errors.count).to eq(1)
+        expect(errors.first.attribute).to eq(:signup_token)
+        expect(errors.first.type).to eq(:expired)
+        # Verify it returned before updating user.
+        expect(user.signup_token).not_to eq(nil)
+      end
```
sadleb, line 147: Can you help me understand what's going on here? I'm probably running out of steam at the end of this painful review and getting lazy, but everywhere else we try not to expose information about the validity of a token for security purposes. Why is this different?

spec/services/register_user_account_spec.rb
rshipp: the last time I tried to use shared_examples, they didn't work like I expected, so I didn't think they'd be appropriate here. I need to be able to set different contexts and run the same "it_behaves_like" specs in that context, but shared examples seem to run in their own contexts, not the contexts of the block running `it_behaves_like`. correct me if that's wrong

spec/services/register_user_account_spec.rb
```diff
@@ -74,5 +74,103 @@
...
+          errors = u.errors
+        end
+        user.reload
+        expect(errors.count).to eq(1)
+        expect(errors.first.attribute).to eq(:reset_password_token)
+        expect(errors.first.type).to eq(:expired)
+        # Verify it returned before updating user.
+        expect(user.reset_password_token).not_to eq(nil)
+      end
+    end
```
sadleb, line 174: Ditto on needing help understanding what is special about this situation and why we give information about why the token is invalid.

spec/services/register_user_account_spec.rb
rshipp: it's not different, here's the relevant code path in the controller for reference:

```ruby
        if updated_user.errors.any? || !updated_user.persisted?
          if updated_user.errors.any? { |e| [:reset_password_token, :signup_token].include? e.attribute }
            # If the token expired, act the same as we do for invalid
            # tokens below. Don't reveal the exact reason for failure.
            return render :bad_link
          end
```

rshipp: to clarify, the register service (which these specs are for) always reveals information about what exactly went wrong. the controller needs this information to determine whether it should render a generic "bad link" page (when tokens are missing, expired, invalid, etc), or specific error messages like "passwords didn't match".

spec/controllers/users/registrations_controller_spec.rb
sadleb: Both seem good. Up to you

spec/models/user_spec.rb
sadleb: I still think we want to make sure the service runs. If a future refactor adds more logic (maybe to send different emails in different situations), we’d want this spec here to make sure they still call this service or force them to refactor this spec to account for the new logic.

I think it also acts as a forcing mechanism to write specs for any changes to this method since there was a already one there. 

Feels the same as all other method specs that check if the underlying service was actually called?

spec/models/user_spec.rb
sadleb: Makes sense. I need to check my specs to see if I had any expiration related dates. I think I may. I might refactor to have them all pull from a constant in a helper class when I do.

spec/services/register_user_account_spec.rb
sadleb: The way I’ve handled this is by having common let variables drive the change in behavior. I’m not sure how to do it cleanly with different subjects though.

PR #727, opened 2021-05-18
Specs and bug fix for user needing reconfirmation for changed email.

sadleb:
### Description
This started off as writing specs for a user whose email changes
and they need to confirm their new email to be able to log in with it:
https://app.asana.com/0/1174274412967132/1200325250500042

I realized that the behavior we had talked about here:
https://github.com/bebraven/platform/pull/685#discussion_r625961906

was never implemented. So in addition to the specs and adding Honeycomb instrumentation,
it implements that behavior so that users trying to log in with their new email
but who haven't confirmed it can now generate a new confirmation email
if they lost it or the token expired.

### Motivation and Context

Copy/pasta from the discussion above:

The scenario is that we just changed their email and sent a confirmation link. They have the unconfirmed_email column set to the new email but the email column is still the old one. If the confirmation token expires, they will be redirected to the login page with no info about why other than something is wrong. They were just told to use the new email to log in with, so they try that. The code uses the email field which is the old email, so it fails to log them in. Since we only redirect to the page to generate a new confirmation token / link if they successfully log in, they just get an “invalid username / password” message. We need to look them up by unconfirmed_email as well to determine if we can send them to the page to generate a new confirmation token / link.

### Screenshots
Changing the confirmation email language to account for the fact that nothing will change if they don't click the button. This is to handle the situation where the email was changed to someone else's incorrectly or for some future user story where users can change their own email and try to use it to gain access to someone else's account. Mimics the language of the password reset email to account for bad-actors, letting you know you can just ignore it. Applies to first time registration and reconfirming an email change.

BEFORE
image

AFTER
image

### Types of changes

### Checklist:


lib/cas_authenticator.rb
```diff
@@ -1,13 +1,11 @@
...
 
     def validate(credentials)
-      @user = User.find_by_email!(credentials[:username])
-      #valid_password?(credentials[:password]) && active_for_authentication?
-      valid_password?(credentials[:password]) && @user.confirmed? && @user.registered?
-    rescue ActiveRecord::RecordNotFound
-      false
+      @user = User.find_by_email(credentials[:username])
+      @user.present? && valid_password?(credentials[:password]) && @user.confirmed? && @user.registered?
     end
```
sadleb, line 9: Unrelated to this PR, but I stared at the `User.find_by_email!` for too long wondering how in the heck this wasn't throwing all sort of exceptions before I noticed the `rescue` below. Feels cleaner to not rely on an exception for a normal code path?

spec/controllers/cas/login_spec.rb
```diff
@@ -46,66 +59,89 @@
...
+        let(:password) { unconfirmed_user.password }
+
+        it 'redirects to registration page with confirmation instructions' do
+          subject
+          expect(current_url).to match(
+            /#{Regexp.escape("/users/registration?login_attempt=true&uuid=#{unconfirmed_user.uuid}")}/
+          )
+          expect(status_code).to eq(200)
+          expect(page).to have_content("Didn't receive email instructions")
+          expect(find('h1').text).to eq('Please confirm your email address')
```
sadleb, line 117: This is the spec for the new behavior for a user trying to login with their `unconfirmed_email` using a valid password for the account.

rshipp (changes requested):
we decided explicitly not to do this, we should discuss again if needed

sadleb:
> we decided explicitly not to do this, we should discuss again if needed

I remember talking about solving the most common issue which would be an expired link another way, but what did we decide to do in the situation where they delete the confirmation email. How would they get access to the account with the new email? 

The two options I can think of other than this are 1) having an engineer go in the database and grab the token so we can manually send a link 2) change the email to something fake and then change it back so that a new email is sent

rshipp:
if someone manages to permanently delete their confirmation email, and the email change is due to a staff typo so the student doesn't know their old email, and the staff has already forgotten what the typo was, we can send them the static link that lets them request a new confirmation email. we could even link that page from login, if we need to. if that page doesn't work with reconfirmation, that should be an easy fix.

sadleb:
> if someone manages to permanently delete their confirmation email, and the email change is due to a staff typo so the student doesn't know their old email, and the staff has already forgotten what the typo was, we can send them the static link that lets them request a new confirmation email. we could even link that page from login, if we need to. if that page doesn't work with reconfirmation, that should be an easy fix.

I should have just gotten on a call to talk through this instead of typing this all out, but I got in too deep trying to think it through so let's talk about this wall of text in today's working session:
-----
The main thing I want to accomplish with all these account creation / mgmt user stories is to reduce the necessity for staff to get involved and allow folks to get into their account themselves in the various scenarios that cause them to not be able to. Without compromising security ofc. Here are some example permutations of how we can get into a situation where they are trying to log in with a user needing reconfirmation but don't have or can't find the confirmation email and we'd like to avoid staff having to do something to get them access:

For all scenarios, they could either delete the email or ignore it at the moment and forget that they got it, so they can't find it when they are actually trying to log in.

1. We had an email with a typo in it and change it to the correct one on our end without them realizing there was a backend issue. 
    1. This happens before the welcome email blast goes out and it seems random b/c we've never communicated their Braven account. They delete or ignore it and promptly forget about it.
    2. This happens after the welcome email blast, which presumably they never got b/c we had the wrong email. Once again, seems random and they ignore/forget or delete.
2. Same typo situation as above but this time we change it b/c they can't log in and open a support ticket. Staff changes it to the correct one an hour or two later. Staff closes the ticket saying their account should work with the proper email after confirming.
    1. They see the confirmation email that night before seeing the support ticket message and delete it b/c they don't realize it's related to the support ticket.
    2. They got busy and don't take action until they show up at LL. They need to get into the account urgently and don't take the time to go dig up the support ticket closing message or confirmation email. In Salesforce everything looks correct so we have to escalate to product support who can use the Admin dash to find out that the new email address is still unconfirmed (and most likely just clicks the "Confirm Account" button).
3. They can't access the email account they had been using (password issues) and the path of least resistance is changing their Braven login to a new email. Open a support ticket to do that. Same type of things as above could happen where b/c it's asynchronous they end up trying to log in with the new email but somehow missed or lost or deleted the confirmation email.

The first suggestion of having staff send them the static "show_resend" link doesn't help with the problem we're trying to solve, allowing them to get access without needing to open a support ticket. Plus, the language on that page would be weird in this context. Most likely support staff would just hit "Confirm Account" on the admin dash.

I don't think that just adding a link to that static page to the login page will reduce support tickets. They would try to log in and it says "invalid username or password". They'd have no clue to try the link to get to the static resend confirmation page. What would that link say "Having trouble logging in and never confirmed your account?" and then we'd change the language on that page from "Confirmation failed" to be even more vague or create a second page with more general language or pass a param to control the language?

The above is more work and I'm not sure it will help solve the problem. Can you remind me of what the security concern is with this approach? They're logging in with valid credentials for an email address that we just want to make sure they actually have access to the inbox. I can't think of a reason why we wouldn't let them resend the email if they used valid credentials?

app/controllers/cas_controller.rb
```diff
@@ -172,13 +174,18 @@ def loginpost
         end
       # DANGER!! This works for our custom CAS authenticator, but probably
       # won't work for other auth methods! If we change auth methods, revisit.
```
sadleb, line 176: TODO: add comments about avoiding having someone who hasn't spent a TON of time learning and thinking through all the password/user stuff a task to change any of this.

lib/cas_authenticator.rb
```diff
@@ -16,5 +14,9 @@ def valid_password?(password)
       BCrypt::Password.new(@user.encrypted_password) == password
     end
 
+    def valid_password_for_unconfirmed_email?(unconfirmed_email, password)
+      @user = User.find_by_unconfirmed_email(unconfirmed_email)
+      @user.present? && valid_password?(password)
+    end
```
sadleb, line 20: TODO: document the fact that this implicitly changes the user. Also, prob write a spec specifically for the authenticator.

lib/cas_authenticator.rb
sadleb: TODO: put `@user.present?` on it's own. Make it so that any future folks adding more conditions don't have to think about user.present and that just immediatelly returns false.

app/controllers/cas_controller.rb
```diff
@@ -172,13 +174,18 @@ def loginpost
...
+            (last_failed_authenticator.respond_to?(:valid_password_for_unconfirmed_email?) && last_failed_authenticator.valid_password_for_unconfirmed_email?(@username, @password))
         # The username/password is correct, but something else is making
         # this account currently invalid for login. Be careful here.
-        user = User.find_by_email(@username)
-        if user.present? && user.registered? && !user.confirmed?
+        user = last_failed_authenticator.user
+        add_honeycomb_context(user)
+
+        if user.present? && user.registered? &&
+           (!user.confirmed? || user.unconfirmed_email == @username)
```
sadleb, line 185: TODO: add language in teh confirmation email about it not taking effect until you click the button. Needs to make sense in the context of a registration confirmation vs an email change confirmation, but if staff chagnes it to an incorrect email or in the future they can change it themselves (and try to get into someone elses account that way) we want to make it clear that nothing will happen until they click the button.

rshipp:
i'm fine with merging this without re-review once you've added the todo comments

PR #733, opened 2021-05-25
Don’t stay on the login form if you actually login

sadleb:
### Description
This started off as a bug fix where in certain situations, like trying several times to login with bad creds, would cause the successful login to just stay on the Log In form with a "You have successfully logged in" message. Task: https://app.asana.com/0/1174274412967132/1200357018381226

In the process of fixing that, I uncovered 2 other bugs:
* When you reset your password, it was automatically logging you in with a "Your password has been successfully changed. You are now logged in message". The PR makes is so that you don't automatically get logged in. The message is now just "Your password has been successfully changed." asking you to log in.
* When you use an invalid confirmation link or try to log in as an unconfirmed user, it shows a page letting you resend the confirmation link. Clicking that button did send the email, but there was no visual indication that anything happened. With this change, it redirects to a page asking you to check your email.

Cancelled hung CI, but the results were:
```
Finished in 2 minutes 53.1 seconds (files took 6.53 seconds to load)
1558 examples, 0 failures, 27 pending
```

Page that you used to be able to get to after trying invalid credentials several times (or clicking the "try logging in instead" link from the invalid confirmation link page and other pages that let you go back to login without passing a `service` param)
image

* Now, if you are a platform admin then after login you go to the root path, otherwise you go to Canvas.
    * Note: If you didn't have a Canvas account, things would behave weird. The only valid use case I can think of for this possible code path is if we create regional accounts that are not admins, who have some narrower role. Fixing that is outside the scope of this PR. [Follow-up task here](https://app.asana.com/0/1174274412967132/1200391779170220).

Page you're now redirected to after resending the confirmation link:
image

Page that used to show after you reset your password. You were logged in at this point. Now the same page is shown but you are not logged in and the "You are now signed in" part of the message isn't there:
image

### Types of changes

### Checklist:


app/controllers/cas_controller.rb
```diff
@@ -156,34 +158,41 @@ def loginpost
...
-          rescue URI::InvalidURIError
-            logger.error("The service '#{@service}' is not a valid URI!")
-            @message = {
-              :type => 'mistake',
-              :message => 'The target service your browser supplied appears to be invalid. Please contact your system administrator for help.'
-            }
-          end
+          @service = helpers.default_service_url_for(user)
+          logger.info("No service param was given, setting it to default for this user: #{@service}.")
+        end
```
sadleb, line 174: It's sort of hard and uncommon to get to this code path b/c we do pass the `service` param around quite a bit in most of the standard code paths, but it is possible and there is never a situation where a successful login should leave you on the form to log in and just say "you're now logged in". This now mimics the redirect that happens in a lot of other flows after you successfully log in and if you're an admin it sends you to platform and otherwise sends you to Canvas.

app/controllers/cas_controller.rb
```diff
@@ -35,6 +35,8 @@ class CasController < ApplicationController
   # PT = Proxy Ticket
   include RubyCAS::Server::Core::Tickets
 
+  GENERIC_LOGIN_FAILED_ERROR_MESSAGE = { :type => 'mistake', :message => 'Incorrect username or password.' }
```
sadleb, line 38: Unrelated to this change. Just trying to force the comments around using the same message for multiple code paths to be true through code.

app/controllers/users/confirmations_controller.rb
```diff
@@ -19,11 +19,15 @@ def show_resend
...
+  # This is used to re-send confirmation instructions if they lost the email
+  # or if the token is invalid.
   def create
     # Don't error out, and don't reveal whether the UUID (or confirmation_token)
-    # was valid. Just try to send an email if the user exists.
+    # was valid. Just try to send an email if the user exists and then tell them
+    # to check their email.
     user = find_user_by_uuid || find_user_by_confirmation_token
     user&.send_confirmation_instructions
+    redirect_to new_user_confirmation_path
```
sadleb, line 30: See the comments on [this commit](https://github.com/bebraven/platform/pull/733/commits/eee6f8b30d3170d0ada4e741bf0540fbe0f78fda) for more context about this bug fix.

app/controllers/users/confirmations_controller.rb
```diff
@@ -104,11 +108,7 @@ def after_resending_confirmation_instructions_path_for(_)
   # auto-log them in there, otherwise do it here in the platform.
   def sign_in_and_get_redirect_path
      username = resource.email
-     if resource.canvas_user_id
-       login_service_url = CanvasConstants::CAS_LOGIN_URL
-     else
-       login_service_url = ::Devise.cas_service_url(request.url, devise_mapping)
-     end
+     login_service_url = helpers.default_service_url_for(resource)
```
sadleb, line 111: This changes the prior behavior slightly since now we use "is admin?" to determine if we should try to send them to Canvas vs Platform. I think it's better to be consistent across all these redirects for the logic. There is a slight bug that I don't think is worth fixing at the moment for folks that we give non-admin roles on platform. Their login will try to take them to Canvas. When we fix that, I want to make sure we're doing it in all the places that redirect.

config/initializers/devise.rb
```diff
@@ -238,7 +238,7 @@
 
   # When set to false, does not sign a user in automatically after their password is
   # reset. Defaults to true, so a user is signed in automatically after a reset.
-  # config.sign_in_after_reset_password = true
+  config.sign_in_after_reset_password = false
```
sadleb, line 241: Here is the original change that made it work this way, which we broke: https://github.com/bebraven/platform/pull/220 -> the reason we want to do this is b/c we want them to immediately be forced to use their password which will hopefully cause them to either update their password manager or write it down however they manage passwords. Basically, we don't want to incentivize using password reset every time to log in b/c it's just a two step process.

spec/controllers/cas/login_spec.rb
```diff
@@ -9,46 +9,6 @@
...
-
-    context 'for registered user' do
-      let!(:valid_user) { create(:registered_user) }
-      it_behaves_like 'login'
-    end
-
-    context 'for user needed reconfirmation of new email' do
-      let!(:valid_user) { create(:reconfirmation_user) }
-      it_behaves_like 'login'
-    end
```
sadleb, line 49: I spent way too much time trying to re-write these specs for the new behavior. The CAS specs are a bit weird. They're routing specs, not controller specs and they use VCR and Capybara stuff. I'm not sure if we can't use standard controller specs for some reason, but it was outside the scope of this PR to make that sort of change. Since this is an edge case, I didn't think it was worth the effort to spend any more time to try and get them working.

rshipp: what do we lose by deleting these, do we have similar tests somewhere else?

rshipp (approved):

spec/controllers/cas/login_spec.rb
sadleb: I don't think we have similar tests for this. We lose tests for this edge case where there is no service param and it gets set automatically to Canvas vs Platform depending on whether you're an admin.

PR #738, opened 2021-06-01
Discord bot MVP

rshipp:
### Description

This is a major PR that adds all "MVP" bot functionality. There are also a couple things changed outside of the bot itself, e.g. in initializers that apply to the platform Rails app.

Basic list of bot features:

* Able to deploy Discord bot to Heroku alongside the Platform webapp (just assign a dyno to the `bot` proc in the Heroku UI)
* Monitoring with Honeycomb, logging, Sentry, and RestClient autoinstrumentation, so we can track Bot uptime / errors / etc
* Integration with Salesforce (Program.Discord_Server_ID__c, Participant.Discord_Invite_Code__c, Contact.Discord_User_ID__c)
  * To start "syncing" a new Program, add the appropriate Discord Server ID to the Program record, invite the bot to the server, assign it the Admin role, and then setup a scheduled task to run `bin/discord_sync`
* Generate unique invite codes for each user in synced programs; save to Salesforce
* Assign users appropriate roles/nick when they join the Discord server, then delete the used invite so it can't be used again
* Create Discord channels/roles automatically from Salesforce data, when a user joins
* Add LC channel-specific permissions overrides, when an LC joins
* Detect when the bot is added to new servers and update local invite caches/etc.

Other changes outside of the bot:

* Fixed missing activesupport import in filter_logging.
* Fixed missing Rails import in SalesforceAPI (we should try to remove that dependency, but I didn't spend any time on that here).
* Removed Rails dependency from restclient_instrumentation.
* Added Sentry->Honeycomb integration. There's a new tag in Sentry events that points directly to the Honeycomb trace waterfall view. Example (log in through the Platform Sentry addon on Heroku): https://sentry.io/organizations/braven-pr/issues/2444477346/?project=5799560&query=is%3Aunresolved.
* New SalesforceAPI functions.

Things to ignore in this PR:

* All the TODOs - they're for later tasks.
* Magic strings/IDs everywhere... they'll get pulled out into constants later.

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200370624758841/f

### Types of changes

### Checklist:


bot.rb
```diff
@@ -0,0 +1,55 @@
...
+require 'rails'
+require_relative 'lib/salesforce_api'
+
+PROGRAM_ID = ENV['BOT_PROGRAM_ID']
+BOT_TOKEN = ENV['BOT_TOKEN']
+SERVER_ID = ENV['BOT_SERVER_ID'].to_i
+GENERAL_CHANNEL_ID = ENV['BOT_GENERAL_CHANNEL_ID'].to_i
+
+# Only run one instance
+exit unless ENV['BOT_ENABLED']
```
sadleb, line 12: Just to make sure I understand, we just wouldn't set this in any env other than prod. In prod, since we're not specifying a formation it will default to one `bot` dyno (looks like anything not named `web` is considered a worker dyno and won't receive HTTP requests from the router). If we accidentally scale the `bot` formation, that would be a problem but we just shouldn't do that. I think this makes sense and works pretty cleanly.

rshipp: It's actually simpler than that, by default the `bot` proc gets assigned 0 dynos, and the command never runs. On whichever app we want to run this, we go in and give it 1, and it starts the bot. This env var is just an extra convenience so we don't have to rely on setting/unsetting BOT_TOKEN or something.

lib/discord_bot.rb
```diff
@@ -0,0 +1,572 @@
...
+require 'filter_logging'
+require 'honeycomb_sentry_integration'
+
+require_relative '../config/initializers/restclient_instrumentation'
+require_relative '../config/initializers/honeycomb_sentry_integration'
+
+PROGRAM_ID = ENV['BOT_PROGRAM_ID']
+BOT_TOKEN = ENV['BOT_TOKEN']
+SERVER_ID = ENV['BOT_SERVER_ID'].to_i
+GENERAL_CHANNEL_ID = ENV['BOT_GENERAL_CHANNEL_ID'].to_i
```
rshipp, line 15: these will get deleted later, they're for development only

lib/honeycomb_sentry_integration.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+# Set up Sentry->Honeycomb integration.
+require 'sentry-ruby'
+
+# Our Honeycomb team name; be sure to change if using a different team.
+HONEYCOMB_TEAM = 'braven'
+
+# Prepend this module to Honeycomb::Client in order to add Honeycomb
+# trace URLs to Sentry events.
```
sadleb, line 9: Whoa, this is super cool!

lib/salesforce_api.rb
```diff
@@ -15,7 +18,7 @@ class SalesforceAPI
   SFContact = Struct.new(:id, :email, :first_name, :last_name)
   SFParticipant = Struct.new(:first_name, :last_name, :email, :role,
                              :program_id, :contact_id, :status, :student_id,
-                             :cohort, :cohort_schedule)
+                             :cohort, :cohort_schedule, :id, :discord_invite_code)
```
sadleb, line 21: For future readers, `:id` is the actual participant ID, not some new discord ID. It would be nicer to be at the beginning, but then we'd have to hunt down everywhere we initialize one of these and re-order the params. Actually, there are only a handful of `SFParticipant.new` calls in the codebase, so maybe worth doing in a follow-up. It's only initialization that matters, since we access the values by attribute after initialization.

lib/salesforce_api.rb
```diff
@@ -170,6 +173,28 @@ def get_participant_id(program_id, contact_id)
...
+                 "WHERE Discord_Invite_Code__c = '#{discord_invite_code}'"
+    response = get("#{DATA_SERVICE_PATH}/query?q=#{CGI.escape(soql_query)}")
+    participant_record = JSON.parse(response.body)['records'][0]
+    if participant_record
+      info = get_participants(participant_record['Program__c'], participant_record['Contact__c'])
+      if info
+        participant = SalesforceAPI.participant_to_struct(info[0])
+        participant.id = participant_record['Id']
+        participant
+      end
```
sadleb, line 190: This works, but I was imagining that you were going to write a `get_participant_info` method similar to `get_contact_info` as a step towards moving away from that APEX based `get_participants` method.

lib/salesforce_api.rb
```diff
@@ -201,6 +226,21 @@ def get_cohort_names(program_id)
     recursively_map_soql_column_to_array('Name', [], initial_api_path)
   end
 
+  def get_cohort_lcs(cohort_name)
+    # TODO: Fix SQLi!
```
sadleb, line 230: fyi: last time I tried to do a nested SOQL query, I think I found out it wasn't supported. We may have to make multiple api calls.

lib/salesforce_api.rb
```diff
@@ -2,6 +2,9 @@
 
 require 'rest-client'
 
+# TODO: Get rid of Rails dependency in this lib.
+require 'rails'
+
```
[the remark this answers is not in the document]
rshipp: It's not only required for discord bot, it just doesn't die when we use it from platformweb because we've already required rails there. The one thing I know depends on this is a `String.present?` call, which is added by something in activesupport. I could change this to just explicitly require whatever adds that method, but I'm not sure how to test enough code in the lib to verify that's all it needs. Unless maybe you have a better way to go about looking for things we use in here that aren't part of Ruby-stdlib?

lib/restclient_instrumentation.rb
```diff
@@ -8,9 +10,16 @@ module RestClientInstrumentation
   def execute &block
     ret_val = nil
 
+    # Use Rails.logger if present, otherwise make our own.
+    if defined?(Rails) && Rails.logger
+      logger = Rails.logger
+    else
+      logger = Logger.new(STDOUT)
```
sadleb, line 17: Have a look at where `./config/environments/production.rb` checks `RAILS_LOG_TO_STDOUT` for how we setup the logger in Rails. I'm not familiar with the `log_formatter` and `ActiveSupport::TaggedLogging` stuff, but we should make sure the logs this outputs more or less match our other stuff, especially that it has timestamps.

Aside, that `RAILS_LOG_TO_STDOUT` env var is now obsolete. It was when we were switching from AWS to Heroku.

sadleb: Also, this code is probably lightning fast, but generally for something like this that only needs to happen once I've gotten accustomed to using the pattern where you create a class method and read or set an instance variable, like this:
```
def self.logger
  @logger ||= begin
     ... your logic here ...
   end
end
```
Then you can call `RestClientInstrumentation.logger('blah')`

The "instance variable in a class method" thing works as a Singleton pattern of sorts b/c a class in Ruby is an instance itself.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+INVITE_MAX_AGE = 0
+
+# Number of times an invite can be used (0 = infinite).
+# Note: We delete invites once they are used successfully, and
+# the Discord API counts invites as "used" even when someone starts
+# to sign up but encounters an error, so it's safer to set this to 0.
+INVITE_MAX_USES = 0
+
+# Global logger.
+LOGGER = Logger.new(STDOUT)
```
sadleb, line 26: I'm assuming that Rails doesn't load this file into the normal app, but is that true? I don't think this line of code would matter, but thinking about it made me question whether we need to exit the bot early if it accidentally gets loaded in the Rails env.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+
+class DiscordBot
+
+  def initialize(
+    token: ENV['BOT_TOKEN'],
+    log_level: ENV.fetch('BOT_LOG_LEVEL', 'INFO'),
```
sadleb, line 32: Nit: in prod we do `ENV.fetch('LOG_LEVEL') { :debug }` Should we be consistent with debug being the default?

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+    @start_time = nil
+  end
+
+  def run
+    # Event handler: The bot is fully connected to the websocket API.
+    @bot.ready do |event|
+      on_ready(event)
+    end
+
+    # Event handler: The bot was added to a new server.
```
sadleb, line 68: Can you add context about why we use a server per program? I remember it being something about making the roles and channels easier to manage vs having a single national server and managing everything there.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+  end
+
+  # Shutdown handler.
+  def shut_down
+    stop_time = DateTime.now.utc
+    Honeycomb.start_span(name: 'bot.stop') do |span|
+      Honeycomb.add_field('start_time', @start_time)
+      Honeycomb.add_field('stop_time', stop_time)
+      Honeycomb.add_field('uptime', stop_time - @start_time) if @start_time
+    end
```
sadleb, line 620: We're in a SIGTERM/SIGINT, do we need to tell Honeycomb to send this right now?

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+  #   rescue StandardError => e
+  #     record_error(e)
+  #   end
+  #
+  # Event handlers are attached in the `run` function above like:
+  #
+  #   @bot.event_name do |event|
+  #     on_event_name(event)
+  #   end
+  #
```
sadleb, line 124: Can you mention the threading aspect as well? I wouldn't have thought about needing to make things thread safe across different event handlers since we're normally in a web request/response world and over there we don't have to think about threads very often.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+    record_error(e)
+  end
+
+  #
+  # Top-Level Utilities
+  #
+  # Functions called directly by one or more event handler.
+  #
+
+  # Initialize/update @invite cache for all servers.
```
sadleb, line 211: I think this comment needs more context and a link to that article about how comparing used invites to new invites is the only way to determine which invite code or person it is joining the server.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+              Honeycomb.add_field('participant.id', participant_id)
+
+              if participant.discord_invite_code
+                # Fetch the existing code if there is one.
+                LOGGER.debug "Found existing invite for participant #{participant_id}"
+                invite_code = participant.discord_invite_code
+              else
+                # Otherwise, create a new one and save it to the Participant record.
+                LOGGER.debug "Creating new invite for participant #{participant_id}"
+                invite_code = create_invite(server_id)
```
sadleb, line 263: Am I understanding it correctly that all of the code in this method is more or less a duplicate of what happens in the nightly sync, solely for the purpose of being able to call this method?

I wish we could have the nightly sync in the platform code startup a bot in the beginning of sync, wait for on ready, start going through it's normal sync logic, and we just add this `create_invite` stuff there. It just needs to make sure and shut it down after the sync and not subscribe to any events other than on ready, right?

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+  def on_ready(event)
+    Honeycomb.start_span(name: 'bot.ready') do |span|
+      @start_time = DateTime.now.utc
+      @servers = @bot.servers
+      send_to_general "Logged in from #{ENV['APPLICATION_HOST'] || 'unknown host'} at #{@start_time}"
+
+      init_invites
+
+      # TODO: Move this somewhere else maybe?
+      # or stop doing it at runtime at all?
```
sadleb, line 135: Is there anyway to trigger this from the normal sync code in the web app? I could imagine refactoring this bot code a bit so that it can run in two modes. 1) when it comes up it just starts listening for member joins and handles roles and channel stuff 2) when it comes up it just waits for on ready and then runs the sync logic to create invite codes, then shuts down.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+      sync_salesforce
+    end
+  rescue StandardError => e
+    record_error(e)
+  end
+
+  def on_server_create(event)
+    Honeycomb.start_span(name: 'bot.server_create') do |span|
+      LOGGER.debug "Joined new server #{event.server.id}"
+      Honeycomb.add_field('server.id', event.server.id)
```
sadleb, line 145: Similar to what we do for restclient, should the discord fields have their own prefix? If not, they should at least have discord_bot in the name, right? Or does all of this go to a different dataset, not mixed in with platform stuff, so we don't have to add `discord` in the field names?

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+
+      # TODO: Move this somewhere else maybe?
+      # or stop doing it at runtime at all?
+      sync_salesforce
+    end
+  rescue StandardError => e
+    record_error(e)
+  end
+
+  def on_server_create(event)
```
sadleb, line 142: How does a server get created? Do we do that manually? Or is the plan to have the code automatically create one per program that is missing one?

I may answer my own question once I get further down in the code.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+      Honeycomb.add_field('user.username', user.username)
+      Honeycomb.add_field('user.discriminator', user.discriminator)
+      Honeycomb.add_field('user.id', user.id)
+      Honeycomb.add_field('server.id', server.id)
+      LOGGER.debug "Member join: '#{user.username}##{user.discriminator}' (#{user.id})"
+
+      # TODO: make sure server.invites isn't cached
+
+      found_invite = false
+      server.invites.each do |invite|
```
sadleb, line 175: Without having read that article, this code is very confusing.

lib/discord_bot.rb
rshipp: Debug is waaaay too much info for the bot (specifically, what Discordrb sends there), it's overkill unless you're actively developing. IIRC we have limits on how much we can send to Papertrail before it starts discarding new logs. I also send invite codes to the logs at debug level, which we might not want?

Actually, I just checked and at INFO level, Discordrb logs every message in all Discord servers, including private channels, and I don't think we want to deal with the prvacy/legal implications there. I might take out the Discordrb logging altogether, or set it with a separate flag or something. Also need to double-check it doesn't read from `ENV['LOG_LEVEL']`, because that would be an issue if we're using DEBUG in our prod apps.

lib/discord_bot.rb
rshipp: This is for the uptime monitoring. We could take it out if we're done trying to improve startup speed, but it doesn't hurt anything to leave in. SIGTERM is a warning-level "stop whenever you get around to it" signal; SIGKILL is the "stop right now" signal. That's why Heroku (and operating systems in general) sends SIGTERM first, waits for whatever timeout interval they have, and then sends SIGKILL. (SIGINT is just the ctrl-c signal, it's only relevant in development.)

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+          @invites[server.id][invite.code] = invite.uses
+
+          # If we found more than one used invite, alert and exit without configuring member.
+          if found_invite
+            LOGGER.warn "Multiple invites used, not sure who's who! Please fix members manually."
+            Honeycomb.add_field('alert.multiple_invites_used', [found_invite.code, invite.code])
+            return
+          end
+
+          found_invite = invite
```
sadleb, line 191: I don't understand this line.

lib/discord_bot.rb
rshipp: Yeah it's all in a different dataset. The recommendation for when to use different datasets is "when events can't be correlated between systems", which I think is very applicable here since one is based on Discord events and the other is HTTP requests. That's why I just put everything in `app`, it makes it easier to read in its standalone dataset.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+
+      Honeycomb.add_field('user.username', user.username)
+      Honeycomb.add_field('user.discriminator', user.discriminator)
+      Honeycomb.add_field('user.id', user.id)
+      Honeycomb.add_field('server.id', server.id)
+      LOGGER.debug "Member join: '#{user.username}##{user.discriminator}' (#{user.id})"
+
+      # TODO: make sure server.invites isn't cached
+
+      found_invite = false
```
sadleb, line 174: Can we set this to `nil` instead? I was confused for a bit b/c I thought this was meant to be a `did_find_invite?` bool

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+  def self.configure_member(member, invite)
+    Honeycomb.start_span(name: 'bot.configure_member') do |span|
+      server_id = member.server.id
+      user = member.instance_variable_get(:@user)
+      Honeycomb.add_field('server.id', server_id)
+      Honeycomb.add_field('user.id', user.id)
+      Honeycomb.add_field('invite.code', invite.code)
+
+      # Try to map this invite to a Participant.
+      # Always fetch latest info from SF, don't rely on a cache.
```
sadleb, line 288: Why? (I could think it through, but I think the comment should say why)

lib/discord_bot.rb
rshipp: The event is just poorly named (Discord API is bad) - it's when a server is "created from the bot's perspective". In other words, when the bot joins a new server.

To answer the question though, yes all of the following steps are manual:
* Creating a new Regional Server from the Braven Discord Server Template
* Inviting the bot to the new Regional Server
* Assigning the bot appropriate permissions in the new Regional Server

I think there is an API call that lets you create a server, so it's possible we could have the bot create servers automatically in the future, but this seems to me like a future improvement not important for Fall Launch.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+      role_names << "Coaching Partner" if participant.role == :"Coaching Partner"
+
+      # 'Tuesday, 6pm' -> 'Tuesday'
+      ll_day = (participant.cohort_schedule || '').split(',').first
+      if ll_day
+        role_names << "Fellow: #{ll_day} LL" if participant.role == :Fellow
+        role_names << "LC: #{ll_day} LL" if participant.role == :"Leadership Coach"
+      end
+
+      if participant.cohort
```
sadleb, line 339: What happens if there is no cohort?

lib/discord_bot.rb
rshipp: Ah yeah, I think that would still work. I just set it to false because I can never remember what things are falsy in Ruby, bc it's different than Python.

lib/discord_bot.rb
rshipp: It just doesn't assign any cohort roles. (If we want different behavior, like an alert, that's a good point to raise though.)

lib/restclient_instrumentation.rb
rshipp: How does a class method set an instance variable, I thought that's explicitly what class methods didn't do... I might have to do some reading on this one.

lib/salesforce_api.rb
rshipp: It worked fine in my testing, but I'll be sure to double-check with a cohort that has more than one LC

lib/salesforce_api.rb
rshipp: I was planning to do something else but ended up just trying to get things working with minimal code change. Lmk if you think I should spend more time refactoring something here. I do want the struct though, it's easier to use than the hash.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+      role_names << "Leadership Coach" if participant.role == :"Leadership Coach"
+      role_names << "Teaching Assistant" if participant.role == :"Teaching Assistant"
+      role_names << "Coaching Partner" if participant.role == :"Coaching Partner"
+
+      # 'Tuesday, 6pm' -> 'Tuesday'
+      ll_day = (participant.cohort_schedule || '').split(',').first
+      if ll_day
+        role_names << "Fellow: #{ll_day} LL" if participant.role == :Fellow
+        role_names << "LC: #{ll_day} LL" if participant.role == :"Leadership Coach"
+      end
```
sadleb, line 337: I think it would be easier to understand this block and the block below if you factored them out into methods like `DiscordBot.get_cohort_schedule_role_name` and `DiscordBot.get_cohort_role_name`.

Also, I'm still having trouble wrapping my head around `role` vs `channel`. Is `role` just a mapping to a channel on a server? It would be helpful if you added comments somewhere in this file talking about roles or linking to an article or documentation that clarifies what they are and how they work with channels.

lib/discord_bot.rb
rshipp: I was also assuming it doesn't load this file, but it's a good thing to check.

lib/restclient_instrumentation.rb
rshipp: Oh but if this part

> we should make sure the logs this outputs more or less match our other stuff, especially that it has timestamps

was about making sure we're not logging things from Rails in different formats, I do want to clarify that the second logger only gets made from the bot, because the Rails app always has Rails loaded. 

And the logger does have timestamps, though I'm not sure it's formatted the same as the Rails logs.

lib/discord_bot.rb
rshipp: I'll figure out how to add this to the comments somewhere, because it is confusing. I'll try to give a short overview -

Roles (like `Admin`, `Fellow`, and `Cohort: Susan Thursday`) define permissions (like "can read messages", "can send messages") that apply to users at a global level. When a user has multiple roles, the highest role (literally, the one closest to the top in the Discord Server settings UI) takes precedence.

So with the global-level permissions for a role, you can say people with the `Fellow` role can send messages, and people with the `Staff` role can edit other people's messages, for moderation.

BUT since we need to say things like "people with the `Cohort: Susan Thursday` role can read messages in the `#cohort-susan-thursday` channel, we need more than the global permissions. So Discord lets you define a completely separate set of permissions to both Channel:Role and Channel:Member combos. Each channel has its own settings, where you can both add role or member relationships, and change individual permissions for each combo. This is what I'm doing down in the `permission_overwrites` stuff in one of the later functions. We assign a permission overwrite for the Channel:Role combo that gives e.g. people in `Cohort: Susan Thursday` read and write access to `#cohort-susan-thursday`. Then we assign a permission overwrite for the Channel:Member combo that gives e.g. Susan management access to `#cohort-susan-thursday`. In any other channel (back at the global level), only `Staff` members get management permissions.

lib/discord_bot.rb
rshipp: No, in order to run this from the webapp we'd have to open a persistent websocket connection from our server, which seems incredibly unsafe. (Or implement all our code twice, once in the bot and once using the oauth flow.)

Since nothing in this sync process blocks any other part of the bot, and it runs in its own thread, I think it's fine to leave it in the on_ready handler? That way we don't even have to schedule a task, it's just guaranteed to run at least once a day during the daily restart, or at every deploy. And we'll be able to run it on-demand once I add the `!sync` command. Unless you think we're likely to run into issues with late-adds and need to streamline this more, in which case we may want to look into Discord Webhooks, and see if we can call a webhook from Salesforce like we call the Platform API for the automatic email change thing.

lib/discord_bot.rb
rshipp: I'm not sure which nightly sync you're talking about, I don't think this code is duplicated anywhere. If you mean the stuff in `app/services/`, this definitely isn't duplicated there.

What would be the benefit of putting this in a scheduled task? It might not be too difficult, but I don't see how it would help us here. It's non-blocking since it's in its own thread, and Heroku already guarantees us that it happens at least once a day due to the restarts.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+          # Otherwise, if the role looks valid, try to create it. This should only happen
+          # for Cohort roles. If the name doesn't start with 'Cohort: ', the role name will
+          # be logged and skipped.
+          unless role
+            template_role = nil
+            # TODO: Factor out contsants.
+            if name.start_with? 'Cohort: '
+              template_role = server.roles.find { |r| r.name == 'Cohort: Template' }
+              Honeycomb.add_field('template_role.name', template_role.name)
+              role = server.create_role(name: name, permissions: template_role.permissions)
```
rshipp, line 422: a note once you get to this part (and for permission_overwrites in the next function) - I decided to copy permissions from a template in the server itself so that it would be easier to manage in the future. This way, if we want to change the permissions for future course runs, we just go edit them in the server settings. The alternative would be to hardcode the permissions we want here (as an ugly integer representing the [bitwise permissions flags](https://discord.com/developers/docs/topics/permissions)), which would mean an engineer would have to push a code change to update the permissions. This makes the code a little longer/more confusing than it would be if we were hardcoding, but I think will save us pain later.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+    Honeycomb.start_span(name: 'bot.get_or_create_roles') do |span|
+      Honeycomb.add_field('role_names.count', role_names.count)
+      LOGGER.debug "Fetching/creating roles"
+
+      roles = []
+      role_names.each do |name|
+        Honeycomb.start_span(name: 'bot.get_or_create_roles.role') do |span|
+          Honeycomb.add_field('role.name', name)
+          # Fetch the role if it already exists.
+          role = server.roles.find { |r| r.name == name }
```
sadleb, line 411: Is there a reason we don't cache roles? Do we expect the list of roles to keep growing over time on a server or do we delete them at some point?

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+          end
+
+          unless role
+            # If there's still no role, something went wrong. Log and skip.
+            Honeycomb.add_field('alert.role_create_error', name)
+            LOGGER.warn "Error with role '#{name}', skipping"
+            next
+          end
+
+          roles << role
```
sadleb, line 436: Should all of this be inside the first if block above? It looks to me like if we skip trying to create a role, we'll still hit this line and then log a `role_create_error` even though we never tried to create it.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+              LOGGER.warn "Skipping unknown role '#{name}'"
+            end
+          end
+
+          unless role
+            # If there's still no role, something went wrong. Log and skip.
+            Honeycomb.add_field('alert.role_create_error', name)
+            LOGGER.warn "Error with role '#{name}', skipping"
+            next
+          end
```
sadleb, line 434: Should this be inside the first if block above? It looks to me like if we skip trying to create a role, we'll still hit this line and then log a role_create_error even though we never tried to create it. Alternatively, put a `next` in the block where we skip?

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+          # TODO: Factor out constants.
+          if role.name.start_with? 'Cohort: '
+            # Cohort channel name is based on the role name, downcased, with spaces replaced
+            # with dashes and all other non-alphabetical characters stripped.
+            name = role.name
+              .gsub(/ /, '-')
+              .gsub(/[^\p{Alpha}-]/, '')
+              .downcase
+            category = server.channels.find { |c| c.type == 4 && c.name == 'Cohorts' }
+            channel = server.channels.find { |c| c.type == 0 && c.name == name }
```
sadleb, line 484: It's hard for me to tie this logic to the rest of the role name logic in this file. I'm wondering if it is worth creating either a `Role` class or a `Channel` class that abstracts away some of the nitty gritty of role name <-> channel relationship / logic.

lib/discord_bot.rb
rshipp: i think you're right, this might be a leftover from before i simplified the part above this. will look into it.

lib/discord_bot.rb
rshipp: Nothing in this MVP deletes them, I was planning to do that in the "kick people out at the end of the program" task. I can cache roles, I just didn't think to. Probably not worth it in the MVP though, it'd only matter if we get so many people joining the server at once that we start hitting API rate limits.

lib/discord_bot.rb
```diff
@@ -0,0 +1,625 @@
...
+          unless role
+            template_role = nil
+            # TODO: Factor out contsants.
+            if name.start_with? 'Cohort: '
+              template_role = server.roles.find { |r| r.name == 'Cohort: Template' }
+              Honeycomb.add_field('template_role.name', template_role.name)
+              role = server.create_role(name: name, permissions: template_role.permissions)
+            else
+              Honeycomb.add_field('alert.role_skip', name)
+              LOGGER.warn "Skipping unknown role '#{name}'"
```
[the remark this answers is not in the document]
rshipp: It's unlikely to happen in prod, as long as the server is set up the way we expect (e.g. it already has the Fellow, TA, LC, etc roles). In development though, it happened all the time, bc I kept forgetting which roles I was using.

lib/salesforce_api.rb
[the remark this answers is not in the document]
sadleb: I wonder how hard it would be to tweak our rspec config to not load rails so we can run just the `spec/lib/salesforce_api_spec.rb` and see if it passes.

Maybe for now, just directly require whatever provides `.present?` if that is the only thing that fails for the discord bot. My hunch is that would drastically improve startup time.

lib/salesforce_api.rb
sadleb: I started to try and write out what it would look like to create a `get_participant_info` that queries for all the fields in the struct and returns one. I see why it's a pain now. Let's punt on that for now, but I do think we should eventually write that method.

lib/restclient_instrumentation.rb
sadleb: > was about making sure we're not logging things from Rails in different formats, I do want to clarify that the second logger only gets made from the bot, because the Rails app always has Rails loaded.
> 
> And the logger does have timestamps, though I'm not sure it's formatted the same as the Rails logs.

The timestamps are the main thing I was worried about, so if that's there it's probably not worth spending time on getting the rest of the format to match Rails. That would only be valuable if we started processing the logs in some way that made it easier to filter when troubleshooting. I don't see us doing anytime soon (if ever) since we have Honeycomb.

sadleb: > How does a class method set an instance variable, I thought that's explicitly what class methods didn't do... I might have to do some reading on this one.

Since a class is itself an instance, an instance variable in a class method is a variable on the single instance of the class. It's sort of related to how Ruby lets you open classes and change their definitions. You're basically just changing the singleton class instance.

lib/discord_bot.rb
sadleb: > Actually, I just checked and at INFO level, Discordrb logs every message in all Discord servers, including private channels, and I don't think we want to deal with the prvacy/legal implications there. I might take out the Discordrb logging altogether, or set it with a separate flag or something. Also need to double-check it doesn't read from `ENV['LOG_LEVEL']`, because that would be an issue if we're using DEBUG in our prod apps.

Yeah, I agree that we should disable it for privay/legal reasons but I think we should put it behind a separate flag so that we can enable it if we need to troubleshoot something and have no other way of understanding what's going on.

Good call on checking the `LOG_LEVEL` thing b/c we are using debug in prod.

lib/discord_bot.rb
sadleb: Oh, I supposed this is a root span so it gets sent immediately when the block exits. Let's just leave it in. I was just confused about what triggered the actual send.

lib/discord_bot.rb
sadleb: Starting to make more sense.

lib/discord_bot.rb
sadleb: I'm talking about the stuff in `app/services`. I forgot that we haven't done the task where we have the nightly sync loop over all current and future programs instead of taking in a specific program ID that we have to manage manually each semester / launch. Once we do that, it will boil down to the same thing though. Loop over all current and future programs, loop over all participants, process them.

My concerns are:
1. Product Support / Launch logistics: if we have two different paths / times that the sync runs instead of just a generic sync that takes care of everything, it's more noise. I'm imagining convos like the following happening every launch: "why don't they have a discord invite link when I just ran the sync? ... Oh, the sync doesn't do that? How do I force it b/c person X needs it? ... Oh, I can't and just have to wait a day? ... Why? ..." 
2. SalesforceAPI limits: I don't have a good sense of how often this is going to happen. If it's usually only once a day, then we'll probably be fine to run the same API calls both here and in the nightly sync to do the looping. But if it's many times in a day, that could be problematic if we have lots of participants.

lib/discord_bot.rb
sadleb: I don't foresee us running into a problem with late adds (and we can cross that bridge when we get there if there is one). I was more concerned about the points made in the comment below: https://github.com/bebraven/platform/pull/738/files?file-filters%5B%5D=.lock&file-filters%5B%5D=.rb#r649272396

lib/discord_bot.rb
sadleb: I don't think we want an alert b/c cohort mapping doesn't happen until LL2 or so. If someone was mistakenly not mapped to a cohort but should be, discord actually might be one of the first things that makes us notice that, so maybe a Honeycomb field indicating it happened so we can query for it?

lib/discord_bot.rb
sadleb: Ok, maybe we should start a wiki page on the discord management / setup stuff and link it at the top of this file so folks can get that higher order context when jumping into this code the first time?

lib/discord_bot.rb
rshipp: If people are joining Discord before LL mapping, that's going to be a problem. We only assign roles when people sign up. If we need to assign more roles later on, we'll have to add something to the bot's sync process to look for Participants with Discord User IDs and re-compute their roles.

lib/discord_bot.rb
rshipp: todo: move sync out of on_ready and instead write a scheduled task that passes an argument to the bot to tell it to just run sync, then exit

lib/discord_bot.rb
rshipp: todo: have sync command re-assign cohort roles only for the server it's run on

lib/discord_bot.rb
```diff
@@ -0,0 +1,650 @@
...
+    case participant.role
+    when :Fellow
+      "Fellow"
+    when :"Leadership Coach"
+      "Leadership Coach"
+    when :"Teaching Assistant"
+      "Teaching Assistant"
+    when :"Coaching Partner"
+      "Coaching Partner"
+    end
```
sadleb, line 413: I'm looking at similar Coaching Partner stuff for the Zoom links and realizing that there may not be a `Coaching Partner` role that comes through. I think the roles come from the `Participant`'s `RecordType` and there isn't a CP record type: 
image

rshipp: 🥴 If that's the case, we'll need to figure out how CPs are tracked in Salesforce, because we do need to be able to invite them and they're treated differently than LCs.

sadleb: Looks like CP's are entered as an LC and have their "Coach Partner Role" field on the LC Participant set to something. I haven't dug in yet though.

lib/discord_bot.rb
rshipp: it does not appear to load the file, i put a `puts` at the top and it never printed anything

lib/discord_bot.rb
rshipp: switched default to debug, and set discordrb log level to warn

lib/discord_bot.rb
rshipp: added comment

lib/discord_bot.rb
rshipp: added comment

lib/discord_bot.rb
rshipp: the reason is just "so we don't use old info" but i added that to the comment.

lib/discord_bot.rb
rshipp: todo: add something about this to a wiki page and link from the code

sadleb (approved):

lib/discord_bot.rb
sadleb: Follow-up to fix this: https://app.asana.com/0/1174274412967132/1200467981732244 -- assigned to me for the moment.

PR #740, opened 2021-06-02
If Canvas email gets out of sync, set it to match Platform/Salesforce

sadleb:
### Description

Sync From Salesforce was failing to reconcile and fix Canvas users
who had their login email accidentally set to something different than
Platform and Salesforce. This was for two reasons:
1. We were grabbing the default notification email when returning Canvas users instead the login email. The login email is the one we care about being in sync.
2.  We weren't calling the service to check if Canvas matched Platform after determining that Salesforce and Platform were in sync

Task: https://app.asana.com/0/1174274412967132/1200197329627692

### Test Plan:
* Run Sync From Salesforce
* Sign-up the Participant using the Account Creation link (Highlander Portal Account Create Link field in SF)
* Make sure you can login
* In Canvas, as an admin, change their login email to something else so it doesn't match by looking them up here: https://braven.instructure.com/accounts/1/users and editing their "LOGIN INFORMATION"
* Then try logging in as that user. It should say the user wasn't found in Canvas (or something like that).
* Then run Sync From Salesforce.
* They should be able to login again and the login information you set in Canvas should get set back

### Types of changes

### Checklist:


app/services/sync_from_salesforce_contact.rb
```diff
@@ -25,6 +28,9 @@ def run!
       return
     elsif @user.email.downcase == @salesforce_contact.email.downcase
       Honeycomb.add_field('sync_from_salesforce_contact.email_changed', false)
+      # Ensure that even if Salesforce and Platform are in sync, the Canvas login email
+      # is as well. NOOP if they are.
+      SyncUserEmailToCanvas.new(@user).run!
```
sadleb, line 33: This is the main change.

app/services/sync_user_email_to_canvas.rb
```diff
@@ -43,22 +43,23 @@ def run!
...
+
+      if CanvasAPI.client.get_user_email_channel(@user.canvas_user_id, @canvas_login_email, comm_channels).present?
         CanvasAPI.client.delete_user_email_channel(@user.canvas_user_id, @canvas_login_email)
-      rescue RestClient::NotFound => e_ok
-        # This shouldn't really happen, but if it does we don't care. The important
-        # part is having the new login as a communication channel.
-        Honeycomb.add_field('alert.confirm_user_account.missing_old_canvas_email_channel', true)
-        Rails.logger.info("Skipping delete! There was no email communication channel for #{@canvas_login_email}")
       end
     end
```
sadleb, line 59: The old code was assuming a platform/salesforce email change that needed to be repeated in Canvas. This PR handles fixing up the communication channels more robustly by adding and deleting based on whether they exist.

lib/canvas_api.rb
```diff
@@ -235,13 +239,13 @@ def get_user_communication_channels(user_id)
...
+    channels = comm_channels || get_user_communication_channels(user_id)
     channels.filter { |c| c['type'] == 'email' }.first['id']
   end
 
-  def get_user_email_channel(user_id, email)
-    channels = get_user_communication_channels(user_id)
+  def get_user_email_channel(user_id, email, comm_channels = nil)
+    channels = comm_channels || get_user_communication_channels(user_id)
     channels.find { |c| c['type'] == 'email' && c['address'] == email}
   end
```
sadleb, line 250: This is just to avoid an extra API call if you have to check for the existence of 2 email channels, like we do in `SyncUserEmailToCanvas`

lib/canvas_api.rb
```diff
@@ -190,7 +191,10 @@ def find_user_by(email:, salesforce_contact_id:, student_id:)
 
     return nil if user.nil?
 
-    LMSUser.new(user['id'], user['email'])
+    # login_id is the email they use to login which may be different from the email they
+    # use for notifications. We care about their login email since that is what matches
+    # Salesforce and Platform email fields.
+    LMSUser.new(user['id'], user['login_id'])
   end
```
sadleb, line 198: This is the other main change.

rshipp (approved):
nothing stands out as wrong to me

PR #741, opened 2021-06-07
Allow editing launched courses

sadleb:
### Description
Allow the editing of launched courses. We didn't initially implement this b/c Fellows/LCs/TAs could have already done work in the course and we needed to think through what could go wrong or what impact editing things could have. [Here is my research](https://docs.google.com/document/d/1zjkTkBOBMNE8uEtx3DDwWLSXzeYjCiDB_iV0q48JOyE/) into how things behave / how we should support this.

Task:
https://app.asana.com/0/1174274412967132/1200093973021830

[Here is a follow-up PR ](https://github.com/bebraven/platform/pull/742) that allows staff to force Fellow data to be deleted.

Notes:
- Modules are the main thing we had to implement code for since the side effects from a user experience perspective are the worst. We need to blow away the associated data.
- Projects/Impact Surveys/etc that already had work done in them can have side effects, but they aren't as bad, so we're just going to leave it up to Designers to be careful. E.g. a submission that was complete may now be missing a new field and need to be resubmitted.
- We don't delete associated data for Projects / Impact Surveys / etc. This way, if the designer messed up when they deleted it the data is still there on the backend so we could (with some work) recover the work once its re-published.

#### Caution message:
image

#### Message if you can't Publish Latest
image

#### Message if you can't Unpublish
image

### Types of changes

### Checklist:


app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -50,16 +99,37 @@ def version_name
...
+  def can_publish_latest?
+    !@course_rise360_module_version.has_student_data?
   end
 
-  def destroy_interactions
-    # Delete interactions for deleted modules, so they don't break grading.
-    Rise360ModuleInteraction.where(canvas_assignment_id: instance_variable.canvas_assignment_id).destroy_all
+  def can_unpublish?
+    !@course_rise360_module_version.has_student_data?
   end
```
sadleb, line 108: This implies that TAs/staff/LCs, anyone without a student role, will have their work blown away silently if no student's had done work as well.

app/views/courses/edit.html.erb
```diff
@@ -4,160 +4,186 @@
...
+<% end %>
+
+<br/>
+<h3>Edit Resources</h3>
+<div class="row">
+  <div class="col-md-6">
+    <%= form_with(model: @course, local: true) do |form| %>
+      <% if @course.errors.any? %>
+        <%= render :partial => 'shared/form_errors', locals: { object: @course, message: 'Course Resources failed to update'} %>
+      <% end %>
```
sadleb, line 151: Needed to pull Resources out of Settings into their own section so that we could allow them to be edited while hiding the ability to edit the other settings.

app/controllers/courses_controller.rb
```diff
@@ -63,7 +63,7 @@ def update
   # DELETE /courses/1
   def destroy
     authorize @course
-    @course.verify_can_edit!
+    raise CourseAdminError.new 'Cannot delete a launched course' if @course.is_launched
```
sadleb, line 66: I thought it best to just prevent this for now. It doesn't seem like an actual user story and would be pretty catastrophic if done accidentally.

app/models/course_rise360_module_version.rb
```diff
@@ -13,4 +15,14 @@ class CourseRise360ModuleVersion < ApplicationRecord
   def canvas_url
     "#{Rails.application.secrets.canvas_url}/courses/#{course.canvas_course_id}/assignments/#{canvas_assignment_id}"
   end
+
+  # Returns true if there is any student data that references this instance.
+  # Ignores whether there is data done by folks with a different role; like a TA
+  # Teacher, or Designer.
+  #
+  # Anytime a student (Fellow) opens a Module, a Rise360ModuleGrade record is created.
+  # It's overkill to check the states and interactions as well.
```
rshipp, line 24: could you reference where this happens, assuming it's fresh in your head? i wouldn't know where to look

sadleb:
> this seems good to me. only one non-blocking comment about a comment

Will do in [follow-up PR](https://github.com/bebraven/platform/pull/742)

PR #743, opened 2021-06-11
Add support for generating Zoom links

sadleb:
### Description
This is a cutover of the Zoom automation we [had in the old code](https://github.com/bebraven/platform/pull/430/). Task: https://app.asana.com/0/1174274412967132/1198943782530007

It implements the following features:
1. A "Generate Zoom Links" tool that allows you to upload a .csv file. You add the names and emails of folks who you need to generate unique Zoom links for. Those links will let them join a Zoom meeting that you *manually* set up ahead of time.
1. Logic to automatically generate unique Zoom links for the Meeting IDs specified on the Salesforce Cohort Schedules on the Program as part of the nightly sync. The links are then stored in the Salesforce Participant record, currently in the "Webinar Access 1" or "Webinar Access 2" fields. Once again, the meeting needs to be set up manually. Then it's configured on the Salesforce Cohort Schedule.
1. The ability to force a sync to regenerate the links for Salesforce Participants. 

### Motivation and Context
For feature 2), one of the links is generally used to give access to the onboarding meeting and the other is used to give access to the recurring meeting for LL's where everyone for that night joins and then they go into breakout rooms. The Zoom prefix stuff is all there to help staff manage the breakout rooms. More context in the Slack threads linked from the Asana task.

For feature 3), if we change the meeting password or put them in a new Cohort Schedule, or change their email then we may need to regenerate their links b/c they are now invalid. Since the logic in the nightly sync is only for a first time setup assuming it won't change, the force option is needed.
 
For feature 1) we need the ability to create Zoom meetings outside of the standard ones for the Accelerator, like a virtual Mock Interview event or some other targeted training event. We also can use it to generate placeholder links for overflow when participants who are not fully confirmed/enrolled need to get in.

#### Manually Setting up a Meeting with the correct settings:
image
image
image

#### Add the meeting to Salesforce for Participants in that Cohort Schedule to have links auto-generated
image
image

#### Generate Zoom links from a .csv

image

#### Generate and sync Zoom links to Salesforce automatically for Participants
image

#### Generated Zoom links showing on the Participant record
image

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
[...]

PR #744, opened 2021-06-15
Discord Bot specs

rshipp:
### Description

Specs and small bugfixes for #738.

### Types of changes

### Checklist:


spec/lib/discord_bot_spec.rb
```diff
@@ -136,44 +136,496 @@
...
+      end
+    end
+
+    context 'with multiple used invites' do
+      before :each do
+        allow(invite1).to receive(:uses).and_return(1)
+        allow(invite2).to receive(:uses).and_return(1)
+      end
+
+      it 'returns nil' do
```
sadleb, line 204: Either here or in the code can you add context about why we alert and exit without configuring them? E.g. "The way we detect who a person is is by watching join events come in and finding the newly used invite. There needs to be a 1-1 mapping to have confidence that the new member is the same person as the one with the newly used invite. If we have two members join before we can match them up, we don't know who is who and need to manually figure that out. So we alert ourselves instead."

spec/lib/discord_bot_spec.rb
```diff
@@ -136,44 +136,496 @@
...
+
+    let(:programs) { create(:salesforce_current_and_future_programs, canvas_course_ids: [1, 2]) }
+    let(:program1) { create(:salesforce_program_record, program_id: programs['records'][0]['Id']) }
+    let(:program2) { create(:salesforce_program_record, program_id: programs['records'][1]['Id'], discord_server_id: nil) }
+    let(:participant1) { SalesforceAPI.participant_to_struct(create(:salesforce_participant, program_id: program1['Id'])) }
+    let(:participant1_id) { "a2Y17000001WLxqXYZ" }
+    let(:participant2) { SalesforceAPI.participant_to_struct(create(:salesforce_participant, program_id: program1['Id'], discord_invite_code: nil)) }
+    let(:participant2_id) { "a2Y17000002WLxqXYZ" }
+    let(:participant3) { SalesforceAPI.participant_to_struct(create(:salesforce_participant, program_id: program2['Id'])) }
+    let(:participant3_id) { "a2Y17000003WLxqXYZ" }
```
sadleb, line 232: Nit: in the future I think the names should reflect what they are used for. For each test below I had to intimately know the logic and be able to go back and forth between the test and these variable definitions to understand the test. I think future devs would have an easier time maintaining this if the variables were things like 
```
program1_with_discord
program2_no_discord
participant1_with_discord_invite
participant2_no_discord_invite
participant3_with_discord_invite
```

spec/lib/discord_bot_spec.rb
```diff
@@ -136,44 +136,496 @@
...
+      before :each do
+        allow(sf_client).to receive(:get_participant_info_by).and_return(nil)
+      end
+
+      it 'exits early' do
+        expect(sf_client).not_to receive(:get_contact_info)
+        expect(sf_client).not_to receive(:update_contact)
+        expect(DiscordBot).not_to receive(:compute_nickname)
+        subject
+      end
```
sadleb, line 388: Can this happen? Does it leave us in a state where all future member joins will fail since we'll now have multiple invites? If so, is there anything more graceful we can do?

spec/lib/discord_bot_spec.rb
```diff
@@ -136,44 +136,496 @@
...
+    end
+
+    context 'with CP' do
+      let(:participant) { SalesforceAPI.participant_to_struct(create(:salesforce_participant_lc)) }
+
+      # TODO! Fix CP code.
+      xit 'returns role name' do
+        expect(subject).to eq("Coaching Partner")
+      end
```
sadleb, line 459: See [this PR](https://github.com/bebraven/platform/pull/743) for a new `SalesforceAPI` class method to check if someone is a CP.

spec/lib/discord_bot_spec.rb
```diff
@@ -136,44 +136,496 @@
...
   end
 
   describe 'self.compute_cohort_schedule_role' do
+    subject { DiscordBot.compute_cohort_schedule_role(participant) }
+
+    context 'with Fellow' do
+      let(:participant) { SalesforceAPI.participant_to_struct(create(:salesforce_participant_fellow)) }
+
+      it 'returns role name' do
+        # Note: Monday is hardcoded in the factory's CohortScheduleDayTime.
```
sadleb, line 470: I think we should explicitly set the `CohortScheduleDayTime` like this instead of relying on hardcoded values in the factories:
```
let(:participant) { SalesforceAPI.participant_to_struct(create(:salesforce_participant_fellow, :CohortScheduleDayTime => 'Monday, 7pm')) }
```
(TODO: add a transient variable if the above syntax doesn't work)

spec/lib/discord_bot_spec.rb
```diff
@@ -136,44 +136,496 @@
...
+        expect(subject).to eq("Fellow: Monday LL")
+      end
+    end
+
+    context 'with LC' do
+      let(:participant) { SalesforceAPI.participant_to_struct(create(:salesforce_participant_lc)) }
+
+      it 'returns role name' do
+        # Note: Monday is hardcoded in the factory's CohortScheduleDayTime.
```
sadleb, line 479: Ditto. Specs get hard to maintain when they rely on hardcoded stuff in the factories.

spec/lib/discord_bot_spec.rb
```diff
@@ -136,44 +136,496 @@
...
+
+    context 'with two LCs' do
+      let(:cohort_lcs) { [
+        {'FirstName__c' => 'TestFirst1', 'LastName__c' => 'TestLast1'},
+        {'FirstName__c' => 'TestFirst2', 'LastName__c' => 'TestLast2'},
+      ] }
+
+      it 'returns role name' do
+        # Note: Monday is hardcoded in the factory's CohortScheduleDayTime.
+        expect(subject).to eq("Cohort: TestFirst1 TestFirst2 Monday")
```
sadleb, line 525: The design doesn't call for an `&` sign in there? E.g. `Bob & Alice Monday` vs `Bob Alice Monday`?

sadleb (approved):
Added some "nitty" feedback. Nothing that would prevent a merge.

spec/lib/discord_bot_spec.rb
rshipp: No, and & isn't allowed in channel names so that'd cause issues.

spec/lib/discord_bot_spec.rb
rshipp: this will happen (e.g. a staff member joins the server with a non-unique invite), and does not make anything else fail

PR #745, opened 2021-06-16
Discord bot: sync_salesforce command

rshipp:
### Description

Add a sync_salesforce command that 
1. Allows linking a Program to a Server
2. Allows running the command with no arguments to sync the currently running program already linked to this Server
3. Syncs all the participants tied to the program

### Motivation and Context

Task: https://app.asana.com/0/0/1200423006044068/f

Screen Shot 2021-06-16 at 2 54 30 PM
Screen Shot 2021-06-16 at 2 56 00 PM
Screen Shot 2021-06-16 at 3 13 16 PM

### Types of changes

### Checklist:


lib/discord_bot.rb
```diff
@@ -186,9 +188,9 @@ def on_member_join(event)
 
       Honeycomb.add_field('user.username', user.username)
       Honeycomb.add_field('user.discriminator', user.discriminator)
-      Honeycomb.add_field('user.id', user.id)
+      Honeycomb.add_field('member.id', event.member.id)
```
sadleb, line 191: What's the difference b/n a user and a member? Are there separate IDs? It's confusing to me, without knowing the discord API stuff, to have some fields be `user.blah` and the ID to be `member.id`

lib/discord_bot.rb
```diff
@@ -22,6 +22,19 @@
 # to sign up but encounters an error, so it's safer to set this to 0.
 INVITE_MAX_USES = 0
 
+# Character(s) prepended to a message that make the bot treat it as a
+# command.
+BOT_COMMAND_KEY = '!'
```
sadleb, line 27: Can you add a comment about the slash commands and the gem not supporting it yet which is why we're doing our own homegrown command char?

lib/discord_bot.rb
```diff
@@ -204,6 +206,30 @@ def on_member_join(event)
...
+      if event.message.content.start_with? BOT_COMMAND_KEY
+        server = event.server
+        Honeycomb.add_field('server.id', server.id)
+        # event.author might be a User instead of a Member, but only if the
+        # Member immediately left the server after sending the message and
+        # before we received the event. And calling this 'member.id' makes
+        # it easier to cross-reference with other events.
+        Honeycomb.add_field('member.id', event.author.id)
+
+        if event.message.author.roles.find { |r| ADMIN_ROLES.include? r.name }
```
sadleb, line 221: I assume non-admin or staff users can't give themselves roles? If so, they could call it `My Fake Admin Role` or something and be able to execute bot commands.

lib/discord_bot.rb
```diff
@@ -265,51 +291,32 @@ def sync_salesforce
...
     end
   end
 
+  # At this point, the message event is guaranteed to 1) have been sent by
+  # a server admin/staff-level user, and 2) have started with the bot command
+  # signifier BOT_COMMAND_KEY.
+  def process_admin_command(event)
+    case event.message.content
+    when /^#{BOT_COMMAND_KEY}#{ADMIN_COMMANDS[:sync_salesforce]}/
+      LOGGER.debug "sync_salesforce command"
```
sadleb, line 306: Add the command to Honeycomb too? So we can see what staff members are running what commands.

lib/discord_bot.rb
```diff
@@ -407,12 +414,158 @@ def self.configure_member(member, invite)
...
+
+  def sync_salesforce_participant(server_id, participant)
+    Honeycomb.start_span(name: 'bot.sync_salesforce_participant') do |span|
+      Honeycomb.add_field('contact.id', participant.contact_id)
+      Honeycomb.add_field('participant.discord_invite_code', participant.discord_invite_code)
+      LOGGER.debug "Processing participant"
+
+      # TODO: Have SFAPI.get_participants use SOQL so it returns the Participant ID directly.
+      participant_id = SalesforceAPI.client.get_participant_id(participant.program_id, participant.contact_id)
+      Honeycomb.add_field('participant.id', participant_id)
```
sadleb, line 552: This should be there now in the struct.

sadleb (approved):
All my comments at `nit` level.

lib/discord_bot.rb
rshipp: I think you're interpreting the `include?` incorrectly, this translates to `if role_name is in the admin roles list`, it doesn't match substrings

lib/discord_bot.rb
rshipp: The IDs, according to our data so far at least, are always the same. For the difference, see https://github.com/bebraven/platform/blob/5bafbf755b15d79d18593dd9f0f4115cd367d29f/lib/discord_bot.rb#L179-L183

lib/discord_bot.rb
sadleb: Oh yeah, I was thinking it would match on a substring. But it’s include on an array so it’s an exact match and we already setup those roles / own them so even if they could create their own roles they would be able to use ours.

lib/discord_bot.rb
sadleb: Kk. I guess we’ll just have to break the muscle memory of querying for `user.blah` in this case.

PR #746, opened 2021-06-17
Discord bot: factor out constants

rshipp:
### Description

Clean up inline magic numbers / strings.

### Motivation and Context

Refactor.

### Types of changes

### Checklist:


lib/discord_bot.rb
```diff
@@ -32,8 +32,27 @@
...
 ADMIN_ROLES = [
   'Admin',
-  'Staff',
+  'Braven Staff',
 ]
+# Other role names.
+TA_ROLE = 'Teaching Assistant'
+CP_ROLE = 'Coaching Partner'
+LC_ROLE = 'Leadership Coach'
+FELLOW_ROLE = 'Fellow'
```
sadleb, line 41: It would be nicer if the SalesforceApi and this code shared constants, but prob not worth the effort. Maybe a comment in both places alerting devs that if they change it, go look in the other spot?

rshipp: They aren't tied to each other. These constants are the role names in Discord, whereas I've used SalesforceAPI constants for all Participant roles.

lib/discord_bot.rb
```diff
@@ -611,11 +628,12 @@ def self.compute_cohort_schedule_role(participant)
     ll_day = (participant.cohort_schedule || '').split(',').first
     return unless ll_day
 
-    # TODO: Factor out constants.
+    # TODO: is there any way to move these interpolated strings to constants?
```
sadleb, line 631: Probably could have a factory method that takes the day and returns the constant string. Not worth it IMO. I’m not concerned about perf here unless we actually notice a problem which likely wouldn’t happen until we get into the hundreds of thousands of calls in a short period of time

sadleb (approved):

lib/discord_bot.rb
rshipp: yeah i'm not concerned about perf, i just don't like having arbitrary magical strings `"LC: #{ll_day} LL"` deep in the code where they're hard to update

lib/discord_bot.rb
sadleb: Kk. If we add a new role on the Salesforce side, like Mentee or something, hopefully the dev will grep for the other Salesforce roles like Fellow and realize Discord, Zoom, etc has SF driven role based logic and factor that into the change.

PR #747, opened 2021-06-18
Discord bot: reconfigure members after cohort mapping

rshipp:
### Description

If a Participant found during the `sync_salesforce` process has a Contact with a `Discord_User_ID__c` set, re-run the `configure_member` code to set up roles again, in case cohort mapping happened since the person signed up and we need to add cohort roles.

### Motivation and Context

Task: https://app.asana.com/0/0/1200469955848426/f

### Types of changes

### Checklist:


lib/discord_bot.rb
```diff
@@ -574,6 +533,24 @@ def sync_salesforce_participant(server_id, participant)
         # Fetch the existing code if there is one.
         LOGGER.debug "Found existing invite for participant #{participant_id}"
         invite_code = participant.discord_invite_code
+
+        # If they already had a code, they may have already signed up too;
+        # check their contact record for a Discord User ID.
+        # TODO: if this is returned by apex query now, don't fetch contact yet.
```
sadleb, line 539: ContactId should be in there. No need to query

lib/discord_bot.rb
```diff
@@ -598,6 +630,14 @@ def create_invite(server_id)
     invite.code
   end
 
+  # Get Member object from a Server ID and User ID.
+  def get_member(server_id, user_id)
+    server = @servers[server_id.to_i]
+    return unless server
+
+    server.members.find { |m| m.instance_variable_get(:@user).id.to_i == user_id.to_i }
```
sadleb, line 638: That’s whack we have to do it this way, but ok.

rshipp: The fact that there's no `:user` accessor on `Member` is the bane of my existence

lib/discord_bot.rb
```diff
@@ -574,6 +533,24 @@ def sync_salesforce_participant(server_id, participant)
...
+          # They have already signed up; reassign roles, in case cohort mapping
+          # happened after they signed up and they don't have cohort roles yet.
+          # This also runs if people change cohorts, and will update their roles
+          # appropriately.
+          LOGGER.debug "Found Discord User ID, re-configuring in case roles changed"
+          Honeycomb.add_field('user.id', user_id)
+
+          member = get_member(server_id, user_id)
+          Honeycomb.add_field('member.id', member&.id)
+          DiscordBot.configure_member_from_records(member, participant, contact) if member
```
sadleb, line 552: Are there any performance or API limit concerns with doing this every sync?

rshipp: yes. Fortunately, the discordrb code seems to be smart enough to do good back-offs and retries when it gets rate-limited by the discord api.

sadleb: If so, is there a way to quickly short circuit bc we know it’s been run with current Salesforce info?

sadleb (approved):
Nothing merge blocking. Just perf / API limit questions and suggestions

lib/discord_bot.rb
sadleb: Submit a PR, lol

lib/discord_bot.rb
rshipp: is there a scenario where we know that? people may change cohorts throughout the semester, if e.g. an LC has to drop out and their Fellows are all reassigned.

lib/discord_bot.rb
rshipp: It seems an odd thing to forget, I'm guessing the omission is intentional. i just don't know the reason.

sadleb: Good point. Such a bizarre thing tho.

lib/discord_bot.rb
sadleb: Can we compute the cohort channel role based on current SF data and then see if they have that role. If so, don’t do any of the other roles?

rshipp: we could, but that would result in more API calls (`number of users * (1 call to check roles + 0 or 1 calls to set roles)` vs `number of users * 1 call to set roles`) , not less

sadleb: Hmm. Yeah, let’s not spend time thinking about it / trying to optimize it more. If we run into actual issues, we can revisit. Thanks for humoring me and linking the rate limit topic

rshipp: 🤷 i did it in one line, so I guess that works

PR #761, opened 2021-06-29
Discord bot: Add create_invite command

rshipp:
### Description

Add an admin command to create a new invite, given a SF Contact ID.

### Motivation and Context

Task: https://app.asana.com/0/0/1200518349398049/f

To help staff more quickly resolve Discord issues if an invite code becomes invalidated, or a late-add didn't get an invite code.

### Types of changes

### Checklist:


lib/discord_bot.rb
```diff
@@ -333,6 +334,10 @@ def process_admin_command(event)
       LOGGER.debug "list_misconfigured command"
       Honeycomb.add_field('command', 'list_misconfigured')
       list_misconfigured_command(event)
+    when /^#{BOT_COMMAND_KEY}#{ADMIN_COMMANDS[:create_invite]}/
+      LOGGER.debug "create_invite command"
+      Honeycomb.add_field('command', 'create_invite')
+      create_invite_command(event)
```
sadleb, line 340: Should we start adding a Honeycomb field identifying the person who ran the command (ideally their email)? I think we're going to run into some noise around getting discord setup and it would be nice to be able to see who is actually doing the stuff so we can help teach them how to avoid whatever issues arise through them not quite doing it properly.

lib/discord_bot.rb
```diff
@@ -748,6 +753,94 @@ def list_misconfigured_command(event)
...
+      message_content << "\n❌ No Program found for this Discord server."
+      message_content << "\nFirst, sync this server with a Salesforce Program ID, like:"
+      message_content << "\n`#{BOT_COMMAND_KEY}#{ADMIN_COMMANDS[:sync_salesforce]} a2Y11000002HY5mEAX`"
+      message.edit(message_content)
+      return
+    end
+    Honeycomb.add_field('program.id', program['Id'])
+
+    message_content << "\n✅ Found Program #{program['Id']}..."
+    message.edit(message_content)
```
sadleb, line 794: At this point, we've copy/pasted this code 4 times almost identically with the exception of one of the messages. Worth consolidating into a method?

sadleb (approved):
Just one nit comment.

lib/discord_bot.rb
rshipp: we never have emails in discord events, but we do always set `app.member.id` to the person who ran the command. I can do some best-effort for adding a display name too, but since the `author` of a message event can be either a member or a user, it's kind of a pain to guarantee we find the right info.

lib/discord_bot.rb
rshipp: there's no pass by reference in ruby, so the function return value would be annoying. it'd have to be something like

```ruby
program, message_content = find_server_program(server, message_content)
message.edit(message_content)
```

Is that ok?

lib/discord_bot.rb
rshipp: addressed in 67a512d

lib/discord_bot.rb
sadleb: We decided it's not worth tryign to refactor more at the moment.

PR #763, opened 2021-06-30
Make it easier to see the users in a Honeycomb query result

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200448671778583

Adds the following fields to every span:
* `app.user.id`
* `app.user.email`
* `app.canvas.user.id`
* `app.salesforce.contact.id`

Also adds the following to the span where the Honeycomb user context is added (usually the controller span). This lets us easily see who can login and with what email:
* `user.registered?`
* `user.confirmed?`
* `app.user.unconfirmed_email`

### Motivation and Context
This let's use group nearly any field in any span we're querying for by users.

Example query for folks synced from Salesforce
image

Drilling down into the above trace, all the user info is there for each Participant
image

Example of looking in the `process_action.action_controller` span and being able to see the user email
image

It's also available if you walk up to the root span
image

### Types of changes

### Checklist:


app/models/user.rb
```diff
@@ -193,6 +193,42 @@ def self.with_signup_token(token)
...
+    honeycomb_id_fields_map.each { |field, value| Honeycomb.add_field(key, value) }
+    add_login_context_to_honeycomb_span()
+  end
+
+  # Adds user fields useful to troubleshoot login issues to the current span
+  def add_login_context_to_honeycomb_span
+    Honeycomb.add_field('user.registered?', registered?)
+    Honeycomb.add_field('user.confirmed?', confirmed?)
+    Honeycomb.add_field('user.unconfirmed_email', unconfirmed_email)
+  end
```
sadleb, line 222: I don't think it hurts to always send these values that help with login troubleshooting. Putting them in the span though, not the whole trace.

app/controllers/application_controller.rb
```diff
@@ -64,27 +64,18 @@ def cas_login_url
...
+  # auto-instrumentation) that would also help
+  #
+  # Note that Honeycomb.add_field() auto-prefixes with 'app'. E.g. 'app.user.id'
+  def add_honeycomb_fields
+
+    # Add the common fields that uniquely identify the user to every span in the trace. It's nice
+    # to be able to group whatever you're querying for by user. Depending on the flow or support
+    # ticket information, one or the other of these common fields may be more convenient to use.
+    current_user&.add_to_honeycomb_trace()
+  end
```
sadleb, line 78: This is the meat of the PR. All controllers automatically send the user context to Honeycomb. They add the 3 unique IDs and the email to the whole trace, and the confirmation/registration status to the span.

In certain places, we need to set that explicitly b/c there is no current user, like in the registration code.

We also need to be careful that services that process a lot of users don't get called from a controller or the user context added to those spans would get overwritten. I think we're good there b/c it's only the module grading rake task and the sync enrollments for program job.

app/controllers/salesforce_controller.rb
```diff
@@ -62,17 +62,21 @@ def update_contacts
...
       @canvas_user_id = contact[:CanvasUserId]
-      Honeycomb.add_field('user.canvas_user_id', @canvas_user_id)
+      Honeycomb.add_field_to_trace('user.canvas_user_id', @canvas_user_id)
 
       @new_email = contact[:Email]
       # Note: we don't send these to Canvas. Rely on users do update their own names.
       @first_name = contact[:FirstName]
       @last_name = contact[:LastName]
       @salesforce_contact_id = contact[:ContactId]
+      Honeycomb.add_field_to_trace('contact.salesforce.id', @salesforce_contact_id)
```
sadleb, line 72: Continue to add these here as well in case we fail to find the user. We can at least see which contact it was for.

app/controllers/users/confirmations_controller.rb
```diff
@@ -138,4 +142,9 @@ def configure_permitted_parameters
     devise_parameter_sanitizer.permit(:create, keys: [:uuid, :confirmation_token])
   end
 
+  def add_honeycomb_context(user)
+    Honeycomb.add_field('user.present?', user.present?)
+    user&.add_to_honeycomb_trace()
+  end
```
sadleb, line 148: You'll see this duplicated a bunch below. We should refactor to a helper, but the scope of this PR had already expanded enough. Let's do it next time we're in here touching this stuff.

app/controllers/users/passwords_controller.rb
```diff
@@ -71,4 +74,10 @@ def after_sending_reset_password_instructions_path_for(_)
     users_password_check_email_path
   end
 
+private
+
+  def add_honeycomb_context(user)
+    Honeycomb.add_field('user.present?', user.present?)
+    user&.add_to_honeycomb_trace()
+  end
```
sadleb, line 82: Do the same thing for password reset that we do for confirmations and registration.

app/services/sync_portal_enrollments_for_program.rb
```diff
@@ -27,9 +27,9 @@ def run
         # the block causes Honeycomb to automatically set the 'error' and 'error_detail' fields.
         Honeycomb.start_span(name: 'sync_portal_enrollments_for_program.sync_participant') do |span|
           span.add_field('app.sync_portal_enrollments_for_program.sync_participant_complete', false)
-          span.add_field('app.user.email', participant.email)
           span.add_field('app.salesforce.contact.id', participant.contact_id)
           span.add_field('app.salesforce.student.id', participant.student_id)
+          span.add_field('app.salesforce.program.id', participant.program_id)
```
sadleb, line 32: Unrelated to this PR, but I wanted to filter who got synced by program and couldn't do it easily.

app/services/sync_portal_enrollments_for_program.rb
```diff
@@ -96,6 +92,7 @@ def sync_portal_enrollment!(user, portal_user, participant)
 
   def find_or_set_up_user!(sf_participant)
     user = User.find_by(salesforce_id: sf_participant.contact_id)
+    Honeycomb.add_field('user.present?', user.present?)
```
sadleb, line 95: Just be consistent with the other stuff and make it so that we can tell when new users are created in a sync.

app/services/sync_from_salesforce_contact.rb
```diff
@@ -18,8 +18,6 @@ def initialize(user, salesforce_contact)
   # and Platform are in sync but somehow the Canvas login email get out of sync
   # (most likely an admin or engineer messing it up).
   def run!
-    Honeycomb.add_field('user.email', @user.email)
-    Honeycomb.add_field('user.unconfirmed_email', @user.unconfirmed_email)
```
rshipp, line 22: do we want to keep unconfirmed_email still?

lib/lrs_xapi_mock.rb
```diff
@@ -30,8 +30,6 @@ class LrsXapiMockError < StandardError; end
   def self.handle_request!(request, endpoint, user)
     Honeycomb.start_span(name: 'lrs_xapi_mock.handle_request!') do |span|
       Honeycomb.add_field('lrs_xapi_mock.path', "/#{endpoint}")
-      Honeycomb.add_field('user.id', user.id)
-      Honeycomb.add_field('user.email', user.email)
```
rshipp, line 34: this means they'll be missing from this span, since this isn't a controller. maybe leave them in? unless you took them out for another reason.

app/controllers/salesforce_controller.rb
```diff
@@ -62,17 +62,21 @@ def update_contacts
       # class here: https://github.com/beyond-z/salesforce/blob/master/src/classes/BZ_SyncContactsToCanvas.apxc
 
       @canvas_user_id = contact[:CanvasUserId]
-      Honeycomb.add_field('user.canvas_user_id', @canvas_user_id)
+      Honeycomb.add_field_to_trace('user.canvas_user_id', @canvas_user_id)
```
rshipp, line 65: did we call this `user.canvas_user_id` instead of `canvas.user.id` bc it's important to have both fields here? and if so, can you add a comment to that effect? don't want people copy/pasting this somewhere that doesn't apply.

app/services/sync_from_salesforce_contact.rb
rshipp: oh i just noticed that's part of the values we always send.

lib/lrs_xapi_mock.rb
sadleb: This is most commonly called from `salesforce_controller#update_contacts` which adds these to the trace, so they'll be there. There may be an edge case where that fails and the nightly sync runs this as part of a job and only sets these on a parent span, but I think it's better to generally not duplicate these in lower level spans and just rely on the top level controller or job to handle adding them?

app/controllers/salesforce_controller.rb
sadleb: Good catch. I'll change.

lib/lrs_xapi_mock.rb
rshipp: I think I'm missing something, why would this be called by salesforce controller? This is the mock LRS lib.

sadleb: Whoops. For that one, the `LrsXapiMockController` calls `LrsXapiMock.handle_request!` so it should be in the trace.

rshipp: ok, then I'm just wondering if there's anything we'd want to query on that's only in the handle_request span, where these values would be useful. if you think there's not bc you moved most of the handle_request stuff to the trace too, then that seems fine

sadleb: Just to clarify, when I say "they're in the trace" - I mean that their on every span in the trace. So where we used to call `Honeycomb.add_field('user.id', user.id)` here, we no longer have to b/c they will be on that same span (and all spans in the trace).

Are you asking if there are other fields outside this PR that would be useful since we happen to be in here?

rshipp: From my understanding, add_field_to_trace only adds fields to the current span, each parent span of the current span, and finally the root trace. It does not add fields to sibling or child spans. Let me know if that's wrong though.

With that in mind, I'm wondering if there are other fields in specifically the hande_request span, that we would want to query on with fields that are now in the trace, but not in this span. (Since you're calling add_field_to_trace from a parent of this span, which will result in them not being in this span).

sadleb: Looks like it adds them to child spans too. Here is [an example trace](https://ui.honeycomb.io/braven/datasets/stagingplatform.bebraven.org/result/4pxdYiZsvSz/trace/m8TUR69Mfdn): 
image

rshipp: Oh weird, I guess it just adds them to every parent/sibling/child temporally after the field was added.

Screen Shot 2021-06-30 at 11 26 52 AM

In that case, my question doesn't apply and this all seems fine.

rshipp (approved):

PR #765, opened 2021-07-01
Sends strings instead of integers for IDs to Honeycomb.

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200523444029190

Integers can get truncated if they get too big. Send strings instead.

Below is what I grepped for. Then I manually looked at the code for each to determine if it was an integer value being sent:
* `Honeycomb.*id` 
* `add_field.*id` 

### Motivation and Context
We saw this when doing Discord stuff where some values that were stored in a 
Ruby integer were being truncated when sent to Honeycomb. Make sure we're always
sending IDs as strings so when the day comes that have a trillion records for something
comes they wont' get truncated.

### Checklist:


config/initializers/lti_authentication.rb
```diff
@@ -55,8 +55,7 @@ def finish_authenticate(span, status, message, url = nil, canvas_id = nil, user
       span.add_field('app.lti_authentication.status', status&.to_s)
       span.add_field('app.lti_authentication.message', message)
       span.add_field('app.lti_authentication.url', url)
-      span.add_field('app.canvas.user.id', canvas_id)
-      span.add_field('app.user.id', user&.id)
+      user&.add_to_honeycomb_trace
```
sadleb, line 58: Unrelated to this PR, but since I was here I changed this b/c:
1) this is the point where we have a user from a valid state param, similar to how we'd have current_user for normal devise session based auth. Go ahead and add the user context to the honeycomb trace.
2) the canvas_user_id is sent as part of `add_to_honeycomb_trace`, so remove that line (the user is looked up using that ID btw)

lib/discord_bot.rb
```diff
@@ -360,7 +360,7 @@ def alert_on_unconfigured_members
       LOGGER.warn("Unconfigured members on server #{server_id}: #{members.map {|m| m.display_name}}")
       Honeycomb.add_field('server.id', server_id.to_s)
       Honeycomb.add_field('alert.unconfigured_members', true)
-      Honeycomb.add_field('alert.unconfigured_members.ids', members.map { |m| m.id })
+      Honeycomb.add_field('alert.unconfigured_members.ids', members.map { |m| m.id.to_s })
```
rshipp, line 363: ids inside of a list don't need to be converted, since honeycomb just converts the list itself to a string. but this won't hurt anything either.

rshipp (approved):

PR #767, opened 2021-07-13
Add confirm button, change confirm failed language

rshipp:
### Description

Make people click a button to confirm their email, and stop allowing confirmation by GET request. Change the language on the confirmation error page to make it clearer that if the user has already confirmed, they should just log in instead.

### Motivation and Context

Tasks:

* Confirm button: https://app.asana.com/0/1174274412967132/1200448671778587/f
* Error page changes: https://app.asana.com/0/1174274412967132/1200448671778585/f

Screen Shot 2021-07-13 at 12 14 42 PM
Screen Shot 2021-07-13 at 12 15 01 PM

### Types of changes

### Checklist:


Javier99620 (changes requested):
-

app/views/devise/confirmations/show.html.erb
```diff
@@ -0,0 +1,18 @@
+<% content_for :title, 'Confirm Your Email' %>
+
+<div class="row">
+  <div class="col-md-10 offset-md-1">
+    <h1>Confirm Your Email</h1>
+    <p>To confirm your email address, click the button below.</p>
+
+    <%= form_with(model: User, url: users_confirmation_confirm_path, html: {method: :post}) do |f| %>
+      <div class="field">
+        <%= f.hidden_field :confirmation_token, value: params[:confirmation_token] %>
```
sadleb, line 10: I assume you tested that this works? Aka that the token comes through nested so that `params.dig(:user, :confirmation_token)` is that value posted here.

app/views/devise/confirmations/show_resend.html.erb
```diff
@@ -3,25 +3,23 @@
 <div class="row">
   <div class="col-md-10 offset-md-1">
     <h1>Confirmation failed</h1>
-    <p>Your confirmation link has expired, has already been used, or was invalid.</p>
+    <p>Your confirmation link has already been used, has expired, or was invalid.</p>
+
+    <p>If you have <strong>already confirmed your account</strong>, try <%= link_to 'logging in', cas_login_path %> instead.</p>
```
sadleb, line 8: I forget, did we make the generic `cas_login_path` work better or should we be trying to pass in a `service` param?

sadleb (approved):
No merge blocking comments assuming that you actually tested that confirmation works.

app/views/devise/confirmations/show.html.erb
rshipp: Not sure if i understand, but the `hidden_field :confirmation_token` inside a `form_with(model: User` will send `user[confirmation_token]`. the `params[:confirmation_token]` reads the value from the `?confirmation_token=` url param, it doesn't set the name sent by the form.

app/views/devise/confirmations/show_resend.html.erb
rshipp: don't remember, I didn't touch anything there

sadleb: It's fine to leave it like it was b/c its outside the scope of this PR, but it looks like in registrations_controller we use `cas_login_path(service: CanvasConstants::CANVAS_URL)` b/c Canvas is our main "service" that someone registering would expect to be sent to. It may be better to do the same here just in case the after login stuff doesn't properly redirect to Canvas instead of the home controller.

rshipp: changed it. another reason to get rid of cas 🙃

PR #772, opened 2021-07-15
Sync grades from Canvas to Platform

rshipp:
### Description

New service and rake task (`rake sync:canvas_grades`) to sync assignment submissions and rubric info from Canvas to Platform.

### Motivation and Context

No task.

Blocking data team from building Periscope dashboards to show grades, since Canvas doesn't provide us the correct info for whatever reason.

### Types of changes

### Checklist:


app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+    canvas_course_ids = program_data['records'].map { |p| p['Canvas_Cloud_Accelerator_Course_ID__c'] }.uniq.compact
+    Honeycomb.add_field('sync_canvas_grades.courses.count', canvas_course_ids.count)
+    Rails.logger.info("Found #{canvas_course_ids.count} courses")
+
+    canvas_course_ids.each do |canvas_course_id|
+      Honeycomb.start_span(name: 'sync_submissions') do
+        Honeycomb.add_field('canvas.course.id', canvas_course_id.to_s)
+        # Sync grades.
+        Rails.logger.info("Syncing grades for canvas_course_id=#{canvas_course_id}")
+        submissions = CanvasAPI.client.get_submission_data(canvas_course_id)
```
sadleb, line 22: We should keep an eye on how resource intensive this is. I'm assuming if there are file uploads, it doesn't actually have the file contents but rather a link to the file or something, but I'm guessing that the entirety of the submission data for a course could get pretty big. Actually, should we send a Honeycomb field with the size of this string/hash?

app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+        Rails.logger.info("Found #{submissions.count} submissions")
+        submissions.each do |submission|
+          Rails.logger.info("Syncing canvas_submission_id=#{submission['id']}")
```
sadleb, line 25: Add to honeycomb too?

app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+          data = {
+            canvas_submission_id: submission['id'],
+            canvas_assignment_id: submission['assignment_id'],
+            canvas_user_id: submission['user_id'],
+            canvas_course_id: canvas_course_id,
+            score: submission['score'],
+            grade: submission['grade'],
+            graded_at: submission['graded_at'],
+            late: submission['late'],
+          }
```
sadleb, line 35: Just double checking, none of this has to be `to_s`'d b/c Honeycomb just sees it as a JSON blob and passes it through as a string?

app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+          CanvasSubmission.upsert(data, unique_by: [:canvas_submission_id])
```
sadleb, line 36: Should we check the result of the `upsert` and send an alert field to Honeycomb if it fails?

app/services/sync_canvas_grades.rb
sadleb: Also, if one course fails to sync due to the API throwing, do we want to keep trying the others or fail the whole sync? I assume fail the whole sync like you've written it until we see that actually happening in the wild and it's for a reason where we actually do want to just log that error but try the rest of the courses.

app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+
+          if submission['rubric_assessment']
+            Rails.logger.info("Found #{submission['rubric_assessment'].length} rubric_assessments")
```
sadleb, line 39: Ditto on Honeycomb?

app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+            submission['rubric_assessment'].each do |canvas_criterion_id, assessment_value|
+              Rails.logger.info("Syncing rating for canvas_submission_id=#{data[:canvas_submission_id]}, canvas_rating_id=#{assessment_value['rating_id']}")
```
sadleb, line 41: Ditto on Honeycomb?

app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+              rating_data = {
+                canvas_submission_id: data[:canvas_submission_id],
+                canvas_criterion_id: canvas_criterion_id,
+                canvas_rating_id: assessment_value['rating_id'],
+                comments: assessment_value['comments'],
+                points: assessment_value['points'],
+              }
+              CanvasSubmissionRating.upsert(rating_data, unique_by: [
```
sadleb, line 49: Ditto on checking the return of the upsert looking for an error?

app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+      end
+
+      # Sync rubrics.
+      Honeycomb.start_span(name: 'sync_rubrics') do
+        Honeycomb.add_field('canvas.course.id', canvas_course_id.to_s)
+        Rails.logger.info("Syncing rubrics for canvas_course_id=#{canvas_course_id}")
+        rubrics = CanvasAPI.client.get_rubrics_data(canvas_course_id)
+        Rails.logger.info("Found #{rubrics.count} rubrics")
+        rubrics.each do |rubric|
+          Rails.logger.info("Syncing canvas_rubric_id=#{rubric['id']}")
```
sadleb, line 64: Ditto on Honeycomb?

app/services/sync_canvas_grades.rb
```diff
@@ -0,0 +1,111 @@
...
+          rubric_data = {
+            canvas_rubric_id: rubric['id'],
+            points_possible: rubric['points_possible'],
+            title: rubric['title'],
+          }
```
sadleb, line 69: Ditto on checking return of upsert looking for errors?

NOTE: I'm going to stop repeating the upsert and Honeycomb question below, but if we decide to do that there are more instances below.

db/migrate/20210719173813_create_canvas_submission_rating.rb
```diff
@@ -0,0 +1,18 @@
+class CreateCanvasSubmissionRating < ActiveRecord::Migration[6.1]
+  def change
+    create_table :canvas_submission_ratings do |t|
+      t.bigint :canvas_submission_id, null: false
```
sadleb, line 4: Is this not unique? Many ratings to a single submission, right?

db/migrate/20210719173813_create_canvas_submission_rating.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string :canvas_criterion_id, null: false
+      t.string :canvas_rating_id, null: false
```
sadleb, line 6: Is this the unique one? Only one rating Id?

db/migrate/20210719182155_create_canvas_rubric_criterion.rb
```diff
@@ -0,0 +1,22 @@
+class CreateCanvasRubricCriterion < ActiveRecord::Migration[6.1]
+  def change
+    # Note Canvas has no table for this, the data is serialized inside
+    # the Rubric model instead. We broke it out to make Periscope stuff
+    # easier.
+    create_table :canvas_rubric_criterion do |t|
+      t.bigint :canvas_rubric_id, null: false
+      t.string :canvas_criterion_id, null: false
```
sadleb, line 8: Unique? Maybe not considering the unique constraint below for the rubric_id/criterion_id pair.

sadleb (approved):
Some questions and suggestions, but nothing merge blocking.

app/services/sync_canvas_grades.rb
rshipp: Added the count of each of the record types (submissions, rubric_assessments, rubrics) to HC. Though we'll never receive that count if it crashes before it can count them. Since the API response is paginated, we can also count the number of requests to see how many pages there were. There is a risk the number of submissions could get overwhelming in a real course, and overflow whatever the max variable size is in Ruby, but I thought that was something to worry about when it happens.

If one course fails, i think we should just fix the error, rather than continuing to sync other things.

app/services/sync_canvas_grades.rb
rshipp: This is inside a loop, we'd have to add a nested span. I don't think it's worth it unless we figure out we need this info later.

app/services/sync_canvas_grades.rb
rshipp: None of this goes to honeycomb, this is just getting sent to our database.

app/services/sync_canvas_grades.rb
rshipp: It'd raise an exception if it failed.

db/migrate/20210719173813_create_canvas_submission_rating.rb
rshipp: Many ratings to a single submission means there will be many duplicate submission IDs in this table.

db/migrate/20210719173813_create_canvas_submission_rating.rb
rshipp: Rating IDs are random numbers from 1-1000, likely to be unique per rubric, but definitely not globally unique. See the comment at the top of the rubric ratings migration for more info.

db/migrate/20210719182155_create_canvas_rubric_criterion.rb
rshipp: As per the code-comment above this, rating/criterion IDs are not unique at all and do not represent a database primary key, they're random numbers from 1-10000 stored in a serialized hash in the Canvas Rubric model.

app/services/sync_canvas_grades.rb
rshipp: inside a loop

app/services/sync_canvas_grades.rb
rshipp: inside a loop

app/services/sync_canvas_grades.rb
rshipp: added

db/migrate/20210719182155_create_canvas_rubric_criterion.rb
rshipp: here's some example IDs for reference 
Screen Shot 2021-07-19 at 4 30 51 PM

db/migrate/20210719173813_create_canvas_submission_rating.rb
sadleb: Ah, ok. Weird that’s it not just an incrementing ID like others

sadleb:
Thanks for the responses. Helpful

PR #777, opened 2021-07-22
Remove some of our middleware classes from stacktraces and logging

sadleb:
### Description
These classes are in the stack for everything and can be a red herring
when you see them.

Note: set `ENV["BACKTRACE"]=1` if you run into an issue were you do want the silenced classes in the traces and logs

### Motivation and Context
These annoy me.

#### postgresql_adapter.rb_rb_BEFORE
image

#### postgresql_adapter.rb_rb_AFTER
image

#### lti_rise360_proxy_BEFORE.png
image

#### lti_rise360_proxy_AFTER.png
image

### Types of changes

### Checklist:


config/initializers/backtrace_silencers.rb
```diff
@@ -3,6 +3,13 @@
 # You can add backtrace silencers for libraries that you're using but don't wish to see in your backtraces.
 # Rails.backtrace_cleaner.add_silencer { |line| /my_noisy_library/.match?(line) }
 
+Rails.backtrace_cleaner.add_silencer { |line|
+  /core_ext\/postgresql_adapter/.match?(line) || \
+  /honeycomb_sentry_integration/.match?(line) || \
+  /honeycomb_trace_propagation/.match?(line) || \
+  /lti_rise360_proxy/.match?(line)
```
rshipp, line 10: can you explain the proxy one more? i see the example screenshot but I don't understand in what scenario real errors would be incorrectly pointing to the proxy code

sadleb: Every request goes through the proxy. It’s middleware as part of Rails routing. For all requests but the ones where it actually proxies the request it’s a NOOP. Any exception that happens as part of a request will have that in the stack trace. There is a 0.001% chance we’d want to look in that code to figure out what is going wrong, so it’s just a distraction if you don’t remember that it’s just always there.

sadleb: In other words, Rails also has a ton of stuff that happens before it hits the controller but we don’t want to see it in the stacktraces bc we’re relying on the Rails routing logic not to have a bug.

rshipp: oh ok, i didn't realize everything went through the proxy code, that's annoying.

rshipp (approved):
didn't know you could do this

PR #778, opened 2021-07-22
Sync current/future programs from SF

rshipp:
### Description

Instead of specifying program ids, just sync all current and future launched programs. New command is `rake salesforce:sync_current_and_future`.

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200325198363900/f

### Types of changes

### Checklist:


sadleb:
Once we turn this on, it’s going to start picking up a bunch of test programs too. I haven’t given much thought to how we should handle that. I’m fine with just getting this in and then iterating to reduce noise and to figure out what changes we may want to make to support our testing needs.

sadleb (approved):

sadleb:
> Once we turn this on, it’s going to start picking up a bunch of test programs too. I haven’t given much thought to how we should handle that. I’m fine with just getting this in and then iterating to reduce noise and to figure out what changes we may want to make to support our testing needs.

Also, I haven’t given any thought to ways we may want to short circuit the sync if a future program isn’t really ready to go. E.g. will this pick up programs that haven’t been launched yet so there is no Canvas course associated? If it does pick those up, how does the current code handle it?

rshipp:
The salesforce function already limits by programs with a course ID. Not sure what a test program looks like, but I do think we need to fix the errors we're currently seeing in syncs, because it's very noisy and leads to people ignoring real errors because they're so used to being inundated with fake ones.

sadleb:
> The salesforce function already limits by programs with a course ID. Not sure what a test program looks like, but I do think we need to fix the errors we're currently seeing in syncs, because it's very noisy and leads to people ignoring real errors because they're so used to being inundated with fake ones.

Task to do that: https://app.asana.com/0/search/1200659281191909/1200660480090168

PR #796, opened 2021-07-29
Add custom error handling

rshipp:
### Description

Add custom error handling, and two new error views. One for generic server errors, and one for Pundit errors. Also adds the HTTP request ID to honeycomb spans (it's already in the Sentry tags), so we can find errors easier.

### Motivation and Context

Tasks: 
* https://app.asana.com/0/1174274412967132/1200663189674232/f
* https://app.asana.com/0/1174274412967132/1200523444029192/f

Screen Shot 2021-07-30 at 1 48 23 PM
Screen Shot 2021-07-30 at 1 49 52 PM

### Types of changes

### Checklist:


app/controllers/home_controller.rb
```diff
@@ -7,13 +7,17 @@ def welcome
...
     authorize :application, :index?
+
     # It's nice to get that page that shows all the routes in dev.
-    raise ActionController::RoutingError, "No route matches path /#{params[:other]}" if Rails.env.development?
-    redirect_to home_welcome_path
+    if Rails.env.development? && Rails.application.config.consider_all_requests_local
+      raise ActionController::RoutingError, "No route matches path /#{params[:other]}"
+    end
+
+    render plain: '404 Not Found', status: 404
```
rshipp, line 21: this render change is not really related, but I thought it made more sense than redirecting to the homepage. This also means things that previously redirected to the homepage inside LTI embeds, now 404 instead.

app/controllers/home_controller.rb
```diff
@@ -7,13 +7,17 @@ def welcome
...
+  # in, return a 404 as expected.
   # This is so that bots can't enumerate paths and objects on our server.
   # Force an actual person to go to github to find that out!
   def missing_route
     authorize :application, :index?
+
     # It's nice to get that page that shows all the routes in dev.
-    raise ActionController::RoutingError, "No route matches path /#{params[:other]}" if Rails.env.development?
-    redirect_to home_welcome_path
+    if Rails.env.development? && Rails.application.config.consider_all_requests_local
```
rshipp, line 17: the change in the `if` logic here makes it so we can test error pages in development by setting `consider_all_requests_local = false`.

app/controllers/application_controller.rb
```diff
@@ -83,6 +85,9 @@ def add_honeycomb_fields
     # will clobber all that and just overwrite them with the LtiLaunch values for those fields
     # in all spans in the trace.
     @lti_launch&.add_to_honeycomb_trace()
+
+    # Add the HTTP request ID.
+    Honeycomb.current_span&.add_field('request.id', request.request_id)
```
rshipp, line 90: https://ui.honeycomb.io/braven/datasets/development/result/bXQmhEoYj3M/trace/wvsv1tBxRqM

app/controllers/concerns/rescuable.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    @request = request
+    render 'errors/internal_server_error', layout: 'lti_canvas', status: 500
+  end
+
+  def capture_error(exception)
+    LOGGER.error e.backtrace.join("\n\t")
+      .sub("\n\t", ": #{e}#{e.class ? " (#{e.class})" : ''}\n\t")
+    Sentry.capture_exception(exception)
+    Honeycomb.client.send(:add_exception_data, Honeycomb.current_span, exception) if Honeycomb.current_span
```
rshipp, line 37: this seemed to be the safest/shortest way. if there's no current span, we won't get the exception in HC, but will still get it in Sentry. I don't think it's worth adding code to handle the no-span case, since we should have a span at this point in applicationcontroller.

app/views/errors/internal_server_error.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Oops!</h1>
+<p>Sorry, something's gone wrong.</p>
+<p>If you need help, please <%= link_to 'contact Braven Support', Rails.application.secrets.support_ticket_url %>,
+and include the following error details:</p>
+<pre>Path: <%= @request.path %>
```
rshipp, line 5: i added the path just because people usually crop out the browser's url bar when they send screenshots, and having that info readily available would be way more helpful. the request id below will let us look things up in sentry and/or honeycomb if we need more info. I opted not to include any other information here since it's easy for us to look it up, and we don't want to disclose too much about server errors to users.

rshipp: also note, I tested and this `path` does not include query params, so we won't leak `state` or anything here.

app/controllers/application_controller.rb
sadleb: Can you add to the comment about why we add to the current span, which is non standard in the codebase. E.g “We add it to the span so that it doesn’t get an ‘app’ prefix and matches the built-in request.id field that the Rails integration uses”

app/controllers/concerns/rescuable.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def handle_error_generic(exception)
+    capture_error(exception)
+
+    @request = request
+    render 'errors/internal_server_error', layout: 'lti_canvas', status: 500
+  end
+
+  def capture_error(exception)
+    LOGGER.error e.backtrace.join("\n\t")
```
sadleb, line 34: This line confuses me. What’s with the `LOGGER` constant? Where is `e` defined?

rshipp: I... am not sure how this worked. I copied it from the discord bot without thinking, tested it, saw the error in the logs, and moved on.

app/controllers/home_controller.rb
sadleb: Seems fine. I can’t think of anything where redirecting to the homepage is behavior we rely on.

app/views/errors/internal_server_error.html.erb
sadleb: Makes sense

app/controllers/concerns/rescuable.rb
rshipp: Fixed now
```
platformweb_1      |   ↳ app/models/user.rb:69:in `admin?'
platformweb_1      | /app/app/controllers/home_controller.rb:6:in `welcome': test message (StandardError)
platformweb_1      | 	/app/vendor/bundle/ruby/3.0.0/gems/actionpack-6.1.4/lib/action_controller/metal/basic_implicit_render.rb:6:in `send_action'
```

rshipp: side note, this seems to bypass all the backtrace filtering, so lmk if you know of a better way to get backtraces into the logs.

sadleb: I think we’d have to dig through the Rails code and find how they do it using the backtrace silencer. Not worth it IMO

sadleb (approved):
Lgtm. I’m excited about this PR

PR #782, opened 2021-07-23
DRAFT: Upgrade to Webpack 5 (`webpacker6` gem)

sadleb:
### Description
This is a full upgrade of basically all npm packages and Ruby gems. It builds and the pure Ruby/Rails stuff seems to work. However, the React stuff doesn't. Looks like the `react-rails` gem may not support webpack 5. There is another `react_on_rails` gem that we should look into. Otherwise, we may just have to wait to upgrade or figure out to cut out these gems.

### Screenshots (if appropriate):

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #798, opened 2021-08-02
Javascript package upgrades (aka npm/yarn) 

sadleb:
### Description
Upgrades as many `node` packages as possible.

Task: https://app.asana.com/0/1174274412967132/1200680236811189

#### Notes:
* I tried upgrade bootstrap from 4 to 5 but they changed a bunch of stuff in the core styles. We would have to audit all of our pages and re-style a number of them. Not worth it at the moment.
* I tried upgrading from webpack v4 to v5, but the `react-rails` gem doesn't support it. There is another `react_on_rails` gem that seems like it should, but I tried that and the component simply didn't load. No errors or anything. Need to revisit react support and upgrading that stuff in order to get to webpack@5 (Note: rails/webpacker@v5x installs webpack@4x. So rails/webpacker@6x is for webpack@5x. I kept getting confused by that)
* CKEditor has a bunch of changes, so we're going to punt on upgrading that until we have some time to really go through their changesets.
* The Heroku buildpack is on node v12x, so I'm sticking with node 12 in dev too (in the Dockerfile).

[See here](https://github.com/bebraven/platform/pull/782) were I tried to upgrade to webpack@5x

### Sanity checks
* Modules
* Projects (submit button, for React)
* Capstone Evaluations
* Content Editor
* Braven Resources
* Module Uploader
* Publish Latest Module

### Motivation and Context
After the Fall launch, upgrades will be riskier and unlikely to get prioritized too often. Let's launch with the latest and greatest where possible. Esp. to get rid of security vulnerabilities. The `yarn audit` is down from 19 issues to 3 after this, and all three are for the same package that I don't think is used in prod runtime code.

image

### Types of changes

### Checklist:


package.json
```diff
@@ -40,44 +40,43 @@
...
     "react": "^16.13.1",
     "react-bootstrap": "^1.5.2",
     "react-burger-menu": "^3.0.4",
     "react-dom": "^16.13.1",
     "react-hot-loader": "4.13.0",
     "react-tabs": "^3.2.1",
     "react_ujs": "^2.6.1",
-    "sass-loader": "^9.0.2",
-    "style-loader": "1.2.1",
+    "sass": "1.32.12",
```
sadleb, line 61: I had to pin this one instead of allowing it to get updated minor versions b/c of a ridiculous number of deprecation warnings for ```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($spacer, 2)

More info and automated migrator: https://sass-lang.com/d/slash-div
```

File: package.json [hunk 2]
```diff
@@ -40,44 +40,43 @@
...
+    "sass-loader": "^10.2.0",
+    "style-loader": "^2.0.0",
```
sadleb, line 63: These are not on the latest, b/c anything newer than this needs webpack v5.

spec/javascript/components/CommandButton.test.js
```diff
@@ -1,3 +1,7 @@
+/**
+ * @jest-environment jsdom
```
rshipp, line 2: do we need this in every new js file now? or what's it about?

rshipp (approved):

spec/javascript/components/CommandButton.test.js
sadleb: There is probably a global way to configure it as the default, which is `node`. This was the first thing mentioned though in the docs for how to configure `jsdom` as the `testEnvironment` in the newest version of `jest`. See https://jestjs.io/docs/configuration#testenvironment-string @rshipp

rshipp: ah, I see it's only in the test files, that's not so bad.

PR #803, opened 2021-08-04
Make our links more accessible.

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200411617157219

#### Confirmation failed - BEFORE
image

#### Confirmation failed - AFTER
image

#### Hustle To Career Project - AFTER
image

#### Link in a Module for reference
image

### Types of changes

### Checklist:


app/assets/stylesheets/shared/style.scss
```diff
@@ -74,12 +74,9 @@ p {
...
 
-a {
-	color: $bv-oceanblue-600;
-}
-
+a,
 a:visited {
-	color: $bv-oceanblue-700;
+  color: $bv-oceanblue-500;
 }
```
sadleb, line 80: I didn't think it was worth trying to find two accessible colors for visited vs not. I think that was just carried over from some legacy Join server styling. I don't think it's actually needed or adds value. Do you disagree?

rshipp: https://webaim.org/blog/wcag-2-0-and-link-colors/ doesn't list :visited style as a requirement, and I don't see other things saying it is in a quick search. Do we just not like the default purple? Color-only indicators aren't accessible anyway so I think it's more a UX issue than an accessibility one.

rshipp (approved):

rshipp:
i noticed in one of your screenshots we say "opens in new tab" for an email link, which doesn't make sense. followup for that here https://app.asana.com/0/1174274412967132/1200701153958445/f

app/assets/stylesheets/shared/style.scss
sadleb: Whoops, that comment is outdated. I originally did a color only fix and backed that out in favor of underlines. I’m just not touching the old colors at this point.

PR #814, opened 2021-08-06
Discord bot: Skip sync for Programs with a server we're not in

rshipp:
### Description

Followup to #812. Root cause is the Fall Dry Run program has a server ID of a server the bot isn't in (we never invited the new bot to that server, only the test bot). Before I delete that server ID from the dry run program, I thought I'd handle that scenario in the bot just in case it happens again.

### Motivation and Context

See #812.

### Types of changes

### Checklist:


lib/discord_bot.rb
```diff
@@ -862,6 +862,16 @@ def sync_salesforce_program(program_id)
 
       # Skip Programs that don't have a Discord Server ID.
       return if server_id == 0
+
+      # Skip Programs that have a Discord Server ID for a server we're not
+      # currently in, so deleting or kicking the bot out of one server doesn't
+      # break every other program.
+      unless @servers.has_key?(server_id)
+        Honeycomb.add_field('sync_salesforce_program.not_in_server', true)
```
sadleb, line 870: Should send an `alert` field?

sadleb (approved):
Approved pending question on alert field vs normal.

lib/discord_bot.rb
rshipp: No, cuz I can imagine a scenario where e.g. we end a pilot/dry run, delete the server, and the program continues to run for a while, and don't want to see alerts every day when there's no action that needs to be taken. I just want the other programs to continue to sync without crashing, and be able to easily see why a program is being skipped if we ever look into it.

PR #819, opened 2021-08-12
Add sidekiq as the default queue adapter

rshipp:
### Description

* Remove `bundle frozen` option, update README with correct rebuild instructions for adding new gems.
* Add sidekiq, and set it as the default queue adapter in production and development
* Keep async adapter in test environment, so we don't need redis as a dependency for CI
* Add sidekiq-scheduler for recurring tasks (currently unused, but nice to have handy?)
* Configure sidekiq to be persistent, allow dynamic schedule modification, and read from the default and low_priority queues (the only ones we seem to use)
* Add a redis container to enable sidekiq job processing in local dev envs
* Add a redis config to enable heroku redis/sidekiq in production
* Enable the sidekiq web interface (admin-only), for easy monitoring
* Globally disable sidekiq retry for failed workers (which we don't use) and ActiveJob jobs (which we do use)

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200663189674250/f

Screen Shot 2021-08-18 at 12 04 22 PM

### Types of changes

### Checklist:


config/application.rb
```diff
@@ -49,10 +49,7 @@ class Application < Rails::Application
     # https://app.asana.com/0/1174274412967132/1200138491722566
     ASYNC_ADAPTER_MAX_THREADS = 3
     PUMA_MAX_THREADS = Integer(ENV.fetch("PUMA_MAX_THREADS") { 5 })
-    config.active_job.queue_adapter = ActiveJob::QueueAdapters::AsyncAdapter.new \
-      min_threads: 1,
-      max_threads: ASYNC_ADAPTER_MAX_THREADS,
-      idletime: 60.seconds
-    end
```
sadleb, line 56: I think we should address the above `TODO` in this PR.

config/routes.rb
```diff
@@ -256,10 +259,16 @@
   # Honeycomb Instrumentation Routes
   post '/honeycomb_js/send_span', to: 'honeycomb_js#send_span'
 
+  # Sidekiq
+  authenticate :user, -> (u) { u.admin? } do
+    mount Sidekiq::Web  => '/sidekiq'
+  end
+
```
sadleb, line 266: What does this do? If the user is an admin, then a new `/sidekiq` route is added but otherwise it's not there? Why do we have to do this differently from other routes where we authenticate in the controller but the route always exists.

config/sidekiq.yml
```diff
@@ -0,0 +1,5 @@
+:dynamic: true
+:max_retries: 0
```
sadleb, line 2: Is this the global setting that makes things not retry by default? If so, and a particular job does want to retry, does that take precedence over this global config? I could envision a future dev trying to create a job that retries and changing this config inadvertently causing all jobs to retry on failures and want to avoid that. Maybe worth a comment / guidance on how to create retrying job but not impact all the others?

sadleb: Oh wait, I just saw `sidekiq_options retry: false` in the `ApplicationJob`. I'm just confused how this `max_retries: 0` setting and that `retry: false` settings play together.

Procfile
```diff
@@ -1,3 +1,4 @@
 web: bin/rails server -p $PORT -e $RAILS_ENV
 release: ./scripts/release-tasks.sh
 bot: bundle exec bin/discord_bot
+sidekiq: bundle exec sidekiq
```
sadleb, line 4: Any idea how this works without having a binary in `bin` like the others?

rshipp: binstubs aren't necessary, rspec doesn't have one either. it just loads the executable from wherever the gems are installed.

config/application.rb
rshipp: forgot to ask about this, do we still need to do anything there? It said we could just stop doing the thread stuff once we stopped using async adapter, but I don't know why

config/routes.rb
rshipp: yep. it's different from the others because this controller is in a gem, not our app, so we can't modify it to add auth or use pundit.

config/sidekiq.yml
rshipp: i didn't try using `max_retries: false` here, but setting it to 0 leaves failed jobs in the `dead` set for 6 months https://github.com/mperham/sidekiq/wiki/Error-Handling. setting `retry:false` in the ApplicationJob makes sure failed jobs are just discarded immediately. also, this config applies to sidekiq Workers as well as ActiveRecord jobs, while the ApplicationJob lines only apply to jobs that inherit from that class.

config/application.rb
sadleb: We just need to remove that `ASYNC_ADAPTER_MAX_THREADS` constant and the places it's used. It should no longer factor into the database connection pool config.

rshipp: ok i removed the `pool` line in the config altogether, is that right?

sadleb: I think we want to:
1. Change the `pool` setting to `<%= Integer(ENV['DB_POOL'] || Platform::Application::PUMA_MAX_THREADS %>`
2. Leave the `PUMA_MAX_THREADS` env var in the example file

Basically, the pool should be set to the `threads` setting in `config/puma.rb` and doing the above accomplishes that (and let's us override the db pool setting with something else if we need to for whatever reason).

rshipp: looks like i removed too much. trying again with puma added back

sadleb: yeah, looks good now

rshipp:
ok the heroku stuff looks all good now

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,9 @@
+require 'sidekiq'
```
sadleb, line 1: Do we want this initializer to run the same in all environments? And in all dyno types?

sadleb (approved):
Approved pending the question about the initializer running the same everywhere

config/initializers/sidekiq.rb
rshipp: it just turns off ssl verify for sidekiq->redis, which won't hurt anything. unless you're asking something else i don't understand

sadleb: I guess I was thinking about 1) the require false and whether it impacted startup time for the server 2) whether it actually started some sidekiq stuff we don’t want to start 

sounds like no on 2) and we prob don’t care about 1) at the moment

rshipp: ah I wasn't thinking about startup time, but that would only impact the CI because Sidekiq already patches itself into Rails in prod/dev without this line. i could've put the require in the spec helper, but i thought it made more sense here.

PR #821, opened 2021-08-20
Add discord message scheduler

rshipp:
### Description

Add a new controller/service/job for scheduling Discord messages.

### Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200663189674250/f

Screen Shot 2021-08-20 at 12 49 16 PM
Screen Shot 2021-08-20 at 12 49 47 PM
Screen Shot 2021-08-20 at 12 49 53 PM

### Types of changes

### Checklist:


config/initializers/constants.rb
```diff
@@ -54,3 +55,51 @@ class CanvasConstants
   ACCOUNT_ADMIN_ROLE_ID = 1
 end
 
+# Discord
+class DiscordConstants
```
rshipp, line 59: @sadleb this stuff sucks, is there a better way to get this functionality? enums?

app/assets/stylesheets/layouts/admin.scss
```diff
@@ -12,3 +12,10 @@
  *
  */
 
+h2 {
+    color: #262426 !important;
+}
+
+a.btn {
+    text-decoration: none;
+}
```
rshipp, line 21: @sadleb i added these overrides bc this seemed like the most appropriate place to put them, and the red headings were stressing me out. lmk if there's a better way.

app/views/discord_schedule/index.html.erb
```diff
@@ -0,0 +1,38 @@
...
+
+  <% @servers.each do |server| %>
+    <div class="row mb-4">
+      <div class="col">
+        <h2 id="<%= server[:name] %>">Server: <%= server[:name] %></h2>
+
+        <% @jobs.filter { |j| j[:server_id] == server[:id] }.each do |job| %>
+          <div class="row">
+            <div class="col-sm-9">
+              <p>In <strong>#<%= job[:channel] %></strong> at <%= job[:at] %></p>
```
rshipp, line 21: unfortunately we lose the TZ data once a job is in the scheduler, so this is going to be a little confusing for people since the times in the list view show up in UTC. they are labelled as UTC though.

app/controllers/discord_schedule_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def create
+    authorize :discord_schedule
+
+    run_at = Time.use_zone(params[:timezone]) { Time.zone.parse(params[:datetime]) }
+    server_id = params[:server_id].to_i
+    channel_name = params[:channel]
+    message_content = params[:message]
+
+    # Schedule the job.
```
sadleb, line 39: Can we just add the timezone as one of the things we pass here so that we can read it back out when we’re displaying the index?

app/views/discord_schedule/index.html.erb
sadleb: See my comment above. But basically, can we add the TZ as an instance variable somewhere that is accesible when showing the list of jobs? I imagine this is going to generate a ton of noise if we can’t show the time in the local timezone. 

If not, maybe we can translate the time stamp to the local time of the user? Is there a way to access that?

config/initializers/constants.rb
sadleb: Yikes. I don’t have any good thoughts at the moment. I’m concerned about maintaining this though for every new discord server and would definitely queue up a task to see if there is a way to make this dynamic by reading from SF programs instead of hardcoding them.

app/views/discord_schedule/index.html.erb
rshipp: only in javascript. i'd rather do that than pass an extraneous timezone into the discord job though

config/initializers/constants.rb
rshipp: options are: 

1. hardcode
2. database
3. start another bot instance on every page load and pull the server names and IDs from there

since this only changes when we add a new region, I thought hardcoding was the most efficient. do you think this is worth adding a table for discord server names/ids?

app/controllers/discord_schedule_controller.rb
rshipp: we'd have to pass it as a param into the discord job, which would then discard it because it doesn't care. seems hacky

config/initializers/constants.rb
rshipp: Actually, I suppose we could probably find a way to pull these from salesforce too, so make that option 2.5. it's slower than a database, but faster than a bot.

app/views/discord_schedule/index.html.erb
rshipp: is this format ok? we're kind of limited with what we can do without installing another library

Screen Shot 2021-08-20 at 3 40 48 PM

config/initializers/constants.rb
sadleb: If we only have to do this when there is a new region, hardcoding seems like the best option for now. Can you just add instructions for how to find those IDs / names for someone (like me) who is unfamiliar with the Discord stuff and finds themselves having to add a new one?

app/assets/stylesheets/layouts/admin.scss
sadleb: We really should fix this in the base styles, but that’s a ton of work. I guess I’m fine with this since it’s only for the admin layout. Can you import the variables file and use the constant for the color though?

config/initializers/constants.rb
rshipp: i can. if we're sticking with hardcoding, do you know a way to do that that's less verbose than what i have here?

app/views/discord_schedule/index.html.erb
sadleb: Format seems fine. 

I also don’t think it’s that hacky to pass it into the job even though it’s only used when listing the jobs. Either way. As long as it shows the local timezone, I think that’s important for usability

app/services/send_discord_message.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def run
+    bot = Discordrb::Bot.new(token: @token, log_mode: :info)
+
+    # Event handler: The bot is fully connected to the websocket API.
+    bot.ready do |event|
+      # Send the message.
+      Honeycomb.start_span(name: 'send_discord_message.on_ready') do
+        Honeycomb.add_field('server.id', @server_id)
```
sadleb, line 22: Do we need to do a `to_s` here?

app/services/send_discord_message.rb
```diff
@@ -0,0 +1,47 @@
...
+
+        # If the channel didn't exist, crash. HC/Sentry will be notified.
+        # We could do extra work to alert staff, but let's not until we need it.
+
+        channel.send_message(@message_content)
+
+        # Bye!
+        bot.stop
+      end
+    rescue StandardError => e
```
sadleb, line 38: I think everything will inherit from `StandardError`, but I'd have to read to make sure that's the case. Either way, if the intention is to catch absolutely everything I think we should just use a plain old `rescue` and not specify to only rescue `StandardError`

app/views/discord_schedule/_form.html.erb
```diff
@@ -0,0 +1,72 @@
...
+          <%= f.label('message', 'Message') %>
+          <%= f.text_area('message', required: true, class: 'form-control', placeholder: 'Hi @everyone... ✨') %>
+          <small id="messageTip" class="form-text text-muted">Tip: Press ⌃-⌘-SPACE to use emoji.</small>
+        </div>
+      </div>
+    </div>
+
+    <div class="row">
+      <div class="col">
+        <p>Be sure to double-check the channel name! Scheduled messages with misspelled channel names will not be sent.</p>
```
sadleb, line 60: Does it handle typing the channel both with and without a leading `#`? If not, I think we should handle that.

config/initializers/constants.rb
sadleb: Let's stick with hard-coding. I don't have any good suggestions to make this less verbose.

app/views/discord_schedule/_form.html.erb
rshipp: yeah you're probably right 😞

app/services/send_discord_message.rb
rshipp: yeah I guess even if it's a syntax error we still want to stop the task from hanging, will update

app/services/send_discord_message.rb
rshipp: yes 🙃

PR #822, opened 2021-08-24
Send both signup_token and Platform user.id to SF Contact

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200832773570726

### Motivation and Context
The Platform `user.id` helps folks make decisions when merging Contacts.
If they know that a Contact has been sync'd to the Platform, they should
choose that as a the primary one when merging duplicates so that the
`user.salesforce_id` in Platform stays valid. When merging, the non-primary
Contacts get deleted and only the primary one remains.

### Types of changes

### Checklist:


app/services/sync_portal_enrollments_for_program.rb
```diff
@@ -154,19 +162,6 @@ def find_or_set_up_user!(sf_participant)
     user
   end
 
-  # Generate token and send to Salesforce.
-  def sync_signup_token!(user)
-    # Only do this process once.
-    return if user.signup_token
```
sadleb, line 160: @rshipp I don't see how this could ever happen so I removed it. Lmk if that was a mistake. This `sync_signup_token!` is only called from the above `find_or_set_up_user!` method and only when we create a brand new `User` record b/c we can't find an existing one using the `Contact.Id` passed in as the `user.salesforce_id`.

sadleb:
Code reviewed this offline. Had a big discussion around not introducing yet another ID that we could start cross referencing in a way were we get the wrong records. I think it fundamentally comes down to the fact that we have 3 systems: 1) Salesforce 2) Platform 3) Canvas, and depending on how far along in the process you are there will be a unique ID in each that are all tied together. I made this field read-only in Salesforce, so hopefully that will prevent things from getting out of sync b/c the only thing that sets it is the code that runs when we save the User record.

PR #823, opened 2021-08-24
Sync error messages

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200660480090168

### Motivation and Context
1. We want product support to be able to fix up issues with Salesforce / Platform / Canvas records without involving engineers.
2. We want the transient Salesforce error to not cause problems that require us to re-sync them for things to work.

### Types of changes

### Checklist:


lib/canvas_api.rb
```diff
@@ -45,21 +46,59 @@ def initialize(canvas_url, canvas_token)
...
+  def retry_timeout &block
+
+    # Original call.
+    block.call()
+
+  # The exception we actually saw was a ReadTimeout, but let's handle OpenTimeout too. See:
+  # https://github.com/rest-client/rest-client/blob/master/lib/restclient/exceptions.rb#L202
+  rescue RestClient::Exceptions::Timeout
+    Honeycomb.add_field('canvas_api.retry_success', false)
+    actual_sleep_time = sleep 0.5
```
rshipp, line 86: this doesn't seem to work in `irb`, sleep just returns an int when I try it. is that what you wanted?

sadleb: From my testing in irb, same. It just returns either 0 or 1. I tried many times. That’s why I’m sending the actual wait time to honeycomb so I can roll this back if it doesn’t end up actually sleeping. From my reading it’s dependent on underlying system stuff but “should” work. Maybe I should first try and see if I can get this code to run in the local container so that we can rule out it being a weird irb thing

sadleb: I think the 0 or 1 might be a rounding thing. But I should dig a bit more

rshipp: yeah it seems to be. that just doesn't seem very useful. wouldn't we want to just see 0.5 in that field instead of an arbitrary "closest integer" value?

rshipp (approved):
barring the sleep time question, this seems fine to me.

PR #836, opened 2021-09-08
Remove state param from URLs wherever possible

rshipp:
### Description

The core of this PR is changing from passing `state` params around everywhere, which we then use both to authenticate the user and to look up context from the LTI launch (such as Canvas course/assignment IDs, and LineItem URLs for sending grades back through LtiAdvantageAPI), to passing `lti_launch_id` params around most places, which we then use **ONLY** to look up context from the LTI launch and **NEVER** to authenticate the user.

The exception to the no-state-param rule is on the Grade Details page, because it's inside an iframe and would cause significant pain to TAs if we made it open in a new tab.

See `lti_launch_controller` for more details on the new auth/redirect flow. 

### Motivation and Context

When people share links that have a `state=` param in the URL, they're unknowingly granting other users full access to their account. We want to prevent anyone from doing that by accident.

Task: https://app.asana.com/0/1174274412967132/1200792611556515/f

### Types of changes

### Checklist:


app/controllers/course_resources_controller.rb
```diff
@@ -16,8 +16,8 @@ def lti_show
     if course&.course_resource
       url = Addressable::URI.parse(course.course_resource.launch_url)
       url.query_values = helpers.launch_query
-      @course_resources_url = url.to_s
-      render :lti_show, layout: 'lti_canvas'
+      course_resources_url = url.to_s
+      redirect_to course_resources_url
```
rshipp, line 20: This is a related behavior change to stop showing the intermediate "click here to open course resources in a new tab" page, since we're already showing that in the lti#redirector form.

app/controllers/lti_launch_controller.rb
```diff
@@ -48,18 +48,87 @@ def launch
     # for each request to authenticate them instead of using session.
     sign_in_from_lti(ll)
 
-    authorize ll
+    authorize ll, policy_class: LtiLaunchPolicy
```
rshipp, line 51: because ll can now be nil, and `authorize nil` will look for `NilClassPolicy` instead of `LtiLaunchPolicy`, we have to specify this explicitly now.

app/controllers/lti_launch_controller.rb
```diff
@@ -48,18 +48,87 @@ def launch
...
+
+    authorize ll, policy_class: LtiLaunchPolicy
+
+    # Make sure the redirect path is safe. We never want to redirect to other sites!
+    # Exclude anything that doesn't start with a slash, since redirecting to
+    # `.evil.com` is valid in Rails. :(
+    return unless path.start_with?('/')
+    # Exclude things that parse as URIs instead of paths, like `//evil.com`.
+    # This may raise URI::InvalidURIError if the path is malformed. That's fine.
+    return unless URI(path).host.nil?
```
rshipp, line 129: this is still dangerous, but it's the best i could come up with. see https://cheatsheetseries.owasp.org/cheatsheets/Ruby_on_Rails_Cheat_Sheet.html#redirects-and-forwards

app/controllers/lti_launch_controller.rb
```diff
@@ -70,11 +139,24 @@ def self.generate_state
...
+
   # Grab the user ID out of the payload and tell devise that they are authenticated for this session.
   def sign_in_from_lti(ll)
-    if user = ll.user
+    user = ll&.user
+    if user
       sign_in user
       Rails.logger.debug("Done signing in LTI-authenticated user #{user.email}")
+    elsif ll.nil?
+      Rails.logger.debug("No LtiLaunch found for the given state")
```
rshipp, line 159: this is a new failure case since previously `ll` could never be nil (LtiLaunch.current would raise instead)

app/models/lti_launch.rb
```diff
@@ -63,7 +78,7 @@ def resource_link_request_message
 
   # The user that this launch is for.
   def user
-    User.find_by(canvas_user_id: request_message.canvas_user_id)
+    User.find_by(canvas_user_id: request_message&.canvas_user_id)
```
rshipp, line 81: fix NoMethodError when request_message doesn't exist

app/policies/lti_launch_policy.rb
```diff
@@ -4,8 +4,15 @@ def login?
   end
 
   def launch?
+    # False if there was no LtiLaunch matching the given state.
+    return false if record.nil?
```
rshipp, line 8: this is a new failure case we need because `record` (which is the LtiLaunch) can now be nil instead of LtiLaunch.current raising an exception.

app/views/project_submissions/edit_submitted.html.erb
```diff
@@ -4,6 +4,7 @@
 <p>Can't edit previous submissions.</p>
 <p>Try editing your
 <%= link_to 'latest submission', new_course_project_version_project_submission_path(
-  @course_project_version,
-  state: @lti_launch.state) %>
+    @course_project_version,
+    lti_launch_id: @lti_launch.id,
+  ), 'data-turbolinks': false %>
```
rshipp, line 9: the turbolinks thing fixes an unrelate bug i found in project resubmissions, where the first time you load the resubmission page it doesn't fetch your answers.

app/views/attendance_event_submissions/_form.html.erb
```diff
@@ -139,7 +139,7 @@
           <% end %>
         <% end %>
 
-        <input type="hidden" name="state" value="<%= @lti_launch.state %>">
+        <input type="hidden" name="lti_launch_id" value="<%= @lti_launch.id %>">
```
rshipp, line 142: some/all of these might not be needed, but i didn't want to test that right now

lib/lti_id_token.rb
```diff
@@ -34,8 +34,8 @@ def self.parse_and_verify(signed_jwt_id_token)
 
   # Parses the payload into the appropriate LtiRequestMessage
   def self.parse(payload)
-    message_type = payload[LtiLaunchRequestMessage::MESSAGE_TYPE_CLAIM]
-    MESSAGE_TYPE_TO_CLASS[message_type].new(payload)
+    message_type = payload&.dig(LtiLaunchRequestMessage::MESSAGE_TYPE_CLAIM)
+    MESSAGE_TYPE_TO_CLASS[message_type].new(payload) if message_type
```
rshipp, line 38: behavior change: used to raise exceptions -> now returns nil

PR #815, opened 2021-08-09
Module Grading enhancements

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200706368146060

Completely broken first pass at refactoring to get feedback.
    
1. Creates a new `grade_module_for_user` service
2. Refactors grading logic out of `grade_modules` and `grade_module_for_user_job` to use this new service
3. Introduces a tentative approach at fixing the following scenarios:
  * Regrade a completed Module with no new interactions as soon as the date change is detected instead of waiting for the due date to pass.
  * Fix accidental 'manual grade overrides' if the computed grade is higher. It's easy to accidentally edit a grade and even if you put it back, it's currently stuck in the "manually graded" mode. This reverts to auto-grading if the auto-grade is higher. There isn't a reason to give a lower manual grade that I can think of.
  * **TODO: what else?**

### Deploy Notes
**TODO:** run script to assign on-time credit to existing folks? I think I can just let grading run and everyone with 100% will get regraded but that's fine?
**TODO**: communicate the hard-coded 10 point thing strongly to Designers when we deploy this and make sure we go do this task soonish.
**TODO:** communicate to regions that we're showing the name of the person who manually graded an assignment so they're not blindsided. It's another opportunity to "train" on the manual grade override feature too.

#### Laptop (1366px screen)
image

#### When new calculated grade was just sent to Canvas
image

#### When grade is manually overridden by staff/TA/LC
image

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #839, opened 2021-09-13
Fix the "grade_manually_overridden?" logic.

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200954533308131

We were setting the `@existing_canvas_grade `when parsing out the
submission but then setting it back to nil in the initializer.

Also fix the short-circuit for `needs_grading?` to only run it once.
`@needs_grading` is a bool so we need to explicitly check for it being
nil to know if we should return the memoized value vs running the logic.

#### Steps to reproduce:
1. Fully complete a module (sometimes other grading scenarios can cause the code in question not to run, so this gets us into a clean state since no more changes can happen)
2. As a TA (a different one from the Canvas API user), manually give the submission a higher grade than the computed one
3. Run the GradeRise360Modules.new.grade_assignment(assignment_id, [user_id]) method for that assignment/user
4. Refresh the submission page and the grade will revert to the computed one before this fix.

### Types of changes

### Checklist:


sadleb:
Foregoing code review b/c we already reviewed it here: https://github.com/bebraven/platform/pull/815 and this is just a small hotfix that doesn't change any of the core logic.

PR #840, opened 2021-09-15
Set all Module submissions to be on Jan 1, 1970

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200954533308126

The Canvas UI will show "Late" in red if the submitted_at time is after
the due date. We create a single placeholder submission the first time
a Fellow opens the Module and use that to post auto-grades too. The
"Submitted At" time has no meaning in terms of their grade since it's
the timestamp of when they opened it. If they open it for the first time
after the due date, it shows "Late", otherwise is always shows on-time.
This can be confusing.

With this change, "Late" will never be shown
and the non-sensical date of Jan 1, 1970 should make support easier
if Fellows needs help understanding their grade and why they didn't
receive on-time credit. It'll be easier to say "ignore that, it doesn't
mean anything" when it's 1970. The on-time credit is given when you reach 
100% completion before the due date, which is explained on the grades page.

Note: the proper fix for this would be to send a new submission to Canvas
everytime they do more work. You can't update just the submited_at time.
That would require a major refactor and may put unecessary load on the server
in terms of API calls and database queries in addition to handling all the
tricky logic to be able to send the proper grade everytime (for example,
we need the submission to determine if the grade was manually overridde
which we'd have to make an API call to get and we'd still have to handle
0'ing out grades when the due date passes if they've never opened it as
well as handling extensions as well as clicking around the Module which can
send new progression statements to lrs_xapi_mock). This solution seemed to
be the right balance of acheiving the intended behavior in a simple way.

### Test Plan
1. Open a Module that is past due for the first time, then go to the Grades tab in the left nav. Click that Module and it should not show Late but instead that it was submitted on Jan. 1, 1970
2. For a Module that's never been opened and is past due, run the nightly grading task. The grade should show 0 instead of "-" after it runs. When you go to the Grades page for it as the Fellow it should say "No Preview Available". Now go open the Module as the Fellow. When you go back to the Grades page it should show a message about "Grade was just updated. It may take a minute to show up in the gradebook" and when you refresh it should show the grade but no "Late" indication. Instead, it'll say submitted on Jan. 1, 1970.
    * Now finish the module and open the Grades page. You should get full engagement credit but no on-time credit. No give an extension and open the grades page again. After the grade is updated you should get on-time credit.
#### Submitted At time in the UI
image

#### Confusing "Late" icon we're trying to avoid
image

#### Another place "Late" shows up before this change
image

#### Another place, from the TA view
image

### Types of changes

### Checklist:


lib/lti_score.rb
```diff
@@ -71,26 +87,26 @@ def self.new_full_credit_submission(canvas_user_id, submission_url)
   #                   Failed, and Pending will cause the scoreGiven to be ignored. FullyGraded
   #                   values will require no action. Possible values are NotReady, Failed, Pending,
   #                   PendingManual, FullyGraded
-  def self.generate(user_id, activity_progress = COMPLETED, grading_progress = FULLY_GRADED, submission = nil, score_given = nil, score_maximum = nil, comment = nil)
+  def self.generate(user_id, activity_progress = COMPLETED, grading_progress = FULLY_GRADED, submission = nil, score_given = nil, score_maximum = nil)
     msg = {
       :userId => user_id.to_s,
-      :timestamp => DateTime.now,
+      :timestamp => Time.now.utc.iso8601(3),
```
sadleb, line 93: On my machine both `DateTime.now` and `DateTime.now.utc` output things in UTC. However, I think the safeset thing is to use `Time.now.utc` to ensure we're using the proper timestamp. See `./app/services/compute_rise360_module_grade.rb` and `./app/models/canvas_submission.rb` for two more examples. Also note the `iso8601(3)` just forces subsecond precision which is what the [LtiAdvantage score docs](https://canvas.instructure.com/doc/api/score.html) say to use.

lib/lti_score.rb
```diff
@@ -41,19 +62,14 @@ def self.new_module_submission(canvas_user_id, submission_url)
     # on the Canvas side next time you try to retrieve the score. I think it's bc
     # of the comments in API docs about only PENDING_MANUAL and FULLY_GRADED causing
     # an associated submission to be created for the score line_item.
-    new_pending_manual_submission(canvas_user_id, submission_url)
+    generate(canvas_user_id, IN_PROGRESS, PENDING_MANUAL, basic_lti_launch_submission(submission_url, NON_SENSICAL_SUBMITTED_AT))
```
sadleb, line 65: I tested out that `IN_PROGRESS` works in all scenarios and it makes more sense than `SUBMITTED` which is what we used to pass for this.

Tested:
1. Grading in nightly task when never opened
2. Grading after opening and creating a submission after running Step 1
3. Grading again in nightly task after due date extension after running Step2
3. Grading before nightly task runs
4. Grading when it's 0 before nightly task runs.

So I'm 99% sure we don't have permission issues!

rshipp (approved):

PR #842, opened 2021-09-21
Add module completed at timestamp to grade details

rshipp:
### Description

Add completed_at timestamp to grade details, if a module was completed after the due date. Might need to adjust the styling later.

### Motivation and Context

People noticed the "LATE" stuff was missing, assumed that meant the module was completed on time and our grading stuff was wrong, and gave Fellows late points back. This is a small change that makes it easier to see when a module was completed.

Task: https://app.asana.com/0/1174274412967132/1201022023697605/f

### Types of changes

### Checklist:


app/services/compute_rise360_module_grade.rb
```diff
@@ -195,6 +198,17 @@ def grade_completed_on_time(interactions)
...
+  # otherwise returns nil.
+  def module_completed_at(interactions)
+    # There shouldn't ever be more than one, but select the earliest
+    # one anyway just in case.
+    interactions.where(progress: 100)
+      .order(:created_at)
+      .first
+      &.created_at
+  end
+
```
sadleb, line 211: @rshipp there is a similar bit of code in `grade_rise360_module_for_user` I was trying to think about how to refactor and consolidate into a method or scope on the `Rise360ModuleInteraction` model but they are just different enough that maybe it's not worth doing.

app/javascript/packs/local_datetime.js
```diff
@@ -1,6 +1,6 @@
 document.addEventListener("DOMContentLoaded", _ => {
   // Convert UTC timestamps to the browser's local time.
-  document.querySelectorAll('span.datetime').forEach(e => {
+  document.querySelectorAll('span.local-datetime').forEach(e => {
     const date = new Date(e.innerText);
     e.innerText = date.toLocaleDateString('default', {
```
sadleb, line 5: Is `job[:at]` where this timestamp is set in UTC? If so, does this convert to the local timezone?

rshipp: yep they're both UTC DateTime objects

rshipp: (but this JS doesn't actually depend on them being UTC, it just parses the string and turns it from whatever it happens to be, into browser-reported localtime)

PR #844, opened 2021-09-21
Hotfix: put state back in CLASS assignments

rshipp:
### Description

Hotfix for #836 

### Types of changes

### Checklist:


app/controllers/lti_launch_controller.rb
```diff
@@ -84,7 +84,10 @@ def launch
     # page, AND making people click a button to open the details in a new tab would
     # be painful for anyone going through the speedgrader, we still put the state
     # param directly into the URL.
-    when /#{rise360_module_grade_path('')}/
+    # TEMP FIX: Do the same for the Attendance Event Submission Answers path, since
+    # that shows up in "CLASS: blah" assignments, and we'll be modifying this soon
+    # anyway.
+    when /#{rise360_module_grade_path('')}/, /#{launch_attendance_event_submission_answers_path}/
```
sadleb, line 90: Should we make an Asana task for this and link it from here @rshipp ? Or rather, is this the task you're referring to? https://app.asana.com/0/1174274412967132/1200877445440509 --> note that at the moment I'm unclear if we're going to switch that page to always show in an iframe or not. Right now, some of them do (the ones published before the change) and some don't. We should make a decision and be consistent.

sadleb (commented):
Generally lgtm. I linked to this PR from the CLASS related Asana task.

app/controllers/lti_launch_controller.rb
rshipp: yeah that's the change i was thinking of. and if there's places that doesn't show up in an iframe, we should fix those ASAP (or do the task you linked) because it'll still be exposing a state param. we checked a bunch of stuff but might have missed things if the published courses are inconsistent.

PR #847, opened 2021-09-21
Hotfix: all the other grade details things

rshipp:
### Description

Hotfix for #836 .

### Types of changes

### Checklist:


app/controllers/lti_launch_controller.rb
```diff
@@ -90,7 +90,13 @@ def launch
     # and nowhere else...
```
sadleb, line 90: Ditto on my comment from the other PR. I think I finally understand what you mean by the "Grade Details" page. You just mean any page that shows up in Speedgrader when viewing a submission? I was thinking you were talking about the "Explain Grade" page for a Module. Maybe just clarify this comment to mention Speedgrader and that these paths represent a "submission" @rshipp ?

sadleb: I can't think of a way to do it, but I wish we could tie this bit of code more directly to the controller action so that when people use an existing action as inspiration to create a new submission page viewed in speedgrader, they don't miss the fact that they need to add their new path/action to this list.

PR #850, opened 2021-09-23
Discord Bot: Add end_program command

rshipp:
### Description

Add a command to "end" a program. It does:

1. confirm!!
2. iterate over all server members
    1. if they don't have any "retained" roles (Staff, Admin, etc)
        1. kick them
3. iterate over all salesforce participants
    1. revoke their invites
4. ~~if there is more than 1 invite remaining:~~
    1. ~~show a message, so staff know to audit invites~~
5. iterate over cohort channels
    1. delete them all
6. iterate over cohort roles
    1. delete them all
7. iterate over remaining channels
    1. maybe skip certain channels, like #rules?
    2. rename channel to #channelname-old
    3. clone channel (permissions, name, etc, not content)
    4. delete old channel
8. delete discord server id from program

Note: Step 4 was REMOVED, because we cycle the general channel, which means all its invites will be deleted. Instead, I'll be updating the docs and cycling all Staff links now to be on the #rules channel, so we don't have to worry about those at all. Followup task for this is https://app.asana.com/0/1174274412967132/1200907761238590/f

### Motivation and Context

Tasks:
* https://app.asana.com/0/1174274412967132/1200592615412360/f
* https://app.asana.com/0/1174274412967132/1200411617157237/f

### Types of changes

### Checklist:


rshipp:
code, docs, template, and all servers updated to address above.

elenabaurkot (approved):

lib/discord_bot.rb
```diff
@@ -1257,6 +1417,82 @@ def self.configure_cohort_channel(member, participant, role)
...
+  # Returns true if deleted, false otherwise.
+  def self.delete_channel(channel, reason=nil)
+    channel.delete(reason)
+    return true
+  rescue Discordrb::Errors::UnknownChannel
+    # Probably another instance of the bot got to it first.
+    LOGGER.debug "Channel already deleted"
+    Honeycomb.add_field('channel_delete_failed', true)
+    return false
+  end
```
sadleb, line 1494: Do we want to add the invite/member/channel IDs to Honeycomb so we can tell which one it was for? Or is that done at a higher level consistently?

spec/lib/discord_bot_spec.rb
```diff
@@ -1383,6 +1544,194 @@
...
+
+      it 'handles exception' do
+        expect {
+          subject
+        }.not_to raise_error(Discordrb::Errors::UnknownChannel)
+      end
+
+      xit 'returns false' do
+        # TODO: why does this fail?
+        expect(subject).to eq(false)
```
sadleb, line 1731: Is this a TODO for this PR or for a follow-up PR? If followup, link the Asana task?

rshipp: neither, I just stopped trying to get it to work because I was spending too much time on it and making no headway, and the test isn't worth the effort. the code works, there's just something weird happening with rspec here. left it in more as just a note to our future selves.

lib/discord_bot.rb
```diff
@@ -848,6 +865,156 @@ def create_invite_command(event)
...
+      return
+    end
+
+    Honeycomb.add_field('end_program_flag', true)
+
+    message_content = "✅ Confirmed program end! Looking up pre-configured Program..."
+    message = event.message.respond(message_content)
+
+    # Reset the global flag back to the initial state.
+    @end_program_flag[server.id] = nil
```
sadleb, line 917: What happens if they enter the command, then accidentally enter it twice instead of once. Both runs will be going simultaneously. Will that lead to problems? If so, should we set the flag to false here to indicate that it's running and then check and skip the logic here if a second command comes into before the first one finishes?

lib/discord_bot.rb
```diff
@@ -1257,6 +1417,82 @@ def self.configure_cohort_channel(member, participant, role)
...
+        # want to delete those. Others, like #bot-commands, are just easier to
+        # keep, and have no risk of exposing student PII because they're staff-only.
+        next if RETAINED_CHANNELS.include? old_channel.name
+        LOGGER.debug "Cycling ##{old_channel.name}"
+
+        channel_count += 1
+
+        # Rename the channel to *-old.
+        channel_name = old_channel.name
+        old_channel.name = "#{channel_name}-old"
```
sadleb, line 1440: This name change takes effect immediately by calling into the API and blocking until it completes? I just want to make sure it's not possible to hit the `create_channel` call below before the old channel was renamed in Discord

lib/discord_bot.rb
rshipp: it's theoretically possible, but I don't think there's any way to do it accidentally. even copy/pasting as fast as I could I didn't end up in that state. the command itself is also pretty resilient, since it's designed to support two instances of the bot running it at the same time. if you think it's worth the effort I can try the `false` thing, but I'd have to do a bunch more testing again.

sadleb (approved):
Nothing merge blocking but I added some comments / questions.

Also, I'm curious what happens if a region forgets to run this before the first invites for the next semester go out? I assume we'd just ask them to run it and blow away everyone, then re-invite the folks? Either way, we should make sure that Annie and Jenna have this solidly in their launch checklist to make sure it was run before we launch the new program. Ideally, this would be run automatically as part of launch.... but that's for a future iteration.

lib/discord_bot.rb
rshipp: there's no way for me to prove that, but probably. if it did call create_channel before the rename completed, it'd just raise an exception and stop, and we'd fix it and try again. in a few hundred channel rename/delete/creates, i never ran into that though.

lib/discord_bot.rb
sadleb: not worth it if you've given it thought and think it'll be fine.

spec/lib/discord_bot_spec.rb
sadleb: kk

rshipp:
> what happens if a region forgets to run this before the first invites for the next semester go out? I assume we'd just ask them to run it and blow away everyone, then re-invite the folks?

It'd be a bad thing, let's never do that. Yes we could run it later and re-invite new people, but they'd have already seen all the old messages. I imagine this will be a product support task, after regions confirm they're done with whatever analytics they want, but we'll figure that out when we get there.

> Either way, we should make sure that Annie and Jenna have this solidly in their launch checklist to make sure it was run before we launch the new program. Ideally, this would be run automatically as part of launch.... but that's for a future iteration.

Yep, all this except this should run close to the program end, wayyyy before the next program launch.

lib/discord_bot.rb
rshipp: actually it's easier to get in that state by running the command 4 times in a small window, which will confirm twice. should still be fine though, that's what all the safe delete/kick classmethods were for.

PR #852, opened 2021-09-28
Specs for recent Module grading refactor / bug fix / features

sadleb:
### Description
Task: https://app.asana.com/0/1174274412967132/1200954533308122

A follow-up PR will have fixes for the floating point math bug: https://app.asana.com/0/1174274412967132/1201022023697611

Core PR that these specs are for:
* https://github.com/bebraven/platform/pull/815

Related PRs:
* https://github.com/bebraven/platform/pull/837
* https://github.com/bebraven/platform/pull/838
* https://github.com/bebraven/platform/pull/839
* https://github.com/bebraven/platform/pull/840
* https://github.com/bebraven/platform/pull/842
* https://github.com/bebraven/platform/pull/843
* https://github.com/bebraven/platform/pull/849

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #854, opened 2021-10-01
BUG: Floating point math issue with grade computation

sadleb:
### Description
Fix floating point math bugs. Task: https://app.asana.com/0/1174274412967132/1201022023697611

Move calculation of the total "grade" inside `ComputedGradeBreakdown` so
that we make sure it matches the displayed grade component breakdown.
To do this, switch from sending a percent to Canvas and instead
send the raw score that we compute. Compute that raw score by
adding up the score for each grade component that we show. This way,
the display of the grade breakdown matches the total score it
sums up to which matches what we store in Canvas, therefore avoiding
re-computing b/c of rounding errors.

Also, sneak in a small fix for specs to use a `Time` object instead of a string
for `completed_at` to match the actual code's behavior. See: https://github.com/bebraven/platform/pull/852#discussion_r719785083

### Test Plan
As a Fellow, open a Module and do some work. View the grade breakdown and make sure it's what you expect. Finish the Module AFTER the due date and make sure the "You completed this module late, at BLAH" message has the correct local time.

Ideally, mimic the score for this Fellow on this Module: https://braven.instructure.com/courses/325/gradebook/speed_grader?assignment_id=3515&student_id=1258 (breakdown below) --> but the specs should take care of that so it's not necessary. Just sanity check that it stops recomputing after this is pushed live:

Here is how your Module grade breaks down:
**Engagement:**
3.6 / 4.0 points
_You completed 89% of the Module._

**Mastery quizzes:**
2.4 / 4.0 points
_You got 60% of the 5 mastery questions correct._

**On Time:**
0.0 / 2.0 points
_Awarded when you complete 100% of the Module before the due date._

**Total Grade:**
6.0 / 10.0 points

It used to show as 5.96 after sending the grade to Canvas. Now it should show 6.0 in Canvas.

### Types of changes

### Checklist:


app/models/canvas_submission.rb
```diff
@@ -28,7 +28,7 @@ def self.parse_attributes(submission, canvas_course_id = nil)
       # always get set to the effective due date for that user correctly. If that turns out not
       # to be the case, we'll have to put logic back in that goes through the assignment overrides
       # and calculates the due date manually.
-      due_at: submission['cached_due_date'],
+      due_at: (submission['cached_due_date'] ? Time.parse(submission['cached_due_date']) : nil),
```
sadleb, line 31: Unrelated to this PR. Any changes in here related to due_dates or completed_at times are just that small change to use Time objects instead of strings.

app/services/compute_rise360_module_grade.rb
```diff
@@ -66,99 +66,91 @@ def run
...
-          GRADE_WEIGHTS[:mastery_quiz] * quiz_grade +
-          GRADE_WEIGHTS[:on_time] * on_time_grade
-        )
-      else
-        # If there are no mastery questions, fold the mastery weight in with the engagement
-        # weight, so that engagement is just worth more.
-        total_grade = (
-          (GRADE_WEIGHTS[:module_engagement] + GRADE_WEIGHTS[:mastery_quiz]) * engagement_grade +
-          GRADE_WEIGHTS[:on_time] * on_time_grade
-        )
```
sadleb, line 92: `ComputedGradeBreakdown` is now responsible for calculating the `total_score` from the component grades, so this logic is there now.

rshipp (approved):

PR #869, opened 2021-10-12
BUG: Rate This Module CSS wasn't being applied if it was navigated to (vs initially loaded)

sadleb:
### Description
Follow-up to: 
* https://github.com/bebraven/platform/pull/867

Task: https://app.asana.com/0/1174274412967132/1201149604158822

All the Rise360 DOM stuff is dynamically AJAX loaded. The Rate This Module iframe won't be there
until you navigate to that "lesson". This change makes sure we're always adding the CSS. Before,
it would only happen if you reloaded the page when it's showing or that's the lesson your bookmark
picked up at when re-opening the module.

Note: this doesn't fix the sizing of the iframe which is now not quite right given that we're no longer iframed inside Canvas. See: https://app.asana.com/0/1174274412967132/1201150384153474

### Testing
Make sure the `rate-this-module` CSS is added when browsing to the lesson with the iframe in it `iframe[src*=rate_this_module_submissions]` in the following scenarios:
* Loading a Module the first time (overview page)
* Picking up where you left off on a Mastery Quiz page
* Picking up where you left off on a Lesson (aka a normal page)
* Picking up where you left off which was the Rate This Module page
* 
Notice how the RTM iframe and all parent nodes up to `<section`class="blocks-lesson">` have the `rate-this-module` class added
image

### Types of changes

### Checklist:


elenabaurkot (commented):
Lgtm!

One question about the "rate_this_module_iframe" variable on line 108 

const rate_this_module_iframe = iframeDocument.body.querySelector('iframe[src*=rate_this_module_submissions]:not(.rate-this-module)');

is this selecting all iframes with a src of 'rate_this_module_submission', but not the if it has a class of 'rate-this-module'? Also as curious about the src*= as opposed to src=

sadleb:
> Lgtm!
> 
> One question about the "rate_this_module_iframe" variable on line 108
> 
> const rate_this_module_iframe = iframeDocument.body.querySelector('iframe[src*=rate_this_module_submissions]:not(.rate-this-module)');
> 
> is this selecting all iframes with a src of 'rate_this_module_submission', but not the if it has a class of 'rate-this-module'? Also as curious about the src*= as opposed to src=

Yup, that's what it's selecting. The `src*=` says to do a wildcard match instead of an exact match. So `<iframe src="https://platformweb/rate_this_module_submissions">` is selected and so is `<iframe src="https://platform.braven.org/rate_this_module_submissions">`

app/javascript/packs/rise360_module_versions.js.erb
```diff
@@ -31,28 +31,13 @@ export async function main() {
   // iframe and all parent nodes up to the top one we need to target
   // in rise360_content.scss (if it's there) so that we can distinguish
   // a Rise360 block with an iframed RTM view vs some random thing iframed in.
-  async function addRateThisModuleCSS() {
+  async function addRateThisModuleCSS(rate_this_module_iframe) {
     const honeySpan = new HoneycombAddToSpan('rise360_module_versions', 'addRateThisModuleCSS');
+    honeySpan.addField('iframe_found', 'true');
```
rshipp, line 36: do we have to send strings to hc in js, or could this be boolean `true` instead?

sadleb: just tested. boolean works. fixed.

PR #876, opened 2021-10-14
Updated waiver policy so only fellows can launch waivers

elenabaurkot:
### Description
Updated the waiver policy so that in order to launch waivers the user must be a student enrolled in the course they are trying to open the waiver for. 

### Motivation and Context
Previously, if an LC opened a waiver, it seemed to be working until they tried to submit and received an error message because they aren't a student and we can't create a grade for them. We want to prevent LCs/TAs from opening waivers all together.

Asana task: https://app.asana.com/0/1174274412967132/1200882227720691/f 

From the screen below, the button to load for waivers should only work for fellows
Screen Shot 2021-10-14 at 9 38 38 AM

When a TA/LC clicks the button to load waivers they receive this error message:
Screen Shot 2021-10-14 at 9 38 49 AM

When a fellow clicks the button to load waivers they should be taken to another button that says "click here to get started"
Screen Shot 2021-10-14 at 9 39 28 AM

### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)


app/policies/waiver_submission_policy.rb
```diff
@@ -18,4 +23,25 @@ def create?
...
+  private
+    # TODO: refactor and use EnrolledPolicy here
+    # https://app.asana.com/0/1174274412967132/1199344732354185
+    # Returns true iff user has student enrollment in any section of course
+    def enrolled_as_student_in_course?
+      # get the id of the course we're in
+      courseId = record.id
+
+      # filter user sections and find sections that correspond with this course
+      courseSections = user.sections.select do |section|
```
rshipp, line 36: nit: can you use `snake_case` for these variables instead of `babyCamelCase`? we try to keep snake in Rails code and camel in our JavaScript.

app/policies/waiver_submission_policy.rb
```diff
@@ -5,6 +5,11 @@ class WaiverSubmissionPolicy < ApplicationPolicy
 
   def launch?
     !!user
```
rshipp, line 7: this line doesn't do anything right now, i think you want
```suggestion
    return false if user.nil?
```
The `!!user` with no return only works if it's the last line in the function and you take advantage of Ruby's implicit return.

rshipp (changes requested):
two comments, the rest looks good

app/policies/waiver_submission_policy.rb
elenabaurkot: For sure, just made the updates!

app/policies/waiver_submission_policy.rb
```diff
@@ -18,4 +23,25 @@ def create?
...
   end
+
+  private
+    # TODO: refactor and use EnrolledPolicy here
+    # https://app.asana.com/0/1174274412967132/1199344732354185
+    # Returns true iff user has student enrollment in any section of course
+    def enrolled_as_student_in_course?
+      # get the id of the course we're in
+      courseId = record.id
```
sadleb, line 33: I think it would be cleaner and more consistent if we create an `initialize` method and set a `@course` instance variable. See `project_submission_policy.rb` for an example of how we do this sort of thing elsewhere.

Also, parroting Ryan's comment about snake_case vs babyCamelCase for `courseId`

app/policies/waiver_submission_policy.rb
```diff
@@ -18,4 +23,25 @@ def create?
...
+
+      # filter user sections and find sections that correspond with this course
+      course_sections = user.sections.select do |section|
+        section.course_id == courseId
+      end
+
+      # loop through user sections for this course
+      #return true if the user has the role "StudentEnrollment" for any sections
+      course_sections.each do |section|
+        return true if user.has_role? "StudentEnrollment", section
```
sadleb, line 43: Can we use the `RoleConstants::STUDENT_ENROLLMENT` constant here?

app/policies/waiver_submission_policy.rb
elenabaurkot: Ah okay, got it, just edited that

app/policies/waiver_submission_policy.rb
elenabaurkot: I updated and added the initialize method, I am not sure if I did this correctly though. I was trying to add the @course and initialize it to record.id, but was receiving an error on the initialize method when I tried that. I changed it to look similar to attendance_event_submission_policy.rb where the @record instance variable is initialized and use that to grab the course after. Let me know if this should be changed @sadleb

PR #881, opened 2021-10-20
Make Discord server Id dynamic 

elenabaurkot:
### Description
Updated the Discord server id to a dynamic id instead of the hard-coded id we had before. This way when a user signs up they are automatically added to the correct Discord server.

### Motivation and Context
Had to update it from the hard-coded discord server id so that users get added to the correct Discord server. 

### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)


app/controllers/discord_signups_controller.rb
```diff
@@ -1,4 +1,6 @@
 require "discordrb"
+require 'salesforce_api'
```
rshipp, line 2: nit: let's be consistent with single/double quotes here. i don't think it matters which one we use

app/controllers/discord_signups_controller.rb
```diff
@@ -11,7 +13,12 @@ def launch
       @discord_user = JSON.parse(response.body)
 
       if @discord_user["email"] && @discord_user["verified"]
-        Discordrb::API::Server.add_member("Bot #{Rails.application.secrets.bot_token}", "722098701878689933", @discord_user["id"], current_user.discord_token)
+        # Get dynamic server id here, based on the user's region
+        participant = SalesforceAPI.client.find_participant(contact_id: current_user.salesforce_id)
+        discord_server_id = participant.discord_server_id
```
rshipp, line 18: can we handle the case where the discord server ID isn't found (is `nil`)? this is unlikely to occur in a real program, but it'd be better to have well-defined behavior there rather than whatever the Discord API does if we pass in a nil. I'd say raising a custom exception is fine for now, and we can improve error handling later if needed.

examples of defining/using custom exceptions:

https://github.com/bebraven/platform/blob/2ae07f820692b2d8c1c61e59c0d608cf76746c4c/app/services/sync_portal_enrollments_for_program.rb#L13-L14

https://github.com/bebraven/platform/blob/2ae07f820692b2d8c1c61e59c0d608cf76746c4c/app/services/sync_portal_enrollments_for_program.rb#L88

rshipp (commented):
two comments, and i'll let @sadleb do the sign-off on this since he did the apex stuff

app/controllers/discord_signups_controller.rb
```diff
@@ -1,7 +1,10 @@
 require "discordrb"
+require "salesforce_api"
+
 #TODO:implement everything --> this is demo
 class DiscordSignupsController < ApplicationController
   layout 'lti_canvas'
+  DiscordServerIdError = Class.new(StandardError)
```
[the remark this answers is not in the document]
elenabaurkot: I was mimicking the courses_controller syntax, but will update to that!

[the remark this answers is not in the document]
rshipp: oh, weird. wonder why it's different there.

app/controllers/discord_signups_controller.rb
```diff
@@ -11,7 +14,14 @@ def launch
       @discord_user = JSON.parse(response.body)
 
       if @discord_user["email"] && @discord_user["verified"]
-        Discordrb::API::Server.add_member("Bot #{Rails.application.secrets.bot_token}", "722098701878689933", @discord_user["id"], current_user.discord_token)
+        # Get dynamic server id here, based on the user's region
+        participant = SalesforceAPI.client.find_participant(contact_id: current_user.salesforce_id)
+        discord_server_id = participant.discord_server_id
+
+        raise DiscordServerIdError, "No Discord Server Id found" if discord_server_id.nil?
```
sadleb, line 21: Can we add the Participant.Id to the error message so that if it does raise we'll be able to quickly track down who the error was for? E.g.
```
  raise DiscordServerIdError, "No Discord Server Id found for Participant.Id = #{participant.id}" if discord_server_id.nil?
```

There's prob some Honeycomb instrumentation we want to do as well, but that's outside the scope of this PR.

lib/salesforce_api.rb
```diff
@@ -20,7 +20,7 @@ class SalesforceAPI
   SFParticipant = Struct.new(:first_name, :last_name, :email, :role,
                              :program_id, :contact_id, :status, :student_id,
                              :cohort, :cohort_schedule, :cohort_id,
-                             :id, :discord_invite_code, :discord_user_id,
+                             :id, :discord_invite_code, :discord_user_id, :discord_server_id,
```
sadleb, line 23: We need to be careful here b/c we're adding a parameter in the middle of the list. Any place were `SFParticipant.new` is called needs to account for this new parameter. Luckily, it looks like there are only 4 places in the code that initialize one of these structs. 3 are in specs and don't go past the 4th or 5th param and the last one you changed below. So we're good, but something to keep in mind going forward!

Here is what I grepped:
image

Pro-tip: The above raw recursive `grep` command is really slow, so I created this alias that excludes a bunch of locations we generally don't want to look in:
```
# grep code only
# alias cgrep='grep -R --exclude "*tmp*" --exclude "*.swp" --exclude "*log*" --exclude "*/packs-test/*" --exclude "*node_modules*" --exclude "*/public/assets/*" --exclude "*/public/packs/*" --exclude "*/cov    erage/*" --exclude "*layout/dashboard.html"'
```

app/controllers/discord_signups_controller.rb
```diff
@@ -11,7 +14,14 @@ def launch
       @discord_user = JSON.parse(response.body)
 
       if @discord_user["email"] && @discord_user["verified"]
-        Discordrb::API::Server.add_member("Bot #{Rails.application.secrets.bot_token}", "722098701878689933", @discord_user["id"], current_user.discord_token)
+        # Get dynamic server id here, based on the user's region
+        participant = SalesforceAPI.client.find_participant(contact_id: current_user.salesforce_id)
```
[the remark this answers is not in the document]
elenabaurkot: Added that comment in - are multiple participants able to have the same salesforce id? @sadleb

app/controllers/discord_signups_controller.rb
elenabaurkot: Just added in :+1:

app/controllers/discord_signups_controller.rb
```diff
@@ -11,7 +14,14 @@ def launch
...
+        discord_server_id = participant.discord_server_id
+
+        raise DiscordServerIdError, "No Discord Server Id found" if discord_server_id.nil?
+
+        # Add user to their Discord Server
+        Discordrb::API::Server.add_member("Bot #{Rails.application.secrets.bot_token}", discord_server_id, @discord_user["id"], current_user.discord_token)
```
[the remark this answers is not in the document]
elenabaurkot: Udpated the add_member to the shorter lines - I also updated the Discord API Request on lines 38-51 not sure if that one looks quite right if you want to double check it. Very interesting about the string freeze, good to know!

lib/salesforce_api.rb
elenabaurkot: Will keep this in mind, and thanks for the grep tip, saving that!

app/controllers/discord_signups_controller.rb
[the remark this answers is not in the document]
sadleb: Yeah, `salesforce_id` is poorly named. It's the `contact_id`. A `Contact` in Salesforce is logically equivalent to a `User` in platform. A Contact/User can be enrolled in multiple programs in different roles. In other words, they can Participate in multiple ways/programs.

sadleb (approved):
Approved pending the change to add `# frozen_string_literal: true` as the first line of the controller.

PR #893, opened 2021-10-28
Smarter Zoom sync logic - regenerate links when thing change

sadleb:
### Description
Task: https://app.asana.com/0/1201131148207877/1201271781592087

**Deploy notes**: 
* In order to support the future user stories to [delete old registrants](https://app.asana.com/0/1201131148207877/1201279111925854) (aka invalidate their link) we need to have the `registrant_id` stored on a `ZoomLinkInfo`. Also, if there aren't `ZoomLinkInfo`s populated then on the next sync the new code will re-sync (aka register) everyone. If nothing has changed, this shouldn't change the links and will result in populating the `ZoomLinkInfo`s. Since this PR doesn't invalidate the old links, the plan is just to let the sync re-sync everyone. 

* Also, deploy this first: [When there are co-LCs, Zoom prefixes should reflect both names](https://app.asana.com/0/1201131148207877/1200813128952649)

### Test Plan
1. Make sure you have a Cohort Schedule with the "Webinar Registration X" fields configured (with different meeting IDs so you can tell if the correct link is created / updated).
    1. See the "Generate Zoom Links" tool for tips on how to create a meeting that will work with the sync.
2. Create an Enrolled Fellow and LC in the same Cohort Schedule and Cohort.
3. Create an Enrolled CP and TA.
4. Run "Sync From Salesforce" (without the Force option)
5. Make sure that the Zoom links generated on each Participant are correct and when they join, the prefixes are what we expect.
6. Update folks names and run the sync. They should be able to join using the same link and their names should be correct
    1. Note: if you do this 4 times for a single Participant, it will start to fail with a message about needing to wait a day. You can only do 3 updates per day for a given email and meeting ID pair.
7. Update folks email addresses (and their name for good measure) and run the sync. The links should change and work.
    1. Note: in this pass, the old links will also work.
    2. If you look at the registrants with the Zoom UI you'll notice that there are two for this person. One with each email. In a future pass, we'd like to delete the original one when adding a new one with a new email.
8. Update the meeting_id on the Cohort Schedule and run the sync. Folks links should change and bring them into the new meeting.
    1. Note: in this pass, the old links will still let them into the previous meeting.
9. Update the meeting password and run the sync. The links still won't work. Run it again with the "Force" option on. Now they should be updated and start working again.

New language in Notes a about "Force Zoom links to be updated"
image

### Types of changes

### Checklist:


app/services/sync_zoom_links_for_participant.rb
```diff
@@ -150,12 +182,13 @@ def zoom_meeting_2_configured?
...
-    Honeycomb.add_field('salesforce.participant.id', @salesforce_participant.id)
-    Honeycomb.add_field('salesforce.participant.status', @salesforce_participant.status)
-    Honeycomb.add_field('salesforce.participant.zoom_meeting_id_1', @salesforce_participant.zoom_meeting_id_1)
-    Honeycomb.add_field('salesforce.participant.zoom_meeting_id_2', @salesforce_participant.zoom_meeting_id_2)
+  def create_or_update_link_infos(registrant1, registrant2)
+    upsert_data = []
+    upsert_data << @new_zoom_link_info_1.attributes.except('id', 'created_at', 'updated_at') if registrant1 && registrant1['error'].blank?
+    upsert_data << @new_zoom_link_info_2.attributes.except('id', 'created_at', 'updated_at') if registrant2 && registrant2['error'].blank?
+    if upsert_data.present?
+      ZoomLinkInfo.upsert_all(upsert_data, unique_by: [:salesforce_participant_id, :salesforce_meeting_id_attribute])
```
rshipp, line 190: upsert skips all the model validations, is that ok here? i see you have a sql constraint on the meeting id attributes, but not the other fields with validations.

sadleb: The `presence: true` validations also have DB `null: false` constraints. But yeah, maybe I should add a DB level check constraint for the length of the Participant.Id to be 18 chars.

sadleb: 18 char DB constraint: https://github.com/bebraven/platform/pull/893/commits/6257017300d78740b89a2879488f72d4ff3a1cfc

app/services/sync_zoom_links_for_participant.rb
```diff
@@ -106,37 +131,44 @@ def cancel_registration(meeting_id)
...
   end
 
   # Note that a force_update doesn't actually update the link unless other info has changed,
   # like the password for the meeting for example.
   def generate_zoom_meeting_1_link?
-    zoom_meeting_1_configured? && (
-      @force_update || @salesforce_participant.zoom_meeting_link_1.blank?
-    )
+    return false unless zoom_meeting_1_configured?
+    return true if @force_update
```
rshipp, line 154: should the return true be first? or is this the behavior you want with `@force_update`?

rshipp (approved):
left some comments, approving bc i don't know if they're relevant

app/services/sync_zoom_links_for_participant.rb
sadleb: This is the behavior we want. Force update will clear out links if they are no longer configured with a meeting in SF

app/services/sync_zoom_links_for_participant.rb
```diff
@@ -24,35 +29,46 @@ def run
...
+    # We may need to clear out the existing links and/or keep one the same b/c the other has changed.
+    # That's why this conditional is complicated.
+    new_join_link1 =  (zoom_meeting_1_configured? ? (generated_link1 || existing_join_link1) : nil)
+    new_join_link2 =  (zoom_meeting_2_configured? ? (generated_link2 || existing_join_link2) : nil)
 
-    new_join_link1 =  (zoom_meeting_1_configured? ? (generated_link1 || old_join_link1) : nil)
-    new_join_link2 =  (zoom_meeting_2_configured? ? (generated_link2 || old_join_link2) : nil)
-
-    links_changed = !(old_join_link1 == new_join_link1 and old_join_link2 == new_join_link2)
+    links_changed = !(existing_join_link1 == new_join_link1 and existing_join_link2 == new_join_link2)
```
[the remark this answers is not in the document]
sadleb: Agree

PR #896, opened 2021-11-01
Added refresh-button to Discord Signups page

elenabaurkot:
### Description
In order for a user to be fully added to the Braven Discord server, they must authorize their account on the Platform Discord Signups page, verify their email and then refresh the page. To prevent users from missing the final step of refreshing the page, we added a refresh button with instructions to remind users to come back and click the button after verifying their email. This was added at two different points in the document because users have to follow a different set of instructions in case they are unable to find the email to verify their account. Since users will see different content on the Discord Signups page based on what step they are at in the process, some of the content was rearranged slightly to make sure it made sense on every screen.

### Motivation and Context
If users don't refresh the Discord Signups page after verifying their email they won't be added to the server. These changes are to try to prevent users from missing that step and ensure they get added smoothly. 

Asana Task: https://app.asana.com/0/0/1201305798562983/f

### Screenshots (if appropriate)
Screen Shot 2021-11-01 at 4 37 22 PM
:


rshipp (approved):

app/views/discord_signups/launch.html.erb
```diff
@@ -27,17 +28,22 @@
...
     <p>Follow these instructions to verify the email associated with your Discord account:</p>
 
-    <p>Step 1: Navigate to the email that you used to sign up for Discord. Locate the Discord confirmation email with the subject "Verify Your Email Address for Discord." </p>
+    <p>Step 1: Navigate to the email that you used to sign up for Discord. Locate the Discord confirmation email with the subject "Verify Your Email Address for Discord."</p>
     <img src="https://platform-prod-assets.s3.amazonaws.com/Discord+Signup+Images/discord-verify-email-subject.png" alt="" class="discord-demo">
 
-    <p>Step 2: Open the email and click the blue "Verify Email" button. If you are successful in finding this email, skip to Step 6. If you cannot find the email, continue to Step 3.</p>
+    <p>Step 2: Open the email and click the blue "Verify Email" button. Then come back here and refresh the page.
+
+    <%= link_to "Refresh", launch_discord_signups_path(lti_launch_id: @lti_launch.id), class: "btn btn-lg btn-primary discord-refresh-btn" %>
```
sadleb, line 44: The Step 2 line is missing a closing `</p>` tag. Also, I think we should avoid using pixel measurements on `discord-refresh-btn`. We should use `rem`s. Here is one way to accomplish something similar looking to what you have that is a little more responsive
```suggestion
    <p>Step 2: Open the email and click the blue "Verify Email" button. Then come back here and refresh the page.</p>

    <div class="d-flex justify-content-center">
       <%= link_to "Refresh", launch_discord_signups_path(lti_launch_id: @lti_launch.id), class: "btn btn-lg btn-primary btn-block discord-refresh-btn" %>
    </div>
```
Using the above, then `discord-refresh-btn` becomes something like
```
.discord-refresh-btn {
  text-decoration: none;
  max-width: 16rem;
} 
```

app/views/discord_signups/launch.html.erb
```diff
@@ -47,7 +53,12 @@
 
     <p>[Optional] Step 5: At the top of the user settings, click the "Resend Verification Email" button. Navigate to your email, open the verification email, and click the blue "Verify Email" button. For further assistance, please submit a <a href="<%=Rails.application.secrets.support_ticket_url%>">help ticket</a> to the Braven Product Support Team.</p>
     <img src="https://platform-prod-assets.s3.amazonaws.com/Discord+Signup+Images/discord-resend-verification-email.png" alt="" class="discord-demo">
+
+    <p>Step 6:Once you're all set up, click the button below to refresh the page.</p>
+
+    <%= link_to "Refresh", launch_discord_signups_path(lti_launch_id: @lti_launch.id), class: "btn btn-lg btn-primary discord-refresh-btn" %></p>
```
sadleb, line 59: Ditto on above comment re: this link. Also notice the stray closing `</p>` tag.

PR #903, opened 2021-11-02
Adjust Discord Signup style

rshipp:
### Description

Adjust styling for Discord Signup page. Some of this will impact other pages, so we should make sure it doesn't break anything.

### Motivation and Context

Task: https://app.asana.com/0/0/1201274835996657/f

### Types of changes

### Checklist:


app/assets/stylesheets/layouts/lti_canvas.scss
```diff
@@ -31,7 +31,7 @@
 // mobile device. If you decide to change the max-width for a certain view,
 // make sure you write your content responsively!
 .bv-lti-container {
-  height: 100vh; // Height of iframe viewport layout, not the actual screen
+  min-height: 100vh; // Height of iframe viewport layout, not the actual screen
```
rshipp, line 34: @sadleb do you think this will break anything / is there a certain page i should check?

app/assets/stylesheets/shared/style.scss
```diff
@@ -74,12 +74,12 @@ p {
...
-a {
+a:not(.btn) {
   color: $bv-oceanblue-600;
   text-decoration: underline;
 }
 
-a:visited {
+a:not(.btn):visited {
   color: $bv-oceanblue-700;
   text-decoration: underline;
```
rshipp, line 84: this is a global adjustment, but shouldn't break anything.

elenabaurkot: this is cool, I didn't know you could use "not" to select the opposite of something

app/assets/stylesheets/layouts/lti_canvas.scss
sadleb: Looking at all the controllers that use it, the main one I'm concerned about is `./app/controllers/linked_in_authorization_controller.rb` -- I think the rest should be fine to be full height, but we'd probably want to test?

rshipp: I'm not seeing any difference in any view, including the linkedin view. What made you more concerned about the linkedin one? I just wanted to check since i don't know what reason this was originally added for or what it does.

app/assets/stylesheets/discord_signups.scss
```diff
@@ -1,14 +1,9 @@
-img.discord-demo {
+img.discord-signup {
   max-width: 100%;
   margin: 0 auto 35px auto;
```
sadleb, line 3: If we do need more bottom margin than the default, can we use `rem` measurements instead of pixels. I know it really doesn't matter in this case, but I'd rather have almost everyone across all CSS stylesheets be in `rem` or `em` to encourage mimicking that.

Also, have we tested how responsive this is? Should we use Bootstrap styles instead? E.g. [`img-fluid`](https://getbootstrap.com/docs/5.0/content/images/)

app/assets/stylesheets/layouts/lti_canvas.scss
sadleb: I think it was originally added just to make any `lti_canvas` iframed page take up 100% of the available space (almost everything was iframed at this point so it was just the root div saying "start by taking all the space").

I'm worried about the linked_in one b/c it's an iframed button that isn't supposed to be 100% height. Rate this module is the other one that came to mind, but that's a different layout so I'm not concerned.

app/assets/stylesheets/discord_signups.scss
sadleb: We're going to revisit this in a separate PR. Want to get this merged.

sadleb (approved):
approved assuming you've sanity checked other pages that use the controller and the min-height change is fine

PR #904, opened 2021-11-02
Updated Discord authorization redirect to redirect to correct page

elenabaurkot:
### Description
This update adds discord_state to the users table and sets the discord_state for the current user when the Discord Signups page is launched. The discord_state is used when the user authorizes their Discord account to validate the authorization request and prevent cross-site request forgery. The lti_launch_id is also added into the discord_state so that it can be added to the redirect url to to send the user back to the correct Platform page with their proper lti_launch_id after the Discord authorization is complete. 

### Motivation and Context
This change is needed so that the user is redirected back to the proper page during the Discord signup process after authorizing their Discord account. 

Asana Task: https://app.asana.com/0/0/1201274835996647/f


app/controllers/discord_signups_controller.rb
```diff
@@ -28,6 +28,12 @@ class DiscordServerIdError  < StandardError; end
   def launch
     authorize :discord_signup
 
+    @state = SecureRandom.hex + ",#{@lti_launch.id}"
```
sadleb, line 31: Can we change this to `@discord_state` just to make it clear that it's not our `lti_launch.state` which is an authentication thingie.

app/controllers/discord_signups_controller.rb
```diff
@@ -72,22 +78,35 @@ def launch
...
+        'https://discord.com/api/v8/oauth2/token',
+        {
+          client_id: Rails.application.secrets.discord_client_id,
+          client_secret: Rails.application.secrets.discord_client_secret,
+          grant_type: 'authorization_code',
+          code: params[:code],
+          redirect_uri: Rails.application.secrets.discord_redirect_uri
+        },
+        content_type: 'application/x-www-form-urlencoded'
+      )
```
sadleb, line 101: When we add code to handle expiration (7 days), let's talk about refactoring this into a private method.

sadleb (approved):
approved pending variable rename

PR #909, opened 2021-11-08
Minor CSS changes for Discord Signup

rshipp:
### Description

Fix the font size of `li`s to be the same as `p`s. Convert the image padding to rem, and add top padding to images to match the bottom padding.

### Motivation and Context

Task: https://app.asana.com/0/0/1201314155101897/f

Screen Shot 2021-11-08 at 2 44 14 PM

### Types of changes

### Checklist:


app/assets/stylesheets/shared/style.scss
```diff
@@ -74,6 +74,10 @@ p {
   margin-bottom: 0.75rem;
 }
 
+li {
+  font-size: 1.2em;
```
rshipp, line 78: see `p` style above

elenabaurkot (approved):
lgtm 👍

app/assets/stylesheets/shared/style.scss
sadleb: I think this is a good idea in general, but we probably want to sanity check Rate This Module after pushing this change since the height of the iframe at various breakpoints was calculated based on the current content and I don't know if this will change the content in there or not. There is likely other places that explicitly adjust the `<li>` font-size to make it look better than what the default used to be and I suspect those will continue to work since they should be defined with a more specific CSS target path, but we may be able to remove some of them now. Prob not worth doing an audit at the moment though. We can just clean them up whenever we happen to be working on any of those views.

sadleb (approved):

app/assets/stylesheets/shared/style.scss
rshipp: there's no `li`s in rate this module, so should be fine. i did check and fix projects, can't think of anything else that would be impacted.

sadleb: Attendance and Capstone Evaluations are the other two places that I can think of that *may* be impacted.

PR #910, opened 2021-11-09
Set expired Discord tokens to nil

elenabaurkot:
### Description
When a user completes the authorization step in the Discord signup assignment, they get an access token that expires in seven days. Once they have the token and refresh the assignment page they should be added to the Braven Discord server and not need the token again, unless they leave the server. This update checks if their discord_token is expired whenever they launch the assignment and if it is expired, it resets the discord_token to nil. This way if the user needs to rejoin the server because they left, they will see the steps to walk them through joining the server again. Previously, we considered refreshing the tokens in a nightly rake task so they did not expire, but to avoid making a lot of unnecessary API calls to the Discord API we decided it would be better to allow the tokens to expire and users can go back through the signup process if necessary. 
 

### Motivation and Context
Every seven days, a user's Discord authorization token expires, when it expires we want the user to see the Discord signup steps again so they can go through the process of being added to the Braven Discord server. 

Asana Task: https://app.asana.com/0/0/1201274835996645/f


app/controllers/discord_signups_controller.rb
```diff
@@ -99,7 +99,11 @@ def oauth
         },
         content_type: 'application/x-www-form-urlencoded'
       )
-      current_user.update!(discord_token: JSON.parse(response.body)['access_token'])
+
+      current_user.update!(
+        discord_token: JSON.parse(response.body)['access_token'],
+        discord_refresh_token: JSON.parse(response.body)['refresh_token']
```
sadleb, line 105: Can you `JSON.parse` the body once into a variable first and then read the values out of there?

app/services/refresh_discord_tokens.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+require 'discordrb'
+
+class RefreshDiscordTokens
+  def initialize
+  end
+
+  # Update discord_token and refresh_token for all users that already have a discord_token
+  def run
+    users = User.where.not(discord_token: nil)
```
sadleb, line 10: Do you have a subtask to re-visit this and decide when to stop refreshing tokens? If so, can you link it in a TODO here? If not, can you make one. It should probably be sometime around when the semester ends, but we'll likely want to discuss.

app/services/refresh_discord_tokens.rb
```diff
@@ -0,0 +1,48 @@
...
+    users.each do |user|
+      refresh_tokens(user)
+    end
+  end
+
+private
+  def refresh_tokens(user)
+    Honeycomb.start_span(name: 'refresh_discord_tokens.refresh_tokens') do
+      user.add_to_honeycomb_span('RefreshDiscordTokens')
```
sadleb, line 19: Can you snake_case the prefix? e.g. `refresh_discord_tokens`

app/services/refresh_discord_tokens.rb
```diff
@@ -0,0 +1,48 @@
...
+        {
+          client_id: Rails.application.secrets.discord_client_id,
+          client_secret: Rails.application.secrets.discord_client_secret,
+          grant_type: 'refresh_token',
+          refresh_token: user.discord_refresh_token,
+        },
+        content_type: 'application/x-www-form-urlencoded'
+      )
+
+      user.update!(
```
sadleb, line 34: If any one user fails, `user.update!` will throw and the refresh operation will fail for all future users in the list. Do we just want to skip and log/send to honeycomb/sentry the failures? 

Also, since this is the second place you're updating the user with these exact attributes, maybe worth adding a `update_discord_tokens(discord_response)` method to the `User` model and calling that? Ditto on the above comment about only calling `JSON.parse(response.body)` once into a variable first.

app/services/refresh_discord_tokens.rb
```diff
@@ -0,0 +1,48 @@
...
+        discord_token: JSON.parse(response.body)['access_token'],
+        discord_refresh_token: JSON.parse(response.body)['refresh_token']
+      )
+    end
+  rescue Discordrb::Errors::UnknownError
+    user.update!(
```
sadleb, line 40: Same question on using `update` vs `update!` and whether we want all future users to stop being processed if for some reason we can't write the database momentarily or something.

Also, should we send the `UnknownError` to Sentry and/or Honeycomb? I don't know how these happen but just swallowing the exception and not being able to see it in any monitoring tools seems bad.

app/services/refresh_discord_tokens.rb
```diff
@@ -0,0 +1,48 @@
...
+      discord_token: nil,
+      discord_refresh_token: nil,
+    )
+  end
+end
+
+# TODO: figure out at what point users don't need tokens refreshed anymore
+# Asana Task: https://app.asana.com/0/1201131148207877/1201340564157240/f
```
sadleb, line 48: Whoops, ignore my above comment. Just got here.

app/services/refresh_discord_tokens.rb
```diff
@@ -0,0 +1,48 @@
...
+        :post,
+        'https://discord.com/api/v8/oauth2/token',
+        {
+          client_id: Rails.application.secrets.discord_client_id,
+          client_secret: Rails.application.secrets.discord_client_secret,
+          grant_type: 'refresh_token',
+          refresh_token: user.discord_refresh_token,
+        },
+        content_type: 'application/x-www-form-urlencoded'
+      )
```
sadleb, line 32: I know the Discord gem handles throttling and retrying, but are we concerned about churning through nearly all of our users and hitting the API in quick succession like this @rshipp ? In other words, should we consider doing this process in batches of our own or doing our own throttling? I assume no, but just want to make sure we've thought about it.

sadleb (changes requested):

app/services/refresh_discord_tokens.rb
sadleb, line 10: Are we worried about loading nearly all the users into memory? If so, maybe consider something like:
```
User.where.not(discord_token: nil).pluck(:id, :discord_refresh_token)
```
and then `User.update_all(the_data)`

app/controllers/discord_signups_controller.rb
elenabaurkot: Made this update in the update_discord_tokens(discord_response) method added to the User model

app/services/refresh_discord_tokens.rb
elenabaurkot: saw your comment below 👍

app/services/refresh_discord_tokens.rb
elenabaurkot: made this update to snake_case

app/services/refresh_discord_tokens.rb
elenabaurkot: Changed this to "user.update"  in the update_discord_tokens(discord_response) method in the User model. I'm sending a honeycomb message, should I send one to sentry too or is just to honeycomb sufficient?

app/services/refresh_discord_tokens.rb
elenabaurkot: 👍

app/services/refresh_discord_tokens.rb
elenabaurkot: I updated this to "user.update" as well and am sending the Honeycomb message if a user isn't updated. I was having a lot of trouble getting the "UnknownError" to send to Honeycomb. This is the error message that comes back in the response {"error": "invalid_grant"} and that's what i was trying to send, but Ryan helped me realize I couldn't get the response with that message in the rescue block because it was out of scope. It still shows up with the error message in Honeycomb, but you have to go to the failed user and search in there to see the actual error message

Screen Shot 2021-11-10 at 6 42 10 PM

app/services/refresh_discord_tokens.rb
elenabaurkot: I'm not sure if this is the same thing, but one of the edge cases we were looking at is to check the rate limits for the refresh tokens. It's not an actual task, just in the google doc within that task, but I could make it a separate task too. I think @rshipp said that if the rate limit ends up being a problem we could not refresh the token and when it runs out, set the discord_token to nil and have the user go through the authorization process again.

app/services/refresh_discord_tokens.rb
elenabaurkot: I haven't used pluck before, but to refresh the tokens I only need the discord_refresh_token for the post request, however when I update the user I am also updating the discord_token so do I need to pluck that too? Also, now that I put the update into the  update_discord_tokens(discord_response) method in the User model is that where I would be running the update_all?

sadleb: If we make sure and implement the task where we stop refreshing the discord tokens at a certain point, I'm not worried about the dataset being too large and therefore being concerned about memory. That said, below is some pseudo code for how I would go about implementing this using `pluck` and also addresses the comments about error handling. The `pluck` approach is useful to get familiar with and think about using when running a query that potentially returns a large dataset.
```
refreshed_token_data = []
users = User.where.not(discord_token: nil).pluck(:id, :discord_refresh_token)
users.each do |user_data|
  user_id, discord_refresh_token = user_data
  begin
    response = refresh_tokens(discord_refresh_token)
    token_data = JSON.parse(response.body)
    refreshed_token_data << user_token_data(user_id, token_data['access_token'], token_data['refresh_token'])
    Sentry.capture_exception(e)
    # TODO: Elena, you mentioned that we've seen this with an "invalid_grant" response. A quick google search seems to indicate that this was a coding error (e.g. not setting the content type header or somethign else wrong with the request). I was thinking this was for an expired token or an invalid token and we should clear out invalid tokens. If we can reproduce trying to use an expired or invalid refresh token, then we should capture that exception and run this code but I don't think we should run it for Unknown error if it was a coding error.
    refreshed_token_data << user_token_data(user_id, nil, nil)
  rescue => e
    Sentry.capture_exception(e)
    # Don't clear out the refresh token b/c we don't know if this is a temporary error or not. As we see
    # errors in the wild, add exception handling for those where we should clear out the existing tokens
    # or handle in some graceful way.
  end

  User.update_all(refreshed_token_data)
end

def user_token_data(user_id, discord_token, discord_refresh_token)
  {
     id: user_id,
     discord_token: discord_token,
     discord_refresh_token: discord_refresh_token
  }
end

def refresh_tokens(discord_refresh_token)
  response = nil
  # Note that by wrapping this in a span, any exceptions that are thrown
  # automatically populate the error_detail field in Honeycomb.
  Honeycomb.start_span(name: 'refresh_discord_tokens.refresh_tokens') do
      response = Discordrb::API.request(
        :oauth2_token,
        nil,
        :post,
        'https://discord.com/api/v8/oauth2/token',
        {
          client_id: Rails.application.secrets.discord_client_id,
          client_secret: Rails.application.secrets.discord_client_secret,
          grant_type: 'refresh_token',
          refresh_token: discord_refresh_token,
        },
        content_type: 'application/x-www-form-urlencoded'
      )
  end
  response
end
```

sadleb (changes requested):

app/services/refresh_discord_tokens.rb
sadleb: See my updated suggestion here: https://github.com/bebraven/platform/pull/910#discussion_r748320090

app/services/refresh_discord_tokens.rb
sadleb: See my updated suggestion here: https://github.com/bebraven/platform/pull/910#discussion_r748320090 -> this relies on letting the exception propagate through the Honeycomb span so that the `error_detail` field gets set.

app/services/refresh_discord_tokens.rb
sadleb: I agree that we should just wait and see if it happens instead of spending time on the front end. As long as the errors that Discord throws end up in Honeycomb we can setup alerts. Alerts will already be setup if the `app.error` field is set, which I think will also happen automatically with the suggested code here: https://github.com/bebraven/platform/pull/910#discussion_r748320090

app/services/refresh_discord_tokens.rb
sadleb: Note that for `upsert_all` to work, we need to add default timestamps at the database level. E.g. `t.timestamps(default: -> { "CURRENT_TIMESTAMP" })`

rshipp:
i'm having trouble following the discussion in the comments here, can we talk about this during the call today?

elenabaurkot:
> i'm having trouble following the discussion in the comments here, can we talk about this during the call today?

That works for me 👍

db/schema.rb
```diff
@@ -637,6 +637,7 @@
     t.string "uuid", default: -> { "gen_random_uuid()" }, null: false
     t.string "discord_token"
     t.string "discord_state"
+    t.integer "discord_expires_at"
```
sadleb, line 640: I think we should use a `datetime` column. When calculating the value from the `expires_in` value I think you can do something like `Time.now.utc + discord_response['expires_in']` and store that. When reading it out I think you can just do 
```
if current_user.discord_expires_at && current_user.discord_expires_at < Time.now.utc
  current_user.update!(discord_token: nil)
end
```

sadleb (changes requested):
See my comment about using a `datetime` column instead of an integer column.

db/schema.rb
elenabaurkot: That makes sense, just made this update 👍

app/controllers/discord_signups_controller.rb
```diff
@@ -28,6 +28,14 @@ class DiscordServerIdError  < StandardError; end
   def launch
     authorize :discord_signup
 
+    # If a user's discord_token expired, reset it to nil
+    discord_expiry_time = current_user.discord_expires_at
+    unless discord_expiry_time == nil
```
sadleb, line 33: You can get rid of this `unless`. The `if current_user.discord_expires_at` part now does that. It basically says "if it's not nil" then "check the time"

app/controllers/discord_signups_controller.rb
```diff
@@ -99,7 +107,15 @@ def oauth
...
-      current_user.update!(discord_token: JSON.parse(response.body)['access_token'])
+
+      discord_response = JSON.parse(response.body)
+      discord_token_expiration = Time.now.utc + discord_response['expires_in']
+      unless current_user.update(
+        discord_token: discord_response['access_token'],
+        discord_expires_at: discord_token_expiration,
+      )
+        Honeycomb.add_field('user.update_failed', current_user)
+      end
```
sadleb, line 118: If this fails, we still consume the token below. Also, with this code we don't get any information about what the actual failure was. I would stick with the `update!` call so that it throws an exception and automatically gets sent to Honeycomb and Sentry. I think this approach made more sense when looping over a bunch and not wanted the whole looping/refresh process to fail. It's different now that it's a single user hitting this page. We want it to fail and then figure out how to handle those failures more gracefully and put in code later if it happens with any frequency in the wild.

app/controllers/discord_signups_controller.rb
elenabaurkot: ah okay, that makes sense and looks nicer too

app/controllers/discord_signups_controller.rb
elenabaurkot: Got it, just made this update 👍

sadleb: Just to clarify, the code should now just be:
```
      current_user.update!(
        discord_token: discord_response['access_token'],
        discord_expires_at: discord_token_expiration,
      )
```

You don't need the `unless` part b/c `update!` either works or raises an exception. If it raises an uncaught exception, that is automatically added to Honeycomb.

elenabaurkot: Okay that makes sense 👍

app/controllers/discord_signups_controller.rb
```diff
@@ -108,11 +108,10 @@ def oauth
 
       discord_response = JSON.parse(response.body)
       discord_token_expiration = Time.now.utc + discord_response['expires_in']
-      unless current_user.update!(
+      current_user.update!(
         discord_token: discord_response['access_token'],
         discord_expires_at: discord_token_expiration,
       )
-        Honeycomb.add_field('user.update_failed', current_user)
       end
```
sadleb, line 115: I think this `end` was for the `unless` that you removed and needs to go too?

elenabaurkot: I think that's correct, just removed

app/controllers/discord_signups_controller.rb
```diff
@@ -28,6 +28,12 @@ class DiscordServerIdError  < StandardError; end
   def launch
     authorize :discord_signup
 
+    # If a user's discord_token expired, reset it to nil
+    discord_expiry_time = current_user.discord_expires_at
```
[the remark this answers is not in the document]
elenabaurkot: Good call, just removed

sadleb (approved):
Assuming you’ve done a bit of testing and have confidence this works, lgtm!

PR #911, opened 2021-11-10
Allow TA's to filter the gradebook in Canvas by their Caseload

sadleb:
### Description
Translates TA Assignments in Salesforce to Canvas sections so that TA's can filter on their Caseload. TA Assignments map TA's to the list of Fellows they are assigned to grade:
Task: https://app.asana.com/0/1201131148207877/1200184617753686

Related Salesforce change to return the list of TA names if TA Assignments exist: 
* https://github.com/beyond-z/salesforce/pull/146

Note: there are [follow-up tasks for all the TODOs](https://app.asana.com/0/1201131148207877/1201368574922350) (esp specs) in this PR. We need to get the end-user-story live and we'll come back and clean all that up once we finish the other user stories that need to be finished by Thanksgiving holiday, [esp this](https://app.asana.com/0/1201131148207877/1200871464995937).

TA Assignments in Salesforce that control the resulting "TA Caseload" Canvas sections
image

Folks with a TA Assignment being added to a second "TA Caseload" section in Canvas after the sync
image

A TA being able to filter in the gradebook to only their "TA Caseload" section
image

### Types of changes

### Checklist:


app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -184,52 +195,55 @@ def enroll_user(canvas_course_id, role, section, limit_privileges_to_course_sect
     )
   end
 
-  # Creates a local DB Section and a Canvas section, but only returns the local one.
+  # Finds or creates a local DB Section and a Canvas section, but only returns the local one.
```
sadleb, line 198: It's helpful to hide whitespace when reviewing the below. 
image

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -184,52 +195,55 @@ def enroll_user(canvas_course_id, role, section, limit_privileges_to_course_sect
...
-                override['course_section_id'] = canvas_section.id
-                override.delete('id')
-                cohort_schedule_overrides << override
-              end
+    course = Course.find_by!(canvas_course_id: canvas_course_id)
+    existing_section = Section.find_by(
+      course_id: course.id,
+      name: section_name,
+    )
+    return existing_section if existing_section.present?
```
sadleb, line 206: I originally changed this method to look up the the local DB `Section` instead of querying Canvas b/c I was trying to have this avoid accidentally getting the TA Caseload sections which would not exist locally. The final implementation doesn't rely on this change though but it seems worth avoiding the extra API call? This would only fail differently if the section was deleted manually from Canvas (which I think we disabled the ability to do with javascript). Plus I had already done my testing with these changes in place. If I put it back to the original code I would need to re-test. Thought? Worth reverting since it's outside the scope of this PR now?

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -272,8 +286,8 @@ def find_canvas_course_section(canvas_course_id, section_name)
     canvas_client.find_section_by(course_id: canvas_course_id, name: section_name)
   end
 
-  def find_user_enrollment(canvas_course_id)
-    canvas_client.find_enrollment(user_id: portal_user.id, course_id: canvas_course_id)
+  def find_user_enrollment(canvas_section_id)
+    canvas_client.find_enrollment(canvas_user_id: portal_user.id, canvas_section_id: canvas_section_id)
```
rshipp, line 290: this seems against the standard of not naming things canvas_xxx_id if they're inside canvas lib

app/services/sync_ta_caseload_for_participant.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
+require 'canvas_api'
```
rshipp, line 2: you use SalesforceAPI below, need to require that too?

lib/canvas_api.rb
```diff
@@ -344,25 +344,53 @@ def get_enrollments(course_id, types=[])
...
+    response = get("/sections/#{canvas_section_id}/enrollments?user_id=#{canvas_user_id}")
+    enrollments = JSON.parse(response.body)
+    return nil if enrollments.blank?
 
+    if enrollments.count > 1
+      # This can only happen if they are enrolled with multiple roles in the same section.
+      # Canvas won't create dupe enrollments for a given section/user/role
+      Honeycomb.add_field(
+        'alert.canvas_api.duplicate_enrollments_for_user',
+        "for canvas_user_id: #{canvas_user_id}, canvas_section_id: #{canvas_section_id}"
```
rshipp, line 369: can you add these as normal fields instead of text inside the alert, so they're easier to look for?

lib/canvas_api.rb
```diff
@@ -344,25 +344,53 @@ def get_enrollments(course_id, types=[])
...
+      )
+    end
+
+    enrollment = enrollments.first
     LMSEnrollment.new(enrollment['id'], enrollment['course_id'],
                       enrollment['type'].to_sym, enrollment['course_section_id'])
   end
```
elenabaurkot, line 377: Just for clarification, it doesn't matter that the first enrollment is being grabbed because the only difference here would be the role and that's not being used here? And everything else should be the same since it's grabbing the same section?

app/services/sync_portal_enrollment_for_account.rb
sadleb: It seemed like we had started moving towards the convention of canvas_xxx in more recently added endpoints. Currently it's inconsistent. I lean towards canvas_xxx if/when we go clean it up to be consistent.

app/services/sync_ta_caseload_for_participant.rb
sadleb: yeah. not absolutely necessary in the real world b/c something else will likely load it first, but better to be safe.

lib/canvas_api.rb
sadleb: There should only be zero or one enrollment which is why i'm sending an alert above if there are more. If we somehow get into a weird state where someone is enrolled as both a TA and Student in the same section, then the behavior is undefined. It was even worse before this change b/c we just grabbed the last enrollment in a course which was way more likely to happen (and still undefined behavior). I could throw an exception but I think since the end-user experience will probably not be impacted too much until we fix the root cause / or fix up the data I thought an alert would be better?

elenabaurkot: Ah okay, so the only thing that will be affected is the enrollment['type'] because they could potentially be added as the wrong type of user?

sadleb: Yup

elenabaurkot: This all seems good to me then - makes more sense to add them and alert then not add them at all

elenabaurkot (commented):
This all looks good to me 👍

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -122,10 +133,11 @@ def sync_enrollment(canvas_course_id, role, section_name, limit_privileges_to_co
     Honeycomb.add_field('canvas.section.id', section.canvas_section_id.to_s)
     Honeycomb.add_field('canvas.section.role', role)
 
-    enrollment = find_user_enrollment(canvas_course_id)
+    enrollment = find_user_enrollment(section.canvas_section_id)
```
sadleb, line 136: Note that this change is b/c the prior method very heavily relied on the fact that we only expected the user to be enrolled in a single section in Canvas per course. Now they can be in the TA Caseload sections, so we're more specific when looking for the existing primary enrollment. It's still possible that they could be enrolled twice in this section with different roles and I'm sending an alert to Honeycomb if that happens. For a given role/section though, I tested that trying to add them a second time doesn't create duplicates. It just returns the same enrollment so this should be pretty reliable.

app/services/sync_portal_enrollment_for_account.rb
sadleb: We discussed this verbally and no one pushed back on reverting this change so we're going to roll with it. It doesn't introduce quite a few small tweaks to the specs though.

rshipp (approved):

PR #914, opened 2021-11-15
Discord edge case updates

elenabaurkot:
### Description
This update is to fix some potential edge case issues. 

1) Previously, if a user launched a second instance of the Discord Signup assignment, they would get a new state for each launch and then if they went back to the first page they launched to authorize the Braven Discord server there would be an error since the state didn't match.  If a user launches multiple instances of the Discord Signup assignment, a new state will only be created on the first launch. This way if they authorize the assignment on either of the launched pages it will work.

2) Previously, if a user was already added to the Braven Discord server, when they refresh the page the Discord Bot "configure_member" command would still run on every refresh. The "user_in_server?" method makes an API call to Discord to check which servers the user is on and checks if any of those servers match the Salesforce participants discord_server_id. It also checks if the user's Salesforce Participant has a discord_user_id. If both of those are true, when the page is refreshed the user will not be added to the user will not be added to the Braven Discord server again and the bot "configure_member" command won't run, since the user is already on the server. 

3) When the user completes the authorization step, if they choose cancel instead of authorize, they won't be added to the Braven Discord server. This update adds an alert when the user is redirected back to Platform letting them know they clicked cancel and to try again. 

### Motivation and Context
To account for possible issues in the Discord signup process and adjust to make it run more smoothly.

Asana Task: https://app.asana.com/0/0/1201274835996646/f

When a user chooses cancel instead of authorize, they will see the alert below
Screen Shot 2021-11-15 at 12 34 58 PM

Screen Shot 2021-11-15 at 1 19 46 PM


app/controllers/discord_signups_controller.rb
```diff
@@ -28,11 +28,17 @@ class DiscordServerIdError  < StandardError; end
   def launch
     authorize :discord_signup
 
-    @discord_state = SecureRandom.hex + ",#{@lti_launch.id}"
-    Honeycomb.add_field('user.discord_state', @discord_state)
-
-    # Save a CSRF token so we can verify in #oauth
-    current_user.update!(discord_state: @discord_state)
+    # raise DiscordServerIdError, "No Discord Server"
```
sadleb, line 31: What does this comment mean?

app/controllers/discord_signups_controller.rb
```diff
@@ -28,11 +28,17 @@ class DiscordServerIdError  < StandardError; end
...
+    user_discord_state = current_user.discord_state
+    if user_discord_state == nil || user_discord_state == ""
```
sadleb, line 33: `user.discord_state.blank?` is the same as `user_discord_state == nil || user_discord_state == ""` and is a little cleaner.

app/controllers/discord_signups_controller.rb
```diff
@@ -100,8 +109,12 @@ def oauth
         content_type: 'application/x-www-form-urlencoded'
       )
       current_user.update!(discord_token: JSON.parse(response.body)['access_token'])
+    rescue Discordrb::Errors::UnknownError => discorderror
+      Sentry.capture_exception(discorderror)
+      if params[:error]
+        redirect_to launch_discord_signups_url(lti_launch_id: lti_launch_id), alert: 'You clicked cancel instead of authorize. Try again and click Authorize to be added to the Braven Discord server.' and return
+      end
```
sadleb, line 116: If it was a different error, meaning no `params[:error]` exists, what should the behavior be? Right now it looks like it just continues on and redirects to the `launch_discord_signups_url` with no message. Should it `raise` instead?

app/controllers/discord_signups_controller.rb
```diff
@@ -127,4 +140,16 @@ def lti_launch_url
   def can_publish_latest?
     false
   end
+
+  def user_in_server?(participant)
+    user_servers_response = Discordrb::API::User.servers("Bearer #{current_user.discord_token}")
+    current_user_discord_servers = JSON.parse(user_servers_response.body)
+
+    current_user_discord_servers.each do |server|
+      if server["id"] == participant.discord_server_id && participant.discord_user_id != nil
```
sadleb, line 149: We can save an API call by putting `return false unless participant.discord_user_id.present?` as the first line of this method.

Also, can you change `server["id"]` to `server['id']`? Small nit, but we should always use string literals when possible.

app/controllers/discord_signups_controller.rb
elenabaurkot: Should I create another error like on line 21 for this except just a generic DiscordError (class DiscordError  < StandardError; end)? Or would this not quite make sense since it could potentially be an error not caused by Discord? @sadleb

app/controllers/discord_signups_controller.rb
elenabaurkot: Did not know you could do that, cool though, I will update to that

app/controllers/discord_signups_controller.rb
elenabaurkot: Whoops forgot to delete this, I threw this in when I was purposely trying to throw an error to test something

app/controllers/discord_signups_controller.rb
sadleb: I would only raise a custom exception like we do with L21 if we know what the error is and can provide a nicer error message or catch it at a higher level and handle it more graceful. For now, if you don't have any other known / expected `Discordrb::Errors::UnknownError`s then I think the code should be this so that the user just sees a generic "something went wrong" error page:
```
      if params[:error]
        Sentry.capture_exception(discorderror)
        redirect_to launch_discord_signups_url(lti_launch_id: lti_launch_id), alert: 'You clicked cancel instead of authorize. Try again and click Authorize to be added to the Braven Discord server.' and return
      else
        raise
      end
```

sadleb: It also might be helpful for troubleshooting to add something like the following in there `Honeycomb.add_field('discord_authorization_cancelled', true)`. That way, we can easily query for this scenario in HC.

elenabaurkot: Ah okay, thanks, I didn't realize you could just raise without putting anything after

sadleb: In ruby, calling raise without any arguments will raise the last error. So this syntax is only for re-raising the caught exception b/c you decided not to handle it.

app/controllers/discord_signups_controller.rb
sadleb: yup!

app/controllers/discord_signups_controller.rb
```diff
@@ -142,14 +143,11 @@ def can_publish_latest?
...
     user_servers_response = Discordrb::API::User.servers("Bearer #{current_user.discord_token}")
     current_user_discord_servers = JSON.parse(user_servers_response.body)
 
-    current_user_discord_servers.each do |server|
-      if server["id"] == participant.discord_server_id && participant.discord_user_id != nil
-        return true
-      end
-    end
-    false
+    current_user_discord_servers.find { |server| server['id'] == participant.discord_server_id }.present?
```
sadleb, line 151: Also, I'm assuming `Discordrb::API::User.servers` returns an empty enumerable object and not `nil` if they are not in any servers. If it can return nil, that needs to be handled b/c `nil.find {}` will throw.

elenabaurkot: Good to know, I will test that out now and see

app/controllers/discord_signups_controller.rb
```diff
@@ -28,11 +28,16 @@ class DiscordServerIdError  < StandardError; end
   def launch
     authorize :discord_signup
 
-    @discord_state = SecureRandom.hex + ",#{@lti_launch.id}"
-    Honeycomb.add_field('user.discord_state', @discord_state)
-
-    # Save a CSRF token so we can verify in #oauth
-    current_user.update!(discord_state: @discord_state)
+    user_discord_state = current_user.discord_state
+    if user.discord_state.blank?
```
sadleb, line 32: Did you mean to use `user_discord_state.blank?`

elenabaurkot: yes 💯

app/controllers/discord_signups_controller.rb
elenabaurkot: Yes it returns an empty array 👍

sadleb (approved):
Lgtm

PR #917, opened 2021-11-16
Add useful Honeycomb fields to the Discord Signup process

elenabaurkot:
### Description
This update adds Honeycomb fields throughout the Discord signup process.

### Motivation and Context
To help debug any errors that may occur in the Discord signup process by sending the information used to signup a Discord user to Honeycomb. 

Asana Task: https://app.asana.com/0/0/1201274835996650/f


app/controllers/discord_signups_controller.rb
```diff
@@ -47,21 +48,23 @@ def launch
     if current_user.discord_token
       response = Discordrb::API::User.profile("Bearer #{current_user.discord_token}")
       @discord_user = JSON.parse(response.body)
+      Honeycomb.add_field('discord_user', @discord_user)
 
       if @discord_user['email'] && @discord_user['verified']
         course = Course.find_by_canvas_course_id!(@lti_launch.course_id)
+        Honeycomb.add_field('course.id', course.id)
```
rshipp, line 55: since course id is a `bigint` field, we have to convert it to a string (`course.id.to_s`) before sending to honeycomb. see https://github.com/bebraven/platform/wiki/Monitoring-&-Troubleshooting:-Honeycomb.io#ids-as-strings. (note this doesn't apply to salesforce IDs since they're already strings)

app/controllers/discord_signups_controller.rb
```diff
@@ -115,11 +118,13 @@ def oauth
 
       discord_response = JSON.parse(response.body)
       discord_token_expiration = Time.now.utc + discord_response['expires_in']
+      Honeycomb.add_field('user.discord_expires_at', discord_token_expiration)
       current_user.update!(
         discord_token: discord_response['access_token'],
         discord_expires_at: discord_token_expiration,
       )
     rescue Discordrb::Errors::UnknownError => discorderror
+      Honeycomb.add_field('discord_authorization_cancelled', true)
```
rshipp, line 127: should this be moved below inside the `if`?

app/controllers/discord_signups_controller.rb
elenabaurkot: good to know, thanks for the article 👍

app/controllers/discord_signups_controller.rb
elenabaurkot: I updated it to inside the `if` statement, I had originally put it above the `if` because I was thinking if it raised a different error would that also cancel the Discord authorization, or not necessarily? @rshipp

rshipp: Ah I'm not sure, I interpreted the `discord_authorization_cancelled` as specifically someone clicked the cancel button. If you think it makes more sense outside the `if`, that's fine too. Can you rename the field to `alert.discord_authorization_canceled` too?

app/controllers/discord_signups_controller.rb
```diff
@@ -27,6 +27,7 @@ class DiscordServerIdError  < StandardError; end
 
   def launch
     authorize :discord_signup
+    Honeycomb.add_field('salesforce.contact.id', current_user.salesforce_id)
```
sadleb, line 30: In `application_controller#add_honeycomb_fields` a bunch of common stuff is already taken care of for `current_user` and `@lti_launch` so I don't think we need to explicitly add fields for those two objects here.

app/controllers/discord_signups_controller.rb
```diff
@@ -47,21 +48,23 @@ def launch
     if current_user.discord_token
       response = Discordrb::API::User.profile("Bearer #{current_user.discord_token}")
       @discord_user = JSON.parse(response.body)
+      Honeycomb.add_field('discord_user', @discord_user)
 
       if @discord_user['email'] && @discord_user['verified']
         course = Course.find_by_canvas_course_id!(@lti_launch.course_id)
+        Honeycomb.add_field('course.id', course.id.to_s)
```
sadleb, line 55: Ditto on above comment. This should already happen automatically. You may want to double check though.

app/controllers/discord_signups_controller.rb
```diff
@@ -47,21 +48,23 @@ def launch
...
         program_id = course.salesforce_program_id
+        Honeycomb.add_field('salesforce.program.id', program_id)
 
         participant = SalesforceAPI.client.find_participant(
           contact_id: current_user.salesforce_id, program_id: program_id
         )
-
+        Honeycomb.add_field('salesforce.participant.id', participant.id)
```
sadleb, line 62: Have a look at `add_to_honeycomb_span` for `SFParticipant`. Maybe use that?

app/controllers/discord_signups_controller.rb
elenabaurkot: Going to leave it inside the `if` for now like you suggested, if an error could be thrown for something outside of the authorization cancellation I guess we probably wouldn't want to send that alert

app/controllers/discord_signups_controller.rb
rshipp: these are on me then, i don't know how to use that stuff yet. could you add some docs on what gets auto-added and when to the HC wiki page when you get a chance @sadleb?

sadleb: Task: https://app.asana.com/0/1201131148207877/1201378722712242

app/controllers/discord_signups_controller.rb
elenabaurkot: It looks like it does 👍

app/controllers/discord_signups_controller.rb
elenabaurkot: Updated to use add_to_honeycomb and removed this field and the program_id field for Honeycomb

app/controllers/discord_signups_controller.rb
```diff
@@ -55,13 +56,12 @@ def launch
...
         )
-
+        participant.add_to_honeycomb_span()
+        
         @discord_server_id = participant.discord_server_id
+        Honeycomb.add_field('salesforce.participant.discord_server_id', @discord_server_id)
         raise DiscordServerIdError, "No Discord Server Id found for Participant.Id = #{participant.id}" if @discord_server_id.nil?
 
-        contact = SalesforceAPI.client.get_contact_info(participant.contact_id)
-        nickname = DiscordBot.compute_nickname(contact)
```
sadleb, line 63: I assume these were removed on purpose bc they’re obsolete?

app/controllers/discord_signups_controller.rb
```diff
@@ -55,13 +56,12 @@ def launch
         participant = SalesforceAPI.client.find_participant(
           contact_id: current_user.salesforce_id, program_id: program_id
         )
-
+        participant.add_to_honeycomb_span()
+        
         @discord_server_id = participant.discord_server_id
+        Honeycomb.add_field('salesforce.participant.discord_server_id', @discord_server_id)
```
sadleb, line 62: Does this need a `.to_s` to avoid a bigint overflow or nah? Check with Ryan.

sadleb: Note to self, we should probably try to figure out how to monkey patch the Honeycomb code so we don’t have to remember to do this

app/controllers/discord_signups_controller.rb
```diff
@@ -115,12 +115,14 @@ def oauth
 
       discord_response = JSON.parse(response.body)
       discord_token_expiration = Time.now.utc + discord_response['expires_in']
+      Honeycomb.add_field('user.discord_expires_at', discord_token_expiration)
```
sadleb, line 118: It’s probably fine to do this here, but I think that next time the launch happens this would be autoinstrumented and that’s the span / trace we’d be interested in. Feel free to ignore this comment. Just musing.

app/controllers/discord_signups_controller.rb
elenabaurkot: I originally removed them because when we use the Discord webhook to call the bot "configure_member" command it takes care of adding the nickname so it is obsolete here, but then realized when running it again that I actually use the contact for that "configure_member" command so I just added it back in further down.

app/controllers/discord_signups_controller.rb
elenabaurkot: Just updated to a string, I'm thinking it does based on the article @rshipp posted above

app/controllers/discord_signups_controller.rb
```diff
@@ -73,11 +73,11 @@ def launch
             "Bot #{Rails.application.secrets.discord_bot_token}",
             @discord_server_id,
             @discord_user['id'],
-            current_user.discord_token,
```
sadleb, line 76: I assume removing nickname here has no impact? I don’t know how this API works but’s it’s weird that it can just figure out what the params are just from the values and not by the position or key name.

app/controllers/discord_signups_controller.rb
```diff
@@ -73,11 +73,11 @@ def launch
...
-            nickname
+            current_user.discord_token
           )
           Honeycomb.add_field('discord_web_hook.add_member', response)
 
+          contact = SalesforceAPI.client.get_contact_info(participant.contact_id)
```
sadleb, line 80: Actually, the contact id is also available on the participant. I don’t think we need to call back into Salesforce to get it.

app/controllers/discord_signups_controller.rb
rshipp: Discord server ID is already a string in Salesforce, so it shouldn't matter. I don't think patching this in honeycomb is desirable bc there's no way for it to know what's an int and what's a bigint.

app/controllers/discord_signups_controller.rb
rshipp: yeah this was just for the nickname, we don't need the contact anymore

app/controllers/discord_signups_controller.rb
elenabaurkot: Removing the .to_s because discord_server_id is a string column in our database and in salesforce so no need to convert

app/controllers/discord_signups_controller.rb
elenabaurkot: Yes because the before this was adding the member with the nickname and the discord_bot "configure_member" call was adding again after so taking this one out

app/controllers/discord_signups_controller.rb
elenabaurkot: Good call 👍

rshipp (approved):

PR #920, opened 2021-11-17
Discord give points

elenabaurkot:
### Description
When students complete the Discord signup assignment and get added to they Braven Discord server, they now received the full 10/10 points in their grade for the assignment. These points will not be weighed in to their actual grade, the assignment is graded for completion. 

### Motivation and Context
Asana Task: https://app.asana.com/0/0/1201339900860401/f

Screen Shot 2021-11-17 at 1 35 26 PM

Submission Details:

Screen Shot 2021-11-17 at 3 21 10 PM


app/controllers/discord_signups_controller.rb
```diff
@@ -161,4 +180,17 @@ def user_in_server?(participant)
...
+  end
+
+  # Only users enrolled as students can create a submission
+  # The LTIAdvantageAPI call to Canvas fails for all other enrollments.
+  def is_enrolled_as_student?(course)
+    course.sections.each do |section|
+      return true if current_user.has_role? RoleConstants::STUDENT_ENROLLMENT, section
+    end
+    false
+  end
```
rshipp, line 195: can you move this method to the User model? there's some similar ones in there, so i think it makes more sense and would be easier to reuse

app/controllers/discord_signups_controller.rb
```diff
@@ -134,6 +144,15 @@ def oauth
     redirect_to launch_discord_signups_url(lti_launch_id: lti_launch_id)
   end
 
+  # Shows a "thank you for submitting" page. Note that since there is no
+  # local discord_signup model, we use completed instead of show b/c it's a static endpoint with
+  # no id.
+  # GET /discord_signups/completed
+  def completed
```
rshipp, line 151: you'll need a matching view for this action. in this case, `app/views/discord_signups/completed.html.erb`

app/controllers/discord_signups_controller.rb
elenabaurkot: Updated this to be in the User model 👍

app/controllers/discord_signups_controller.rb
elenabaurkot: That makes sense, added in the view and added the screenshot of how it comes up in the submission details to the screenshots section

app/models/user.rb
```diff
@@ -228,6 +228,13 @@ def honeycomb_id_fields_map
       'canvas.user.id': canvas_user_id.to_s,
     }
   end
+
+  def is_enrolled_as_student?(course, user)
+    course.sections.each do |section|
+      return true if user.has_role? RoleConstants::STUDENT_ENROLLMENT, section
```
rshipp, line 234: instead of passing in a user, since this is already on the user model, you can just call `has_role?` directly without an object

elenabaurkot: Ah that makes more sense, didn't realize I could call it without user, just made this update

app/models/user.rb
```diff
@@ -228,6 +228,13 @@ def honeycomb_id_fields_map
       'canvas.user.id': canvas_user_id.to_s,
     }
   end
+
+  def is_enrolled_as_student?(course)
```
sadleb, line 232: Can we put this up near the `ta_for?` and `lc_for?` methods. It’s a similar concept.

app/controllers/discord_signups_controller.rb
```diff
@@ -83,6 +85,14 @@ def launch
           client.execute do |builder|
             builder.content = "!configure_member #{@discord_user['id']} #{contact['Id']}"
           end
+
+          if current_user.is_enrolled_as_student?(course)
```
sadleb, line 89: If I recall correctly, we have to do this bc LtiAdvantage doesn’t work for TA enrollments. Is that correct? If so, can you mention that in a comment? If it works for both, I think it’s better if we submit a score for both students and TA roles bc it may be useful to help us know if it’s been done when a TA submits a ticket about discord access.

app/models/user.rb
elenabaurkot: Yes, just updated and moved right above those two methods.

app/controllers/discord_signups_controller.rb
elenabaurkot: I was getting an error when doing it for TA so I believe that is correct, I just added in a comment here to mention that.

sadleb (approved):
Lgtm

PR #922, opened 2021-11-18
Salesforce Sync: Show Fellows in Attendance list before creating their Canvas account

sadleb:
### Description
Task: https://app.asana.com/0/1201131148207877/1200871464995937

In order for Fellows to show up in the attendance list before they've created their Canvas account, we need to have the local Platform role/section mappings in place. I've refactored the three services that do the following to be more self contained and have the logic specific to each separated (haven't renamed them though):
- sync program (sync_portal_enrollments_for_program)
- sync participant (sync_portal_enrollment_for_account)
- sync contact (sync_from_salesforce_contact)

This made it easier to remove the short-circuit in the Sync Program service that would stop if the user wasn't in Canvas, but instead to run all three sync services for everyone and have each decide what needs to be done. In the Sync Participant service, we can now manage the local Section/Role mappings for all Enrolled Participants regardless of Canvas account creation status.

_Note_: this gets us a step closer to being able to support syncing individual Participants on-demand instead of an entire program

### Test Plan
Note: I'm going to split the testing into categories below with some proposed steps for each, but you can likely test each category in parallel by doing steps from each at the right moments. Read through ALL the categories below first to understand the various things we need to test so that you can jump between the categories and test the proper things with Participants without having to create new ones for each category.

#### Prerequisites:
- A Program with a Cohort Schedule that has a Zoom meeting configured.
- A published Attendance Event assignment (aka ATTEND assignment) in the Accelerator course for this Program with a due date in the next couple days.
- A fully Enrolled LC Participant in the Cohort Schedule and in their own Cohort
- A fully Enrolled TA Participant
- Run the sync and use the sign-up links to create Canvas accounts for these Participants

#### Cat 1 (new feature) Enrolling a new Participant but not signing up for your Canvas account -> they still show up in Attendance list
1. Sign-in to Canvas with the LC account and go to the LC Playbook. Hit "Take Attendance" and see that there is no-one in the list for the ATTEND assignment.
2. Create a fully Enrolled Fellow Participant in the Cohort Schedule (but not any Cohort). Also create a TA Assignment for this Fellow Participant and the TA Participant used in the sanity testing below
3. (regression test) Run the sync but don't signup to create the Canvas account yet -> LC still doesn't see anyone in the list.
4. Now add them to the same Cohort as the LC and run the sync (still no Canvas account) -> now the LC should see them in the attendance list.
5. Have the LC take attendance for them and submit it
6. Now use the sign-up link from Salesforce to sign the Fellow up and create their Canvas account
7. They should have full credit for ATTEND assignment
8. Load the attendance list again as the LC. Their previous submission should pre-populate. In other words, they should see that they also marked off the attendance for that Fellow.
9. Now in Salesforce switch them to a new Cohort and run the sync
10. (regression test) When the LC loads attendance they should no longer be in the list. Also, they should be moved from the old Cohort Canvas section to the new one in the Accelerator course

#### Cat 2 (regression) Changing someone's email in Salesforce controls their login email
1. Create a new fully Enrolled Fellow Participant and run the sync (don't use the "send signup email" option).
2. They shouldn't get any email sent as part of the sync, but they should show up in the Platform admin dash after (and it shows Enrollments even though they're not on Canvas yet).
3. Change their email in SF, they should NOT get a "Braven login instructions" email (don't think it will update in Platform dash either)
4. Run the sync. They should still NOT get a "Braven login instructions" email, but now it should be updated in Platform dash.
5. Now use the sign-up link to create an account. Make sure the signup flow is using the most up-to-date email in SF
6. Now change email in SF again, this time they get a "Braven login instructions" email in real-time (don't confirm it yet)
7. Run sync, they should NOT get a second "Braven login instructions" email
8. Can still log in with old email, but not new one.
9. Now go confirm the new email. Can login with new one but not old one.

#### Cat 3 (regression) Mucking up your email in Canvas gets corrected
1. Use a fully Enrolled Fellow Participant who has created their account and can login to Canvas properly (e.g. after running the Cat 2 steps).
2. In Canvas, go to Account -> Settings. On the right under "Ways to Contact" click "+Email Address", add a new one, and confirm it.
3. Now click the "star" icon on the new email to change it to your default.
4. Delete the other email too (the one in our systems)
5. Logout and try logging in with the email you just seemingly changed it to. It shouldn't work.
6. You should still be able to login with the email in our systems.
7. Run the sync and nothing should change. You still have Canvas setup to send notifications to some new email but your login is still tied to what's in Salesforce.
8. Two scenarios:
    1. Scenario 1) Imagine a rogue staff member or LC or TA tries to help you out and logs into Canvas with their account that has the TA role (and Staff Account role actually). 
        1. * RECORD SCRATCH HALT * sound: Turns out TA's can't actually edit the login in Canvas. Phew. A Canvas Account Admin could, but only prod support and instructional designers have that so I don't think it's worth testing this flow out, BUT, if they did go edit the login manually in Canvas, the next sync should auto-correct it back to what's in Salesforce. To change it for good, they need to use Scenario 2 below
    2. Scenario 2) Imagine you contact support to change the login email to the new one. Go make that change in SF to match the one you manually tried to setup in Canvas
        1. From here, Cat 2) Step 6) and on is how things work.
        2. TODO: i'm actually not 100% sure what the behavior will be for the "Ways to Contact" changes made above. Maybe document what happens and let's talk about what should happen?

#### Cat (ChAll)
1. sanity check TA permissions -> a TA should be able to Act As User
2. sanity check a TA Caseload -> if there is a TA Assignment in Salesforce, a corresponding TA Caseload(their name) Canvas section should be created on sync
3. sanity check that Zoom links still sync. Make sure and test that if you Drop a Participant, their links are cleared / cancelled successfully
4. sanity check that when you're initially in a Cohort Schedule but no Cohort with a Canvas account, then after you are mapped to a Cohort and the sync runs you are moved from Cohort Schedule Canvas section to the Cohort Canvas section and that the due dates carried over from the schedule
5. sanity check that being able to login works case insensitively -> e.g. Brian@bebraven.org logs in the same as brian@bebraven.org 

### Types of changes

### Checklist:


app/controllers/salesforce_controller.rb
```diff
@@ -65,7 +65,6 @@ def update_contacts
       Honeycomb.add_field_to_trace('canvas.user.id', @canvas_user_id.to_s)
```
sadleb, line 65: This file is not important to review. Just small tweaks while I was in there that confused me as I was trying to understand change how the `SyncFromSalesforceContact` service was used/initialized.

app/models/user.rb
```diff
@@ -247,6 +255,25 @@ def password_required?
...
+
+  # Override Devise's method in devise/models/confirmable.rb
+  # to directly change the email instead of setting the unconfirmed_email
+  # column unless they are registered.
+  # We want to be able to change their email at any time without
+  # notifying them before they've actually created their account.
+  def postpone_email_change?
+    return false unless registered?
+    super
+  end
```
sadleb, line 275: Here is the relevant line we're changing the behavior of: https://github.com/heartcombo/devise/blob/main/lib/devise/models/confirmable.rb#L58

app/models/user.rb
```diff
@@ -247,6 +255,25 @@ def password_required?
...
   end
 
+  # Override Devise's method in devise/models/confirmable.rb
+  # to only send reconfirmation emails for registered users.
+  # We want to be able to change their email at any time without
+  # notifying them before they've actually created their account.
+  def reconfirmation_required?
+    return false unless registered?
+    super
+  end
```
sadleb, line 265: Here is the relevant line we're changing the behavior of: https://github.com/heartcombo/devise/blob/main/lib/devise/models/confirmable.rb#L53

app/services/sync_from_salesforce_contact.rb
```diff
@@ -1,56 +1,154 @@
...
+    @user.save! if @user.changed?
+
+    @user
+  end
+
+private
+
+  # Finds or creates a new user if one doesn't exist depending on the
+  # @create_new_user_if_missing setting.
+  def find_or_create_user()
```
sadleb, line 55: Refactoring `sync_portal_enrollments_for_program#find_or_set_up_user!` to this service since creating platform Users should be the responsibility of the Contact sync, not the Program sync.

app/services/sync_from_salesforce_contact.rb
```diff
@@ -1,56 +1,154 @@
...
+  # to work until they confirm the new email in case it was an accident or mistake.
+  # We only update Canvas if both Salesforce and Platform are in sync but somehow the
+  # Canvas login email get out of sync (most likely an admin or engineer messing it up).
+  def run
+    Honeycomb.add_field('salesforce.contact.id', @salesforce_contact.id)
     Honeycomb.add_field('salesforce.contact.email', @salesforce_contact.email)
 
-    if @user.unconfirmed_email&.downcase == @salesforce_contact.email.downcase
+    @user = find_or_create_user()
+    return nil unless @user.present?
```
sadleb, line 31: This behavior used to be in the program sync.

sadleb, line 31: This behavior used to be in the program sync.

app/services/sync_from_salesforce_contact.rb
```diff
@@ -1,56 +1,154 @@
...
+    @user = new_user
+  end
+
+  # Returns the email that should be stored in Platform based on the sync logic where the
+  # the Saleforce Participant is the source of truth, but they have to confirm any changes
+  # after creating their account. Also handles syncing that email to Canvas in case it was
+  # manually changed over there somehow (developer screw up or staff with admin access
+  # changing the login email which is different from the primary email used for Canvas
+  # communications / notifications).
+  def get_synced_email
```
sadleb, line 112: This method is simply a refactoring the original `run` method logic into here sine now the `run` does more than just syncs the Contact's email. There is no real change here.

Same core logic applies, but the `update!` now happens as a `save!` at the end if anything changed (and won't send a reconfirmation email always). The reason I changed that is so that it'll be easier to add future "sync" logic to this service and just have the database update once.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -1,48 +1,57 @@
...
+      Honeycomb.add_field('sync_portal_enrollment_for_account.complete?', false)
+
+      # Find or create local users and make sure they are in sync with Salesforce (and Canvas)
+      create_new_user = (sf_participant.status == SalesforceAPI::ENROLLED)
+      sync_contact_service = SyncFromSalesforceContact.new(@sf_contact, create_new_user)
+      @user = sync_contact_service.run
+
+      SyncZoomLinksForParticipant.new(sf_participant, @force_zoom_update).run
+
+      logger.info("Started sync enrollment for #{sf_participant.email}")
```
sadleb, line 36: The change here is that now:
1. the Zoom sync is run here instead of from the Program sync so that we can eventually just call this Participant sync directly without needing to sync the whole program
2. we just blindly call the Contact sync and it handle the logic of what needs to happen internally. Before, we were selectively calling the Contact sync only if we thought something changed that required a sync.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -85,6 +94,11 @@ def course_section_name
   end
 
   def drop_enrollment!
+    if @user.blank?
+      Honeycomb.add_field('sync_portal_enrollment_for_account.skip_reason', 'Dropped Participant never synced before')
+      return
+    end
+
```
sadleb, line 101: Just moving this check from the Program sync to here. It used to short-circuit when creating the user's if they weren't being Enrolled. Same thing is now happening here. The Contact sync won't create a new user when not Enrolling and this becomes a NOOP.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -106,68 +120,85 @@ def complete_enrollment!
...
+    Honeycomb.start_span(name: 'sync_portal_enrollment_for_account.sync_primary_enrollment') do
+      Honeycomb.add_field('canvas.course.id', canvas_course_id.to_s)
+      Honeycomb.add_field('canvas.section.role', role)
+
+      # Even before someone has registered and created their Canvas account, we create the Canvas
+      # section they would get enrolled in and ensure that they are in the local Section for it
+      # so that features which rely on the local Section roles work from day 1. E.g. attendance
+      section_name = section_name.blank? ? SectionConstants::DEFAULT_SECTION : section_name
+      section = find_or_create_section(canvas_course_id, section_name)
+      user.add_role role, section
```
sadleb, line 152: This whole PR is literally in service of this line * grimacing *

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -1,48 +1,57 @@
...
-# A helper service to perform the logic that takes folks who are confirmed as participants in the
-# program in Salesforce and creates their Canvas accounts. Also, moves/drops/completes their
-# enrollment when things change.
-# TODO: rename to SyncFromSalesforceParticipant
+# Sync's a Salesforce Participant to Platform and Canvas.
+# TODO: rename to SyncParticipant or SyncSalesforceParticipant
 class SyncPortalEnrollmentForAccount
 
-  def initialize(user:, portal_user:, salesforce_participant:, salesforce_program:)
-    @portal_user = portal_user
```
sadleb, line 10: I'd love feedback / more thought about this particular change. In the Program sync, we used to call into the CanvasAPI to find the user by their email, falling back to try to find them by their SIS ID if not found. We'd then pass that "portal_user" all the way through the layers, but we mostly just used it for the `canvas_user_id`. Can you think of any reason that we should call into Canvas every time to get that user instead of just relying on the local `User#canvas_user_id` ?

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -106,68 +120,85 @@ def complete_enrollment!
...
+      user.add_role role, section
+
+      Honeycomb.add_field('canvas.section.name', section.name)
+      Honeycomb.add_field('canvas.section.id', section.canvas_section_id.to_s)
+
+      unless @user.has_canvas_account?
+        Honeycomb.add_field('sync_portal_enrollment_for_account.skip_reason', 'No Canvas account yet')
+        logger.info("Skipping sync primary enrollment for #{user.email}. No Canvas account.")
+        return
       end
```
sadleb, line 161: This is where the short-circuit has been moved to from where it used to be in the Program sync so that we could add the local Section/roles before they had a Canvas account.

app/services/sync_portal_enrollment_for_account.rb
```diff
@@ -106,68 +120,85 @@ def complete_enrollment!
...
+      logger.info('Removing user enrollments from canvas')
+      enrollments&.each { |enrollment|
+        canvas_client.delete_enrollment(enrollment: enrollment)
+      }
+    end
+
+    # Even if they had no Canvas account, they may have local Section roles
+    # which we still need to remove since we give them local roles as soon as they
+    # become an Enrolled Participant in Salesforce.
+    user.remove_all_section_roles()
```
sadleb, line 136: TODO: what if they come back another semester? We should limit this to only sections in this course.

app/services/sync_from_salesforce_contact.rb
```diff
@@ -1,56 +1,154 @@
...
+    # Be careful to only add this to the span b/c when this is run as part of program sync
+    # there is not a single "user" associated with this trace. We're instrumenting information
+    # about multiple users in that case.
+    @user.add_to_honeycomb_span('sync_from_salesforce_contact')
+
+    @user.email = get_synced_email()
+
+    # Note: we don't send these to Canvas at the moment. Rely on users to update
+    # their own names or implement that later.
+    @user.first_name = @salesforce_contact.first_name
```
sadleb, line 42: TODO: Populate the `first_name` in the struct from the preferred first name in the JSON. We want to use that everywhere.

app/models/user.rb
```diff
@@ -55,6 +55,10 @@ def student_section_by_course(course)
     sections_with_role(RoleConstants::STUDENT_ENROLLMENT).find_by(course: course)
   end
 
+  def remove_all_section_roles
```
sadleb, line 58: TODO; fix to scope down to only course

app/services/sync_from_salesforce_contact.rb
sadleb: I think we need to do this in the SFParticipant struct too b/c the sync goes off that so it would just sync it back. Task since this is too much scope creep for this already massive PR: https://app.asana.com/0/1201131148207877/1201399664994349

app/services/sync_portal_enrollment_for_account.rb
sadleb: We discussed on a call and agree that we should be using the ID and not the email. If the code breaks in new places for any edge cases (like manually deleting a Canvas user in Canvas itself), then we can fix those when they arise.

PR #927, opened 2021-11-23
Added specs for Discord signup launch

elenabaurkot:
### Description
Added specs for launch in the Discord signups controller. 

### Motivation and Context
To help keep the Discord Signup code maintainable.


spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+require 'discordrb'
+require 'discordrb/webhooks'
+require 'salesforce_api'
 
 RSpec.describe DiscordSignupsController, type: :controller do
   render_views
 
+  let(:lti_advantage_api) { double(LtiAdvantageAPI) }
+  let(:sf_client) { instance_double(SalesforceAPI) }
+  let(:discordrb_api_user) { instance_double(Discordrb::API::User) }
```
sadleb, line 11: I never paid attention to `double` vs `instance_double` but it looks like we should lean towards using `instance_double`s b/c they have a little more reliable behavior: https://www.ombulabs.com/blog/rspec/ruby/spy-vs-double-vs-instance-double.html 

We'll probably want to update our guidance somwhere if we all agree to that and maybe do a mass find/replace at some point b/c currently we're arbitrarily choosing one vs the other.

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+
+    shared_examples 'renders initial steps' do
+      it 'renders initial steps' do
+        subject
+        expect(response.body).to include('Step 1: Sign In or Register for Discord')
+      end
+    end
+
+    context 'with TA Enrollment' do
+      let(:params) { { lti_launch_id: lti_launch.id, state: lti_launch.state } }
```
sadleb, line 44: Doesn't look like this `params` variable is used, right?

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+      before(:each) do
+        user.add_role RoleConstants::TA_ENROLLMENT, section
+        subject
```
sadleb, line 47: Maybe we the `subject` in the `before` block elsewhere, but I usually call it in the test itself. For example, would this run the `get launch` twice in the context's below since `it_behaves_like 'renders initial steps'` also calls `subject`? If so, we'd be slowing down the tests unecessarily.

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+        end
+      end
+    end
+
+    context 'with a student participant' do
+      let(:participant) { SalesforceAPI.participant_to_struct(create(:salesforce_participant_fellow)) }
+
+      before(:each) do
+        user.add_role RoleConstants::STUDENT_ENROLLMENT, section
+        subject
```
sadleb, line 70: Ditto on the comments above about using `subject` in the before. I think this would unnecessarily call it twice.

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
   render_views
 
+  let(:lti_advantage_api) { double(LtiAdvantageAPI) }
+  let(:sf_client) { instance_double(SalesforceAPI) }
+  let(:discordrb_api_user) { instance_double(Discordrb::API::User) }
+
+  let(:course) { create :course }
+  let(:user) { create :fellow_user }
+  let(:section) { create :section, course: course }
+  let(:sf_program_struct) { SalesforceAPI.program_to_struct(create(:salesforce_program_record)) }
```
sadleb, line 16: Is this used?

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+        allow(LtiLaunch).to receive(:from_id)
+          .with(user, lti_launch.id)
+          .and_return(lti_launch)
+      end
+
+      it_behaves_like 'renders initial steps'
+
+      it 'does not modify discord_state' do
+        expect(user.discord_state).to eq('teststate')
+      end
```
sadleb, line 91: Is the `subject` being called here? I'm not sure this test is actually testing anything. If you put the following in here, does it pass?
```
expect(LtiLaunch).to receive(:from_id)
          .with(user, lti_launch.id)
          .and_return(lti_launch)
```

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+    end
+
+    context 'without discord_state' do
+      let(:participant) { SalesforceAPI.participant_to_struct(create(:salesforce_participant_fellow)) }
+
+      before(:each) do
+        user.add_role RoleConstants::STUDENT_ENROLLMENT, section
+        user.discord_state = nil
+        subject
+        user.reload
```
sadleb, line 101: Ditto on above comments.

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+
+      it_behaves_like 'renders initial steps'
+    end
+
+    context 'with an expired discord_token' do
+      let(:participant) { SalesforceAPI.participant_to_struct(create(:salesforce_participant_fellow)) }
+
+      before(:each) do
+        user.add_role RoleConstants::STUDENT_ENROLLMENT, section
+        user.discord_expires_at = "2021-11-20 18:07:10.024855000 +0000"
```
sadleb, line 134: Is this hardcoded b/c you were trying to get an expiration in the past? If so, you can just do something like `user.discord_expires_at = Time.now.utc - 1.week`

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+      let(:participant) { SalesforceAPI.participant_to_struct(create(:salesforce_participant_fellow)) }
+
+      before(:each) do
+        user.update!(discord_token: 'faketoken')
+        stub_request(:get, "#{Discordrb::API.api_base}/users/@me").to_return(body: discord_user)
+        user.add_role RoleConstants::STUDENT_ENROLLMENT, section
+      end
+
+      context 'with a discord user that has no email' do
+        let(:discord_user) { { id: "fakeid1", email: nil, verified: false }.to_json }
```
sadleb, line 167: Should we create a `discord_user` factory similar to how we have factories for Salesforce and Canvas API responses?

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
 require 'rails_helper'
+require 'discordrb'
+require 'discordrb/webhooks'
+require 'salesforce_api'
 
 RSpec.describe DiscordSignupsController, type: :controller do
   render_views
 
+  let(:lti_advantage_api) { double(LtiAdvantageAPI) }
```
rshipp, line 9: this line can be deleted

spec/controllers/discord_signups_controller_spec.rb
rshipp: it's used on line 28 in the subject

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -1,9 +1,292 @@
...
+              subject
+              expect(LtiScore).to have_received(:new_full_credit_submission).once
+              expect(lti_advantage_api_client).to have_received(:create_score).once
+            end
+          end
+
+          context 'without student' do
+            before :each do
+              user.remove_role RoleConstants::STUDENT_ENROLLMENT, section
+              user.remove_role RoleConstants::STUDENT_ENROLLMENT
```
sadleb, line 266: Don't we get a fresh user for each test? I'm confused why we have to remove the student roles.

spec/controllers/discord_signups_controller_spec.rb
rshipp: i'd value readability/maintainability over speed in some of these cases. yes, some of them will call the subject twice, but that's pretty negligible if we save a couple lines of code in a few `it` blocks.

spec/controllers/discord_signups_controller_spec.rb
sadleb: That's the `params` attribute for the `get` method I think. If it was used there it would be `get :launch, params: params` instead.

spec/controllers/discord_signups_controller_spec.rb
rshipp: I'd lean against doing that, unless we start using it in other places. it's just a small json blob we use in one action in one controller.

spec/controllers/discord_signups_controller_spec.rb
rshipp: i can't use `let` because we modify `user` explicitly in a `before` block in a parent context, and `before` blocks happen after `let`.

spec/controllers/discord_signups_controller_spec.rb
rshipp: ah that's an oversight then, we should change line 28

spec/controllers/discord_signups_controller_spec.rb
sadleb: I guess I always think about the `before` block as the test setup and the `it` block is where the test is actually run, so it's less intuitive to me. Then again, that's probably just me and why I haven't fully started to leverage `it { some_matcher }` type tests.

spec/controllers/discord_signups_controller_spec.rb
sadleb: Sold

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: just removed 👍

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: no, I think I was before, but took it out now

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: Removed this on line 44, but I can't figure out what's wrong with line 28

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: Ah I see what you're saying, I updated the subjects all to be inside the 'it' statements now

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: I removed this variable on line 44, but I can't figure out what's wrong with line 28

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: Ah I see what you mean, I updated the subject to be in it statements for all of them instead of before

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: 👍

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: No I think I was before, but just removed it

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: I see what you're saying that this test isn't actually doing anything, I added in the subject, but am having a hard time getting it to pass after adding this in

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: Yes, just updated to this 👍

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: Just removed 👍

spec/controllers/discord_signups_controller_spec.rb
elenabaurkot: this is working and passing now, much thanks to @rshipp

rshipp (changes requested):
something weird must have happened during that last `pull --rebase` we did, the commits are all messed up. lmk if it takes you more than a few minutes to figure out and we can fix it in a call tomorrow.

elenabaurkot:
I think I was able to fix pull --rebase, and it looks OK to me, but let me know if anything looks off

elenabaurkot:
> looks like it deleted my commit, but the code appears to be in there, so it's probably fine

Whoops, I am not entirely sure I went about fixing this correctly sorry about that @rshipp

rshipp (approved):

PR #934, opened 2021-11-30
Prevent multiple programs from being synced to same Discord Server

elenabaurkot:
### Description
This update prevents an admin from running the Discord "sync_salesforce" command with a new program if there is already a different program synced to the Discord server and it prevents the running of "sync_salesforce" if there is more than one program linked to the same Discord server. Previously, if someone tried to run the Discord Bot sync_salesforce command without arguments it would just sync the first program it found (even if there was more than one linked) and if the command was run with a program id, it would sync that program to the Discord server even if another program was already synced.

### Motivation and Context
To prevent two programs from being synced to the same Discord server- this would cause problems because the Discord bot would just run commands for whichever program it finds first and ignore the other. 

Asana Task: https://app.asana.com/0/1201131148207877/1201229358750471/f

Message Discord Bot sends when trying to sync a new program to a Discord server that already has another program synced to it:
Screen Shot 2021-11-30 at 9 01 31 AM

Message Discord Bot sends when trying to run sync_salesforce if more than one program is synced to the Discord server:
Screen Shot 2021-11-30 at 9 01 40 AM


lib/discord_bot.rb
```diff
@@ -491,6 +491,24 @@ def sync_salesforce_command(event)
...
 
+    # Look for a current/future Program with this Discord Server ID.
+    programs = SalesforceAPI.client.get_current_and_future_accelerator_programs
+    relevant_programs = programs.select { |program| program['Discord_Server_ID__c'].to_i == server.id.to_i }
+
+    if relevant_programs.length > 1
+      message_content = "\n❌ This server has more than one program linked to it."
+      message_content << "\n Make sure only one program is linked to the server "
+      message_content << "before running the sync_program command."
+      message_content << "\n IDs of Programs linked to this server:"
```
rshipp, line 502: can you remove the space after the `\n` for all these? it just makes the message look a little off

lib/discord_bot.rb
```diff
@@ -491,6 +491,24 @@ def sync_salesforce_command(event)
...
+      relevant_programs.each {|program|
+        message_content << "\n #{program['Id']}"
```
rshipp, line 504: for this one specifically, can you either remove the space after the `\n`, or add an asterisk `\n* ` so this gets rendered as a bulleted list?

lib/discord_bot.rb
```diff
@@ -491,6 +491,24 @@ def sync_salesforce_command(event)
...
+      }
+      message = event.message.respond message_content
+      return
```
rshipp, line 507: since you're returning immediately, there's no need to do `message =`, as that variable will never be used

lib/discord_bot.rb
```diff
@@ -491,6 +491,24 @@ def sync_salesforce_command(event)
...
+    end
+
+    program = relevant_programs[0]
```
rshipp, line 510: prefer using `array.first` over `array[0]`, for readability

lib/discord_bot.rb
```diff
@@ -491,6 +491,24 @@ def sync_salesforce_command(event)
...
+    # Look for a current/future Program with this Discord Server ID.
+    programs = SalesforceAPI.client.get_current_and_future_accelerator_programs
+    relevant_programs = programs.select { |program| program['Discord_Server_ID__c'].to_i == server.id.to_i }
+
+    if relevant_programs.length > 1
+      message_content = "\n❌ This server has more than one program linked to it."
+      message_content << "\n Make sure only one program is linked to the server "
+      message_content << "before running the sync_program command."
+      message_content << "\n IDs of Programs linked to this server:"
+      relevant_programs.each {|program|
```
rshipp, line 503: for multi-line blocks, use `do`/`end` in place of `{ }`. this is mostly a style convention, but there are some reasons https://stackoverflow.com/questions/3680097/ruby-multiline-block-without-do-end

lib/discord_bot.rb
```diff
@@ -499,6 +517,16 @@ def sync_salesforce_command(event)
     elsif args.count == 1
       # Sync with the given Program.
       program_id = args.first
+
+      # We want to prevent a program from being synced to the Discord server
+      # if there is already a different program linked to the server
+      if !program.nil? && program_id != program['Id']
+        message_content = "\n❌ A program is already linked to this server."
+        message_content << "\n Please remove Program #{program['Id']} before adding Program #{program_id}"
```
rshipp, line 525: this guidance is probably fine for dev/testing, but we don't want regions/prod support to be manually removing discord server IDs or running end_program before they're ready. maybe replace this last line with something like 

```
Two programs cannot be running in a single server at the same time. You *must* end the previous program (#{program['Id']}) before beginning a new one. Ending a program is destructive and cannot be undone.
```

rshipp (commented):
mostly minor changes requested, looks good overall

lib/discord_bot.rb
elenabaurkot: ah yes, just removed 👍

lib/discord_bot.rb
elenabaurkot: Added the asterisks in, doesn't actually make a bullet, but the asterisks are showing

lib/discord_bot.rb
elenabaurkot: got it, that makes sense

lib/discord_bot.rb
elenabaurkot: good to know, I didn't realize there were any actual differences other than style

sadleb: TIL. def good to know!

lib/discord_bot.rb
```diff
@@ -491,6 +491,24 @@ def sync_salesforce_command(event)
     Honeycomb.add_field('sync_salesforce_command.args.count', args.count)
     Honeycomb.add_field('sync_salesforce_command.args', args)
 
+    # Look for a current/future Program with this Discord Server ID.
+    programs = SalesforceAPI.client.get_current_and_future_accelerator_programs
```
sadleb, line 495: Hmm, what if the ID is set on another program that is not current or future? E.g. Program for "Fall 2021 - University1" ends and we're gearing up for "Spring 2022 - University 1". We haven't run the command to end and kick everyone from the server for Fall. Then we try to sync the spring program. This wouldn't prevent that, right?

sadleb: If this is true, we may want to write a more direct query in `SalesforceAPI` for this. Something like what `get_participant_info_by(discord_invite_code:)` does

elenabaurkot: Ah good point, did not think of that, I will look into a query that can ended programs too (maybe just ones within the last year?)

rshipp: we don't care about ended programs, because the sync will never sync an ended program. we only care that there is one and only one program returned by current_and_future that matches this server id.

rshipp: > We haven't run the command to end and kick everyone from the server for Fall so it still has the Discord Server ID set. Then we try to sync the Spring program for the server. This wouldn't prevent that, right?

this is true, but seems out of scope for this command. if we want to enforce uniqueness on discord server id across all programs, can we do that from salesforce?

sadleb: kk, got it. Maybe just add a comment about that? I can see others reading this and not knowing that larger context and thinking there is a bug. I assume there will never be a reason we'd change the sync to do stuff with ended programs? In other words, we can rely on that assumption here and don't need to do anything to help prevent future refactors / enhancements from breaking due to this assumption?

rshipp: yeah we'll never need to sync ended programs, if someone's changing cohort mappings after a program ends, they should... stop doing that.

i do think enforcing uniqueness of discord server id on Salesforce program side is reasonable and we should queue up a task for you or SF team to handle that.

sadleb: kk, enforcing uniqueness on the Salesforce side is a 30 second change, so let's do that. Right now, I can't turn on the "unique" constraint on SF b/c there are duplicates for the test programs so I'll queue up a task.

sadleb: https://app.asana.com/0/1201131148207877/1201453841518466

sadleb (approved):
since we're going to enforce uniqueness using a Salesforce constraint, this lgtm, but defer to @rshipp

rshipp (approved):

PR #938, opened 2021-12-02
Heroku Connect: read-only access to Salesforce data with Rails models

sadleb:
### Description
Adds the infrastructure to expose Salesforce data to our Rails app using read-only models. The models point to tables created and managed using the [Heroku Connect add-on](https://devcenter.heroku.com/categories/heroku-connect).
* More info [in the wiki here](https://github.com/bebraven/platform/wiki/Salesforce#heroku-connect).
* Dev setup instructions in the [README here](https://github.com/bebraven/platform#heroku-connect).

A follow-up PR will be created to hook-up the `app/services/sync_salesforce_program` to get a list of `SFParticipant`s using these models instead of the `SalesforceAPI` as well as to [remove the obsolete student ID](https://app.asana.com/0/1201131148207877/1201453841518462) from the code.

### Motivation and Context
The in addition to the motivating factors mentioned in the wiki and in the comments on the [HerokuConnect::HerokuConnectRecord](https://github.com/bebraven/platform/blob/highlander/app/models/heroku_connect/heroku_connect_record.rb) class, 
this is laying the groundwork to implement [a more real-time sync](https://app.asana.com/0/1201131148207877/1200769578679244)

### Deployment Notes
* Configure the `HEROKU_CONNECT_DATABASE_URL` in our CI environment to point at the production following database.
* Configure the `HEROKU_CONNECT_DATABASE_URL` in our production environment to point at the production database, using a read-only user.

Here is a proof of concept, populating the existing `Salesforce::SFParticipant` struct using these 8 Rails models instead of calling into the API
image

### Types of changes

### Checklist:


config/database.yml
```diff
@@ -1,26 +1,67 @@
...
+# env.
 development:
-  <<: *default
-  database: platform_development
+  primary:
+    <<: *default
+    database: platform_development
+    host: <%= ENV['DATABASE_HOST'] %>
+    username: <%= ENV['DATABASE_USERNAME'] %>
+    password: <%= ENV['DATABASE_PASSWORD'] %>
```
sadleb, line 37: Even though these three options are now repeated, I think it's clearer. Before, they would be pulled into the `production` config but ignored b/c `<%= ENV['DATABASE_URL'] %>` take precedence deep in the Rails code. I spent several more minutes than I should have trying to understand this even though I had done a deep dive into that stuff back when we first set this up.

app/controllers/discord_signups_controller.rb
```diff
@@ -44,7 +44,7 @@ def launch
     # Show Staff/Faculty/Test TAs a different page, so they can't use this.
     # Note: We will likely adjust how this is determined later. See the Slack
     # discussions in this task for more info: https://app.asana.com/0/0/1201314155101899/f
-    if participant.role == SalesforceAPI::TEACHING_ASSISTANT && participant.volunteer_role != 'Teaching Assistant'
+    if participant.role == SalesforceAPI::TEACHING_ASSISTANT && participant.volunteer_role != SalesforceAPI::TEACHING_ASSISTANT
```
sadleb, line 47: Slightly outside the scope of this PR, but as I was exposing this using the `HerokuConnect::Candidate#candidate_role` method, I thought it was more consistent with the `SFParticipant::role` to have it be a symbol. It's only used in a couple places.

app/models/heroku_connect/cohort.rb
```diff
@@ -0,0 +1,44 @@
...
+  def zoom_prefix
+    ret = dlrs_lc1_first_name__c
+    ret << " #{dlrs_lc1_last_name__c[0].upcase}." if dlrs_lc1_last_name__c.present?
+    if dlrs_lc_total__c.to_i > 1 && dlrs_lc_firstname__c.present?
+      # Co-LCs
+      ret << " / #{dlrs_lc_firstname__c}"
+      ret <<  " #{dlrs_lc_lastname__c[0].upcase}." if dlrs_lc_lastname__c.present?
+    end
+    ret
+  end
```
sadleb, line 43: This is hard to read. Maybe have a look at the specs instead of the guts here.

spec/factories/heroku_connect/cohort.rb
```diff
@@ -0,0 +1,23 @@
...
+    createddate { Time.now.utc.strftime("%F %T") }
+    isdeleted { false }
+    sequence(:name) {|i| "C#{i} Tues (Heroku Connect)" }
+    sequence(:dlrs_lc1_first_name__c) {|i| "TestLC1First_#{i}"}
+    sequence(:dlrs_lc1_last_name__c) {|i| "TestLC1Last_#{i}"}
+    dlrs_lc_total__c { 1 }
+
+    association :program, factory: :heroku_connect_program
+    association :cohort_schedule, factory: :heroku_connect_cohort_schedule
+
```
elenabaurkot, line 14: Is this the way they do the naming on Salesforce for the LCs?

elenabaurkot (commented):
This looks good to me 👍

app/models/heroku_connect/heroku_connect_record.rb
```diff
@@ -0,0 +1,108 @@
...
+  # This is what lets the normal ActiveRecord relationships, like has_many and belongs_to
+  # work. See here for more info:
+  # https://devcenter.heroku.com/articles/heroku-connect-database-tables#system-columns
+  def self.primary_key
+    'sfid';
+  end
+
+  # By default, ActiveRecord select all columns. Change that to only select the ones
+  # we care about for this model considering there can be like 100 columns when we care about 3.
+  # We could easily blow up the memory usage if we load it all.
```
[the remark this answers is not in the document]
sadleb: This is filtering out rows with the `isdeleted` column set to true. I haven’t dug into how something gets this set to true, but my guess is they do a soft delete for a period of time which allows you to restore deleted records before they are permanently deleted. I just noticed that this column was there for nearly all tables, so I thought it made sense to not return “deleted” rows.

spec/factories/heroku_connect/cohort.rb
sadleb: Are you asking about the `dlrs_lc_xxx` stuff? If so, the DLRS prefix stands for something like dynamic lookup rollup summary. It’s an add-on they use to aggregate values from related records and store the result in a field. It’s simply a convention to help folks know where the data comes. We can make the fields we add whatever we want, so they are simply what the Salesforce folks chose when implementing these fields. 

Lmk if you were asking something else though!

app/controllers/discord_signups_controller.rb
rshipp: this will fail, volunteer role is a string and the constant is a symbol.

app/models/heroku_connect/cohort.rb
rshipp: i think we've changed the zoom prefix stuff in salesforce a few times. what's the plan to maintain both? or can we delete the formula from salesforce?

app/controllers/discord_signups_controller.rb
rshipp: i think i see where this is happening now, 
https://github.com/bebraven/platform/pull/938/files#diff-9b5e4430d50744a450ab293276a32cc75045acf8f327b8f1af3574b9f711793cR70
I'm still worried this will break other things. Did you look in the discord bot code for things that use participant structs / volunteer role?

sadleb: It's actually this line that makes it work since the existing stuff isn't hooked up to the new HerokuConnect models yet: https://github.com/bebraven/platform/pull/938/files#diff-13fdf5ec29a50b6bd27aa212b859d2d5177a4954a45e180f4338df03adf35d14R488 --> I think I looked for all possible references, but maybe I'll back that change out in this PR and tackle it more holistically in the follow-up one which will require a ton of testing and change all the constants to symbols and not just this one.

app/models/heroku_connect/cohort.rb
sadleb: I think we should delete the formula from SF. https://app.asana.com/0/1201131148207877/1201502404869771

app/controllers/discord_signups_controller.rb
sadleb: Actually, on second thought I think I should leave this change in this PR. I'm pretty confident that everything still works after grepping through the code and looking at all instances of:
- `'Teaching Assistant'`
- `TEACHING_ASSISTANT`
- `'Coach Partner'`
- `COACH_PARTNER`
- `volunteer_role`
- `SFParticipant`

I'm planning on switching the other three constants over to symbols in the follow-up PR and that PR will touch a lot more of the existing code and be harder to ensure nothing breaks without a massive testing effort. So getting this refactor out of the way here in order to make the new `HerokuConnect` models match and use symbols seems like a good approach?

sadleb:
Code review sign-off over Slack: https://bebraven.slack.com/archives/GL25HFFGE/p1639503024050800?thread_ts=1638906155.032900&cid=GL25HFFGE

PR #944, opened 2021-12-06
Generate Zoom Links CSV Validations

elenabaurkot:
### Description
This update checks the CSV file a user uploads to generate zoom links and if it finds any errors it let's the user know what they are and does not run the GenerateZoomLinksJob.

### Motivation and Context
To let a user know if they have uploaded an invalid CSV file that will not actually work to generate zoom links.

Asana Task: https://app.asana.com/0/1201131148207877/1201402263050098/f

Example error messages:
Screen Shot 2021-12-06 at 4 37 02 PM
Screen Shot 2021-12-06 at 4 36 15 PM


app/controllers/zoom_controller.rb
```diff
@@ -25,16 +32,73 @@ def generate_zoom_links
...
+    regex_header_validation(/^first.*name$/i, csv_headers[1], 'B', 'first_name')
+    regex_header_validation(/^last.*name$/i, csv_headers[2], 'C', 'last_name')
+  end
+
+  def regex_header_validation(regex, user_header_field, col_letter, header)
+    if regex.match(user_header_field.strip).nil?
+      raise GenerateZoomLinksError, "Your CSV file should have the headers 'email', 'first_name' and 'last_name' in that order. You have the header '#{user_header_field}' in column #{col_letter} instead of '#{header}'. Please edit this in your CSV and try again."
+    end
+  end
+
```
elenabaurkot, line 87: Added this comment to see if I should keep this meeting ID validation in app/services/generate_zoom_links as well?   Wanted to see what you guys thought. Also, in there it validates a meeting is 10 or 11 digits, but this link says it can be 9, 10, 11 (https://support.zoom.us/hc/en-us/articles/201362193-Joining-a-Zoom-video-call) wanted to double check that as well.  Also the todo below probably wouldn't be too tricky to add in here as well if we want that updated, but can also leave it for a separate PR if that's better.

config/environments/development.rb
```diff
@@ -12,7 +12,7 @@
   config.eager_load = false
 
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
```
sadleb, line 15: I don't think you want to check this in. I usually add a `TODO: tmp` or something when doing dev only changes that I need to back out and always search for `TODO` in my PR before opening it for review. I also try not to commit these sorts of things when possible.

elenabaurkot: Did not want to check this in, will definitely add that in from now on, thanks!

app/controllers/zoom_controller.rb
```diff
@@ -25,16 +40,88 @@ def generate_zoom_links
...
+  def check_meeting_settings(meeting_info)
+    # Zoom host must be licensed
+    host_zoom_account = ZoomAPI.client.get_zoom_user(meeting_info['host_id'])
+    if host_zoom_account['type'] != 2
+      raise GenerateZoomLinksError,'The meeting host does not have a licensed Zoom acount, please try again with a host that has a licensed account.'
+    end
+
```
elenabaurkot, line 103: Should "on-prem" accounts be allowed to be meeting host as well (seems like they have similar permissions to licensed accounts) or do we only use licensed accounts so this isn't something we need to worry about?

app/controllers/zoom_controller.rb
```diff
@@ -25,16 +40,88 @@ def generate_zoom_links
...
+  # I think it makes sense to put in here and let the user know if it's not going to work immediately
+  def validate_meeting(meeting_id)
+    unless /^\d{9,11}$/.match(meeting_id.strip)
+      # currently have 10 or 11 digit num in app/services/generate_zoom_links.rb
+      # this link says 9, 10 or 11
+      # https://support.zoom.us/hc/en-us/articles/201362193-Joining-a-Zoom-video-call
+      raise GenerateZoomLinksError, "Meeting ID '#{meeting_id}' format is invalid. It must be a 9, 10 or 11 digit number."
+    end
+  end
 end
```
elenabaurkot, line 127: Added this comment to see if I should keep this meeting ID validation in app/services/generate_zoom_links as well? Also, in there it validates a meeting is 10 or 11 digits, but this link says it can be 9, 10, 11 (https://support.zoom.us/hc/en-us/articles/201362193-Joining-a-Zoom-video-call) wanted to double check that as well.

app/controllers/zoom_controller.rb
```diff
@@ -25,16 +40,88 @@ def generate_zoom_links
...
+  # Moving this here from app/services/generate_zoom_links.rb (still have it in there for now too)
+  # Not sure if I should leave this in both places, but if it doesn't pass here, thinking it
+  # shouldn't be an issue there, so I am thinking I should take it out there.
+  # I think it makes sense to put in here and let the user know if it's not going to work immediately
+  def validate_meeting(meeting_id)
+    unless /^\d{9,11}$/.match(meeting_id.strip)
+      # currently have 10 or 11 digit num in app/services/generate_zoom_links.rb
+      # this link says 9, 10 or 11
+      # https://support.zoom.us/hc/en-us/articles/201362193-Joining-a-Zoom-video-call
+      raise GenerateZoomLinksError, "Meeting ID '#{meeting_id}' format is invalid. It must be a 9, 10 or 11 digit number."
```
elenabaurkot, line 124: Added this comment to see if I should keep this meeting ID validation in app/services/generate_zoom_links as well? Also, in there it validates a meeting is 10 or 11 digits, but this link says it can be 9, 10, 11 (https://support.zoom.us/hc/en-us/articles/201362193-Joining-a-Zoom-video-call) wanted to double check that as well.

app/controllers/zoom_controller.rb
```diff
@@ -17,6 +20,18 @@ def init_generate_zoom_links
   def generate_zoom_links
     authorize :Zoom
 
+    validate_meeting(params[:meeting_id])
+
```
elenabaurkot, line 24: Not sure if I need this validate_meeting method anymore since I have the rescue statement below, it seems if meetings have the wrong number of digits they also return the NotFound error handled below so should I just remove this?

app/controllers/zoom_controller.rb
sadleb: I think any account that allows us to use the API to generate links should be fine to allow. I don’t think we have any of those licenses so we can’t test, but it doesn’t hurt to allow it in case we ever add any. If it breaks, will fix it at that point

config/initializers/constants.rb
```diff
@@ -55,3 +55,10 @@ class CanvasConstants
   STAFF_ACCOUNT_ROLE_ID = 142
   ACCOUNT_ADMIN_ROLE_ID = 1
 end
+
+# Adding this error as a constant here because it is referenced from the rescuable
+# concern in /app/controllers/concerns/rescuable.rb. Since ZoomController inherits from AppicationController
+# the rescuable concern can't access custom exception classes defined inside ZoomController and
+# creates errors. With this in an initializer we can rescue from GenerateZoomLinksError
+# instead.
+class GenerateZoomLinksError < StandardError; end
```
sadleb, line 64: I don't think you need this and I think it could cause bugs. In `rescuable` and other places outside of the `ZoomController`, like in the specs, you just need to reference it as `ZoomController::GenerateZoomLinksError`. E.g.
```
    rescue_from ZoomController::GenerateZoomLinksError, :with => :handle_generate_zoom_links_error
```

app/controllers/zoom_controller.rb
sadleb: Yeah, I think you can remove this. It was a short-term solution that was easy instead of doing the `# TODO: use API calls to make sure the meeting is set up properly.` in `app/services/generate_zoom_links.rb`

app/controllers/zoom_controller.rb
```diff
@@ -25,16 +40,88 @@ def generate_zoom_links
...
-    strip_converter = ->(field) { field.strip }
+    strip_converter = ->(field, field_info) {
+      columns = ['first', 'second', 'third']
+
+      if field_info.index > 2
+        raise GenerateZoomLinksError, "You have too many columns in your CSV file, you should have 3 (email, first_name, last_name) and you have #{field_info.index + 1}."
+      end
+
+      # check for empty fields
+      if field_info.header.nil? || field_info.header.strip == ""
```
sadleb, line 51: `field_info.header&.strip.blank?` seems cleaner?

app/controllers/zoom_controller.rb
```diff
@@ -25,16 +40,88 @@ def generate_zoom_links
...
+        raise GenerateZoomLinksError, "You have an empty header in your CSV file. You left the header in the #{columns[field_info.index]} column blank."
+      elsif field.nil? || field == ""
```
sadleb, line 53: `field.blank?`

app/controllers/zoom_controller.rb
```diff
@@ -25,16 +40,88 @@ def generate_zoom_links
...
+        raise GenerateZoomLinksError, "You have an empty cell in your CSV file. You left the cell in row #{field_info.line} under the #{field_info.header} header blank."
+      end
+
+      # Ensure all fields under the email header (in column 1 - index 0) are formatted as emails
+      if field_info.index == 0
+        if /.+(@).+(\.).+/.match(field).nil?
```
sadleb, line 59: Can we piggyback off an existing email validation regex, namely? `Devise.email_regexp` ? It's defined in './config/initializers/devise.rb ' and was [introduced here](https://github.com/heartcombo/devise/pull/4001) if you want more info/context on it.

app/controllers/zoom_controller.rb
sadleb: See my above comments where I started to reply with a suggestion for what to do, but I think we should discuss on a call.

app/controllers/zoom_controller.rb
sadleb: We discussed on a call. We're going to move the job and all validation inside the `GenerateZoomLinks.run`.  The `run` will first validate everything and throw if something about the overall meeting or .csv is bad. Then it will kickoff the job to actually register each participant. For any rows that fail, the failures will go in the email sent after the job completes.

app/controllers/zoom_controller.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
+require 'zoom_api'
```
sadleb, line 3: I think you can get rid of this now that you're using the `GenerateZoomLinks` service

app/services/generate_zoom_links.rb
```diff
@@ -31,32 +39,62 @@ def run
...
+    # to copy/paste a value and end up with a trailing space.
+    strip_converter = ->(field, field_info) {
+      columns = ['first', 'second', 'third']
+
+      if field_info.index > 2
+        raise GenerateZoomLinksError, "You have too many columns in your CSV file, you should have 3 (email, first_name, last_name) and you have #{field_info.index + 1}."
+      end
+
+      if field_info.header.blank?
+        raise GenerateZoomLinksError, "You have an empty header in your CSV file. You left the header in the #{columns[field_info.index]} column blank."
```
sadleb, line 68: It's a little odd that these two checks happen in the strip converter instead of using some higher level method, but it seems fine and not worth putting more time into.

app/services/generate_zoom_links.rb
```diff
@@ -68,12 +106,69 @@ def add_registered_participants_to_meeting()
   end
 
   def validate_meeting(meeting_id)
-    unless /^\d{10,11}$/.match(meeting_id)
-      raise GenerateZoomLinksError, "Meeting ID '#{meeting_id}' format is invalid. It must be a 10 or 11 digit number."
+    begin
+      Integer(meeting_id)
+      meeting_info = ZoomAPI.client.get_meeting_info(meeting_id)
+    rescue ArgumentError
+      raise GenerateZoomLinksError,'Meeting ID must be an integer.'
```
sadleb, line 113: I'm wondering if this error message can be more helpful. If we have regional folks without much of a math background they may see this and have no idea what is actually wrong and still open a Halp ticket. Maybe "Meeting ID '#{meeting_id}' format is invalid. It should be all numbers. Make sure you have the correct meeting ID."

app/services/generate_zoom_links.rb
```diff
@@ -68,12 +106,69 @@ def add_registered_participants_to_meeting()
...
+    rescue RestClient::NotFound
+      raise GenerateZoomLinksError,'Meeting not found, make sure you have the correct meeting ID.'
+    end
+    meeting_info
+  end
+
+  def check_meeting_settings(meeting_info)
+    meeting_end_time = Time.zone.parse(meeting_info['start_time']).utc + (meeting_info['duration'] * 60)
```
sadleb, line 121: I looked at the specs to see if that had en example of what Zoom sends, but it just uses `Time.now`. What does one of these strings look like? I'm just wondering about the need to use `Time.zone.parse` vs `Time.parse`

Also, I'm concerned this won't work properly for recurring meetings. E.g. the `start_time` may be later today so this check would fail, but the final recurrence of the meeting already happened so it's actually ended. Wondering if this API endpoint would be safer? https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/pastmeetingdetails

app/services/generate_zoom_links.rb
```diff
@@ -68,12 +106,69 @@ def add_registered_participants_to_meeting()
...
+    if Time.now.utc > meeting_end_time
+      raise GenerateZoomLinksError,'This meeting has already ended, participants cannot be registered for past meetings.'
+    end
+
+    # Zoom host must be licensed (type 2) or on-prem (type 3)
+    # https://marketplace.zoom.us/docs/api-reference/zoom-api/users/user
+    host_zoom_account = ZoomAPI.client.get_zoom_user(meeting_info['host_id'])
+    if host_zoom_account['type'] != 2 && host_zoom_account['type'] != 3
```
sadleb, line 129: Can we create constants for these, like
```
class ZoomAccountType
  LICENSED=2
  ON_PREM=3
end
```

Then you would reference them as `ZoomAccountType::LICENSED` from inside this class or `GenerateZoomLinks::ZoomAccountType::LICENSED` from outside?

app/services/generate_zoom_links.rb
```diff
@@ -68,12 +106,69 @@ def add_registered_participants_to_meeting()
...
+      raise GenerateZoomLinksError,'The meeting host does not have a licensed Zoom acount, please try again with a host that has a licensed account.'
     end
 
-    # TODO: use API calls to make sure the meeting is set up properly.
-    # E.g. that "Registration = required" is turned on, email notifications are off, etc.
-    # https://app.asana.com/0/1174274412967132/1200659281191965
+    # Registration should be set to required
+    if meeting_info['settings']['approval_type'] == 2
```
sadleb, line 134: Ditto on using a constant instead of 2.

app/services/generate_zoom_links.rb
```diff
@@ -68,12 +106,69 @@ def add_registered_participants_to_meeting()
...
+  end
+
+  def regex_header_validation(regex, user_header_field, col, header)
+    unless regex.match?(user_header_field)
+      raise GenerateZoomLinksError, "Your CSV file should have the headers 'email', 'first_name' and 'last_name' in that order. You have the header '#{user_header_field}' in the #{col} column instead of '#{header}'."
+    end
+  end
+
+  def validate_emails(meeting_host_email)
+    participants_to_register.map.with_index do |participant, i|
```
sadleb, line 158: The strip_converter (which should probably be renamed at this point to something like `validation_converter`) already loops over every row and does row level validation. Now you're looping over them again. Does it make sense to do these checks in the converter?

The is basically the opposite of the above comment about checking things in the `strip_converter` that should more intuitively be checked at the top-level b/c they aren't row-level issues while these are row level issues.

app/services/generate_zoom_links.rb
elenabaurkot: I think the reason I did it here for the second check was because field.strip was throwing a NoMethodError if the fields were blank, should I be rescuing from that instead?

app/services/generate_zoom_links.rb
elenabaurkot: That makes sense, just updated 👍

app/services/generate_zoom_links.rb
elenabaurkot: Should it just be a plain constant like `NO_REGISTRATION_REQUIRED=2` or a class like the other 
` class ZoomApprovalType
    NO_REGISTRATION_REQUIRED=2
  end`

app/services/generate_zoom_links.rb
elenabaurkot: Yes I think that does make more sense, going to rename and update the strip converter to include this

app/services/generate_zoom_links.rb
sadleb: You check field.blank? below, so I don’t think that would throw anymore.

app/services/generate_zoom_links.rb
sadleb: Up to you. Both ways seem fine if there is only one approval type we care about.

app/services/generate_zoom_links.rb
elenabaurkot: It looks like Time.zone.parse and Time.parse return the same thing so I will use Time.parse. Also, trying to use the past meeting details end point is giving some issues because if the meeting hasn't happened yet it comes up as does not exist, which as long as the meeting does exist when we get info that's fine, but for recurring meetings it just gives the past meeting info for one occurrence, but no indication of when the actual last meeting is. I think using the meeting info end point I can check the final occurrence end time/duration and that seems to be working.

elenabaurkot: The only issue I could see arising would be if they scheduled a recurring meeting with no fixed time because then technically the meeting never ends, but I feel like it wouldn't make sense to do that because they will have a meeting time so I don't know if that would be an issue really.

sadleb: For the past meeting details endpoint, if nothing is returned b/c it hasn't happened yet then we would just say the validation passes b/c we know it hasn't ended.

As "for recurring meetings it just gives the past meeting info for one occurrence, but no indication of when the actual last meeting is" -- are you saying that if at least one occurrence has happened, but there are still more occurrences in the future that the endpoint still returns past meeting info? I would have expected this endpoint to only return info if the meeting had ended and there were no more future recurrences. If that was true, then the validation for whether the meeting has ended is just "does the pastmeetingdetails endpoint return info? If so, fail the validation and say that it's ended. If not, it's not a past meeting so pass the validation."

sadleb: Oh, I think I understand the confusion. I’m suggesting that you use that past meeting details endpoint only to validate if the meeting has ended. You’ll still need to use the other current meeting details and bdpoibt to validate the other settings

elenabaurkot: I am not sure I totally understand, with the past meeting details endpoint for a recurring meeting it seems to show the last past meeting, it doesn't give any info on how many more until it ends so I don't think it will solve if the meetings are done because it gives info even when more meetings are set. I updated to use get meeting info and for recurring meetings I can just check if there's any future occurrences and if there aren't any I can assume it's ended and it seems to be working.

app/services/generate_zoom_links.rb
elenabaurkot: I realized I can check the headers in the comment I added in lines 166-172 instead of what I had here- the only issue I am seeing if I do it that way, is if someone were to have white space an extra column the error gets a little funky and starts saying the column is blank under header blank  where otherwise it might say theres a blank header/extra column although I'm not sure any of those errors will be super helpful if someone makes the mistake of adding extra spaces in an empty column? but I can use that commented code and remove here if we think it makes more sense

app/controllers/zoom_controller.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
-
 require 'csv'
 
 class ZoomController < ApplicationController
   layout 'admin'
 
+  class GenerateZoomLinksError < StandardError; end
```
rshipp, line 7: since you're always raising and catching `GenerateZoomLinks::GenerateZoomLinksError` defined in the service, it's safe to remove this unused definition now.

app/views/errors/generate_zoom_links_error.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Generate Zoom Links Error</h1>
+
+<% if @exception.message %>
+  <p><%= @exception.message %></p>
+<% end %>
+
+<p>To return to the previous page and try again, <%= link_to 'click here', '/generate_zoom_links' %>.</p>
```
rshipp, line 7: we should use the path helper instead of hardcoding, like 
```suggestion
<p>To return to the previous page and try again, <%= link_to 'click here', generate_zoom_links_path %>.</p>
```
(or whatever the real path helper for this path is, I'm not sure)

app/controllers/zoom_controller.rb
elenabaurkot: good call, just removed this

app/views/errors/generate_zoom_links_error.html.erb
elenabaurkot: Updated to use the path helper 👍

sadleb (approved):
lgtm!

PR #948, opened 2021-12-14
Discord: sync server channels/roles to local db

rshipp:
### Description

New rake task to sync Discord server channels and roles to the local db, as a prerequisite for building more Discord Scheduler features. Note the ambiguous ID column names in the new tables.

Will need to schedule this on Heroku once it's deployed in order for it to have any effect.

### Motivation and Context

Task: https://app.asana.com/0/0/1201314155101895/f

### Types of changes


app/models/discord_server_channel.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class DiscordServerChannel < ApplicationRecord
+  # Note discord_server_id refers to local `DiscordServer.id`, not
+  # the external Discord `server.id`.
```
sadleb, line 5: Do we have a need for the external ID? If we do, would we just call that `discord_external_server_id`? I'm thinking about the `HerokuConnect` models where we have an `id` and a `sfid`. We don't really care about the local 'id' so we set `sfid` to the primary key and use that for joins. We could do something similar here and add a `DiscordServer#discord_id` which is the external ID and use that as the primary key. Then we could have this column be the external ID to use as the foreign key and there wouldn't be ambiguity.

app/services/sync_discord_servers.rb
```diff
@@ -0,0 +1,86 @@
...
+      # Loop over servers in the local db, and update their channel/role info
+      # from the Discord API.
+      DiscordServer.all.each do |server|
+        sync_server(server)
+      end
+    end
+  end
+
+  def sync_server(server)
+    Rails.logger.info("Starting sync for discord_server_id=#{server.discord_server_id}")
```
sadleb, line 22: To clarify, this one is the external discord server id? If so, I'm realizing that the naming ambiguity is going to be pretty confusing.

app/services/sync_discord_servers.rb
```diff
@@ -0,0 +1,86 @@
...
+    Honeycomb.start_span(name: 'sync_discord_servers.sync_server') do
+      Honeycomb.add_field('discord_server.id', server.id.to_s)
+      Honeycomb.add_field('discord.server.name', server.name)
+      Honeycomb.add_field('discord.server.id', server.discord_server_id)
```
sadleb, line 26: Is it worth using the pattern I started using recently for other things were we add a `add_to_honeycomb_span` method to the model and any consumer that want's the info just calls that so we have consistent fields for any consumer that wants to add that instrumentation?

app/services/sync_discord_servers.rb
```diff
@@ -0,0 +1,86 @@
...
+      sync_server_roles(server)
+    end
+  end
+
+  def sync_server_channels(server)
+    response = Discordrb::API::Server.channels("Bot #{Rails.application.secrets.discord_bot_token}", server.discord_server_id)
+    channels = JSON.parse(response.body)
+    channels.each do |channel_data|
+      next if channel_data['name'].start_with? COHORT_CHANNEL_PREFIX
+      next unless channel_data['type'] == TEXT_CHANNEL
```
sadleb, line 38: I get skipping non-text channels, but why skip cohort channels?

app/services/sync_discord_servers.rb
```diff
@@ -0,0 +1,86 @@
...
+      else
+        DiscordServerChannel.create!(
+          # Note since channel.discord_server_id field name is ambiguous, it
+          # will always refer to the local DiscordServer id.
+          discord_server_id: server.id,
+          discord_channel_id: channel_data['id'],
+          position: channel_data['position'],
+          name: channel_data['name'],
+        )
+      end
```
sadleb, line 57: Should we wrap this in a span and instrument it? Would we need to be able to easily query for a particular channel to understand what happened with the sync for it?

app/services/sync_discord_servers.rb
```diff
@@ -0,0 +1,86 @@
...
+        )
+      else
+        DiscordServerRole.create!(
+          # Note since role.discord_server_id field name is ambiguous, it
+          # will always refer to the local DiscordServer id.
+          discord_server_id: server.id,
+          discord_role_id: role_data['id'],
+          name: role_data['name'],
+        )
+      end
```
sadleb, line 83: Ditto on above question. Wrap in span and instrument this role sync or nah?

elenabaurkot (commented):
lgtm 👍

app/services/sync_discord_servers.rb
rshipp: We don't want or need cohort channels in our database for this feature.

app/services/sync_discord_servers.rb
rshipp: Doubt it, this code is unlikely to break, and low priority if it does. We already have activerecord autoinstrumentation.

app/services/sync_discord_servers.rb
rshipp: This is external yes, but why confusing? The IDs are visually distinguishable, the internal one will look like `5` and the external will look like `12987309128370912730978`.

app/models/discord_server_channel.rb
rshipp: yes we need the external ID, and since it's a string, I don't really want to use it as the primary key. I'm open to renaming the external ID or this table/model, but I can't think of any names that are any less confusing, and don't think it's worth spending too much time on. `discord_id` is ambiguous, and `discord_external_server_id` is confusing and long. I'm also not clear what you're suggesting in that last sentence.

app/services/sync_discord_servers.rb
rshipp: We only use this once, so I don't think so for now.

app/models/discord_server_channel.rb
sadleb: I guess I was hoping that we could easily have `DiscordServerChannel.discord_server_id` hold the value of `DiscordServer.discord_id` as the foreign key reference so that it's not ambiguous whether it's the internal or external id. If we were concerned about using a string as a foreign key for performance reasons, we could limit the length of the string to something reasonable (the default is 255 but you can choose a smaller `limit`). However, I looked into how to write the migration and it seems overly complicated so I agree that it's not worth spending more time in since this `discord_server_id` should never be used directly and is just for the associations to work. E.g. say we had `dsc = DiscordServerChannel.find(some_channel_id)`. It's expected that we'd use `dsc.discord_server.discord_id` and not `dsc.discord_server_d` to get the external id

app/services/sync_discord_servers.rb
sadleb: See my response to the above, but yeah, not confusing beyond the initial migrations b/c we'd never access something like `channel.discord_server_id` directly.

app/services/sync_discord_servers.rb
sadleb: I think we should start keeping this in mind going forward though. Anytime we're adding a second place in the code where we instrument some attributes of a model, let's not pick and choose. Instead, let's create a `add_to_honeycomb_span` method that loops over the attributes.

sadleb (approved):

app/models/discord_server_channel.rb
rshipp: I didn't add the extra alias only because I don't expect to ever reference from a channel/role back up to the server, but I can throw it in too. I think it's unlikely we'll ever need to read `channel.discord_server` at all, the references are just there to enable `server.channels`.

PR #947, opened 2021-12-14
Account Creation: Remove student_id from our code.

sadleb:
### Description
Task: https://app.asana.com/0/1201131148207877/1201453841518462

The SIS_ID in Canvas used to be `"BVSFID#{salesforce_contact_id}-SISID#{student_id}"` where `student_id` was their actual university student ID.

With this change, it's now: `"BVSFID#{salesforce_contact_id}-SISID#{platform_user_id}"`

We don't currently use the SIS_ID anywhere, but it needs to be unique per Canvas user and sometimes it's nice to be able to easily see either the Contact ID or Platform User ID when you're looking at a user in Canvas.

### Motivation and Context
We used to use it as a fallback way to lookup a user when we couldn't find them by email. Now we store both the Contact.ID
and User.canvas_user_id which are unique identifiers for a "person" and no longer need to rely on student id. It adds a lot of complexity to try and get that in the new Heroku Connect stuff (plus it may not exist or may be wrong). Also, there are PII concerns around having this in more places than we need.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #949, opened 2021-12-14
SYNC: recurring scheduled jobs support + ParticipantSyncInfo model

sadleb:
### Description
One more step closer to a [more real-time sync](https://app.asana.com/0/1201131148207877/1200769578679244). This PR adds support for
1. Recurring scheduled jobs that use Sidekiq along with options to prevent multiple from being run simultaneously
2. Caching the last successfully sync'd values using a `ParticipantSyncInfo` model

#### ParticipantSyncInfo model
The `ParticipantSyncInfo` has a column for every data point that if changed, would require a new sync to run. It also includes a `ParticipantSyncInfo::Diff` class that acts like a `ParticipantSyncInfo` but allows you to compare the current values to the last saved values. When we cutover the nightly sync to a "every X minutes" sync, the plan is to use an efficient query to get all the current values in Salesforce, compare them to the last successfully synced values, and only run a sync on those Participants that had changes.

The `ParticipantSyncInfo::SyncScope` module is responsible for this efficient query. It joins all the related tables and selects out the columns that a `ParticipantSyncInfo` stores. In a proof of concept against production, we were able to select the data from all 1837 current and future Participants in ~2 - 2.3 seconds.

In this PR, there is a `ParticipantSyncInfo.run_sync_poc` that does this select, compare, save operation adding a bunch of Honeycomb instrumentation and logs so we can start monitoring it's performance / stats in prod. It does not hook this up to the actual sync logic yet though. That will be in a follow-up PR using this prototype as a starting point:
* https://github.com/bebraven/platform/pull/946

**UPDATE:** follow-up PR is here:
* https://github.com/bebraven/platform/pull/950

#### Recurring Scheduled Jobs infrastructure
The second part about this PR is that it adds a recurring job / cron-like functionality for jobs. See `app/jobs/sidekiq_jobs/recurring_job.rb` for a ton of detail about that. The headline, is that when this is pushed a new scheduled job will start running every 10 minutes by default (configurable using the `SALESFORCE_SYNC_EVERY` env var) and call `ParticipantSyncInfo.run_sync_poc`. Further, this job only allows one instance of it to be running at a time so if the previous job hasn't finished before the next scheduled one starts, the next one will fail and we'll have to wait for the following scheduled time for it to try again.

Note: it may be helpful to reference the initial commit of the sidekiq infrastructure which this PR modifies:
* https://github.com/bebraven/platform/pull/819

Sentry / Sidekiq integration example. Note: this is for a job that failed to run b/c it was configured to only allow one at a time using `sidekiq-unique-jobs` and a second one tried to run while the first one was still running.
image

Honeycomb / Sidekiq integration (using custom middleware). Notice how there are traces for both a `RecurringJob` (that failed due to unique locks) as well as a normal `ActiveJob` with consistent fields.
image

   .... here are those fields

Testing the `sidekiq-unique-jobs` locking. This is a scheduled job that sleeps for 1 second, 20 times. The frequency of the scheduled job was set shorter than that so a new job was attempted to be queued while the first one was running. Notice how it fails and throws (which also makes it to Sentry and Honeycomb with the above integrations) 
image

### Types of changes

### Checklist:


app/models/heroku_connect/candidate.rb
```diff
@@ -52,11 +51,16 @@ class Role
...
   def candidate_role
-    coach_partner_role__c || record_type.name
+    HerokuConnect::Candidate.calculate_candidate_role(coach_partner_role__c, record_type.name)
+  end
+
+  # Heroku Connect can't sync formula fields so we're reimplementing the logic of
+  # "Volunteer_Role__c" here (which is now called Candidate Role in the UI).
+  def self.calculate_candidate_role(candidate_role_select, record_type_name)
+    candidate_role_select || record_type_name
   end
```
sadleb, line 65: You'll see this pattern in a couple places below. I think it still makes sense to have a `candidate_role` method directly on this `HerokuConnect::Candidate` model, but we also need it when running the sync using `ParticipantSyncInfo` and we already have the values needed to calculate it and don't need to query the database again. This just allows both spots to share the logic.

app/models/heroku_connect/participant.rb
```diff
@@ -65,29 +67,40 @@ def full_name
...
 
-    return []
+  # Note: candidate_roles, like Coach Partner, use a Leadership Coach record type
+  # in Salesforce so we need to check that as well.
+  def is_lc?
+    role == Role::LEADERSHIP_COACH &&
+    (
+      candidate_role == HerokuConnect::Candidate::Role::LEADERSHIP_COACH ||
+      candidate_role == HerokuConnect::Candidate::Role::LC_SUBSTITUTE
+    )
```
sadleb, line 92: Slight change to the old logic that just checked `!= SalesforceAPI::COACH_PARTNER`. We expect that the list of other roles could grow so let's explicitly look for what we consider an LC and assume the rest are non-LCs.

lib/salesforce_api.rb
```diff
@@ -24,14 +24,6 @@ class SalesforceAPI
...
   SFContact = Struct.new(:id, :email, :first_name, :last_name)
 
-  # TODO: get rid of student_id. We used to use it as a fallback way to lookup
-  # a user when we couldn't find them by email. Now we store both the Contact.ID
-  # and User.canvas_user_id which are unique identifiers for a "person" and no
-  # longer need to rely on student id. It adds a lot of complexity to try and get
-  # that in the new Heroku Connect stuff (plus it may not exist or may be wrong)
-  # See app/models/heroku_connect/participant.rb#to_struct for another spot to change
-  # Task: https://app.asana.com/0/1201131148207877/1201453841518462
-
```
sadleb, line 34: Done here https://github.com/bebraven/platform/pull/947 but forgot to remove comment.

app/jobs/sync_salesforce_changes_job.rb
```diff
@@ -0,0 +1,54 @@
...
+# Note: the Discord sync is separate.
+#
+# This is run as a scheduled CronJob every X minutes.
+# Only one of these jobs can be run at a time, so make sure and choose X to be
+# longer then the expected run-time, otherwise the alerts will be noisy (but
+# it should still work fine, the job will fail and the next one should work).
+#
+# See CronJob for more info about how this differs from our normal ActiveJobs
+# that are not scheduled.
+class SyncSalesforceChangesJob  < CronJob
```
sadleb, line 14: TODO: move to new top-level folder called `sidekiq_jobs` and rename.

app/jobs/cron_job.rb
```diff
@@ -0,0 +1,50 @@
...
+# Sentry and Honeycomb are integrated with these jobs using
+# the sentry-sidekiq gem and our custom Sidekiq::HoneycombMiddleware
+# module in honeycomb_sidekiq_integration.rb.
+# The root Honeycomb span for all Sidekiq jobs is: name=sidekiq.job
+# and the sidekiq.class field holds the actual class name.
+#
+# A very common requirement for scheduled jobs is for them to be unique,
+# meaning only one should be running at a time for a given set of parameters / args.
+# We use this gem to provide that functionality:
+# https://github.com/mhenrixon/sidekiq-unique-jobs
```
sadleb, line 29: TODO: if the on_conflict stuff starts raising errors, and we come to this code, make it easy to find the other on_conflict options so we can try other strategies if twe need to, like retry or whatever.

app/jobs/sync_salesforce_changes_job.rb
```diff
@@ -0,0 +1,54 @@
...
+                  lock_timeout: 0,
+                  # Once a lock is gotten, expire it after this time interval. This is REALLY
+                  # important b/c if the server restarts or crashes while a lock is held
+                  # and it's not released, then all future jobs wouldn't run if there was
+                  # no lock_ttl (time to live) specified.
+                  lock_ttl: JOB_INTERVAL,
+                  lock_prefix: 'sync',
+                  on_conflict: {
+                    client: :raise,
+                    server: :raise
```
sadleb, line 43: Change this to `log`. It should be smart enough to handle if we change the interval to be lower and start actually skipping runs.

lib/honeycomb_sidekiq_integration.rb
```diff
@@ -0,0 +1,55 @@
+require 'honeycomb-beeline'
+require 'active_job/queue_adapters/sidekiq_adapter'
+
+# Taken from here: https://github.com/honeycombio/beeline-ruby/issues/29
+# and modified to handle ActiveJob workers too.
+module Sidekiq
```
sadleb, line 6: Look into being able to subscribe the notification infrastructure of Rails instead of using this.

lib/honeycomb_sidekiq_integration.rb
```diff
@@ -0,0 +1,68 @@
...
+        yield
+        return
+      end
+
+      # Note: for ActiveJobs there is auto-instrumentation configured in config/initializers/honeycomb.rb
+      # For non-ActiveJobs that subclass our SidekiqJob base class there isn't auto-instrumentation b/c
+      # sidekiq doesn't publish any ActiveSupport::Notification events. Either way, by wrapping all of
+      # jobs in a manual span like this instead of relying on auto-instrumentation, uncaught Exceptions
+      # now magically populate the root error and error_detail fields so we don't have to go digging
+      # down into 'perform.active_job.exception' and 'perform.active_job.exception_object' fields.
```
sadleb, line 32: This comment addresses: https://github.com/bebraven/platform/pull/949/files#r770785138

lib/honeycomb_sidekiq_integration.rb
sadleb: nope (not easily at least, and we'd lose nice error handling)

app/jobs/active_jobs/application_job.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  # Most jobs are safe to ignore if the underlying records are no longer available
+  # discard_on ActiveJob::DeserializationError
+
+  include Sidekiq::Worker::Options
+  sidekiq_options retry: false
+
+  # Note: see lib/honeycomb_sidekiq_integration.rb for information about instrumentation.
+  # All jobs run later using Sidekiq have a sidekiq.job root span and sidekiq.class field
+end
```
sadleb, line 16: I moved this file to a subdirectory so the diff got messed up, but the only change is removing the `around_perform` stuff and relying on the `honeycomb_sidekiq_integration` for instrumentation.

app/models/heroku_connect/candidate.rb
elenabaurkot: I think I am a little confused by this part. From my understanding, the candidate role method checks if the candidate has a coach partner role set and if so it returns that, otherwise it just returns the record_type.name. When the candidate role method returns the value, does it set it somewhere? How does this make it so we don't need to query the database again?

elenabaurkot (commented):
lgtm 👍

app/models/heroku_connect/candidate.rb
sadleb: The `coach_partner_role__c` is a column on the `candidate` so that doesn't require another query. But on L58 when we pass `record_type.name` that's an association to the `record_type` table, so it has to query the DB to get the `name` out of that table. In `ParticipantSyncInfo`, that fancy `SyncScope` does all the joins to other tables in a single query and stores them as the attributes. So `ParticipantSyncInfo.role` would already hold this `record_type.name` without having to go back to the DB.

sadleb: See here: https://github.com/bebraven/platform/pull/949/files#diff-cff3519e635f0e5690ef55369dd9400a2aadf62a6a5b547d5f7efd30a70b989cR230

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
```diff
@@ -0,0 +1,75 @@
...
+class SyncSalesforceChangesJob  < RecurringJob
+
+  # Time duration in seconds before a sync job is considered dead and the lock
+  # should be available for another job to run. This is REALLY important b/c
+  # if the server restarts or crashes while a lock is held and it's not released,
+  # then all future jobs wouldn't run if there was no lock_ttl (time to live) specified.
+  #
+  # This should be configured as the the maximum amount of time that we expect a sync to
+  # take in the worst case. If we set it lower than an actual sync run, we'll end up
+  # with two syncs running at once which could / would duplicate work.
```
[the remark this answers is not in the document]
sadleb: By default, two jobs can run at once. We use `lock: :while_executing` below on L51 in to turn on the `sidekiq-unique-jobs` functionality and only allow one of these to run at a time. With this functionality turned on, if one jobs is running and another tries to run it just skips and logs it b/c we set the `on_conflict: { client: :log, server: :log }` setting below on L61. This `SALESFORCE_SYNC_MAX_DURATION` basically says "if one job is running so that no other jobs can run, after 15 minutes consider it dead and let new jobs run". This is a failsafe for a server restart or crash that causes the running job to never release the lock so that other jobs can run.

app/models/heroku_connect/candidate.rb
elenabaurkot: Ah okay nice that makes sense, thanks

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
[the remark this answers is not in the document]
elenabaurkot: Ah okay so after this time interval the lock is gone and a new job can run, which is where it would duplicate if it were to overlap

.env.example
```diff
@@ -55,6 +55,8 @@ PUMA_MAX_THREADS=5
 WEB_CONCURRENCY=0
 # End Puma
 SALESFORCE_HOST=test.salesforce.com
+SALESFORCE_SYNC_EVERY=10.minutes
+SALESFORCE_SYNC_MAX_DURATION=15.minutes
```
rshipp, line 59: i know we already had one of these, but let's not make a habit of adding more env vars that need to be eval'd. we don't lose that much by just hardcoding a unit, and making people set `120` instead of `2.hours`.

app/jobs/active_jobs/application_job.rb
rshipp: oh, that's not what i was thinking when i suggested different folders, but i guess it's fine if rails still knows where to find them all.

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+# Responsible for syncing information for ALL Participants in Current and
+# Future Salesforce Programs to Platform, Canvas, Zoom, etc.
+# Note: the Discord sync is separate.
+#
+# This is run as a scheduled job every ENV['SALESFORCE_SYNC_EVERY'] seconds.
```
rshipp, line 7: this comment explicitly mentions seconds, but isn't that inaccurate/meaningless since the env var contains its own units?

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
```diff
@@ -0,0 +1,75 @@
...
+  # with two syncs running at once which could / would duplicate work.
+  # Note: the longest sync with the old code was 52 minutes at the time of writing.
+  # This was primarily b/c each CanvasAPI call was taking on the order of 400ms and
+  # we were calling it for every Enrolled Participant regardless of whether there
+  # were changes. The new sync logic won't call the API if there are no changes,
+  # so we don't expect the max to be anywhere near that high. I'm arbitrarily
+  # choosing 15 min for now and we should tune this once the new sync code is
+  # live and we gather data.
+  SALESFORCE_SYNC_MAX_DURATION = begin
+    max_duration = Integer(ENV['SALESFORCE_SYNC_MAX_DURATION']) if ENV['SALESFORCE_SYNC_MAX_DURATION']
```
rshipp, line 31: how does Integer('15.minutes') work? is the example .env just incorrect and those vars are supposed to be integer seconds?

app/models/heroku_connect/participant.rb
```diff
@@ -65,46 +67,29 @@ def full_name
...
+  # TODO: remove the following three methods from SalesforceAPI and cutover all usage to these ones.
+  # https://app.asana.com/0/1201131148207877/1201515686512765
 
-    if ta_assignments.exists?
-      return ta_assignments.each.map { |tassign| "#{tassign.ta_participant.full_name}" }
-    end
+  # Note: at the time of writing, staff members are also setup with a TA
+  # role. In the future, we may want to distinguish staff from actual TAs.
+  def is_teaching_assistant?
+    role == Role::TEACHING_ASSISTANT
```
rshipp, line 76: i can't find where `Role::TEACHING_ASSISTANT` is defined, what is this?

lib/honeycomb_sidekiq_integration.rb
```diff
@@ -0,0 +1,70 @@
...
+    # @param [Hash] job the full job payload
+    #   * @see https://github.com/mperham/sidekiq/wiki/Job-Format
+    # @param [String] queue the name of the queue the job was pulled from
+    # @yield the next middleware in the chain or worker `perform` method
+    # @return [Void]
+    def call(worker, job, queue) # rubocop:disable MethodLength
+
+      # This is mostly for dev in case you don't configure Honeycomb
+      if Honeycomb.client.nil?
+        yield
```
rshipp, line 23: just bc i have no idea what this code is doing, do we need a `yield if block_given?` here? or does it just always yield?

app/jobs/active_jobs/application_job.rb
sadleb: In `application.rb` the following is what makes Rails ignore the subdirectory layout and treat them both as though they are in the `jobs` directory:
```
  autoloader.collapse('app/jobs/sidekiq_jobs')
  autoloader.collapse('app/jobs/active_jobs')
```

.env.example
sadleb: Yeah, I agree. I'll change these to be absolute seconds. I did this in the first place b/c I always get annoyed when I'm looking up docs on a config value and they don't mention the units and I have to hunt them down. I was like "oooh, Rails has nice abstractions so you don't have to care about the units". But even though we're being careful to throw if these can't be evaluated as Integers I think it opens up the likelihood that a future dev (or myself) forgets about this and mimics an existing ENV var usage pattern but forgets to sanitize the values. The security tradeoff isn't worth the convenience.

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
sadleb: Yeah. See my first response above. I'm going to change these to actually take seconds.

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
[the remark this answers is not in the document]
sadleb: Yup, exactly.

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
sadleb: That's a bug. I was supposed to `eval` first and then cast. I'm going to switch to only take seconds though per the above comments.

app/models/heroku_connect/participant.rb
sadleb: It's above on L54.

lib/honeycomb_sidekiq_integration.rb
sadleb: This was a copy/paste from the link at the top of the file. I also don't know why it yields, but yeah, `yield if block_given?` is safer. I'll change it (or maybe read up on what is passed to this `call` method and what this block is and add comments)

.env.example
sadleb: fyi: I changed `SALESFORCE_SYNC_MAX_DURATION` to expect just the integer number of seconds, but `SALESFORCE_SYNC_EVERY` is part of the `sidekiq.yml` and the scheduler gem uses this [fugit](https://github.com/floraison/fugit) gem to parse that. Using an int didn't work. So it's inconsistent with the other var, which isn't ideal, but at least there are no `eval`s introduced by this PR which is the main thing I'm trying to avoid.

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
sadleb: It's now expected to be specified in seconds (even though it could technically be some other unit).

lib/honeycomb_sidekiq_integration.rb
sadleb: This always get's passed a block and should yield. More info: https://www.rubydoc.info/github/mperham/sidekiq/Sidekiq/Middleware

elenabaurkot (commented):
lgtm!

rshipp (approved):

PR #950, opened 2021-12-17
SYNC: cutover sync services to use ParticipantSyncInfo and HerokuConnect models

sadleb:
### Description
TODO:

TODO: I keep getting the following in the console. I spent a long time troubleshooting but can't seem to get rid of it. I want to try running something in prod that uses the ParticipantSyncInfo class regularly (like every 10 min) but isn't mission critical to see if we get this error in prod too:

PG::UnableToSend (SSL SYSCALL error: EOF detected)
Troubleshooting notes:

tried adjusting idle_timeout and reaping_frequency options in database.yml
tried passing in TCP keep alive params to Postgres using the variables option in database.yml
tried the reconnect option in database.yml
tried printing out information about the connections in the pool. When this error happens, the is_active? method (which calls Postgres with something like SELECT; just to see if it responds returns false
If I call verify! on the dead connection, it revives it.
Normally, verify! is only called when teh connection is first established. Then it's supposed to just sit there open for the life of the thread. But somehow, it's dying and not being considered idle and not being reaped.
TODO: try the dead_connection_timeout option? Basically, just reap the connection every X seconds and re-establish it?

### Types of changes

### Checklist:


sadleb:
Closing this. We decided to go another route using SIS Imports instead. Here is the PR: https://github.com/bebraven/platform/pull/993

PR #951, opened 2021-12-20
Discord Scheduler: Add dynamic channel selection

rshipp:
### Description

Instead of a free text field for channel, use a channel dropdown that only shows channels in the currently selected server.

The code to allow "all cohort channels" messages is a little brittle, but I think it's worth it. I added HC tracking so we can delete it if we find out it's not.

Screen Shot 2022-01-04 at 10 11 41 AM

Screen Shot 2022-01-04 at 10 11 16 AM
Screen Shot 2022-01-04 at 10 09 49 AM

### Motivation and Context

Task: https://app.asana.com/0/0/1200935044334830/f

### Types of changes

### Checklist:


app/javascript/packs/discord_schedule.js
```diff
@@ -0,0 +1,39 @@
...
+    const serverID = e.target.value;
+    const channelSelect = document.querySelector('select#channel_id');
+    serverSelect.disabled = true;
+    channelSelect.disabled = true;
+
+    if (serverID === '') {
+      // Remove all old channels from the dropdown.
+      while (channelSelect.options.length > 0) {
+        channelSelect.remove(0);
+      }
```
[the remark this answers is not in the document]
rshipp: no, just the first one. the while loop removes all of them. i got this from here https://www.javascripttutorial.net/javascript-dom/javascript-add-remove-options/

app/controllers/discord_server_channels_controller.rb
```diff
@@ -0,0 +1,13 @@
+class DiscordServerChannelsController < ApplicationController
+
+  include DryCrud::Controllers::Nestable
+  nested_resource_of DiscordServer
+
+  # TODO: why do i need this????
+  prepend_before_action :set_parent
```
sadleb, line 7: Weird. Maybe try this in `config/routes` and see if it makes a difference?
```
  resources :discord_servers, only: [:index, :new, :create, :destroy] do
    resources :discord_server_channels, only: [:index]
  end
```
instead of
```
  resources :discord_servers, only: [:index, :new, :create, :destroy] do
    get '/channels', to: 'discord_server_channels#index'
  end
```

app/javascript/packs/discord_schedule.js
```diff
@@ -0,0 +1,39 @@
...
+          }
+          // Add an empty option to the top and make the field required.
+          channelSelect.add(new Option('', ''));
+          channelSelect.required = true;
+          // Add all this server's channels to the dropdown.
+          channels.forEach(channel => {
+            channelSelect.add(new Option(`#${channel.name}`, channel.id));
+          });
+          // Add a shortcut to send a message to all cohort channels.
+          channelSelect.add(new Option(`All cohort channels`, 0));
```
sadleb, line 31: Is it possible to send a special string like 'ALL_COHORT_CHANNELS' instead of `0`?

app/javascript/packs/discord_schedule.js
```diff
@@ -0,0 +1,39 @@
...
+
+    if (serverID === '') {
+      // Remove all old channels from the dropdown.
+      while (channelSelect.options.length > 0) {
+        channelSelect.remove(0);
+      }
+      channelSelect.add(new Option('Choose a server first', ''));
+      serverSelect.disabled = false;
+    } else {
+      fetch(`/discord_servers/${serverID}/channels.json`).then(response => {
```
sadleb, line 17: I suppose this is fine for a first pass since there could end up being many servers in the future and we wouldn't want to pre-load all channels for all servers in JS. If it's too slow/clunky, we can do a future pass to optimize.

sadleb: Would the non-AJAX approach be to render all server / channel options in HTML before displaying the page? Did you explicitly decide not to do it that way for some reason?

app/javascript/packs/discord_schedule.js
rshipp: I could make it use `cohort-` like I do in the Ruby, but there's no way to share constants between languages, so it seems equally hacky to me.

app/views/discord_schedule/index.html.erb
```diff
@@ -21,7 +21,13 @@
         <% @jobs.filter { |j| j[:server_id] == server[:id] }.each do |job| %>
           <div class="row">
             <div class="col-sm-9">
-              <p>In <strong>#<%= job[:channel] %></strong> on <span class="local-datetime"><%= job[:at] %></span></p>
+              <% if job[:channel] == 'cohort-' %>
```
sadleb, line 24: Is there an existing constant we should use?

app/controllers/discord_server_channels_controller.rb
rshipp: Changing it to this made no difference, i still need the set_parent
```ruby
    resources :discord_server_channels, only: [:index], path: '/channels'
```

app/views/discord_schedule/index.html.erb
rshipp: Not without require-ing the discord bot library inside a view, and I don't even know if that's a thing you can do

app/javascript/packs/discord_schedule.js
sadleb: fyi: we can share Rails variables with JS code using `.erb` like we do in `app/javascript/packs/rise360_module_versions.js.erb` and `app/javascript/packs/google_tag_manager.js.erb`

sadleb: I just don't like `0` b/c of it's traditional usage to mean `false` -- when I see special values set to 0 I immediately think `false`

app/javascript/packs/discord_schedule.js
rshipp: Another approach would be to have a JSON blob in the view that had all the data we need, but yes then we'd still need JS to control the dropdown options. Loading the channels from an API is future-proof and way cleaner. I don't foresee there ever being a speed issue with this, even if it took a couple seconds, people are used to that kind of slowness in UIs. That's why I made the selects disabled while loading, so there's no chance of clicking things before they're ready.

app/javascript/packs/discord_schedule.js
rshipp: The 0 here was because the rest of the values are integer IDs, and 0 is a valid integer value but not a valid ID, so it's easy to handle separately. All these values come through as strings anyway though, so if I can figure out how to get a Ruby constant in there I might as well use that.

app/controllers/discord_server_channels_controller.rb
sadleb: What do you mean by not work? When I change the code to this:
image

It seems to work?
image

image

app/javascript/packs/discord_schedule.js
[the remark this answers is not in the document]
elenabaurkot: ah okay that makes sense

app/controllers/discord_server_channels_controller.rb
rshipp: Without `before_action :set_parent`, it never calls the `Nestable` code, which means the list of channels it returns is *all* channels, instead of just the channels for the given server. You can see this in the log screenshot you posted, where it selects all channels instead of having a WHERE clause.

sadleb: gotcha. I vaguely remember there being a bug with this but I can't find any asana tasks or TODOs so I have no idea where we left off. I think we really just don't have any places other than `custom_content_versions_controller` and `project_submission_answers_controller` that have an `index` and are nestable. Both look like they just don't use the DryCrud provided list.

rshipp: ah, then maybe that part of nestable was never working. I'll just fix it in there instead.

app/controllers/concerns/dry_crud.rb
```diff
@@ -101,7 +101,7 @@ module Nestable
       # associated controller's index, show, edit, etc methods.
       included do
         delegate :parent_resource_classes, to: 'self.class'
-        before_action :set_parent
+        prepend_before_action :set_parent
```
sadleb, line 104: Whoa, was the bug fix really this easy? I feel like I spent a ton of time trying to fix this awhile back and gave up.

rshipp: Yeah it was calling the code already, just not in the right order. So when Nestable's `models_list` checked `if @parent.present?`, it failed and fell back to the `Model.all` behavior.

elenabaurkot (commented):
Looks good to me!

sadleb (approved):
lgtm!

PR #952, opened 2021-12-20
Generate new Platform signup link when invalid

elenabaurkot:
### Description
When viewing a user from the Platform Dashboard, if they user hasn't registered yet there is an option to send them a new signup link. If the Platform signup link has expired it won't work. This update changes it so that if the link is expired, the "Send New Sign-up Email" button will say "Generate Valid Sign-up Link and Send New Sign-up Email" instead. When the button is clicked, a new signup token is generated and saved to the database and updated on the Salesforce Contact and a valid sign-up email is sent. 

### Motivation and Context
So that Admin users can easily generate valid sign-up links to send users. 

Asana Task: https://app.asana.com/0/1174274412967132/1201142798375307/f

When the user has a valid signup link they see a button that says: 
Screen Shot 2021-12-20 at 3 40 01 PM

When they have an invalid link the button says:
Screen Shot 2021-12-20 at 3 40 08 PM


app/controllers/users_controller.rb
```diff
@@ -80,12 +80,14 @@ def send_confirm_email
...
-  # sign-up token and store it in Salesforce too.
-  # https://app.asana.com/0/1174274412967132/1201142798375307
   def show_send_signup_email
     authorize @user
+
+    if !@user.signup_period_valid?
+      raw_signup_token = @user.set_signup_token!
+      @user.sf_save_signup_token!(@user, raw_signup_token)
+    end
+
```
sadleb, line 90: I don't think we want to automatically regenerate it just by loading the User admin dash for a user. I think it would be ok to do it automatically if they actually click the "Send Signup Email" button though. So maybe just move this to `def send_signup_email` below?

Also, I don't think we should add this `sf_save_signup_token!(@user, raw_signup_token)` method to the `User` model. We generally try to avoid doing API calls in models b/c models are meant for local database access. E.g. if we had API calls in there, it may be easy to overlook that and put something in a database transaction that calls out to the network which is BAD (and which we've done in the past).

I would just make this:
```
    if !@user.signup_period_valid?
      raw_signup_token = @user.set_signup_token!

       # Set new User signup token on the Salesforce Contact record
       # Note: call it with the raw token, *not* the encoded one from the database b/c that's
       # what is needed in the Account Create Link.
        SalesforceAPI.client.update_contact(user.salesforce_id, {'Signup_Token__c': raw_signup_token })
    end
```

app/models/user.rb
```diff
@@ -215,6 +215,13 @@ def set_signup_token!
     raw
   end
 
+  def sf_save_signup_token!(user, raw_signup_token)
+    # Set new User signup token on the Salesforce Contact record
+    # Note: call it with the raw token, *not* the encoded one from the database b/c that's
+    # what is needed in the Account Create Link.
+    SalesforceAPI.client.update_contact(user.salesforce_id, {'Signup_Token__c': raw_signup_token })
+  end
+
```
sadleb, line 224: In addition to my notes above about not adding this method here, fyi that IF we did, the method could just be 
```
def sf_save_signup_token!(raw_signup_token)
  SalesforceAPI.client.update_contact(salesforce_id, {'Signup_Token__c': raw_signup_token })
end
```

It's an instance method on a user, so `salesforce_id` is the accessor to this `user`'s value.

app/views/users/show.html.erb
```diff
@@ -29,7 +29,11 @@
     <div class="card-body">
       <%= link_to "Edit", edit_user_path(@user), class: "card-link btn btn-warning" %>
       <%= link_to "Send Confirmation Email", send_confirm_email_user_path(@user), class: "card-link btn btn-success", method: :post if @user.registered? && !@user.confirmed? %>
-      <%= link_to "Send New Sign-up Email", send_new_signup_email_user_path(@user), class: "card-link btn btn-success" unless @user.registered? %>
+      <% if !@user.signup_period_valid? %>
+        <%= link_to "Generate Valid Sign-up Link And Send New Sign-up Email", send_new_signup_email_user_path(@user), class: "card-link btn btn-success" unless @user.registered? %>
+      <% else %>
+        <%= link_to "Send New Sign-up Email", send_new_signup_email_user_path(@user), class: "card-link btn btn-success" unless @user.registered? %>
+      <% end %>
```
sadleb, line 36: Above I said that it's probably fine to automatically generate a new signup token when sending the email, but on further thought, there may be invalid signup tokens that we need to regenerate but we don't want to send them this sign-up email through platform. Instead, we just want Salesforce to have the right once so that an automated email has a working signup link. Maybe check-in with Alexis and Jenna on this design and explore whether it would be better to have a new "Regenerate Signup Link" button that only shows if the token is invalid. It's a little more code/work so we should only do it if there really is a valid use case to regenerate but not as part of sending this platform email. If it's just a "hmm, that might be nice" response -> let's just stick with this approach b/c I think this is going to be rare now and we shouldn't spend too much time.

app/models/user.rb
elenabaurkot: ah okay, good to know

app/views/users/show.html.erb
elenabaurkot: That's actually what I had originally because @rshipp had mentioned we may not want to send the email always, but switched it over to this because we weren't sure when then might happen and another comment in the todo code made me think this made more sense lol, but it wouldn't be too difficult to switch it back to that if we think that makes more sense

rshipp: yeah my initial reaction was we might want that flexibility, but on second thought, flexibility no one's explicitly asked for, at the cost of more code, is usually a bad idea.

sadleb: agree

elenabaurkot: Ultimate decision in standup is to leave the button to automatically generate a new signup token when clicked -- no need for a separate button to generate a new signup token and not send an email at that point in time.

spec/controllers/users_controller_spec.rb
```diff
@@ -237,20 +248,54 @@
...
+        end
+
+        context 'with an invalid signup token' do
+          let(:user) { create :unregistered_user_with_invalid_signup_token }
+          let(:new_token) { 'new token'}
+
+          before(:each) do
+            allow(sf_client).to receive(:update_contact)
+          end
+
```
elenabaurkot, line 276: Is the test on lines 277-282 valid? I feel like I shouldn't have to call user.set_signup_token! on line 280, since it does it in the send_signup_email method, but the user's token doesn't get reset unless I add it here. I am also checking if sf_client it received update_contact, but is that actually testing since I told sf_client to be allowed receive update_contact on line 274? Just because it's allowed to receive it does it mean it actually will or is this still testing if it actually gets it?

rshipp: maybe you need a `user.reload` on line 280 instead? if it's just line 281 that's failing

elenabaurkot: that was it thank you! @rshipp

spec/controllers/users_controller_spec.rb
```diff
@@ -49,15 +49,26 @@
...
+        context 'with a valid signup token' do
+          let(:user) { create :unregistered_user_with_valid_signup_token }
+
+          it 'shows the "Send New Sign-Up Email" button' do
+            run_show
+            expect(response.body).to match(/<a.*href="\/users\/#{user.id.to_s}\/send_new_signup_email">Send New Sign-up Email\<\/a\>/)
+          end
+        end
+
+        context 'with an invalid signup token' do
```
sadleb, line 66: How are we making the sign-up token invalid? Maybe I'm blind b/c the test passes, but I just don't see it.

spec/controllers/users_controller_spec.rb
```diff
@@ -237,20 +248,53 @@
...
+
+          before(:each) do
+            allow(sf_client).to receive(:update_contact)
+          end
+
+          it 'generates a new, valid signup token and sends an email' do
+            run_send_signup_email
+            expect(sf_client).to have_received(:update_contact)
+            user.reload
+            expect(user.signup_token).not_to eq('starter token')
```
sadleb, line 280: I feel like it would be cleaner to save the original token in a variable first and then check that it was changed after instead of hard-coding this value that's in the factory. E.g. 
```
            invalid_token = user.signup_token
            run_send_signup_email
            expect(sf_client).to have_received(:update_contact)
            user.reload
            expect(user.signup_token).not_to eq(invalid_token)
```

spec/factories/users.rb
```diff
@@ -28,6 +28,17 @@
...
+      sequence(:salesforce_id) { |i| "003x%011dAAQ" % i }
+      sequence(:signup_token) { "starter token" }
+      sequence(:signup_token_sent_at) { "Mon, 07 Dec 2020 22:25:35 +0000" }
+    end
+
+    factory :unregistered_user_with_valid_signup_token do
+      sequence(:salesforce_id) { |i| "003x%011dAAQ" % i }
+      sequence(:signup_token_sent_at) { DateTime.now }
+    end
+
```
sadleb, line 41: You can nest these inside `unregistered_user` so that you don't have to repeat `sequence(:salesforce_id) { |i| "003x%011dAAQ" % i }` See below with a couple more `TODO` comments inline:

```
factory :unregistered_user do
      sequence(:salesforce_id) { |i| "003x%011dAAQ" % i }

      factory :unregistered_user_with_invalid_signup_token do
        sequence(:signup_token) { "starter token" }
       # TODO: does this work b/c this hardcoded date happens to be far enough in teh past that the token has expired? If so, I think it's cleaner to do something like this instead:
        sequence(:signup_token_sent_at) { 1.year.ago }
        sequence(:signup_token_sent_at) { "Mon, 07 Dec 2020 22:25:35 +0000" }
      end

      factory :unregistered_user_with_valid_signup_token do
        # TODO: why isn't the signup_token set there too?
        sequence(:signup_token_sent_at) { DateTime.now }
      end
  end
```

spec/controllers/users_controller_spec.rb
elenabaurkot: So I think the unregistered user doesn't have an sign-up token so it was invalid by default, but I guess it would probably be better/make more sense to have a test user with an explicitly invalid token so I will update that

spec/factories/users.rb
elenabaurkot: This is good to know, just made these updates

spec/controllers/users_controller_spec.rb
elenabaurkot: updated to this 👍

sadleb (approved):
lgtm!

PR #958, opened 2022-01-04
SECURITY: prevent most common passwords

sadleb:
### Description
Follow-up to: 
* https://github.com/bebraven/platform/pull/955

Just prevent using one of the most common passwords. The list was [taken from here](https://github.com/danielmiessler/SecLists/blob/master/Passwords/Common-Credentials/10k-most-common.txt) and cut down to only those 8 chars or more since we already prevent passwords less than 8 chars (2086 passwords in the final dictionary)

Requiring certain characters is explicitly against password strength best practices:
* https://en.wikipedia.org/wiki/Password_strength#Guidelines_for_strong_passwords
* https://spycloud.com/new-nist-guidelines/
    
This approach of just preventing the most common passwords is the bare minimum until we can one day cutover to a 3rd party auth service, like auth0.

Here is what trying to use "password" as the password now shows:
image

Mobile

### Types of changes

### Checklist:


app/services/check_password_complexity.rb
```diff
@@ -0,0 +1,36 @@
...
+class CheckPasswordComplexity
+
+  def initialize(password)
+    @password = password
+  end
+
+  # Returns true if the password is complex enough and false if not. "complex" here just
+  # means it's not one of the most common passwords greator than or equal to 8 characters.
+  def run
+    !password_dictionary.include?(@password)
```
rshipp, line 25: since the list appears to be all-lowercase, i think a small additional improvement here might be
```suggestion
    !password_dictionary.include?(@password.downcase)
```

sadleb: Good call. Added that and a spec for it.

elenabaurkot (commented):
lgtm 👍

rshipp:
@sadleb heads up just in case you didn't see the CI failure

sadleb:
> @sadleb heads up just in case you didn't see the CI failure

Thanks @rshipp - I missed that

PR #959, opened 2022-01-05
Performance optimization for nightly SyncCanvasGrades task.

sadleb:
### Description
By default, the `per_page` pagination setting for the CanvasAPI is 10. This PR increases it it to 100 (the max) for the nightly grade sync to reduce API calls and speed up the overall sync.

I tested with the Lehman Fall 2021 course and here was the difference:

* With per_page=100
`t1 = Time.now; proj_submissions = CanvasAPI.client.get_submission_data(329); t2 = Time.now; puts "### took #{t2-t1} time to fetch #{proj_submissions.count} submissions"`
`### took 168.1101254 time to fetch 9570 submissions`

Sooo, 95 API calls. Currently, the code  is 10 per page, so 950 API calls.
here is the same script results with the default per_page of 10

* With per_page=10
`### took 712.2104292 time to fetch 9570 submissions`

A 4 fold improvement.

For reference, here is the last trace of the year for the above Lehman course: https://ui.honeycomb.io/braven/datasets/platform.braven.org/result/efAgHsaYDa4/trace/gx4V5RdYR8e?span=7575d4f7119b23e6 which took a little over 10 minutes.

* Here is a span of a test course sync with these changes (33s total, ~20s of API calls):
https://ui.honeycomb.io/braven/datasets/stagingplatform.bebraven.org/result/8jAMPa1Smra/trace/hLQLCaRpkqg?span=b38736b56340e1fc

* Compared to a span in prod for the same test course without these changes (91s total, ~85s of API calls):
https://ui.honeycomb.io/braven/datasets/platform.braven.org/result/eynB7Qf7M74/trace/epktY8WPjqT?span=2aba92275a0750dd

I noticed this when working on: https://app.asana.com/0/1201131148207877/1201598672390752

### Types of changes

### Checklist:


app/services/sync_canvas_grades.rb
```diff
@@ -22,34 +22,37 @@ def run
...
-            submission['rubric_assessment'].each do |canvas_criterion_id, assessment_value|
-              Rails.logger.debug("Syncing rating for canvas_submission_id=#{data[:canvas_submission_id]}, canvas_rating_id=#{assessment_value['rating_id']}")
-
-              # Sometimes the rating id is nil. Only seen this on one test user,
-              # so maybe an isolated case? Send some stats to Honeycomb just in case.
-              unless assessment_value['rating_id']
-                Honeycomb.add_field('sync_canvas_grades.no_rating_id', true)
-                Honeycomb.add_field('sync_canvas_grades.assessment_value', assessment_value)
-                next
+          Honeycomb.start_span(name: 'sync_submission') do
```
sadleb, line 25: When linking to the trace I noticed some random fields on the top-level `` span like `sync_canvas_grades.assessment_value` and `sync_canvas_grades.no_rating_id`. This puts each submission inside its own span so the instrumentation makes sense.

Note that each submission was already creating a new `sql.active_record` span when we do the `upsert`, so this doesn't really add more to browse through when analyzing one of these traces.

Turn whitespace off when reviewing this. It's just a 3 line change.

elenabaurkot (commented):
Lgtm!

app/services/sync_canvas_grades.rb
rshipp: i don't see a problem with this change, but i think leaving those fields at the top level was intentional. the no_rating_id field is a flag that should never happen, were you seeing that in production?

app/services/sync_canvas_grades.rb
```diff
@@ -22,34 +22,37 @@ def run
...
+            Honeycomb.add_field('canvas_submission.id', submission['id'])
+            Rails.logger.debug("Syncing canvas_submission_id=#{submission['id']}")
+            data = CanvasSubmission.parse_attributes(submission, canvas_course_id)
+            CanvasSubmission.upsert(data, unique_by: [:canvas_submission_id])
+
+            if submission['rubric_assessment']
+              Honeycomb.add_field('canvas_submission.rubric_assessment.count', submission['rubric_assessment'].length)
```
rshipp, line 32: this doesn't seem to follow our conventions, since `canvas_submission.rubric_assesment` doesn't correlate with anything in our database.

lib/canvas_api.rb
```diff
@@ -164,6 +164,7 @@ def get_assignment_submissions(course_id, assignment_id, filter_out_unopened_ass
   # https://canvas.instructure.com/doc/api/submissions.html#method.submissions_api.for_students
   def get_submission_data(course_id)
     response = get("/courses/#{course_id}/students/submissions", {
+      'per_page': 100,
```
rshipp, line 167: this part seems fine, i'm guessing 100 is the max we can do?

app/services/sync_canvas_grades.rb
sadleb: These `rubric_assessment`s are turned into `canvas_submission_rating` models in our database. I guess `canvas_submission_rating.count` follows the convention more (but not really)? I'll just put it back to `sync_canvas_grades.rubric_assessments.count` b/c this isn't a priority at the moment.

lib/canvas_api.rb
sadleb: Yeah, 100 is the max.

app/services/sync_canvas_grades.rb
sadleb: Yeah, it's happening in prod. Toggle through the `sync_submissions` spans on this trace to see some examples: https://ui.honeycomb.io/braven/datasets/platform.braven.org/result/efAgHsaYDa4/trace/gx4V5RdYR8e?span=7575d4f7119b23e6

sadleb: Note that the `sync_canvas_grades.rubric_assessments.count` field was also at the top-level which made no sense b/c it was the count of the last submission processed that had assessments

app/services/sync_canvas_grades.rb
rshipp: yeah anything with `.count` should probably just use the local-namespace convention

PR #962, opened 2022-01-05
Discord Scheduler: allow role mentions

rshipp:
### Description

Allow mentioning roles by name, instead of only `@everyone`.

### Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1201314155101891/f

### Types of changes

### Checklist:


spec/services/schedule_discord_message_spec.rb
```diff
@@ -5,7 +5,7 @@
 require 'discord_bot'
 
 RSpec.describe ScheduleDiscordMessage do
-  describe '#run' do
+  describe '.run' do
```
elenabaurkot, line 8: What is the difference between using describe '#run' vs '.run'?

elenabaurkot (commented):
Lgtm!

spec/services/schedule_discord_message_spec.rb
rshipp: `#blah` is the syntax used for controller actions, like `#new`. i think we just have it copied in other places in our specs by accident. these names are just by convention though, so it doesn't break anything either way.

app/helpers/discord_schedule_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+  ROLE_MENTION_REGEX = /<@&(\d+)>/
+
+  def render_discord_message(message)
+    # For now, this is just the inverse of
+    # ScheduleDiscordMessage.convert_role_mentions.
+    message.scan(ROLE_MENTION_REGEX).flatten.each do |discord_role_id|
+      # This is a role mention.
+      role = DiscordServerRole.find_by(discord_role_id: discord_role_id)
+      # Skip invalid role IDs, instead of crashing.
+      next unless role
```
sadleb, line 11: Worth adding a Honeycomb field?

app/services/schedule_discord_message.rb
```diff
@@ -35,10 +36,28 @@ def run
...
 
       # Schedule the job.
       SendDiscordMessageJob.set(wait_until: run_at).perform_later(server_id, channel_name, message_content)
     end
   end
+
+  def self.convert_role_mentions(server, message)
+    if message.include? '@'
+      # Possible role mention(s).
+      server.roles.sort_by { |r| -r.name.length }.each do |role|
```
sadleb, line 49: This seems fine, but I had to look up `sort_by` usage to understand the `-r` syntax. Another option is `server.roles.sort_by { |r| r.name.length }.reverse` which may be more readable?

app/services/schedule_discord_message.rb
```diff
@@ -35,10 +36,28 @@ def run
...
+        # Try each role, starting with the one with the longest name.
+        # Otherwise we might see `@Fellow: Tuesday` and assume the mention was
+        # for the `@Fellow` role instead of the `@Fellow: Tuesday` role.
+        if message.include? "@#{role.name}"
+          # To mention a role in a Discord message, use the syntax:
+          # <@&ROLE_ID>.
```
sadleb, line 55: This comment is a bit confusing. If I understand correctly, to mention a role as the end user you just use the normal `@rolename` syntax, but internally we convert that to `<@&ROLE_ID>` for rendering purposes?

app/helpers/discord_schedule_helper.rb
```diff
@@ -0,0 +1,18 @@
+module DiscordScheduleHelper
+  ROLE_MENTION_REGEX = /<@&(\d+)>/
```
sadleb, line 2: It took me a minute to figure out what this was. Maybe add a comment here about how this is the format used in `convert_role_mentions`?

app/helpers/discord_schedule_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+  def render_discord_message(message)
+    # For now, this is just the inverse of
+    # ScheduleDiscordMessage.convert_role_mentions.
+    message.scan(ROLE_MENTION_REGEX).flatten.each do |discord_role_id|
+      # This is a role mention.
+      role = DiscordServerRole.find_by(discord_role_id: discord_role_id)
+      # Skip invalid role IDs, instead of crashing.
+      next unless role
+
+      message = message.sub("<@&#{discord_role_id}>", "@#{role.name}")
```
sadleb, line 13: Is this when we're actually sending the message content to discord? Or is this just to show it in the platform UI? If it's us creating the message to send to discord, I don't get why we convert the original `@rolename` to `<@&ROLE_ID>`?

UPDATE: I googled and found out that `<@&ROLE_ID>` is actually the expected format for a role mention in a discord bot: https://stackoverflow.com/questions/52042695/discord-bot-role-mentioning. Maybe mention that somewhere and link to the docs?

spec/services/schedule_discord_message_spec.rb
elenabaurkot: ah okay, so is '.run' the standard convention for when you describe actions in a service and '#run' is standard for controllers?

sadleb (commented):
I eventually figured out what is going on here, so there's no need to respond to each comment but I'm leaving them there to highlight what was confusing for a first time reader. I think the key was just not knowing that `<@&ROLE_ID>` is the format expected by Discord when sending a message from a bot. I initially assumed you were just doing some validation / sanitization on the user input.

app/services/schedule_discord_message.rb
rshipp: ah yeah that is better

spec/services/schedule_discord_message_spec.rb
rshipp: if I'm describing a controller action, I use `#`. if I'm describing anything else, including a non-action method in a controller, I try to remember to use `.`.

sadleb: Actually, the convention is `#` for instance methods and `.` for class methods: https://www.betterspecs.org/

app/helpers/discord_schedule_helper.rb
rshipp: nah this isn't interesting, i just wanted to make sure entering weird stuff that accidentally matched the role-mention format wouldn't crash things.

spec/services/schedule_discord_message_spec.rb
elenabaurkot: Good to know, thanks 👍

app/helpers/discord_schedule_helper.rb
rshipp: updated comments to hopefully clarify

sadleb (approved):

PR #966, opened 2022-01-06
Set Discord token to nil when receiving unauthorized error after authorizing

elenabaurkot:
### Description
Several users had issues when authorizing Braven to access their Discord account. They received a token, but then the Discord assignment wouldn't load and they have a Discordrb::Errors::UnknownError (with 401:unauthorized error detail). This update rescues the error, resets the user's Discord token to nil alerts the user to try again. 

### Motivation and Context
Asana Task: https://app.asana.com/0/1201131148207877/1201625170018356/f

Screen Shot 2022-01-06 at 5 00 13 PM


app/controllers/discord_signups_controller.rb
```diff
@@ -64,7 +64,18 @@ def launch
...
+      begin
+        response = Discordrb::API::User.profile("Bearer #{current_user.discord_token}")
+      rescue Discordrb::Errors::UnknownError => discorderror
+        Sentry.capture_exception(discorderror)
+
+        current_user.update!(discord_token: nil)
+        Honeycomb.add_field('alert.discord_token.reset', true)
+
+        lti_launch_id = @discord_state.split(",")[1]
+        redirect_to launch_discord_signups_url(lti_launch_id: lti_launch_id), alert: 'Something went wrong, please try again.' and return
```
rshipp, line 76: i think we already have `@lti_launch.id` here, does that work instead of parsing it out of the state?

app/controllers/discord_signups_controller.rb
```diff
@@ -64,7 +64,18 @@ def launch
     end
 
     if current_user.discord_token
-      response = Discordrb::API::User.profile("Bearer #{current_user.discord_token}")
+      begin
+        response = Discordrb::API::User.profile("Bearer #{current_user.discord_token}")
+      rescue Discordrb::Errors::UnknownError => discorderror
+        Sentry.capture_exception(discorderror)
```
sadleb, line 70: Note that the `Discordrb::Errors::UnknownError` will not end up in Honeycomb like this. We'll still have the underlying `RestClient::Exception` but if we want the Discord error we'd have to do something like this so that the error bubbles through a span so that Honeycomb automatically puts it in the `error` and `error_detail` fields:
```
begin
  Honeycomb.start_span(name: 'discord_signups_controller.get_discord_user') do
    response = Discordrb::API::User.profile("Bearer #{current_user.discord_token}")
  end
```
Note that if for some reason, we didn't want to wrap this in a span then we'd have to manually add them like this [example here](https://github.com/bebraven/platform/blob/highlander/app/controllers/salesforce_controller.rb#L74)

app/controllers/discord_signups_controller.rb
elenabaurkot: good call, that does work

app/controllers/discord_signups_controller.rb
rshipp: There's nothing valuable there. We already have RestClient autoinstrumentation for the underlying request.

spec/controllers/discord_signups_controller_spec.rb
```diff
@@ -152,6 +152,30 @@
...
+      end
+
+      it 'redirects to discord signups launch page' do
+        expect(subject).to redirect_to(launch_discord_signups_path(lti_launch_id: lti_launch.id))
+      end
+
+      it 'shows alert that something went wrong' do
+        subject
+        expect(flash[:alert]).to match /Something went wrong, please try again./
+      end
```
elenabaurkot, line 176: I was originally trying to add in another test here with the shared example it 'renders initial steps', but it only shows the redirect- when redirecting is there a way to check what the page will render after the redirect? (not sure if it's really necessary, just curious)

app/controllers/discord_signups_controller.rb
elenabaurkot: @sadleb I updated to what you recommended here, @rshipp does this look alright to you for handling the `Discordrb::Errors::UnknownError` with Honeycomb

elenabaurkot: @rshipp didn't refresh page and missed your original comment - if there's nothing valuable there should I put this back to how it was before?

rshipp: doesn't matter either way, as long as it works

sadleb (approved):
lgtm

PR #967, opened 2022-01-06
Discord Server Sync: Remove deleted channels/roles

rshipp:
### Description

In addition to adding/updating channels/roles, we should also remove them from our local db when they've been deleted from the Discord server.

### Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1201314155101891/f

### Screenshots

Screen Shot 2022-01-06 at 4 34 54 PM

### Types of changes

### Checklist:


app/services/sync_discord_servers.rb
```diff
@@ -33,6 +33,16 @@ def sync_server(server)
   def sync_server_channels(server)
     response = Discordrb::API::Server.channels("Bot #{Rails.application.secrets.discord_bot_token}", server.discord_server_id)
     channels = JSON.parse(response.body)
+
+    # Remove deleted channels.
+    deleted_channels = server.channels.map { |c| c.discord_channel_id } - channels.map { |c| c['id'].to_s }
+    puts server.channels.map { |c| c.discord_channel_id }.to_s
+    puts channels.map { |c| c['id'].to_s }.to_s
```
elenabaurkot, line 40: I don't know if you wanted to keep the puts here or not, but other than that looks good to me 👍

app/services/sync_discord_servers.rb
```diff
@@ -33,6 +33,16 @@ def sync_server(server)
   def sync_server_channels(server)
     response = Discordrb::API::Server.channels("Bot #{Rails.application.secrets.discord_bot_token}", server.discord_server_id)
     channels = JSON.parse(response.body)
+
+    # Remove deleted channels.
+    deleted_channels = server.channels.map { |c| c.discord_channel_id } - channels.map { |c| c['id'].to_s }
```
sadleb, line 38: Is there any possibility that the `channels.map` is temporarily empty or incorrect causing us to destroy the local ones accidentally? If that did happen, is it fine b/c the next successful sync would add them back? It doesn't actually destroy anything that we need to safeguard, right?

app/services/sync_discord_servers.rb
```diff
@@ -33,6 +33,16 @@ def sync_server(server)
...
+    puts server.channels.map { |c| c.discord_channel_id }.to_s
+    puts channels.map { |c| c['id'].to_s }.to_s
+    Rails.logger.info("Deleting channels: #{deleted_channels}")
+    Honeycomb.add_field('sync_discord_servers.deleted_channels', deleted_channels)
+    DiscordServerChannel.where(discord_channel_id: deleted_channels).destroy_all
```
sadleb, line 43: Should we wrap this all in a `if deleted_channels.present?` block?

app/services/sync_discord_servers.rb
```diff
@@ -61,6 +71,14 @@ def sync_server_channels(server)
   def sync_server_roles(server)
     response = Discordrb::API::Server.roles("Bot #{Rails.application.secrets.discord_bot_token}", server.discord_server_id)
     roles = JSON.parse(response.body)
+
+    # Remove deleted roles.
+    deleted_roles = server.roles.map { |c| c.discord_role_id } - roles.map { |c| c['id'].to_s }
+    Rails.logger.info("Deleting roles: #{deleted_roles}")
+    Honeycomb.add_field('sync_discord_servers.deleted_roles', deleted_roles)
+    DiscordServerRole.where(discord_role_id: deleted_roles).destroy_all
```
sadleb, line 79: Ditto on the q's above. 1) anything to safeguard against if there is a temp blip? 2) wrap in `if blah.present?` block?

spec/services/sync_discord_servers_spec.rb
```diff
@@ -97,6 +97,30 @@
         expect(DiscordServerChannel.find_by(name: text_channel[:name])).not_to eq(nil)
       end
     end
+
+    context 'with deleted channel' do
```
sadleb, line 101: Are there any tests that cover the case where the local db channels match the server channels? I'm just thinking about making sure this works in that case: `DiscordServerChannel.where(discord_channel_id: deleted_channels).destroy_all`

app/services/sync_discord_servers.rb
rshipp: whoops, nope

app/services/sync_discord_servers.rb
rshipp: anything is possible, but it's not likely. that's like asking if technically, could a salesforce api call to get participants return an empty list. and correct, we wouldn't really care, the discord scheduler would just be broken until the next day or we rerun the task.

spec/services/sync_discord_servers_spec.rb
rshipp: there's some directly above this. but `Model.where(something: []).destroy_all` translates to `delete from table where something in ()`, which does nothing.

app/services/sync_discord_servers.rb
rshipp: why? an extra branch to remove a no-op SQL call doesn't seem worth it

sadleb (approved):
Thanks for answering my questions. lgtm

PR #970, opened 2022-01-07
Zoom: Reduce ZoomAPI::TooManyRequestsErrors for meeting Hosts

sadleb:
### Description
The Zoom sync has special logic for meeting hosts to prevent
errors from being raised. Instead, it sets the link in Salesforce to:
'NONE: cannot generate Zoom link for a Host of the meeting'

However, there is a bug where since it doesn't save a ZoomLinkInfo
for these, it keeps trying again on every sync and catching the same
`ZoomAPI::HostCantRegisterForZoomMeetingError`. After 3 times, that error
changes to `ZoomAPI::TooManyRequestsErrors` which get raised to Sentry
and set in Honeycomb. We now avoid this by skipping the sync entirely
if we've already set the link to the above NONE message.

Note: since this is a one time thing, there could be edge cases where
the person is a Host but then things change, like a new meeting ID is
used or they are removed as a Host, and the sync will never regenerate
the link. The guidance here will be to clear out the NONE message from
the Participant record and the sync will start processing them again.

Example traces/errors:
Sentry: https://sentry.io/organizations/braven-pr/issues/2904568915/?project=5639961&query=is%3Aunresolved
Honeycomb: https://ui.honeycomb.io/braven/datasets/platform.braven.org/result/EYi2jzuqYgz/trace/zLugzMrm4fR?span=b6ae6ff5ff382d75

### Motivation and Context
This is just noise.

### Types of changes

### Checklist:


sadleb:
Foregoing a code review b/c it's a small tweak, nothing fundamental. I sanity tested a normal Zoom sync to make sure I didn't break anything. It's hard to test the new Host behavior in staging though b/c I need a real user so I'm relying on the specs for that and we'll see if it works in prod.

PR #972, opened 2022-01-10
SYNC: Run the job once a minute and gracefully exit early if it hits the max duration

sadleb:
### Description
Change the sync proof of concept to:
1. exit out and stop processing folks if we're past the `SALESFORCE_SYNC_MAX_DURATION`
1. save each participant as they're synced.
    
This means that we can still leave the max duration short-ish (to handle a job crashing or a server restart during a job) and increase the frequency of the sync so that the next job will start soon and pick up where the previous one left off. I'm somewhat arbitrarily going with 
1. sync every minute
1. stop processing a sync after 10 minutes and let the next one continue

### Motivation and Context
When reviewing the data in production from this:
* https://github.com/bebraven/platform/pull/949

I found that there are certain times when a lot of changes happen (like 400 Participants have a change probably from a roster or other bulk upload). Even with the 2 second per person sync, this exceeded the 15 minute `SALESFORCE_SYNC_MAX_DURATION` across all programs and we ended up with two jobs running in parallel processes the same folks.

Also, based on prod benchmarking, the sync can be more like 4 seconds (and up to 7s) per Participant with changes. Bumping that in the PoC too.

### Types of changes

### Checklist:


rshipp (approved):

elenabaurkot (commented):
lgtm!

PR #973, opened 2022-01-11
Hotfix: SidekiqUniqueJobs::OnConflict::LogHoneycomb

sadleb:
### Description
**UPDATE**: this only seems to be the case if the lock_ttl (aka SALESFORCE_SYNC_MAX_DURATION)
is less than the frequency of the recurring job (aka SALESFORCE_SYNC_EVERY). This means that two
jobs are running in parallel and they conflicted but b/c the first job's lock had expired the second job
was allowed to run. The solution here is probably to alert or fail if the setting for the max duration is
less than the setting for the recurring schedule. Closing this PR b/c it shouldn't happen in prod.

----
The Honeycomb field (and the log) was wrong. Both the running
job and the one being skipped were logged and instrumented in Honeycomb
as having been skipped. This is a small hack so we can tell which is which
(though I'm leaving the buggy log alone and just fixing Honeycomb b/c
we don't really looks at logs and it's not worth fixing)

### Motivation and Context
Don't waste folks time trying to figure out why a job is being instrumented as having been skipped
when it wasn't.

Follow-up to this PR where I'm trying to gather more real data about how the sync job runs during key launch times (aka Fellow invite communications are happening today):
* https://github.com/bebraven/platform/pull/972

Here is what it looks like now in Honeycomb. The first job was actually skipped, the second one
wasn't but was the job that conflicted with the skipped job.
image

### Types of changes

### Checklist:


sadleb:
Will ask for a retroactive code review b/c I want this code to be live when comms go out (and ideally before any roster uploads or bulk actions are taken that cause sync stuff to run).

PR #975, opened 2022-01-13
Nightly rake task: give zeros to unsubmitted past due LTI assignments

elenabaurkot:
### Description
This creates a rake task that will be run nightly to check for unsubmitted LTI assignments that are past due and give them zeros. As of now it does not check whether or not the assignment is muted because the Canvas API doesn't seem to return that accurately for LTI assignments. 

### Motivation and Context
To simplify the grading process for TAs.

Asana Task: https://app.asana.com/0/1201131148207877/1201598672390752/f


app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+        Rails.logger.info("Exit early: no current/future accelerator programs with a Canvas course ID set.")
+        return
+      end
+
+      courses = Course.where(canvas_course_id: canvas_course_ids)
+      courses.each do |course|
+        get_assignments_and_submissions(course)
+      end
+    end
+  end
```
sadleb, line 36: Make everything below here private?

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def get_assignments_and_submissions(course)
```
sadleb, line 38: This is called `get_assignments_and_submissions` but it actually gets them and grades them which is confusing.

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+    # Get unsubmitted submissions for all of the above assignments
+    submissions = CanvasAPI.client.get_unsubmitted_assignment_data(course.canvas_course_id, assignment_ids)
+
+    assignment_ids.each do |assignment_id|
+      grade_unsubmitted_submissions(submissions, assignment_id, course.canvas_course_id)
+    end
+  end
+
+  def grade_unsubmitted_submissions(submissions, canvas_assignment_id, canvas_course_id)
+    submissions_to_grade = Hash.new
```
sadleb, line 52: Nit: we usually just use `{}` in this codebase instead of `Hash.new`

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    # Get unsubmitted submissions for all of the above assignments
+    submissions = CanvasAPI.client.get_unsubmitted_assignment_data(course.canvas_course_id, assignment_ids)
+
+    assignment_ids.each do |assignment_id|
+      grade_unsubmitted_submissions(submissions, assignment_id, course.canvas_course_id)
+    end
+  end
+
+  def grade_unsubmitted_submissions(submissions, canvas_assignment_id, canvas_course_id)
```
sadleb, line 51: Nit: I like that you called the service `grade_unsubmitted_assignments` to be consistent with `grade_rise360_modules.rb`, but something like `zero_out_grades` might be more intuitive to a reader?

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+    submissions_to_grade = Hash.new
+    submissions.each do |submission|
+      # Allowing for 10 minute buffer past assignment cached_due_date in case time sync is a little off
```
sadleb, line 54: Nit: I've always heard of this concept as "clock skew" or "timing skew". Another dev might be more likely to understand what this means if you use the word "skew" in there.

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+      next unless submission['cached_due_date'] && submission['cached_due_date'] < Time.now.utc - 600
```
sadleb, line 55: Instead of `600` its a bit more readable to use `10.minutes`.

sadleb, line 55: For the rise360module stuff, we do `CanvasSubmission.parse(raw_submission)` and then use the `is_graded?` and `due_in_past?` helpers for what is essentially the same logic. I'm thinking we should do that here too just so both go through the same code path and we'd fix any bugs in one, in both places. If you do this, and Ryan agrees that it's a good idea to give a 10 minute buffer (which really shouldn't come into play), I would move the buffer offset into the `due_in_past?` method too.

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      if canvas_assignment_id == submission['assignment_id']
+        submissions_to_grade[submission['user_id']] = 0
```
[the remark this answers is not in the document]
sadleb: docs for reference on what types of grade values it can take: https://canvas.instructure.com/doc/api/submissions.html#method.submissions_api.update

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+    end
+
+    if submissions_to_grade.count > 0
```
sadleb, line 62: We tend to use `present?` in this codebase even though it's functionally the same.

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+  end
+
+  def grade_unsubmitted_submissions(submissions, canvas_assignment_id, canvas_course_id)
+    submissions_to_grade = Hash.new
+    submissions.each do |submission|
+      # Allowing for 10 minute buffer past assignment cached_due_date in case time sync is a little off
+      next unless submission['cached_due_date'] && submission['cached_due_date'] < Time.now.utc - 600
+
+      if canvas_assignment_id == submission['assignment_id']
```
sadleb, line 57: If you're having trouble cleaning putting the submissions into buckets per assignment ahead of time (e.g. something like the below hash) I would at least move this to the first line of the loop as `next unless submission['assignment_id'] == canvas_assignment_id`

```
assignments_to_submissions = {
  id_1 => [submissions_for_this_id],
  id_2 => [submissions_for_this_id],
}
```

lib/canvas_api.rb
```diff
@@ -171,6 +171,13 @@ def get_submission_data(course_id)
     get_all_from_pagination(response)
   end
 
+  def get_unsubmitted_assignment_data(course_id, assignment_ids)
+    query_params = "per_page=100&assignment_ids[]=#{assignment_ids.join("&assignment_ids[]=")}&student_ids[]=all&workflow_state=unsubmitted"
+    response = get("/courses/#{course_id}/students/submissions?#{query_params}")
+    
+    get_all_from_pagination(response)
```
sadleb, line 178: Something like the following would return the submissions bucketed into assignments which may be a better API to use in general?
```
get_all_from_pagination(response).to_h { |s| [ s['assignment_id'], s ] }
```

app/services/grade_unsubmitted_assignments.rb
sadleb: I just added an example of how you could bucket the submissions returned from the API to make this easier to do: https://github.com/bebraven/platform/pull/975/files#r784352647

app/services/grade_unsubmitted_assignments.rb
elenabaurkot: Discussed on Zoom and decided to leave public to allow for testing

app/services/grade_unsubmitted_assignments.rb
elenabaurkot: updated to call this `grade_unsubmitted_assignments` instead

app/services/grade_unsubmitted_assignments.rb
elenabaurkot: Updated this method name to `zero_out_grades`

app/services/grade_unsubmitted_assignments.rb
elenabaurkot: Updated this to use due in past and I moved the 10 minute buffer over to the CanvasSubmission model's `due_in_past?` method

app/services/grade_unsubmitted_assignments.rb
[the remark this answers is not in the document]
elenabaurkot: Updated this to use `'0%'`

app/services/grade_unsubmitted_assignments.rb
elenabaurkot: Went with option 2 here and bucket the assignments in the API

spec/services/grade_unsubmitted_assignments_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+      allow(CanvasAPI).to receive(:client).and_return(canvas_client)
+      allow(canvas_client).to receive(:update_grades)
+    end
+
+    context 'with submissions due in past' do
+      let(:submissions) { [submission_due_tomorrow, submission_due_yesterday] }
+
+      it 'calls update_grades and sets submission scores to zero' do
+        expect(canvas_client).to receive(:update_grades)
+        subject
```
elenabaurkot, line 176: For this test I was trying to see if there was a way to also check if the grade was changed to zero `expect(submission_due_yesterday['grade'].to eq(0)` but was having a lot of trouble figuring out if there was a way to do this since the submission gets parsed (`@canvas_submission = CanvasSubmission.parse(submission)`) and I couldn't figure out how to check the grade on the parsed submission - not sure if there is a way to do this that's worth it?

elenabaurkot:
When manual testing the one case I couldn't get to work  is if a cohort has a due date in the future, and one student from that cohort has a due date in the past it seems to override the single student due date and go with the cohort due date so it won't grade their assignment until the cohort due date. The reverse of that works (if a cohort is due in the past and a student is due in the future it will wait until the student's due date to grade and not go with the cohort due date). I am thinking it isn't a problem because it doesn't make much sense that the whole cohort would have one due date and one student from that cohort would have an earlier due date, but just wanted to double check.

rshipp:
That makes sense to me, I think the only valid case is for one student to have a later due date than the rest of their cohort, not earlier.

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,79 @@
...
+    if assignment_ids.empty?
+      Rails.logger.info("Skip grading assignments for canvas_course = #{course.canvas_course_id}; no assignments to grade")
+      return
+    end
+
+    # Get unsubmitted submissions for all of the above assignments
+    submissions_by_assignment = CanvasAPI.client.get_unsubmitted_assignment_data(course.canvas_course_id, assignment_ids)
+
+    submissions_by_assignment.each do |assignment_id, submissions|
+      zero_out_grades(submissions, assignment_id, course.canvas_course_id)
```
sadleb, line 52: nit: most of our methods take parameters from least specific to most specific. In this case, that would be `zero_out_grades(course.canvas_course_id, assignment_id, submissions)`

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  end
+
+  def zero_out_grades(submissions, canvas_assignment_id, canvas_course_id)
+    submissions_to_grade = {}
+    submissions.each do |submission|
+      @canvas_submission = CanvasSubmission.parse(submission)
```
sadleb, line 59: I don't think this needs to be an instance variable. aka `canvas_submission` instead of `@canvas_submission` should be fine here.

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -0,0 +1,79 @@
...
+      next unless @canvas_submission.due_in_past?
```
sadleb, line 60: i think we use the `@canvas_submission.needs_zero_grade?` method here which also checks if it `is_graded?`. Even though we shouldn't have any submissions with grades since we used `get_unsubmitted_assignment_data` it's possible that a TA or staff member manually gave someone a grade for someone reason even though there is no submission. Actually, I think this is expected for the group capstone project where one person submits the project for the group and the TAs go give everyone the same grade.

lib/canvas_api.rb
```diff
@@ -171,6 +171,27 @@ def get_submission_data(course_id)
...
+    all_submissions.map { |s|
+      # If there isn't already a submission for an assignment id,
+      # add a key value pair with the {assignment id: [submissions]}
+      # If there are already submissions for an assignment id, add that submission to the list
+      if submissions_by_assignment[s['assignment_id']].nil?
+        submissions_by_assignment[s['assignment_id']] = [s]
+      else
+        assignment_submissions = submissions_by_assignment[s['assignment_id']]
+        assignment_submissions.push(s)
+        submissions_by_assignment[s['assignment_id']] = assignment_submissions
```
sadleb, line 189:
```suggestion
        submissions_by_assignment[s['assignment_id']] << s
```

elenabaurkot: This looks much nicer, updating 👍

app/services/grade_unsubmitted_assignments.rb
elenabaurkot: updated the ordering from least specific to most

app/services/grade_unsubmitted_assignments.rb
elenabaurkot: That makes sense, just made this update 👍

sadleb (approved):
Lgtm! After we deploy we’ll need to schedule the rake task. Fingers crossed on this one bc it’s a pretty big change. After we have everything turned on, we should sanity check that things are working properly in a real course for the first couple days.

PR #980, opened 2022-01-27
Zero out grades: skip attendance events and manually graded assignments

elenabaurkot:
### Description
When an assignment due date passes, this rake task will run and set un-submitted, past-due assignments to zero. This is an update to this previous PR: [Nightly rake task: give zeros to unsubmitted past due LTI assignments #975](https://github.com/bebraven/platform/pull/975) to filter out assignments that are attendance events and assignments that have "post_manually" set to true, this way if a TA is still working on grading them it won't set those assignments to zero. In order for un-submitted assignment that are set to "post manually" to be given zeros when a TA is done grading, they must edit the "grade posting policy" in assignment settings and set it back to post automatically (see screenshots below). 

### Motivation and Context
Asana Task: https://app.asana.com/0/1201131148207877/1201598672390752/f

Screen Shot 2022-01-27 at 10 08 52 AM

Screen Shot 2022-01-27 at 10 09 08 AM


sadleb:
@elenabaurkot in the PR description can you link to the actual PR and not the commit? E.g.
* https://github.com/bebraven/platform/pull/975

Commit SHAs can get overwritten (that's what a rebase does) and then the link would break. Unlikely to happen once something is merged to the prod branch, but better to be consistent with PR links anyway.

Also, can we change the title of the PR to something like: "Zero out grades: skip attendance events and manually graded assignments"? The fact that we have to change the `assignment_filter` is a detail of the implementation and a dev skimming the list of PRs wouldn't know what this PR was even related to unless they clicked into it (or had intimate knowledge of this piece of code).

sadleb:
FYI: I made a subtask to make sure we take the next steps with the team on the guidance to set the "grade posting policy" back to automatic when grades are fully done and posted: https://app.asana.com/0/1201131148207877/1201734458795759/f

app/services/grade_unsubmitted_assignments.rb
```diff
@@ -73,7 +73,9 @@ def zero_out_grades(canvas_course_id, canvas_assignment_id, submissions)
 
   def assignment_filter(assignment)
     return false unless assignment['submission_types'].include?('external_tool')
-    return false unless assignment['published'] == true
+    return false unless assignment['published']
+    return false if CourseAttendanceEvent.where(canvas_assignment_id: assignment['id']).exists?
+    return false if assignment['post_manually']
     assignment['id']
   end
```
sadleb, line 80: I think this new logic is sufficiently non-intuitive that it merits a comment. Can we add a some explanation of the logic behind skipping attendance events (e.g. if it's unsubmitted it means attendance wasn't taken and that needs to happen. We should get an LC or staff member to submit it.) and the logic behind skipping those with `post_manually` set (e.g. Projects are set to `post_manually` so that TAs can spend some time manually grading them. When done, they are expected to post the grades and set the grade posting policy pack to automatic.).

sadleb (approved):
Approving in advance of adding a bit of comments to the `assignment_filter` a bit.

app/services/grade_unsubmitted_assignments.rb
elenabaurkot: Updated to add an explanatory comment in for the assignment filter

PR #981, opened 2022-01-27
Fix ConnectionTimeoutError: update Sidekiq thread count to match Puma

elenabaurkot:
### Description
We were getting the error "ActiveRecord::ConnectionTimeoutError: could not obtain a connection from the pool within 5.000 seconds" because our Sidekiq concurrency didn't match our DB thread count. This update makes it so they match. 

### Motivation and Context
Asana Task: https://app.asana.com/0/1201131148207877/1201695425068140/f
Article outlining problem: https://dev.to/appaloosastore/active-record-sidekiq-pools-and-threads-18d5


config/sidekiq.yml
```diff
@@ -3,6 +3,10 @@
   - default
   - low_priority
 
+# If you change the number of threads here make sure to also change the PUMA_MAX_THREADS
+# number to match it in ENV and application.rb
```
elenabaurkot, line 7: Unsure if this comment should also tell them to change the number 5 set for the pool in database.yml and to change the DB_POOL environment variable (I don't currently have that in my ENV file, but if someone does it might cause the same issue?) `pool: <%= Integer(ENV['DB_POOL'] || Platform::Application::PUMA_MAX_THREADS || 5) %>`

sadleb: `ENV['DB_POOL']` isn't used. I added that for a similar reason we added `SIDEKIQ_MAX_THREADS` below - just in case we needed a quick fix to override the pool count. E.g. imagine something in a single Puma thread spinning off it's own threads and hammering the database causing issues. We could bump the DB_POOL count to add more connections until we got to the bottom of why a single Puma thread was using more than one db connection.

All that to say --> I think we should actually change the line below to the following to match and not mention `ENV['DB_POOL']` in the comment.
```
:concurrency: <%= Integer(ENV['DB_POOL'] || ENV['SIDEKIQ_MAX_THREADS'] || 5 )%>
```

sadleb:
@elenabaurkot can we change the PR title to something like: "Fix ConnectionTimeoutError: update Sidekiq thread count to match Puma"? I'd like to have our list of PRs readable by a dev who doesn't know every detail of the codebase. In other words, "Fix ConnectionTimeoutError" would be meaningful to any dev when scanning the list of PRs but "sidekiq thread count to match puma" might not mean anything to them unless they had spent time with both.

config/sidekiq.yml
elenabaurkot: Ah okay, got it, just made this update 👍

elenabaurkot:
> @elenabaurkot can we change the PR title to something like: "Fix ConnectionTimeoutError: update Sidekiq thread count to match Puma"? I'd like to have our list of PRs readable by a dev who doesn't know every detail of the codebase. In other words, "Fix ConnectionTimeoutError" would be meaningful to any dev when scanning the list of PRs but "sidekiq thread count to match puma" might not mean anything to them unless they had spent time with both.

That definitely makes sense, just updated the title to this 👍

sadleb (approved):
lgtm

PR #983, opened 2022-01-31
Refactor: Enrollment checks in policies

rshipp:
### Description

Refactor policies that do some kind of "enrollment" check, to increase maintainability and reduce the number of SQL calls per HTTP request.

### Motivation and Context

Tasks:
* https://app.asana.com/0/1174274412967132/1199344732354185/f
* https://app.asana.com/0/1174274412967132/1201078306275560/f

### Types of changes

### Checklist:


app/models/user.rb
```diff
@@ -102,17 +104,31 @@ def admin?
...
     course.sections.each do |section|
       return true if has_role? RoleConstants::STUDENT_ENROLLMENT, section
     end
     false
   end
 
+  def is_enrolled_as_ta?(course)
+    return false if course.nil?
+    course.sections.each do |section|
+      return true if has_role? RoleConstants::TA_ENROLLMENT, section
```
rshipp, line 127: check: is this how we want to do this check? or should it be specifically the ta section?

rshipp: move this back to the attendance policy

app/models/user.rb
```diff
@@ -139,6 +155,45 @@ def lc_for?(target_user, course)
...
   end
 
+  # True if the user is enrolled in a course that the supplied `record`
+  # references. E.g.:
+  #
+  #     user.can_access?(rise360_module_version)
+  #     user.can_access?(project_submission)
+  #
+  # `record` can be any model instance that responds to `course` or `courses`,
+  # except a User or a Section, where "can_access" would be meaningless.
```
rshipp, line 165: add some comments about how this isn't just for submissions / you must add course associations

app/models/user.rb
```diff
@@ -102,17 +104,31 @@ def admin?
...
+  # If it's your own submission, or you're a TA/LC for the submission owner in that course.
   def can_view_submission_from?(target_user, course)
-    return ta_for?(target_user, course) || lc_for?(target_user, course)
+    return ta_for?(target_user, course) || lc_for?(target_user, course) || target_user == self
   end
 
+  # Note: we can refactor these is_enrolled_as methods to use associations if
+  # we find out the joins would be a significant efficiency/scalability
+  # improvement. It probably won't matter unless the number of sections per
+  # course is really big.
```
rshipp, line 115: just refactor this now to student_roles -> student_sections -> student_courses

app/policies/attendance_event_submission_policy.rb
```diff
@@ -15,6 +15,7 @@ def launch?
 
     # Is the user a TA in any section in the course?
     course = record.instance_of?(Course) ? record : record.course
+    return false if course.nil?
     course.sections.each do |section|
       return true if user.has_role? RoleConstants::TA_ENROLLMENT, section
     end
```
sadleb, line 21: Similar to `has_many :student_sections` can we add a `has_many :ta_sections` association so we can use that here and avoid the possible 70 DB calls?

app/models/user.rb
```diff
@@ -35,6 +35,13 @@ class User < ApplicationRecord
 
   # All sections where this user has any role.
   has_many :sections, -> { distinct }, through: :roles, source: :resource, source_type: 'Section'
+  # All courses from above sections.
+  has_many :courses, -> { distinct }, through: :sections
+
+  # Student-specific associations, for efficiency.
+  has_many :student_sections, -> { where("roles.name = ?", RoleConstants::STUDENT_ENROLLMENT) },
+    through: :roles, source: :resource, source_type: 'Section'
```
sadleb, line 43: Fanciest association I've seen to date.

I feel like I've always used scopes in the past when I had to do something like this not realizing that an association could also be scoped down. Note to self here to keep this possibility in mind.

app/models/user.rb
```diff
@@ -139,6 +145,58 @@ def lc_for?(target_user, course)
...
+        # behavior (and also prevents people from creating a submission in a
+        # course they're not enrolled in).
+        return true if courses.include?(record.course)
+      elsif record.respond_to?(:courses)
+        return true if (courses & record.courses).any?
+      else
+        # Doesn't implement `.course` or `.courses`, so we don't know what to
+        # do. Raise an exception to encourage the developer to add this missing
+        # association.
+        raise NotImplementedError.new "#{record.class} model does not implement course or courses; consider adding the appropriate association?"
```
sadleb, line 193: nice

app/policies/course_attendance_event_policy.rb
```diff
@@ -9,21 +9,8 @@ def unpublish?
 
   # TODO: Move this to AttendanceEventSubmission policy in the next PR
   def launch?
-    return true if is_enrolled?
+    return true if user&.courses&.include?(record&.course)
```
sadleb, line 12: Imo it's better to still have this in an `is_enrolled?` method like follows b/c this one-liner takes a minute to understand:
```
def is_enrolled?
  user&.courses&.include?(record&.course)
end

File: app/policies/enrolled_policy.rb [hunk 8]
```diff
@@ -0,0 +1,23 @@
+# A special parent policy to simplify writing child policies that check for
+# enrollment in a course, before allowing user access to a record referenced
+# in that course.
+class EnrolledPolicy < ApplicationPolicy
+  def show?
+    # Admins can always see the record.
+    return true if user&.admin?
```
sadleb, line 7: Can `user` be nil? I assume so b/c of the `user&.` usage. What does this `show?` method do for a `nil` user? Looks like it raises `'You are not enrolled in this course.'`. Is that what we want? It's different than the `record.nil?` behavior.

app/policies/enrolled_policy.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    # Can't show something that doesn't exist.
+    return false if record.nil?
```
sadleb, line 10: Can user be nil? I assume so b/c of the `user&.` usage. What does this `show?` method do for a nil user? Looks like it raises "'You are not enrolled in this course.'". Is that what we want? It's different than this `record.nil?` behavior.

app/models/custom_content.rb
```diff
@@ -6,6 +6,9 @@ class CustomContent < ApplicationRecord
   has_many :custom_content_versions
   alias_attribute :versions, :custom_content_versions
 
+  has_many :course_custom_content_versions, through: :custom_content_versions
+  has_many :courses, -> { distinct }, through: :course_custom_content_versions
```
sadleb, line 10: Looks like there are two `def courses` below which is ... 😳 I'm surprised this association take precedence or maybe it doesn't?

app/models/user.rb
```diff
@@ -102,15 +109,14 @@ def admin?
     has_role? :admin
   end
 
+  # If it's your own submission, or you're a TA/LC for the submission owner in that course.
   def can_view_submission_from?(target_user, course)
-    return ta_for?(target_user, course) || lc_for?(target_user, course)
+    return ta_for?(target_user, course) || lc_for?(target_user, course) || target_user == self
```
sadleb, line 114: Nit: can we put the `target_user == self` check first? Its the fastest check and *maybe* the most common use-case?

app/models/user.rb
```diff
@@ -139,6 +145,58 @@ def lc_for?(target_user, course)
...
+  # If `record` is a submission (or submission-like object, that has a `.user`
+  # reference), this additionally checks whether the user has permission to
+  # view submissions from `record.user`.
+  #
+  # When writing models that you intend to pass into this method, be sure to
+  # always add a :course or :courses association, as appropriate! We'll raise
+  # an exception if you don't, to help you remember.
+  def can_access?(record)
+    Honeycomb.start_span(name: 'user.can_access?') do
+      Honeycomb.add_field('can_access.record.class', record.class.to_s)
```
sadleb, line 166: Note to other reviewers. `nil.class.to_s` is the string "NilClass"

app/models/user.rb
```diff
@@ -139,6 +145,58 @@ def lc_for?(target_user, course)
...
+
+      return false if record.nil? || record.is_a?(User) || record.is_a?(Section)
+
+      # Check enrollment.
+      if record.respond_to?(:course)
+        Honeycomb.add_field('course.id', record.course.id.to_s)
+
+        # For submissions, we additionally check whether the user can view the submission.
+        if record.respond_to?(:user)
+          Honeycomb.add_field('can_access.record.user.id', record.user.id.to_s)
```
sadleb, line 176: Above you name the field `course.id` but here its not `user.id`? I'm leaning towards using `can_access.record.course.id` above but could be convinced the opposite too.

app/models/user.rb
```diff
@@ -139,6 +145,58 @@ def lc_for?(target_user, course)
...
+        end
+
+        # Note, if by some weird edge case a user is trying to view their own
+        # submission from a course they are no longer enrolled in, this will
+        # return false and they will not be able to access it. This is desired
+        # behavior (and also prevents people from creating a submission in a
+        # course they're not enrolled in).
+        return true if courses.include?(record.course)
+      elsif record.respond_to?(:courses)
+        return true if (courses & record.courses).any?
```
sadleb, line 188: Note to other reviewers: `courses & record.courses` is an "intersection" operator.

That said, @rshipp does `courses.intersect?(record.courses)` work? Looks like they added that in Ruby 3.1 when you don't care about the actual result: https://blog.saeloun.com/2021/05/18/ruby-adds-array-intersect.html#after

app/policies/accelerator_survey_submission_policy.rb
```diff
@@ -18,41 +12,10 @@ def launch?
...
-  # https://app.asana.com/0/1174274412967132/1199344732354185
-  # Returns true iff user has any type of enrollment in any section of course
-  def is_enrolled?
-    record.course.sections.each do |section|
-        RoleConstants::SECTION_ROLES.each do |role|
-          return true if user.has_role? role, section
-        end
-    end
-    false
+    show?
```
sadleb, line 19: Should we preserve this Asana task link here? https://app.asana.com/0/1201131148207877/1199512664295642

app/policies/discord_signup_policy.rb
```diff
@@ -27,15 +27,8 @@ def unpublish?
...
   # Returns true iff user has any type of enrollment in any section of course
   def is_enrolled?
-    record.sections.each do |section|
-      RoleConstants::SECTION_ROLES.each do |role|
-        return true if user.has_role? role, section
-      end
-    end
-    false
+    user&.courses.include?(record)
   end
```
sadleb, line 33: I was trying to understand why this didn't inherit the `EnrolledPolicy` and thought it was because there was no `show` action, but I'm realizing its b/c it's different enrollment logic. Similar to the attendance one where it checks if you're a TA in any section, this just checks if you're enrolled in any section.

Looking through the other policies that don't subclass `EnrolledPolicy` I'm realizing that it's going to be pretty confusing to try and mimic another feature's code pattern and know whether to use `EnrolledPolicy` or not.

I'm thinking it may be better if we have it be a module that we can include in policies that have to do enrollment checks and it just provides helper methods that each policy can call instead of implementing the `show?` method for you. E.g.

```
module EnrolledPolicy
  def can_access?
    # the logic currently in `show?`
  end

  # Returns true iff user has any type of enrollment in any section of course
  def enrolled_in_course?
    user&.courses.include?(record)
  end

  def enrolled_in_course_as(role)?
    # have the attendance stuff use this
  end

  def enrolled_as_student?
    enrolled_in_course_as(RoleConstants::STUDENT_ENROLLMENT)
  end

  # etc
end
```

Then this class would be:
```
class DiscordSignupPolicy < ApplicationPolicy
  include EnrolledPolicy

  def launch?
    # Requires logged-in user enrolled in the current course.
    enrolled_in_course?
  end
  ...
end
```

And something that currently inherits from `EnrolledPolicy` would turn into:
```
class CapstoneEvaluationSubmissionPolicy < ApplicationPolicy
  include EnrolledPolicy

 def show?
    # Admins can always see the "thank you for submitting" #show page.
    # This might be weird, so revisit if we decide we don't want this.
    can_access?
  end
end
```

app/models/custom_content.rb
rshipp: weird, idk what happened there.

app/models/user.rb
rshipp: I didn't use `user.id` because that'd be overwritten by your global honeycomb user propagation stuff. Using non-namespaced wherever possible lets us filter things more easily. But I can put record in the local namespace if you think that's better here.

app/models/user.rb
rshipp: i think that would work, but it looks like we're not on whatever version of ruby has this
```
irb(main):001:0> [1, 2, 3].intersect?([3, 2, 1])
(irb):1:in `<main>': undefined method `intersect?' for [1, 2, 3]:Array (NoMethodError)
Did you mean?  intersection
```

app/policies/accelerator_survey_submission_policy.rb
rshipp: i think we've long-since solved this issue? but lmk if that's not true

app/models/user.rb
sadleb: I think `can_access.record.course.id` on L190 is better.

app/models/user.rb
sadleb: Dang, yeah, we're on Ruby 3.0 not 3.1

app/policies/accelerator_survey_submission_policy.rb
sadleb: i think you're right. i just marked it complete.

sadleb (approved):
lgtm structurally but I don't think I would be able to catch any actual logic errors through a review so we'll have to rely on testing.

elenabaurkot (commented):
lgtm 👍

PR #988, opened 2022-02-10
Rise360 Modules: show message to update browser if content is blank

sadleb:
### Description
After 10 seconds show a message asking folks to use a supported, up-to-date, browser if the Module is blank.

Asana task: https://app.asana.com/0/1201131148207877/1201387757929157

### Motivation and Context
Older browsers (we've especially noticed this with Safari 13.1) throw console errors like the following:
`TypeError: Argument 1 ('listener') to MediaQueryList.addListener must be a function" in main.bundle.js:348:82630`

This leads to a completely blank page b/c Rise360 loads all the content dynamically, but fails

We've had an influx of folks opening tickets b/c of this. Hopefully showing a message after a bit will reduce that and allow them to solve it on their own.

Note that during my testing of this PR I discovered that older versions of the published Modules worked on older Safari. Rise360 must have broken support at some point in the last 6-ish months.

Desktop:
image

Mobile:

Honeycomb field that comes through:
`js.app.rise360_module_versions.handleBlankRise360Content.blank_content_found`
image

### Types of changes

### Checklist:


rshipp (approved):
i wish we had visual regression testing for this, but since we don't, this seems reasonable

sadleb:
> i wish we had visual regression testing for this, but since we don't, this seems reasonable

Note that I also tested adding this div to a page that actually does load content and it’s not visible, so this should be safe even if it takes longer than 10 seconds to load. It’ll show and then go away.

PR #991, opened 2022-02-10
Capstone Evaluations grade breakdown

elenabaurkot:
### Description
This update makes it so that a 'GROUP PROJECT: Capstone Challenge: Teamwork' is published when the 'TODO: Complete Capstone Evaluations' assignment is published. In the new teamwork assignment, admins will have the ability to see who has completed their capstone evaluations, who still needs to and to calculate and publish the teamwork grades for fellows based on their peers responses to the Capstone Evaluation assignment. After publishing grades, if fellows submit new Capstone Evaluations, admins can publish again to update the grades of the users' the new submission effects. 

### Motivation and Context
To make it easier to calculate fellow teamwork grades without having to do it manually. 

Asana Task: https://app.asana.com/0/1174274412967132/1199210519637836/f

Fellow view for the teamwork assignment before grades are published:
Screen Shot 2022-02-10 at 3 08 52 PM

Fellow view for the teamwork assignment after grades are published:
Screen Shot 2022-02-10 at 3 56 03 PM

Fellow view for the teamwork grade after they are published:
Screen Shot 2022-02-10 at 3 55 44 PM

TA/LC 'not authorized' view of the teamwork assignment:
Screen Shot 2022-02-10 at 4 03 31 PM

Admin view of the teamwork assignment when there are new Capstone Evaluation submissions to be graded:
Screen Shot 2022-02-16 at 2 50 30 PM

Admin view of the teamwork assignment when there are no new Capstone Evaluation submissions to be graded:
Screen Shot 2022-02-16 at 2 50 50 PM


app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
+<%# Student View %>
+<% if @student_user %>
+  <h1>Peer Review Grade</h1>
```
sadleb, line 3: "Peer Review" is outdated language. That's what we originally called them in that Draft PR, but now they are referred to as "Capstone Evaluations". Can we update everywhere we mention "Peer Review" to "Capstone Evaluation" ?

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
...
+
+  <% if @submissions_have_been_graded && !@total_score.nan? %>
+    <h2>Total Score</h2>
+
+    <p>Your Peer Review total score is: <%= @total_score %> / <%= @question_scores.count * 10 %></p>
```
sadleb, line 8: `@total_score` is a decimal, but the total is an integer. Can we change all the scores in this to be decimals (with 1 decimal place? E.g. show things like `7.0 / 40.0` and `7.0 / 10.0` instead?

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
...
+
+    <h2>Score Breakdown</h2>
+
+    <ul>
+      <% @question_scores.each do |question, score| %>
+        <li><%= question %>: <%= score %> / 10</li>
```
sadleb, line 14: Can we create a constant in the model for the `10` points that each question is worth and use that here? E.g. `CapstoneEvaluationQuestion::POINTS_POSSIBLE`? I think the existing `POINTS_POSSIBLE` constant should actually be `TOTAL_POINTS_POSSIBLE`.

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
+<%# Student View %>
+<% if @student_user %>
```
sadleb, line 2: It's a little inconsistent to use `@student_user` here but call `current_user.has_role? :admin` below. Can we get rid of `@student_user` altogether and just call `current_user.is_enrolled_as_student?(@course)` here?

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
+<%# Student View %>
```
sadleb, line 1: So that the browser tab shows the right thing:
```suggestion
<% content_for :title, 'Capstone Challenge: Teamwork' %>
<%# Student View %>
```

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
...
+      <% @question_scores.each do |question, score| %>
+        <li><%= question %>: <%= score %> / 10</li>
+      <% end %>
+    </ul>
+  <% else %>
+    <p>Your Peer Review grade has not yet been calculated. Once all your peers give a review, the resulting score is aggregated and a final "grade" is given here. You will be able to see the breakdown of the score you got for each question. E.g. "Actively contributed to team success".</p>
+  <% end %>
+<%# Admin view %>
+<% elsif current_user.has_role? :admin %>
+  <div class="container">
```
sadleb, line 22: I think you can remove the `container` div. There should already be one there provided by the `lti_canvas` layout.

sadleb:
In the PR description, I think one of the screenshots labelled "Fellow view for the teamwork grade after they are published:" is out of date.

app/models/course.rb
```diff
@@ -61,6 +61,10 @@ def surveys
     survey_versions.map { |v| v.survey }
   end
 
+  def students
+    sections.map { |s| s.students }.flatten.uniq
+  end
```
sadleb, line 66: We did this approach early on when we were all learning Ruby/Rails b/c we didn't know how to do associations/scopes all that well. Ideally, we'd create a `has_many :students` association for this doing something similar to how we get student sections from a user below:
```
has_many :student_sections, -> { where("roles.name = ?", RoleConstants::STUDENT_ENROLLMENT) },
    through: :roles, source: :resource, source_type: 'Section'
```

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+  def launch
+    authorize @course, policy_class: CapstoneEvaluationSubmissionPolicy
+
+    @capstone_evaluation_submissions = all_capstone_evaluation_submissions
+    @question_scores = grade_capstone_eval_questions(@capstone_evaluation_submission_answers)
+
+    @total_score = @question_scores.map { |k, v| v }.sum
+  end
+
+  def score
```
sadleb, line 28: Mentioned this on our call, I think all this scoring logic should go in a `grade_capstone_evaluations` service.

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+    authorize :CapstoneEvaluationSubmission
+    new_capstone_evaluation_submissions = CapstoneEvaluationSubmission.where(course: @course, new: true)
```
sadleb, line 30: It might be cleaner to add a `graded` and `ungraded` scopes on `CapstoneEvaluationSubmission` like:
```
scope :graded, -> { where(new: false) }
scope :ungraded, -> { where(new: true) }
```

Then we can add a `capstone_evaluation_submissions` association on the `Course` model like:
```
has_many :capstone_evaluation_submissions
```
Finally, these would allow you to do things like the following which feels cleaner
```
@course.capstone_evaluation_submissions.graded
```

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    # Total by summation.
+    submission_answers.each do |answer|
+      question_scores[answer.question.text] += answer.input_value.to_f
+    end
+
+    # Divide to get the mean.
+    question_scores.each do |k, v|
+      # divide the number of total answers a user received for all questions, by the number of questions
+      submission_count = submission_answers.count / CapstoneEvaluationQuestion.all.count
```
sadleb, line 77: Instead of doing `CapstoneEvaluationQuestion.all.count` inside the loop, we should do it outside so it's only called once. E.g.
```
all_questions_count = CapstoneEvaluationQuestion.all.count
question_scores.each do |k, v|
  submission_count = submission_answers.count / all_questions_count
  ...
end
```

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+      question_scores[k] = question_scores[k] / submission_count
+    end
+    question_scores
+  end
+
+private
+  def set_course
+    @course = Course.find_by!(canvas_course_id: @lti_launch.request_message.canvas_course_id)
```
sadleb, line 85: Nit: use: `Course.find_by!(canvas_course_id: @lti_launch.course_id)`

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  # We only want to display scores for submissions that have been included in a fellow's grade
+  # once a submission is graded and included on the teamwork grade, new is set to false
+  def graded_capstone_evaluation_submissions
+    CapstoneEvaluationSubmission.where(
+      course: @course,
+      new: false
+    )
```
sadleb, line 94: If we do the above scopes, this would become:
```
@course.apstone_evaluation_submissions.graded
```

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+  def set_capstone_evaluation_submission_answers
+    @capstone_evaluation_submission_answers = policy_scope(
+      CapstoneEvaluationSubmissionAnswer.where(submission: graded_capstone_evaluation_submissions)
+    )
+  end
+
+  def set_submissions_have_been_graded
+    @submissions_have_been_graded = false
+
+    all_capstone_evaluation_submissions = CapstoneEvaluationSubmission.where(course: @course)
```
sadleb, line 106: `@course.capstone_evaluation_submissions` using the above recommended scopes.

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
...
+    <div class="row">
+      <div class="col-md-12">
+        <h1>Compute Capstone Evaluation Results</h1>
+
+        <% if !@submissions_have_been_graded %>
```
sadleb, line 27: What if we didn't show a different message depending on whether they had ever been published? We always just show the lists and tell them what to do to publish any new ones? Does knowing that it's the first time add any value?

I do see value in showing a different message if there are no new submissions to grade though b/c everything is up-to-date.

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
...
+            <ul>
+              <% @capstone_evaluation_submissions.each do |submission| %>
+                <li><%= submission.user.full_name %></li>
+              <% end %>
+            </ul>
+            <br>
+          <% end%>
+        <% else %>
+        <%# Show after evaluations have been graded at least once  %>
+          <% students_with_new_submissions = @capstone_evaluation_submissions.filter_map { |s| s.user if s.new }%>
```
sadleb, line 55: Ditto on not doing this `filter_map` in the view. The view should just access something like `students_with_new_submissions` (either through a helper_method or instance variable)

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
...
+        <% end %>
+
+        <% remaining_students = @course.students.uniq - @capstone_evaluation_submissions.map { |s| s.user } %>
+        <% unless remaining_students.empty? %>
+          <p>Students who have not yet submitted Capstone Evaluations:</p>
+          <ul>
+            <% remaining_students.each do |student| %>
+              <li><%= student.full_name %></li>
+            <% end %>
+          </ul>
```
sadleb, line 85: Can we move this above the "Students who have submitted Capstone Evaluations that are published" section. Folks will care more about this list than the one that is all set.

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
...
+          <ul>
+            <% students_with_published_submissions = @capstone_evaluation_submissions.filter_map { |s| s.user if !s.new }%>
+            <% students_with_published_submissions.each do |student| %>
+              <li><%= student.full_name %></li>
+            <% end %>
+          </ul>
+          <br>
+        <% end %>
+
+        <% remaining_students = @course.students.uniq - @capstone_evaluation_submissions.map { |s| s.user } %>
```
sadleb, line 78: Ditto on not doing this logic in the view.

app/views/capstone_evaluation_results/launch.html.erb
elenabaurkot: Discussed in meeting -> updating to "Capstone Evaluation Teamwork Grade" to differentiate from the actual Capstone Evaluation submission assignment

app/views/capstone_evaluation_results/launch.html.erb
elenabaurkot: That makes sense, made this update

app/views/capstone_evaluation_results/launch.html.erb
elenabaurkot: updated to this 👍

app/controllers/capstone_evaluation_results_controller.rb
elenabaurkot: Updated the scoring logic to be in a the `grade_capstone_evaluations` service

elenabaurkot:
> In the PR description, I think one of the screenshots labelled "Fellow view for the teamwork grade after they are published:" is out of date.

This is actually something I meant to ask about and forgot, right now this is the actual view because the fellow still doesn't actually submit anything for this assignment and we also don't submit anything for them because we discussed there is sometimes an issue when we submit for them if it's not from their launched assignment. Currently, the code is just using the Canvas API to grade the assignment, so when they view the grade this is what they see, but if they launch the actual teamwork assignment that is when they see their grade breakdown. I don't know if there's a way to change the submission view without submitting something or to put in something to let a fellow know they should go to the assignment to see the grade breakdown because it doesn't seem super intuitive @sadleb

db/migrate/20220208211907_add_new_to_capstone_evaluation_submissions.rb
```diff
@@ -0,0 +1,5 @@
+class AddNewToCapstoneEvaluationSubmissions < ActiveRecord::Migration[6.1]
+  def change
+    add_column :capstone_evaluation_submissions, :new, :boolean, default: true
```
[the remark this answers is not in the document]
elenabaurkot: Looked into indexes a little and just wanted to clarify: adding the index just makes the search faster/optimizes performance because it only needs to search for that specific column in a table- I don't actually have to change anything in the code when I add an index, just add the index to the table?  @sadleb

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,90 @@
...
+              <% end %>
+            </ul>
+
+            <p>To publish these, click the "Compute Capstone Evaluation Results" button</p>
+            <br>
+          <% end%>
+
+          <p>Students who have submitted Capstone Evaluations that are published:</p>
+          <ul>
+            <% students_with_published_submissions = @capstone_evaluation_submissions.filter_map { |s| s.user if !s.new }%>
```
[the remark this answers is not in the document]
elenabaurkot: Made this update to add the grading logic in the launch view to the grade_capstone_evaluations service instead and exposed the service in an instance variable 👍

app/controllers/capstone_evaluation_results_controller.rb
elenabaurkot: Ah okay that makes sense, updated to this

app/views/capstone_evaluation_results/launch.html.erb
elenabaurkot: Yes, moving up 👍

app/views/capstone_evaluation_results/launch.html.erb
elenabaurkot: Updated this to show a message on whether or not there are new submissions ready to be published - took away if else for whether or not they have been published

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,82 @@
...
+      <% if @grade_capstone_eval_service.students_with_new_submissions.empty? %>
+        <p>There are no new submissions ready to be published.</p>
+      <% else %>
+        <p>Number of new submissions ready to be published: <%= @grade_capstone_eval_service.students_with_new_submissions.count %></p>
+        <p>To publish grades, click the 'Compute Capstone Evaluation Teamwork Results' button.</p>
+      <% end %>
+
+      <%= form_with(url: score_capstone_evaluation_results_path, method: "post",local: true) do |form| %>
+        <%= form.hidden_field :lti_launch_id, value: @lti_launch.id %>
+        <%= form.submit "Compute Capstone Evaluation Teamwork Results", class: 'btn btn-primary', data: { confirm: 'Are you sure? This will send Capstone Evaluation Teamwork grades to Canvas for all students.' } %>
```
elenabaurkot, line 37: Didn't think of this until now, but should I only show this button when there are new Capstone Evaluation submissions to grade? Because if there aren't any, then I guess there's no need for the button at that point (see final view in PR 'Admin view of the teamwork assignment when there are no new Capstone Evaluation submissions to be graded')

rshipp: will it break anything if they click it when there aren't any new ones? if not, i think it's fine to still show the button. it's slightly less code to maintain that way, and we have an easy way to force-regrade if something's broken.

elenabaurkot: No - it just redirects to the same page with an alert 'there are no new submissions to grade' so I will just leave it if that's the case 👍

sadleb:
Noting that we need a deploy / release plan to replace the current Teamwork assignments with the LTI one. We need to do this for both the Accelerator Course template and all launched spring 2022 courses. We may have to write a script or rake task that we run after the code is live

app/services/grade_capstone_evaluations.rb
```diff
@@ -0,0 +1,50 @@
...
+      .uniq
+
+    grades = {}
+    users_to_grade.each do |user|
+      submission_answers_for_user = CapstoneEvaluationSubmissionAnswer.where(
+        submission: @all_capstone_eval_submissions,
+        for_user: user
+      )
+
+      question_scores = CapstoneEvaluationResultsController.new.grade_capstone_eval_questions(submission_answers_for_user)
```
rshipp, line 25: can you move this controller method inside this (or another) service? instantiating an instance of a controller is really non-standard, i think we should avoid that.

app/policies/capstone_evaluation_submission_answer_policy.rb
```diff
@@ -0,0 +1,26 @@
+class CapstoneEvaluationSubmissionAnswerPolicy < ApplicationPolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    raise Pundit::NotAuthorizedError, 'must be logged in' unless user
+    raise Pundit::NotAuthorizedError, 'no submission specified' unless record
```
rshipp, line 6: looks like you only ever pass in a Course or a Symbol as the record from the controller, so this `no submission specified` message will be confusing. since the symbol will never be nil, maybe change it to `no course specified`?

app/policies/capstone_evaluation_submission_answer_policy.rb
```diff
@@ -0,0 +1,26 @@
+class CapstoneEvaluationSubmissionAnswerPolicy < ApplicationPolicy
```
rshipp, line 1: is this policy ever used? i don't see it anywhere

app/services/grade_capstone_evaluations.rb
```diff
@@ -0,0 +1,50 @@
...
+      total_score = question_scores.map { |k, v| v }.sum
+
+      grades[user.canvas_user_id] = total_score
+    end
+
+    CanvasAPI.client.update_grades(@course.canvas_course_id, @lti_launch.assignment_id, grades)
+
+    @new_capstone_eval_submissions.each do |submission|
+      submission.update!(new: false)
+    end
```
rshipp, line 35: nit: you can also do 
```suggestion
    @new_capstone_eval_submissions.update_all(new: false)
    end
```
It will skip callbacks/validation https://apidock.com/rails/ActiveRecord/Base/update_all/class, but I think that's OK in this case since those submissions should already be guaranteed valid, since they're in the DB?

app/views/capstone_evaluation_results/launch.html.erb
```diff
@@ -0,0 +1,82 @@
...
+        <p>Students who have not yet submitted Capstone Evaluations:</p>
+        <ul>
+          <% @grade_capstone_eval_service.remaining_students.each do |student| %>
+            <li><%= student.full_name %></li>
+          <% end %>
+        </ul>
+        <br>
+      <% end%>
+
+      <% if !@submissions_have_been_graded %>
```
rshipp, line 51: nit: prefer `unless` over `if !` in most cases

app/services/grade_capstone_evaluations.rb
elenabaurkot: Updated this method to be inside the grade_capstone_evaluations service 👍

app/policies/capstone_evaluation_submission_answer_policy.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    raise Pundit::NotAuthorizedError, 'must be logged in' unless user
+    raise Pundit::NotAuthorizedError, 'no submission specified' unless record
+    @user = user
+    @record = record
+  end
+
+  class Scope
```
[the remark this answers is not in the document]
elenabaurkot: It's used in the `set_capstone_evaluation_submission_answers` method inside the `capstone_evaluation_results_controller`

app/policies/capstone_evaluation_submission_answer_policy.rb
elenabaurkot: I think you're right - just using the scope for this policy

app/services/grade_capstone_evaluations.rb
elenabaurkot: ah okay nice, just updated to this

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  before_action :set_capstone_evaluation_submission_answers, only: [:launch]
+  before_action :set_submissions_have_been_graded, only: [:launch]
+  before_action :set_grade_capstone_eval_service_instance, only: [:launch, :score]
+
+  def launch
+    authorize @course, policy_class: CapstoneEvaluationSubmissionPolicy
+
+    @capstone_evaluation_submissions = all_capstone_evaluation_submissions
+    @question_scores = @grade_capstone_eval_service.grade_capstone_eval_questions(@capstone_evaluation_submission_answers)
+    @total_score = @question_scores.map { |k, v| v }.sum
```
sadleb, line 24: I think I'm getting confused about what we need for when Fellows view this vs when admins view it. The admin view needs `@capstone_evaluation_submissions` but not the Fellow view, right? The Fellow view needs `@question_scores` but not the admin, right?

Ditto on the above `set_blah` actions. Some are only needed for the admin view but not Fellow, like `set_submissions_have_been_graded`

Maybe it's not worth the complexity of conditionally querying the database and setting the instance variables based on whether they are an admin vs fellow but it does give me pause that a future refactor might accidentally expose something to a Fellow that only admins should be able to see.

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+  end
+
+  def all_capstone_evaluation_submissions
+    @course.capstone_evaluation_submissions
+  end
+
+  def new_capstone_evaluation_submissions
+    @course.capstone_evaluation_submissions.ungraded
+  end
```
sadleb, line 72: Nit: do we need these methods in the controller? Can this all be inside the service and just accessed through `attr_reader`s or methods on the service?

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def set_grade_capstone_eval_service_instance
+    @grade_capstone_eval_service = GradeCapstoneEvaluations.new(
+      @course,
+      @lti_launch,
+      new_capstone_evaluation_submissions,
+      all_capstone_evaluation_submissions
```
sadleb, line 79: I think the service can handle returning new vs all submissions without having to pass them in. It already has the `@course`

app/policies/capstone_evaluation_submission_answer_policy.rb
```diff
@@ -0,0 +1,26 @@
...
+    def initialize(user, record)
+      @user = user
+      @record = record
+    end
+
+    def resolve
+      return CapstoneEvaluationSubmissionAnswer.all if user.admin?
+
+      # Get all submissions for the given user
+      return CapstoneEvaluationSubmissionAnswer.where(for_user: user).graded
```
sadleb, line 23: The comment doesn't match the code. This returns all graded answers. 

I'm still a bit confused how this scope is used though. We pass graded submissions in when we call it, but completely ignore the graded submissions and just return all graded answers (for any course) for the user? This actually may be merge blocking. If they are a returning Fellow they could have answers from a previous course and those would be returns too I think?

app/policies/capstone_evaluation_submission_answer_policy.rb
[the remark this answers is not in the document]
elenabaurkot: Removed CapstoneEvaluationSubmissionAnswerPolicy - querying proper submissions using .where( ) and scopes instead

app/controllers/capstone_evaluation_results_controller.rb
elenabaurkot: Updated these instance variables to call methods from the grade_capstone_evaluation service instead to get what is needed for the different views

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  # once a submission is graded and included on the teamwork grade, new is set to false
+  def graded_capstone_evaluation_submissions
+    @course.capstone_evaluation_submissions.graded
+  end
+
+  def set_capstone_evaluation_submission_answers
+    @capstone_evaluation_submission_answers = policy_scope(
+      CapstoneEvaluationSubmissionAnswer.where(submission: graded_capstone_evaluation_submissions)
+    )
+  end
```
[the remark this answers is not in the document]
elenabaurkot: Updated to remove this and query for the proper submission answers (graded or all submission) in the `grade_capstone_eval_questions` in the `grade_capstone_evaluations` service

app/controllers/capstone_evaluation_results_controller.rb
elenabaurkot: Updated to get the submissions using `@course`

app/policies/capstone_evaluation_submission_answer_policy.rb
elenabaurkot: Removed this policy

sadleb:
> Noting that we need a deploy / release plan to replace the current Teamwork assignments with the LTI one. We need to do this for both the Accelerator Course template and all launched spring 2022 courses. We may have to write a script or rake task that we run after the code is live

We talked about this and the plan is just to manually go change the assignment in the template and all launched courses to External Tool, open in new tab, and paste the launch URL in there.

app/services/grade_capstone_evaluations.rb
```diff
@@ -0,0 +1,114 @@
...
+    question_scores = {}
+    CapstoneEvaluationQuestion.all.each do |question|
+      question_scores[question.text] = 0.0
+    end
+
+    if all_submissions == true
+      submissions = all_capstone_eval_submissions
+    else
+      submissions = all_capstone_eval_submissions.graded
+    end
```
sadleb, line 46: Instead of loading all `capstone_evaluation_submissions` for all users, I think you can just get the ones for the target user. E.g.
```
    if all_submissions == true
      submissions = @course.capstone_evaluation_submissions.where(user: user)
    else
      submissions = @course.capstone_evaluation_submissions.where(user: user).graded
    end
```

sadleb:
> > In the PR description, I think one of the screenshots labelled "Fellow view for the teamwork grade after they are published:" is out of date.
> 
> This is actually something I meant to ask about and forgot, right now this is the actual view because the fellow still doesn't actually submit anything for this assignment and we also don't submit anything for them because we discussed there is sometimes an issue when we submit for them if it's not from their launched assignment. Currently, the code is just using the Canvas API to grade the assignment, so when they view the grade this is what they see, but if they launch the actual teamwork assignment that is when they see their grade breakdown. I don't know if there's a way to change the submission view without submitting something or to put in something to let a fellow know they should go to the assignment to see the grade breakdown because it doesn't seem super intuitive @sadleb

Follow-up PR to address this:
* https://github.com/bebraven/platform/pull/999

sadleb (approved):
Approving in advance of addressing the [one last comment](https://github.com/bebraven/platform/pull/991#discussion_r815019274) about not using `all_capstone_eval_submissions `

app/services/grade_capstone_evaluations.rb
sadleb: ActiveRecord actually might be smart enough to not actually do the database query until you've added in the `for_user_id` stuff below, but I still think it's clearer to operate on the `submissions` only for this user and not all users.

elenabaurkot: Discussed on call-  goal is really to get all Capstone Evaluation Submission Answers submitted for the user for the course (to either calculate their grade from the evaluations their peers gave or show them their grade) so updated to the following:
```
if all_submissions == true
      submission_answers_for_user = CapstoneEvaluationSubmissionAnswer.where(for_user: user)
else
      submission_answers_for_user = CapstoneEvaluationSubmissionAnswer.where(for_user: user).graded
end
```

PR #999, opened 2022-02-22
Cap evals breakdown viewable from grades

elenabaurkot:
### Description
Previously, students could only view their Capstone Evaluation Teamwork grade breakdown by launching the assignment. Now they can view it by going to the grades and selecting the assignment to see the grade breakdown in the submission details.

### Motivation and Context
Asana Task: https://app.asana.com/0/1201131148207877/1201827229050414/f

https://github.com/bebraven/platform/pull/991#issuecomment-1036597720

Screen Shot 2022-02-22 at 12 43 19 PM


lib/canvas_api.rb
```diff
@@ -120,15 +120,17 @@ def update_course_page(course_id, wiki_page_id, wiki_page_body)
...
-      'to create basic_lti_launch submission types instead'
-    #  body = {
-    #    'submission[submission_type]' => 'basic_lti_launch',
-    #    'submission[user_id]' => canvas_user_id,
-    #    'submission[url]' => lti_launch_url
-    #  }
-    #  post("/courses/#{canvas_course_id}/assignments/#{assignment_id}/submissions", body)
+    # raise NotImplementedError, 'This is going to lead to headaches. It''ll sort of work ' \
+    #   'in certain situations, but fail with permission errors in others. Use LtiAdvantage ' \
+    #   'to create basic_lti_launch submission types instead'
```
elenabaurkot, line 125: Wasn't sure if I should leave this comment in or not in case it is relevant for when we create a submission in a different way

app/services/grade_capstone_evaluations.rb
```diff
@@ -20,6 +20,15 @@ def run
...
 
       grades[user.canvas_user_id] = total_score
+
+      # Create a submission for the Capstone Evaluations Teamwork assignment for the user
+      # we are giving a grade, so they are able to see the grade breakdown from grades
+      CanvasAPI.client.create_lti_submission(
+        @course.canvas_course_id,
+        @lti_launch.assignment_id,
+        user.canvas_user_id,
+        'https://platformweb/capstone_evaluation_results/launch'
```
sadleb, line 30: This URL needs to not be hardcoded. It will break in prod. E.g. something like `launch_capstone_evaluation_results_url(protocol: 'https')`

lib/canvas_api.rb
```diff
@@ -119,16 +119,21 @@ def update_course_page(course_id, wiki_page_id, wiki_page_body)
...
-    #    'submission[url]' => lti_launch_url
-    #  }
-    #  post("/courses/#{canvas_course_id}/assignments/#{assignment_id}/submissions", body)
+    body = {
+       'submission[submission_type]' => 'basic_lti_launch',
+       'submission[user_id]' => canvas_user_id,
+       'submission[url]' => lti_launch_url
+     }
+
+     post("/courses/#{canvas_course_id}/assignments/#{assignment_id}/submissions", body)
```
sadleb, line 136:
```suggestion
     response = post("/courses/#{canvas_course_id}/assignments/#{assignment_id}/submissions", body)
     JSON.parse(response.body)
```
We generally do this so that the method returns a hash instead of a raw json string and it's easier to work with (even if you don't currently use the response, i think following the pattern is good).

Also, the reason for doing it on two separate lines is just so the stacktrace would make it clear whether it was a JSON parsing error vs an error doing the post.

spec/services/grade_capstone_evaluations_spec.rb
```diff
@@ -30,15 +30,37 @@
...
+        capstone_evaluation_question_id: i + 1,
+        input_value: 8
+      ) }
+    end
+
     before(:each) do
       allow(CanvasAPI).to receive(:client).and_return(canvas_client)
+      allow(canvas_client)
+        .to receive(:create_lti_submission)
+        .and_return(nil)
```
sadleb, line 51: Nit: I don't think you actually need the `and_return(nil)` part. I know we do that in some places, but I think that's the default and we just did that without thinking.

app/services/grade_capstone_evaluations.rb
```diff
@@ -20,6 +20,15 @@ def run
       total_score = question_scores.map { |k, v| v }.sum
 
       grades[user.canvas_user_id] = total_score
+
+      # Create a submission for the Capstone Evaluations Teamwork assignment for the user
+      # we are giving a grade, so they are able to see the grade breakdown from grades
```
sadleb, line 25: Did you test what happens if the button is pressed multiple times by an admin? Does it just keep creating new submissions? That's probably fine as long as it works properly from the Fellow perspective.

elenabaurkot: If they press it again and there aren't any new submissions to grade it redirects them with an alert saying `No new submissions to grade.`

app/services/grade_capstone_evaluations.rb
elenabaurkot: updated to use the rails_url instead

lib/canvas_api.rb
elenabaurkot: Good to know, updated to this 👍

spec/services/grade_capstone_evaluations_spec.rb
elenabaurkot: Removed .and_return(nil) and tests are still passing so that make sense 👍

lib/canvas_api.rb
```diff
@@ -119,16 +119,22 @@ def update_course_page(course_id, wiki_page_id, wiki_page_body)
 
   # Submits an assignment for a user that when viewed, launches the specified
   # lti_launch_url as the submission to view.
+  #
+  # WARNING: This method should only be used in the event that fellows do not ever need to
+  # create a submission for an assignment. If this method is used, LTI Advantage API cannot
+  # be used to submit the assignment again. Once this method is used to create a submission,
+  # future submissions will only be able to be made using this method again, so fellows will
+  # never be able to create a submission themselves.
+  # Use LtiAdvantage 'to create basic_lti_launch submission types instead'
```
sadleb, line 128: Can we also add a WARNING in the `capstone_evaluation_results_controller.rb` b/c that's where someone in the future would go if they were going to add a `create` or `edit` action that tried to post a submission back using `LtiAdvantage`?

sadleb (approved):
lgtm pending adding a warning comment in the place a dev would most likely be making the mistake of trying to use `LtiAdvantage` to post a submission

lib/canvas_api.rb
elenabaurkot: Updated to add in the extra warning 👍

PR #1002, opened 2022-02-25
Correct wrong discord account

elenabaurkot:
### Description
This update adds the ability for a Fellow to change the account they have associated with the Braven Discord Server. If a user joins the Braven Discord server, but then decides they want to use a different email/account for Discord, on the Discord Signups completed assignment page there is a button that allows them to reset the assignment and do it again with the correct account. This doesn't remove their old account from the Braven Discord user so someone will need to do this manually if we want the old account removed. 

### Motivation and Context
Asana Task: https://app.asana.com/0/1201131148207877/1201776442167386/f

Makes it easier for users to update the account they're using themselves, without having to submit a request. 

Completed Assignment Page with "Reset Assignment" button:
Screen Shot 2022-02-25 at 2 14 56 PM

Notice shown on redirect, after clicking the reset button: 
Screen Shot 2022-02-25 at 2 15 37 PM


app/views/discord_signups/launch.html.erb
```diff
@@ -80,7 +80,13 @@
     <ul>
       <li>Discord Login Email: <strong><%= @discord_user["email"] %></strong></li>
       <li>Discord Username: <strong><%= @discord_user["username"] %>#<%= @discord_user["discriminator"] %></strong></li>
-    </ul>
+    </ul><br>
+
+    <p>If you want to change the email associated with your Braven Discord account, click the button below to reset this assignment. Then go thorugh the assignment again and signup with the email you would like to use.</p>
+    <div class="d-flex justify-content-center">
```
elenabaurkot, line 86: Should I add language here saying something like "If you would like to have your old account removed from the Braven Discord server please open a support ticket"?

rshipp: no, we should just handle it from our end. there's a few open tickets for that already

app/controllers/discord_signups_controller.rb
```diff
@@ -177,6 +177,15 @@ def completed
     render layout: 'lti_canvas'
   end
 
+  def reset_assignment
+    authorize :discord_signup
+
+    current_user.update!(discord_token: nil)
+    Honeycomb.add_field('user.discord_token.reset_to_nil', true)
```
rshipp, line 184: can you change this field name to `discord_signups.reset_token` or something like that, so that it's namespaced by the place in the code to look for it rather than a database object?

app/views/discord_signups/launch.html.erb
```diff
@@ -80,7 +80,13 @@
     <ul>
       <li>Discord Login Email: <strong><%= @discord_user["email"] %></strong></li>
       <li>Discord Username: <strong><%= @discord_user["username"] %>#<%= @discord_user["discriminator"] %></strong></li>
-    </ul>
+    </ul><br>
+
+    <p>If you want to change the email associated with your Braven Discord account, click the button below to reset this assignment. Then go thorugh the assignment again and signup with the email you would like to use.</p>
```
rshipp, line 85: Can we rephrase this to not mention email at all? e.g. 
```suggestion
    <p>If you want to change the Discord account you use for Braven, click the button below to reset this assignment. Then go through the assignment again and be sure to log in to the correct Discord account.</p>
```
might need to run this by Alexis and see if she has any input

app/controllers/discord_signups_controller.rb
elenabaurkot: That makes sense, updated to this 👍

app/views/discord_signups/launch.html.erb
elenabaurkot: Updated to this for now and sent to Alexis to double check if it sounds alright

app/controllers/discord_signups_controller.rb
rshipp: looks like the new field name is still based off the model name. take a look at the naming conventions here: https://github.com/bebraven/platform/wiki/Honeycomb:-Monitoring-&-Troubleshooting#field-names

we differentiate between things that represent object values (e.g. `user.email`), and things that represent local variables/flags (e.g. `my_service.did_something`). in this case the field you're setting represents a local flag indicating the token was reset, so it falls into the second category, but the name you're using here is referencing a database object (`user`). it should instead reference the service/controller name (`discord_signups`)

elenabaurkot: ah okay, that actually makes sense now, updated to `discord_signups.reset_token` thanks for the explanation @rshipp

app/controllers/discord_signups_controller.rb
```diff
@@ -177,6 +177,15 @@ def completed
     render layout: 'lti_canvas'
   end
 
+  def reset_assignment
+    authorize :discord_signup
+
+    current_user.update!(discord_token: nil)
```
sadleb, line 183: Do we care about leaving the old user in there or should we do something to try and remove it from the server @rshipp ?

rshipp: discussed this in the eng call, we should leave it in

rshipp (approved):

PR #1003, opened 2022-02-25
Update ruby deps

rshipp:
### Description

Update all gems with `bungle update`, pin devise_cas_authenticable to v1 since there's a v2 out now that breaks everything.

Note: This PR requires a container **rebuild**.

### Motivation and Context

Fix security issues

### Types of changes

### Checklist:


spec/rails_helper.rb
```diff
@@ -115,7 +115,7 @@ def should_not condition
 # Patch HttpStreamingResponse to make rack-proxy compatible with webmocks
 # Note: we "load" here b/c it was already "required" before the Rack stuff loaded
 # so it seemed to be getting clobbered.
-load 'support/http_streaming_response_patch'
+load 'support/http_streaming_response_patch.rb'
```
rshipp, line 118: see the heroku CI results for the first commit in this branch. @sadleb any idea what's going on here? this has me a little worried that there might've been some change we're missing that will break other things at runtime.

sadleb: Super weird. I googled around a bit and found one article mentioning that `load` needs the extension. See the "DIFFERENCE BETWEEN LOAD AND REQUIRE" section here: https://codedecoder.wordpress.com/2013/09/23/require-and-load-in-ruby-loaderror-cannot-load-such-file/

So either this wasn't working before, but didn't fail. Or it was working and something changed, either with the default configs or something core to the ruby code. I guess if we grep for any other places in the code that we use `load` instead of `require` and add the extension, we should be ok @rshipp ?

rshipp: looks like that's the only one
```
$ grep -r "load ['\"]" app config spec lib
spec/rails_helper.rb:load 'support/http_streaming_response_patch.rb'
lib/tasks/db.rake:    load "#{Rails.root}/db/dummies.rb"
```

sadleb (approved):
lgtm. We should be on standby when we merge to do some quick sanity testing after this deploys.

PR #1004, opened 2022-03-01
Sync canvas assignment overrides (due dates)

rshipp:
### Description

Sync Canvas assignment override data into the local platform db.

Prerequisite work for a larger task on creating a due date management tool.

### Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1201634733776799/f

### Types of changes


app/services/sync_canvas_dates.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'canvas_api'
+require 'salesforce_api'
+
+class SyncCanvasDates
```
rshipp, line 6: naming question: i changed my mind partway through when I got more information, so the naming is inconsistent. do we want to call this stuff "Dates", or "AssignmentOverrides"? I'm currently leaning towards the latter, but thought i'd ask before making the change.

db/migrate/20220301210843_create_canvas_assignment_override.rb
```diff
@@ -0,0 +1,29 @@
...
+      t.timestamps(default: -> { "CURRENT_TIMESTAMP" })
+
+      # Add a unique constraint that enforces uniqueness for [override, user]
+      # and [override, section] pairs. Canvas AssignmentOverrides will always
+      # only have a section or 1+ users, never both. Unique indexes treat null
+      # as non-matching (e.g. two rows (1, null) and (1, null) will not violate
+      # a uniqueness constraint), so we use COALESCE to get around that.
+      t.index 'canvas_assignment_override_id, coalesce(canvas_section_id, -1), coalesce(canvas_user_id, -1)',
+        unique: true,
+        name: 'index_canvas_assignment_overrides_unique_1'
```
rshipp, line 26: this took me a while to figure out, but works now

app/services/sync_canvas_dates.rb
elenabaurkot: I think AssignmentOverrides makes more sense

app/models/canvas_assignment_override.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class CanvasAssignmentOverride < ApplicationRecord
+  # Note we can NOT use canvas_assignment_override_id as a PK here, because
+  # it's not unique in this table. See the
```
elenabaurkot, line 5: Is the canvas_assignment_override_id not unique because if overrides are used with student Ids, multiple students with the same override will have the same canvas_assignment_override_id?

rshipp: correct

db/migrate/20220301210843_create_canvas_assignment_override.rb
rshipp: > When you use `coalesce(canvas_section_id, -1)` does this mean it will either evaluate to the `canvas_section_id` if given one or to -1/null if it doesn't have a `canvas_section_id`?

if `canvas_section_id` is `NULL`, the `coalesce` will return `-1`; otherwise, it returns the `canvas_section_id`. 

> Since CanvasAssignmentOverrides only have a section or users, do one of either the `canvas_section_id` or `canvas_user_id` have to evaluate to -1?

correct, one of them will always be -1.

> Also, could a CanvasAssignmentOverride be added for a section and then another CanvasAssignmentOverride be added for a specific student within that section?

yes, in fact that will always be the case when individual students are granted extensions, because there were already section-level due dates (which are defined as overrides). if you want to answer "what is a certain student's due date", you have to look up both the user-level and section-level overrides, and return the latest applicable date.

elenabaurkot:
lgtm 👍

app/services/sync_canvas_dates.rb
sadleb: No strong opinion on the service name. I'm good with either Dates or AssignmentOverrides, but for the model I think we should stick with what you have, CanvasAssignmentOverrides. If you change this to SyncCanvasAssignmentOverrides I imagine a new dev would grep for "Date" or "Dates" if they didn't know what an override was and was trying to find the code. So a comment mentioning that overrides are dates would hopefully let them find this without too much trouble.

app/services/sync_canvas_dates.rb
```diff
@@ -0,0 +1,48 @@
...
+  def run
+    Honeycomb.start_span(name: 'sync_canvas_dates.run') do
+      canvas_course_ids = SalesforceAPI.client.get_current_and_future_accelerator_canvas_course_ids
```
sadleb, line 9: Should we start using HerokuConnect for new code that just reads SF info? E.g. add a new scope here https://github.com/bebraven/platform/blob/highlander/app/models/heroku_connect/program.rb#L42 to return this list?

app/services/sync_canvas_dates.rb
```diff
@@ -0,0 +1,48 @@
...
+      Honeycomb.add_field('sync_canvas_dates.courses.count', canvas_course_ids.count)
+      Rails.logger.info("Found #{canvas_course_ids.count} courses")
```
sadleb, line 11: nit: debug instead of info?

app/services/sync_canvas_dates.rb
```diff
@@ -0,0 +1,48 @@
...
+      canvas_course_ids.each do |canvas_course_id|
+        SyncCanvasDates.sync_course(canvas_course_id)
+      end
+    end
+  end
+
+  def self.sync_course(canvas_course_id)
+    Honeycomb.start_span(name: 'sync_canvas_dates.course') do
+      Honeycomb.add_field('canvas.course.id', canvas_course_id.to_s)
+      Rails.logger.info("Syncing canvas_course_id=#{canvas_course_id}")
```
sadleb, line 22: nit: "Syncing **dates** for..." ?

app/services/sync_canvas_dates.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      # Note: We're assuming here that the assignment-level due/lock/unlock dates
+      # are not set, and that all dates are configured via section or user-level
+      # overrides. If that ends up not being true, we'll have to change some of
+      # this to account for that.
```
sadleb, line 27: Slight possibility we could start using the "Everyone" assignment level dates to solve this user story: https://app.asana.com/0/1201131148207877/1200733149189730

app/services/sync_canvas_dates.rb
```diff
@@ -0,0 +1,48 @@
...
+      overrides = CanvasAPI.client.get_assignment_overrides_for_course(canvas_course_id)
```
sadleb, line 28: If syncing runs for a course that has no dates set yet, does `get_assignment_overrides_for_course` return an empty array or nil? If nil, this will error. We ran into this in the past for something I can't recall at the moment.

db/migrate/20220301210843_create_canvas_assignment_override.rb
sadleb: What happens if someone does enter a date into the "Everyone" field (aka the top-level override for the whole assignment)? Will anything break? Even though we don't currently use or expect that to be set, it's very easy to do and therefore will happen.

sadleb (commented):
Waiting to hear back about what happens if the Everyone override happens to be set and whether this gracefully handles a course that has no dates set yet before approving.

app/services/sync_canvas_dates.rb
rshipp: that'll break stuff for this PR. let's talk about this on monday then

db/migrate/20220301210843_create_canvas_assignment_override.rb
rshipp: has no effect on the index (bc it'd just be `(override_id, -1, -1)`, but does break other assumptions

rshipp: for posterity, the "Everyone Else" due date for assignments does not show up as an override, it only shows up as a due date for the assignment. (so it also won't show up in our synced data at all, which is fine for now and easy to change later if we find out we need it)

app/services/sync_canvas_dates.rb
rshipp: https://github.com/bebraven/platform/pull/1004#discussion_r826410632

rshipp:
merging based on meeting discussion

PR #994, opened 2022-02-17
Fix project radio buttons in safari

rshipp:
### Description

Safari doesn't fire focus/blur events for radio buttons when you click on/off them. As a workaround, use the `change` event instead. 

(Test this live: http://jsfiddle.net/6ok8jyh1/3/. From https://stackoverflow.com/questions/25272926/safari-click-focus-on-radio-button.)

Change event: https://www.w3schools.com/jsref/event_onchange.asp

### Motivation and Context

Thread: https://bebraven.slack.com/archives/C02H6RE7DB5/p1645108879444919

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have added neccessary labels
- [ ] My PR title follows the convention in the guide
- [ ] I have assigned myself to the PR
- [ ] I have requested reviews from team members
- [ ] I have created a task for reviewers on PM tool
- [ ] All new and existing tests passed.

PR #1006, opened 2022-03-02
Fix project autosave retry bug

rshipp:
### Description

Fix the long-standing retry bug that occurs when you click "Retry" on an input that has not been changed since the last save.

Since we use onchange now instead of onblur (#994), we no longer need to manually check for changed input values.

### Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1201881062197879/f

### Types of changes


sadleb (approved):
Lgtm

PR #1008, opened 2022-03-03
Project autosave fixes/debug info

rshipp:
# Description

I'll update this tomorrow - it's a day's worth of troubleshooting in Safari. Tries to auto-detect and fix autosave issues, and sends some extra info to Honeycomb for debugging.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1201868723621550/f


sadleb (approved):
Let's talk about this more when we get a chance b/c I don't understand it, but I don't see anything glaring that could break.

PR #1009, opened 2022-03-07
Project autosave followup

rshipp:
# Description

Followup for #1008. Add more spans, change timers. Remove non-timer events so we can look at browser stats and get a better idea who we have to repair for.

# Motivation and Context

Task:  https://app.asana.com/0/1201131148207877/1201868723621550/f

# Screenshots

PR #1010, opened 2022-03-10
Project Autosave: Add more debug info

rshipp:
# Description

Followup to #1009, #1008. Add more debug info to try and see if we're actually fixing things, or checking the wrong thing to detect the problem, or what.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1201868723621550/f

Had another ticket some in with a weird trace that ran the 10k repair strategy, but seems to have been working.


app/javascript/packs/project_answers.js
```diff
@@ -286,10 +297,17 @@ honeySpan.sendNow();
 */
 [2000, 5000, 10000].forEach(delay => {
     setTimeout(() => {
-        if(getOneSupportedElement().onchange === null) {
+        const input = getOneSupportedElement();
+        if(input.onchange === null) {
```
elenabaurkot, line 301: If someone clicks into an input, doesn't add an answer and then clicks off of it could the `input.onchange === null` without it being an actual issue?

rshipp: if `onchange` is null it means the event listener hasn't been attached to the element. this not asking whether the event has fired. by default, all event handlers on every element will be null. when you attach a handler, the `on${event}` attribute on the element will be set to the handler function. e.g., try this in your browser console:

```javascript
const element = document.querySelector('body');
console.log(element.onclick);
element.onclick = (e) => { console.log('in event handler'); console.log(e); };
console.log(element.onclick);
```

app/javascript/packs/project_answers.js
```diff
@@ -265,6 +273,9 @@ const getOneSupportedElement = () => {
 // Send an event outside of main so we know the JS is running at all.
 const honeySpan = new HoneycombAddToSpan('project_answers', 'script');
 honeySpan.addField('run', true);
+// Bump this version number any time you need to see what version of the JS they're using.
+// TODO: come up with a better way to get this globally.
```
sadleb, line 277: This is ugly but easy so I'm good with it.

sadleb (approved):

PR #1011, opened 2022-03-11
Silence noisy ActiveRecord::RecordNotUnique alert

rshipp:
# Description

Patch Honeycomb's `beeline` gem to partially revert the change introduced in https://github.com/honeycombio/beeline-ruby/pull/166.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1201902273750962/f

The noise (or at least some of it) was coming from here: https://github.com/bebraven/platform/blob/c866fd497699135cbcc6cf326a49de34cdee88a7/app/controllers/rise360_module_versions_controller.rb#L48

Since `create_or_find_by` raises and handles an exception by design, we never want to see those coming through as errors. I'm assuming here that if there are SQL exceptions that we do want to see errors for, we'll let them propagate up far enough that they get stuck in the span. This should be pretty safe.

Example trace before this patch: https://ui.honeycomb.io/braven/datasets/platform.braven.org/result/k7u4ybcLSiP/trace/mjKQmnG345e?span=a8b50d1ea1844c4f

Example trace after this patch: https://ui.honeycomb.io/braven/datasets/development/result/mvUDsdaBZCD/trace/wTDVUATzn7i?span=aec6986c71fd4165


config/initializers/honeycomb_beeline270_patch.rb
```diff
@@ -0,0 +1,36 @@
...
+          error, error_detail = payload[:exception]
+          # START BRAVEN PATCH
+          unless name == 'sql.active_record'
+            # Don't copy `sql.active_record.error` to `error`, because it
+            # clutters up the data with noise from create_or_find_by queries
+            # that raised and handled their own exception.
+            span.add_field("error", error) if error
+            span.add_field("error_detail", error_detail) if error_detail
+          end
+          # /END BRAVEN PATCH
```
rshipp, line 31: this is the patch; the rest of this file is duplicated from the source listed in the references at the top of this file

sadleb (approved):
Lgtm. I sort of assume they'll change this upstream at some point b/c of how normal these exceptions should be.

PR #1012, opened 2022-03-14
Updated attendance to show most recently submitted attendance data

elenabaurkot:
# Description
Previously, if attendance was taken by someone, other people could only see the attendance by masquerading as the user who took attendance and then launching the attendance event. This update makes it so that any time attendance is launched, it shows the most recent attendance data taken for the given `CourseAttendanceEvent`. We only create new CourseAttendanceEvent submissions if it's the first time attendance is being taken, or if the user taking attendance is different than the previous user who took it- otherwise we just update the attendance. 

# Motivation and Context
To make it easier to take attendance, so users don't have to figure out who took attendance and masquerade as them (or look into Periscope) to be able to view the attendance data. 

Task: https://app.asana.com/0/1201131148207877/1201862534993707/f


sadleb (commented):
some suggestions

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -71,6 +78,15 @@ def edit
   def update
     authorize @attendance_event_submission
 
+    # If we are updating attendance and a new user is submitting, we want to create a new AttendanceEventSubmission
+    # If it is the same user who previously submitted we will just update their answers
+    if @attendance_event_submission.user_id != current_user.id
+      @attendance_event_submission = AttendanceEventSubmission.new(
+        user: current_user,
+        course_attendance_event: @course_attendance_event,
+      )
```
sadleb, line 87: I think you need to use `create!` here. In the initial launch, we would create a new one, then authorize, then save it. Now that we're creating the new one on the fly here, we need to in effect do the same thing.

elenabaurkot: ah okay that's good to know - did not consider that!

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -45,15 +45,25 @@ def launch
...
+
+    # This @attendance_event_submission could be for either this user or another user.
+    # Explicitly authorize this user and let a new submission get created for this user
+    # in the update method if we started from another user's submission
+    AttendanceEventSubmissionPolicy.new(current_user, @attendance_event_submission).launch?
     authorize @attendance_event_submission
+
     @attendance_event_submission.save! # Do this after the authorization so we don't add an unauthorized .new record
     redirect_to edit_attendance_event_submission_path(
       @attendance_event_submission,
```
sadleb, line 69: @elenabaurkot I just realized that I have the same question about the `edit` policy that I did for the `launch` one. If it happens to be a submission for another user, are we authorizing the right thing? On second thought, maybe just calling `authorize @attendance_event_submission` set's the `user` inside the policy to the `current_user`. So the thing I said above about using `AttendanceEventSubmissionPolicy.new(current_user, @attendance_event_submission).launch?` is wrong and we were already good on the policy side of things?

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -80,13 +75,14 @@ def edit
...
     authorize @attendance_event_submission
+    course_attendance_event = CourseAttendanceEvent.find(@attendance_event_submission.course_attendance_event_id)
 
     # If we are updating attendance and a new user is submitting, we want to create a new AttendanceEventSubmission
     # If it is the same user who previously submitted we will just update their answers
     if @attendance_event_submission.user != current_user
       @attendance_event_submission = AttendanceEventSubmission.create!(
         user: current_user,
-        course_attendance_event: @course_attendance_event,
+        course_attendance_event: course_attendance_event,
```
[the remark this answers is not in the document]
elenabaurkot: ah okay nice, going to use that, thanks 👍

spec/controllers/attendance_event_submissions_controller_spec.rb
```diff
@@ -417,11 +464,33 @@
...
       end
+
+      context 'when the current user who created the attendance event submission is different than the current user' do
+        before(:each) do
+          attendance_event_submission.update(user: other_user)
+        end
+
+        it 'creates a submission' do
+          expect { subject }.to change(AttendanceEventSubmission, :count).by(1)
+        end
```
sadleb, line 475: If it's not too hard, can you try and test that the new submission created is owned by the user doing the update?

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -45,14 +45,21 @@ def launch
...
-    )
+    if all_attendance_submissions.empty?
+      @attendance_event_submission = AttendanceEventSubmission.new(
+        user: current_user,
+        course_attendance_event: @course_attendance_event,
+      )
+    else
+      @attendance_event_submission = all_attendance_submissions.last
+    end
+
```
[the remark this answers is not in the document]
elenabaurkot: updated to this (minus the authorization update based on [this comment](https://github.com/bebraven/platform/pull/1012#discussion_r826317593))

spec/controllers/attendance_event_submissions_controller_spec.rb
elenabaurkot: Yes, updated to add that test in 👍

PR #1014, opened 2022-03-16
Fellows can't submit capstone evals for fellows without a canvas_user_id

elenabaurkot:
# Description
This is a fix for the [Capstone Evaluations Grade Breakdown PR ](https://github.com/bebraven/platform/pull/991) where we made an update to allow admin users to grade capstone evaluation submissions by launching the 'GROUP PROJECT: Capstone Challenge: Teamwork' assignment and clicking a button that creates a teamwork grade for fellow's based on the evaluations of their teammates. While demo-ing the new feature we ran into an [error](https://ui.honeycomb.io/braven/datasets/platform.braven.org/result/AeT6JCGQBLb/trace/5iPQ2n4nQ7b?span=b6be38584b392349)  that occurred when an admin user clicks the button to create grades for users that did not have a canvas_user_id, but had capstone evaluations submitted for them. The error occurred because fellows were able to submit capstone evaluations for users in their section that didn't have a `canvas_user_id` most likely because they never actually signed up for canvas. When the admin tried to create grades for these fellows that did not actually exist on Canvas, a 401 unauthorized error occurred. This update makes it so that fellows can only submit capstone evaluations for other fellows that have a canvas_user_id. 

# Motivation and Context
To ensure the Capstone Evaluation grading tool works properly and is not trying to submit grades for users that are not on Canvas. 

Task: https://app.asana.com/0/1174274412967132/1199210519637836/f


app/controllers/capstone_evaluation_submissions_controller.rb
```diff
@@ -42,7 +42,9 @@ def set_eval_users
       student_section = current_user.student_section_by_course(@course)
       if student_section&.students&.exists?
         # You're enrolled as a student and there are other students in this course
-        @eval_users = student_section.students.where.not(id: current_user.id)
+        @eval_users = student_section.students.filter do |student|
+          student.id != current_user.id && student.canvas_user_id != nil
+        end
```
rshipp, line 47: question, that you may have already tried - does `where.not(id: current_user.id, canvas_user_id: nil)` work here? other than that, this seems like a great solution to me

sadleb (approved):
Approving in advance of addressing Ryan's suggestion.

app/controllers/capstone_evaluation_submissions_controller.rb
elenabaurkot: I tried this out originally and it didn't work- just gave it another try and still a no go, but looked it up and it seems to work if I do it like this

```
        @eval_users = student_section.students.where.not(id: current_user.id)
          .and(student_section.students.where.not(canvas_user_id: nil))
```

Another suggestion I saw was using your suggestion and chaining on `.to_sql` at the end, but that didn't seem to work for me either

I updated to the first method I pasted here, do we think that is better than the original .filter? @rshipp @sadleb

app/controllers/capstone_evaluation_submissions_controller.rb
```diff
@@ -43,6 +43,7 @@ def set_eval_users
       if student_section&.students&.exists?
         # You're enrolled as a student and there are other students in this course
         @eval_users = student_section.students.where.not(id: current_user.id)
+          .and(student_section.students.where.not(canvas_user_id: nil))
```
sadleb, line 46: Just read some stuff on this, and yeah, looks like the SQL query for multiple attributes in a `where.not` clause changed from what we want to something different [further reading here](https://www.bigbinary.com/blog/rails-6-deprecates-where-not-working-as-nor-and-will-change-to-nand-in-rails-6-1) (we’re on rails 6.1 btw). Can you try the following and if that doesn’t work go back to original `filter`? 

‘’’
 @eval_users = student_section.students.where.not(id: current_user.id).where.not(canvas_user_id: nil)
 ‘’’

app/controllers/capstone_evaluation_submissions_controller.rb
sadleb: Just read some stuff on this, and yeah, looks like the SQL query for multiple attributes in a `where.not` clause changed from what we want to something different. [further reading here](https://www.bigbinary.com/blog/rails-6-deprecates-where-not-working-as-nor-and-will-change-to-nand-in-rails-6-1) (we’re on rails 6.1 btw). Can you try the following and if that doesn’t work go back to original `filter`? 

```
 @eval_users = student_section.students.where.not(id: current_user.id).where.not(canvas_user_id: nil)
 ```

elenabaurkot: that worked! @sadleb

PR #1015, opened 2022-03-16
Raise error if 'require authentication to join' is on in Zoom settings

elenabaurkot:
# Description
In certain situations, when adding registrants to a Zoom meeting, if the Zoom settings has the "require authentication to join" box checked it will cause the error `{"code":1001,"message":"User does not exist: test-user@test.com"}` (it seems to depend on who the host is). This update raises an error and tells the user generating Zoom links to make sure this "require authentication to join" box is unchecked in the Zoom settings. 

# Motivation and Context
To make sure users know why the generate Zoom links tool may be failing. 

Task: https://app.asana.com/0/1201131148207877/1201941919088518/f

# Screenshots
Screen Shot 2022-03-16 at 5 14 30 PM

"Require authentication to join" box should be unchecked:
Screen Shot 2022-03-16 at 5 16 10 PM


sadleb (approved):
Approving in advance of the `nit` suggestion.

PR #1017, opened 2022-03-21
Render errors/not_authorized using HTML format

elenabaurkot:
# Description
When a user makes a request for a material in an unexpected format they should now see an unauthorized error - previously we were getting a missing template error because it was trying to render the `errors/not_authorized` view with the wrong format (css instead of html).

# Motivation and Context
To prevent incorrect missing template error, when users should be seeing a not authorized error.

Task: https://app.asana.com/0/1201131148207877/1201817062473286/f


rshipp (approved):
this seems good to me

PR #1018, opened 2022-03-21
Handle Invalid LtiLaunch

elenabaurkot:
# Description
This update handles when an LtiLaunch expires, by raising and LtiInvalidError and sending a user back to the login page. This prevents misleading errors from occuring like `NoMethodError` when trying to access attributes of the LtiLaunch when it is no longer valid. 

# Motivation and Context
To prevent misleading error messages when a user's LtiLaunch is invalid.

Task: https://app.asana.com/0/1201131148207877/1201925454385521/f

# Screenshots
Screen Shot 2022-03-21 at 3 52 56 PM


app/models/lti_launch.rb
```diff
@@ -29,9 +36,16 @@ def self.from_id(user, id)
...
+
+    if launch.nil?
+      raise LtiInvalidError
+    end
+
     # ONLY return this LtiLaunch if it's owned by the current user.
     if user && launch&.user == user
       launch
+    else
+      raise LtiInvalidError
```
rshipp, line 48: i'm a little worried about this behavior change from returning `nil` if nothing's found, to raising an exception. i did a search for `from_id` in the code and it came up with this:

```
$ grep from_id -r app
app/models/lti_launch.rb:  def self.from_id(user, id)
app/controllers/discord_signups_controller.rb:    LtiLaunch.from_id(current_user, current_user.discord_state&.split(',')&.last&.to_i).nil?
app/helpers/lti_helper.rb:    lti_launch ||= LtiLaunch.from_id(current_user, params[:lti_launch_id].to_i) if params[:lti_launch_id]&.to_i
app/helpers/lti_helper.rb:    lti_launch ||= LtiLaunch.from_id(current_user, lti_launch_id) if lti_launch_id
```

have you already / can you check out those locations, and whatever might be calling them, and make sure an exception doesn't break anything in unexpected ways?

elenabaurkot: I realized this after making the PR some specs were failing - I think I fixed up how it would effect the discord_signups_controller now because it used it to see if it needs a new discord state, so I updated it to this 

```
def needs_new_discord_state?
    return true if current_user.discord_state.blank?

    begin
      LtiLaunch.from_id(current_user, current_user.discord_state&.split(',')&.last&.to_i)
    rescue LtiLaunch::LtiInvalidError
      return true
    end
    false
  end
```

going to look into the lti_helper now too

sadleb: If there are some places where we want it to return `nil` b/c that code needs to handle it directly instead of letting the top level `ApplicationController` handle it in `Rescuable` I think it would be cleaner to have the following methods similar to how `ActiveRecord` does it. You'd have to grep for all calls to `from_id` and `from_state` to make sure you changed everything correctly.
```
def self.from_state(state)
  # no change from old code
end

def self.from_state!(state)
  lti_launch = LtiLaunch.from_state!
  raise LtiInvalidError if lti_launch.nil?
  lti_launch
end

### same pattern for from_id
```

app/helpers/lti_helper.rb
```diff
@@ -21,8 +21,12 @@ def get_lti_launch
...
     # No particular reason for this choice, just make sure to document if you
     # change it later.
-    lti_launch = LtiLaunch.from_state(params[:state]) if params[:state]
-    lti_launch ||= LtiLaunch.from_id(current_user, params[:lti_launch_id].to_i) if params[:lti_launch_id]&.to_i
+    begin
+      lti_launch = LtiLaunch.from_state(params[:state]) if params[:state]
+      lti_launch ||= LtiLaunch.from_id(current_user, params[:lti_launch_id].to_i) if params[:lti_launch_id]&.to_i
+    rescue LtiLaunch::LtiInvalidError
+      lti_launch = nil
+    end
```
sadleb, line 29: Looking at where this `get_lti_launch` is used, I think we're basically undoing the whole point of raising the exception. A bunch of places used to just get nil and then other exceptions would be thrown, but we generally want the exception to be thrown and handled in rescuable so that other code isn't hit leading to noisy Honeycomb and Sentry stuff

PR #1021, opened 2022-03-22
Get most recent attendance submission for user for course attendance event

elenabaurkot:
# Description
Reverted this PR: [Updated attendance to show most recently submitted attendance data](https://github.com/bebraven/platform/pull/1012) that got the showed the most recent attendance because it was getting the most recent attendance submission for a specific course event. The most recent submission for a course attendance event could have been for any section so it may not show the actual attendance. By reverting, we went back to the method that any user should only have one attendance submission for any course attendance event, but in the reverted update users were able to submit more than one (if they submitted and then another user submitted for the same event, the original submitted could submit a new one  and create a new attendance event submission because it only checked if the user submitting was the last user to submit and if not it created a new one). Since there are duplicate attendance submissions for a given user for a specific course attendance event, this update makes sure we are getting their most recent submission. 

# Motivation and Context
To make sure we are showing accurate attendance data.


sadleb (approved):
Lgtm

PR #1022, opened 2022-03-24
Update attendance

elenabaurkot:
# Description
Previously, if attendance was taken by someone, other people could only see the attendance by masquerading as the user who took attendance and then launching the attendance event. We made an update in this PR [Updated attendance to show most recently submitted attendance data](https://github.com/bebraven/platform/pull/1012)  that showed the most recent attendance because it was getting the most recent attendance submission for a specific course attendance event. However, we had to revert it because the most recent submission for a course attendance event is only for a given section so it may not show the actual attendance for the section you are looking at. This PR updates to show the most recent attendance without having to masquerade as the user who took it by getting the most recent attendance for any fellow in the section being showed for the course attendance event being viewed. 

# Motivation and Context
To make it easier to take attendance, so users don't have to figure out who took attendance and masquerade as them (or look into Periscope) to be able to view the attendance data.

Task: https://app.asana.com/0/1201131148207877/1201862534993707/f


app/models/attendance_event_submission_answer.rb
```diff
@@ -2,6 +2,9 @@
 
 class AttendanceEventSubmissionAnswer < ApplicationRecord
   belongs_to :attendance_event_submission
+  # TODO: it works without this, but should I have this here to match the has_many in the CourseAttendanceEvent model?
+  has_one :course_attendance_event, through: :attendance_event_submission
```
elenabaurkot, line 6: Everything seems to work without this association, but should I have this here to match the has_many in the CourseAttendanceEvent model?

rshipp: oh, you did this in a different (but completely valid) way than what I mentioned. having this association certainly doesn't hurt anything, but i think the only time we care about having matching inverse associations is if they're direct (not `through:`) with a `belongs_to` on one side and a `has_many`/`has_one` on the other side. personally i lean towards leaving this in, because it's always more convenient when you don't have to add new associations when writing a feature.

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -66,6 +67,17 @@ def edit
...
+    unless @fellow_users.blank?
+      answer_ids = @fellow_users.map do |fellow|
+        @course_attendance_event.attendance_event_submission_answers.where(for_user_id: fellow.id)
+          .order(:updated_at)
+          .pluck(:id)
+          .last
+      end
+
+      @answers = AttendanceEventSubmissionAnswer.where(id: answer_ids)
+    end
```
sadleb, line 80: If `@fellow_users.blank?` this won't run and `@answers` will be nil. Doesn't look like the view handles that?

I don't think we actually need L71. We can just run this regardless of whether there are no `@fellow_users`. E.g. in the console this mimics that to show that no exceptions happen:
```
answer_ids = [].map do |fellow| 1234 end;
@answers = AttendanceEventSubmissionAnswer.where(id: answer_ids)
@answers.find_by(for_user: User.first)
```

elenabaurkot: ah okay that makes sense, just removed that 👍

sadleb (approved):
lgtm!

app/models/attendance_event_submission_answer.rb
elenabaurkot: Okay I will leave it in there - and I wasn't sure if I was doing the association you had mentioned or if I had reversed it, but I think associations make a lot more sense to me than they did before now so that's good at least

PR #1023, opened 2022-03-25
Specs cleanup: compile Webpacker once per run / fix deprecation warnings

sadleb:
# Description
Speeds up specs by only compiling Webpacker once.

This reduced `bundle exec rspec` on my local dev env from 7min, 40sec to about 6min. Curious to see what happens in the CI env.

Update: This doesn't seem to have much of an effect on the test run time in the CI environment, but I think it's still useful to have dev specs be faster. 

There was a failing spec that made no sense and I think was the result of the slight timing change. I cleaned up some deprecation warnings while troubleshooting and it led me to find what I think is a fix for that long standing "bug" where we get the following in the logs but have no clue where it was coming from:
`NoMethodError: undefined method []' for nil:NilClass`.

# Motivation and Context
Task: https://app.asana.com/0/1201131148207877/1201973902482807

# Screenshots
A missing LtiLaunch for an endpoint called with a `state` param or header now properly returns a 403 forbidden error.
image

It used to be a 500 error ([Sentry example](https://sentry.io/organizations/braven-pr/issues/2556551183/?project=5639961&query=is%3Aunresolved&statsPeriod=14d))
image


spec/feature/project_answers/project_answers_spec.rb
```diff
@@ -67,13 +67,17 @@
...
+      expect(page).to have_field(project_submission_answers.first.input_name)
       expect(page).to have_field(project_submission_answers.first.input_name,
         with: project_submission_answers.first.input_value, wait: 10)
     end
 
     it 'saves the input value' do
       # Wait for data to be fetched so it doesn't overwrite
       # the answer we set.
+      # Do it in two steps just for better granularity if the first one fails.
+      expect(page).to have_field(project_submission_answers.first.input_name)
```
sadleb, line 80: For some reason in CI this example was failing but it works locally. I tried putting some logging in temporarily and it started passing. Just mimicking what examples further down in this file do by checking in two steps. Maybe it's a timing thing?

spec/capybara_helper.rb
```diff
@@ -19,14 +19,14 @@
...
-  chrome_opts = chrome_shim ? { "goog:chromeOptions" => { "binary" => chrome_shim } } : {}
+  chrome_opts = chrome_shim ? { "goog:chromeOptions" => { "binary" => chrome_shim }, "goog:loggingPrefs" => {browser: 'SEVERE'} } : {}
   puts "### chrome_opts = #{chrome_opts}"
 
   Capybara.register_driver :chrome do |app|
     Capybara::Selenium::Driver.new(
       app,
       browser: :chrome,
-      desired_capabilities: Selenium::WebDriver::Remote::Capabilities.chrome(chrome_opts)
+      capabilities: Selenium::WebDriver::Remote::Capabilities.chrome(chrome_opts)
```
sadleb, line 29: Gets rid of this warning:

```
WARN Selenium [DEPRECATION] [:desired_capabilities] :desired_capabilities as a parameter for driver initialization is deprecated. Use :capabilities with an Array value of capabilities/options if necessary instead.
```

spec/capybara_helper.rb
```diff
@@ -19,14 +19,14 @@
 if chrome_shim
   Selenium::WebDriver::Chrome.path = chrome_shim
 
-  chrome_opts = chrome_shim ? { "goog:chromeOptions" => { "binary" => chrome_shim } } : {}
+  chrome_opts = chrome_shim ? { "goog:chromeOptions" => { "binary" => chrome_shim }, "goog:loggingPrefs" => {browser: 'SEVERE'} } : {}
```
sadleb, line 22: Was tinkering with getting rid of the following deprecation error. This change isn't actually what fixed it but I think it's good to leave this here in case we ever want to change it to `ALL` to get all console logs instead of the default which seems to be `SEVERE`

```
WARN Selenium [DEPRECATION] Manager#logs is deprecated. Use Chrome::Driver#logs instead.
```

spec/feature/project_answers/project_answers_spec.rb
```diff
@@ -67,13 +68,17 @@
 
     it "fetches the most recent answers" do
       # Check that data is fetched
+      # Do it in two steps just for better granularity if the first one fails.
+      expect(page).to have_field(project_submission_answers.first.input_name)
```
sadleb, line 72: For some reason in CI this example was failing but it works locally. I tried putting some logging in temporarily and it started passing. Just mimicking what examples further down in this file do by checking in two steps. Maybe it's a timing thing?

config/initializers/lti_authentication.rb
```diff
@@ -69,9 +69,12 @@ def finish_authenticate(span, status, message, url = nil, canvas_id = nil, user
...
         Rails.logger.debug(message)
         success!(user)
-      else
+      elsif status == :forbidden
         Rails.logger.warn(message)
-        handle_failure(status)
+        custom!([403,{'Content-Type' => 'text/plain'},["403 Forbidden"]])
+      else # this is a code bug if we get here
+        Rails.logger.error(message)
+        fail!('Unknown server error')
```
sadleb, line 77: Fixes the following error that was happening for non-success `status`es.
```
LtiAuthentication::WardenStrategy couldn't find LtiLaunch using state value
    Rack app ("POST /honeycomb_js/send_span" - (127.0.0.1)): #<NoMethodError: undefined method `[]' for nil:NilClass>
```

elenabaurkot (commented):
lgtm 👍

rshipp (approved):
nothing stands out to me as an issue here

PR #1024, opened 2022-03-25
Remove ability to publish Post-Accelerator Survey to LC playbook

elenabaurkot:
# Description
This update makes it so that the Post-Accelerator Survey cannot be published in an LC playbook course since it is not used in there. We remove the add button to prevent this from happening. 

# Motivation and Context
To avoid accidentally publishing the Post-Accelerator Survey in LC playbook courses. 

# Screenshots
View of editing a Post-Accelerator Survey in an LC playbook course from course management:
Screen Shot 2022-03-25 at 3 12 10 PM

View of editing a Post-Accelerator Survey in an Accelerator course from course management:
Screen Shot 2022-03-25 at 3 11 55 PM


app/controllers/courses_controller.rb
```diff
@@ -121,4 +123,17 @@ def create_params
     params.require('course').permit(:name)
   end
 
-end
+  def set_course_type
```
elenabaurkot, line 126: I was considering doing this like this instead:
```
    if @course.name.include?("LC")
      # This is the LC Playbook course.
      @lc_mode = true
    end
```

But I didn't know if we should really be using Salesforce to check this instead of just checking if 'LC' is in the course name?

app/controllers/courses_controller.rb
```diff
@@ -121,4 +122,11 @@ def create_params
...
+    if @course.name.include?("LC")
+      # This is the LC Playbook course.
+      @lc_mode = true
+    end
```
elenabaurkot, line 129: Previously, I was using this (used in Capstone Evaluation Submissions controller), but updated to the above instead:
```
    # Determine the Accelerator course for this LC Playbook course.
    # Will return `nil` if this is not an LC Playbook course.
    accelerator_canvas_course_id = SalesforceAPI.client.get_accelerator_course_id_from_lc_playbook_course_id(
      @course.canvas_course_id,
    )

    if accelerator_canvas_course_id.present?
      # This is the LC Playbook course.
      @lc_mode = true
    end

```

Is what I have now OK or should we really be using Salesforce to check this instead of just checking if 'LC' is in the course name? I mostly updated to the way I have it now because it was less code and easier for tests, but not much more needs to be done for tests if the other way is better/preferred.

sadleb: We should use Salesforce. Don't want this breaking if we ever implement publishing to a launched course and and a launched "2025 Spring Braven Accelerator - LCSC" course. 

That said, I don't think we should use the existing `get_accelerator_course_id_from_lc_playbook_course_id` method. When the sync stuff goes live, we're going to start requiring all courses to be associated with a Program. In that PR I added an `is_lc_playbook?` method to `Course`. See [here](https://github.com/bebraven/platform/pull/993/files#diff-434c70e8f67701891105f1af973edd3f69d4cd17338275df30d1ed778a17c16dR54).

I think you should basically copy that code and use it. I can resolve the merge conflict on my next rebase.

sadleb: If there is no salesforce program set on the course, I think we should default to turning off the ability to publish this.

elenabaurkot: Good call - did not think about the fact that literally anything with 'lc' would return true. Updated this to use the is_lc_playbook_course? I had to bring over a couple of the other methods to use that too

sadleb (approved):
lgtm!

PR #1027, opened 2022-04-04
Add basic copy overrides (due dates) feature

rshipp:
# Description

Bare-bones implementation of a new feature to copy Canvas Assignment Overrides (due dates) from one course/section to another course/section.

This PR does not implement the part we talked about for preventing someone from copying dates to a section that has existing dates (though Canvas does raise an error if you try), or to wipe out dates from a section to start fresh.

This PR requires a container **restart** to pull in the new yarn package.

# Motivation and Context

Task: https://app.asana.com/0/1178906757852153/1201634733776799/f

# Screenshots

Copy view:

Screen Shot 2022-04-04 at 3 28 26 PM

Preview view:

Screen Shot 2022-04-04 at 3 26 32 PM

Index view, for local debugging purposes:

Screen Shot 2022-04-04 at 3 31 19 PM


app/javascript/packs/canvas_assignment_overrides.js
```diff
@@ -0,0 +1,114 @@
...
+            toDateElem.value = '';
+            fromDateTextElem.dataset['date'] = null;
+
+            const courseID = fromCourseElem.value;
+            console.log({fromCourseElem, courseID});
+            const apiURL = `/courses/${courseID}/canvas_assignment_overrides.json`;
+            fetch(apiURL).then(response => {
+                response.json().then(canvasAssignmentOverrides => {
+                    const override = canvasAssignmentOverrides.find((o) => {
+                        return o.canvas_section_id == sectionID && o.assignment_name === 'ATTEND: Learning Lab 1';
```
rshipp, line 69: todo: factor this string out somewhere

app/javascript/packs/canvas_assignment_overrides.js
```diff
@@ -0,0 +1,114 @@
...
+    // Dynamic due date based on from_section.
+    fromSectionElem.onchange = (e) => {
+        const sectionID = e.target.value;
+        if (sectionID === '') {
+            // Reset all the dynamic page elements.
+            fromDateTextElem.innerText = 'Please choose a course and section';
+            dateOffsetTextElem.innerText = '0';
+            dateOffsetElem.value = '0';
+            toDateElem.value = '';
+            fromDateTextElem.dataset['date'] = null;
```
rshipp, line 55: i want to factor out this reset stuff somehow, but not sure yet how to do it efficiently since each time we have to reset a slightly different chunk of them

app/models/canvas_assignment_override.rb
```diff
@@ -10,6 +10,44 @@ class CanvasAssignmentOverride < ApplicationRecord
...
+    course_attendance_event&.attendance_event or
+      course_custom_content_version&.custom_content_version&.custom_content or
+      course_rise360_module_version&.rise360_module_version&.rise360_module
+  end
+
+  # Return a hash in the format expected by Canvas.
+  # Only for use with section overrides, not user-level overrides.
+  # See also lib/canvas_api.rb:create_assignment_overrides.
+  def to_canvas_hash
+    {
```
rshipp, line 40: raise if section id is nil

app/policies/canvas_assignment_override_policy.rb
```diff
@@ -0,0 +1,2 @@
+class CanvasAssignmentOverridePolicy < ApplicationPolicy
+end
```
elenabaurkot, line 2: When you have an empty policy is this just because there aren't any actions that require restrictions/permissions?

app/views/canvas_assignment_overrides/index.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.array! @canvas_assignment_overrides, partial: "canvas_assignment_overrides/canvas_assignment_override", as: :canvas_assignment_override
```
elenabaurkot, line 1: I think I am a little confused exactly how this part works. Is this turning the @canvas_assignment_overrides data to json and then displaying each one using the partial `app/views/canvas_assignment_overrides/_canvas_assignment_override.json.builder`? Also, since this view is `index.json` instead of `index.html.erb` does it still align with the `index` action in the `canvas_assignment_overrides_controller`? (and if so, how does it have access to the `@canvas_assignment_overrides` instance variable because `set_canvas_assignment_overrides` method doesn't have a `before_action` for `index`)

app/policies/canvas_assignment_override_policy.rb
rshipp: no this inherits from ApplicationPolicy, which requires admin for everything

app/views/canvas_assignment_overrides/index.json.jbuilder
rshipp: > Is this turning the @canvas_assignment_overrides data to json and then displaying each one using the partial `app/views/canvas_assignment_overrides/_canvas_assignment_override.json.builder`? 

Yes, though I'm just copying this code from elsewhere in the repo so I'm not sure what all of it means either.

> Also, since this view is `index.json` instead of `index.html.erb` does it still align with the `index` action in the `canvas_assignment_overrides_controller`? 

Yes, this is what gets rendered if the requested type was JSON (e.g. if you visit `/courses/1/canvas_assignment_overrides.json` instead of `/courses/1/canvas_assignment_overrides`). But it still uses the same `index` action in the controller.

> (and if so, how does it have access to the `@canvas_assignment_overrides` instance variable because `set_canvas_assignment_overrides` method doesn't have a `before_action` for `index`)

This is related to the commend above that `before_action` - our DryCrud::Nestable concern sets a variable for us with the same name, that gets used in the index. Brian mentioned I should rename the one I'm creating manually to make this less confusing.

app/services/copy_canvas_assignment_overrides.rb
```diff
@@ -0,0 +1,88 @@
...
+      Honeycomb.add_field('copy_canvas_assignment_overrides.new_assignments.count', new_assignments.count)
+
+      # note: @canvas_assignment_overrides has already been dup-shifted, so direct
+      # modification is fine. It's also explicitly *not* saved to the database,
+      # so don't run .save or .update here.
+      translated_overrides = @canvas_assignment_overrides.map do |override|
+        new_assignment = new_assignments.find { |a| a['name'] == override.assignment_name }
+        # Skip assignments we can't find in the new course.
+        unless new_assignment
+          Rails.logger.warn("Skipping canvas_assignment_id=#{override.canvas_assignment_id}, name=#{override.assignment_name}")
```
sadleb, line 32: Add Honeycomb field?

app/services/copy_canvas_assignment_overrides.rb
```diff
@@ -0,0 +1,88 @@
...
+          next
+        end
+
+        # Translate assignment and section IDs.
+        override.canvas_assignment_id = new_assignment['id']
+        override.canvas_section_id = @section.canvas_section_id
+
+        # Convert to the format expected by the Canvas API.
+        override.to_canvas_hash
+      end.compact
```
sadleb, line 42: Note that this is a `filter_map` method that operates in much the same way where you don't have to call `compact` at the end

app/services/copy_canvas_assignment_overrides.rb
```diff
@@ -0,0 +1,88 @@
...
+            override_hash[:assignment_name] = new_assignment&.dig('name')
+            override_hash
+          end
+
+          # We're creating these right now, so no need for an upsert.
+          CanvasAssignmentOverride.insert_all!(override_hashes)
+        end
+      end
+
+      Honeycomb.add_field('copy_canvas_assignment_overrides.bulk_copy_complete?', true)
```
sadleb, line 85: I've generally been using `success?, true` for these types of fields. E.g.  `Honeycomb.add_field('copy_canvas_assignment_overrides.success?', true)`

app/services/copy_canvas_assignment_overrides.rb
```diff
@@ -0,0 +1,88 @@
...
+          # from Canvas.
+          override_hashes = override_hashes.map do |override_hash|
+            new_assignment = new_assignments.find { |a| a['id'] == override_hash[:canvas_assignment_id] }
+            # Use safe accessors so we fall back to nil if something's missing.
+            override_hash[:assignment_name] = new_assignment&.dig('name')
+            override_hash
+          end
+
+          # We're creating these right now, so no need for an upsert.
+          CanvasAssignmentOverride.insert_all!(override_hashes)
```
sadleb, line 81: If we re-run this service for a given course, will this fail? Or will it create dupes?

db/schema.rb
```diff
@@ -98,6 +98,7 @@
     t.date "all_day_date"
     t.datetime "created_at", precision: 6, default: -> { "CURRENT_TIMESTAMP" }, null: false
     t.datetime "updated_at", precision: 6, default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.string "assignment_name"
```
sadleb, line 101: Can this be `nil`? If not, can we add a DB constraint?

app/javascript/packs/canvas_assignment_overrides.js.erb
```diff
@@ -0,0 +1,112 @@
...
+                        const dueDate = new Date(override.due_at).toLocaleDateString();
+                        fromDateTextElem.dataset['date'] = override.due_at;
+                        fromDateTextElem.innerText = dueDate;
+                        return;
+                    }
+
+                    // Override or date is missing.
+                    fromDateTextElem.innerText = 'Unable to find a due date for the assignment in that course/section.';
+                });
+            });
```
sadleb, line 80: Does it make more sense to do this logic in Ruby and have the JS just call something like `/courses/${courseID}/anchor_assignment.json` ? I guess it's nice to have the `canvas_assignment_overrides.json` endpoint in case we want to do more fancy stuff, future enhancements on the JS side so maybe it's not worth trying to make this JS simpler esp since it's already written code.

app/javascript/packs/canvas_assignment_overrides.js.erb
```diff
@@ -0,0 +1,112 @@
...
+                const dateOffset = toDate.diff(fromDate, 'day');
+
+                if (dateOffset) {
+                    // If dateOffset is 0 or NaN, don't set these.
+                    dateOffsetTextElem.innerText = dateOffset;
+                    dateOffsetElem.value = dateOffset;
+                }
+            } catch (e) {
+                console.log(e);
+                dateOffsetTextElem.innerText = 'ERROR!';
```
sadleb, line 106: We're small enough that if this happens the staff member will likely just work directly with us to figure out what's going on and so it doesn't matter if we send to Honeycomb. But I wonder if it's worth adding that error handling anyway just so that folks who copy/mimic this code will know how to properly send errors to Honeycomb?

sadleb (approved):
Nothing merge blocking imo, esp since we still need to write the tests and can take any of the feedback into account in that PR.

db/schema.rb
rshipp: this one's nullable

app/services/copy_canvas_assignment_overrides.rb
rshipp: it'll fail on the canvas line, before it gets here

app/services/copy_canvas_assignment_overrides.rb
```diff
@@ -0,0 +1,88 @@
...
+
+require 'canvas_api'
+
+# Canvas Assignment Overrides represent assignment due dates for different
+# sections/users.
+class CopyCanvasAssignmentOverrides
+
+  class NoTranslatedOverridesError < StandardError; end
+
+  def initialize(course:, section:, canvas_assignment_overrides:)
```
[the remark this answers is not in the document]
rshipp: my opinion is the opposite of this, positional args are almost always less readable. but i changed it anyway

app/services/copy_canvas_assignment_overrides.rb
rshipp: this is inside a loop, what honeycomb field are you suggesting?

sadleb (approved):

PR #1035, opened 2022-04-07
Enhance add_alert() to send a Slack channel it's for.

sadleb:
# Description
Enhance `add_alert()` to be able to send the Slack channel it's targeting. Also add a `add_support_alert()` convenience method for alerts that Product Support should be aware of but engineers can generally ignore. By default, `add_alert()` goes to the engineering alerts channel

# Motivation and Context
We want top-level catch-all triggers for any new alert added, but we also want to be able to distinguish which Slack channel they are intended for. It's too hard to maintain the alert -> channel mapping using the existing `alert_name` and `alert_class`. Instead, developers should explicitly choose if they want to send an alert that's intended for Product Support, otherwise it defaults to the engineering alerts which engineers should dig into / fix.

# Screenshots
image


app/services/clone_course.rb
```diff
@@ -105,7 +105,7 @@ def update_salesforce
         { salesforce_field_name => @destination_course.canvas_course_id }
       )
     else
-      Honeycomb.add_alert('unknown_course_type',
+      Honeycomb.add_support_alert('unknown_course_type',
```
sadleb, line 108: I went through all the places we call `add_alert` at the moment, which is just sync code for now, and took my best guess at whether it should go to engineers or just product support. The changes below are the only ones I though were product support only worthy at the moment.

Note: We have a task to cutover all alerts to this framework: https://app.asana.com/0/1201131148207877/1202052488545511

elenabaurkot (commented):
lgtm 👍

PR #1038, opened 2022-04-08
Don't fail if a Zoom Meeting ID is entered with whitespace

sadleb:
# Description
It's easy to copy/paste the Meeting ID into the Cohort Schedule with whitespace from the UI instead
of without it from the URL. Let's just handle that by stripping whitespace from it. Before this, the error would be `URI::InvalidURIError`.

# Motivation and Context
Got a bunch of alerts about this failing and it wasn't immediately clear what the problem was. Instead of raising, we can just handle it seamlessly.

# Screenshots


rshipp (approved):

elenabaurkot (commented):
lgtm 👍

PR #1041, opened 2022-04-08
Clean Attendance Sections dropdown - sort and exclude empty ones

sadleb:
# Description
If you're an admin or have `CanTakeAttendanceForAll` role, you can select which Section to take attendance for. However, old Sections with no users left assigned a role to them still show up. This excludes those sections from the list

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202091130082329

# Screenshots
Before `C2 Tue Virtual()` had no users:
Take Attendance For Section

After:
image


app/models/section.rb
```diff
@@ -33,6 +33,21 @@ class Type
   scope :cohort_or_cohort_schedule, -> { where(section_type: [Type::COHORT, Type::COHORT_SCHEDULE]) }
   scope :teaching_assistants, -> { where(section_type: Type::TEACHING_ASSISTANTS) }
 
+  # Filter down to only sections with at least one User assigned a role for it.
+  scope :with_users, -> { joins(:roles).distinct }
```
sadleb, line 37: Works b/c an `INNER JOIN` doesn't return rows with no association.

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -139,7 +139,8 @@ def set_course_attendance_event
 
   def set_all_attendance_sections
     if current_user.can_take_attendance_for_all?
-      @all_attendance_sections = Section.cohort_or_cohort_schedule.where(course: @accelerator_course)
+      @all_attendance_sections = Section.cohort_or_cohort_schedule.with_users
+        .where(course: @accelerator_course).order_by_name
```
sadleb, line 143: Query ends up looking something like this:

```
SELECT DISTINCT "sections".* FROM "sections" 
  INNER JOIN "roles" ON "roles"."resource_type" = 'Section'
  AND "roles"."resource_id" = "sections"."id" 
WHERE "sections"."section_type" IN ('cohort', 'cohort_schedule') 
  AND "sections"."course_id" = 2
```

app/models/section.rb
```diff
@@ -33,6 +33,21 @@ class Type
...
+  # From https://stackoverflow.com/a/25042119/12432170.
+  scope :order_by_name, -> {
+    sort_by{ |section|
+      section.name.gsub(/\d+/) { |s|
+        # Left-pad numbers with zeroes.
+        # 8 is arbitrarily longer than the numbers we'll see in titles.
+        "%08d" % s.to_i
+      }
+    }
+  }
```
sadleb, line 49: Copied from a similar `order_by_title` in `models/course_attendance_event.rb`

app/models/section.rb
sadleb: TODO: look up uniq vs distinct. Uniq is ruby in memory.

PR #1048, opened 2022-04-13
T-1 Fix error message to be more friendly to user on security_error.h…

joe-cena-git:
# Description
This Pull Request fixes the user error message to be more friendly to the user.
It removes the "Security Error" heading and changes the wording of the error message slightly.

# Motivation and Context
Task: [T-1 As Participant, I get scared when I see Security Error and may not see the instructions to re-launch the Discord Assignment](https://app.asana.com/0/1201916906514465/1201387757929170) 

# Screenshots
Not sure how to reproduce this error, but the markup is a simple update.


app/views/errors/security_error.html.erb
```diff
@@ -1,5 +1,4 @@
-<h1>Security Error</h1>
-<p>Sorry, something's gone wrong. Try returning to your Canvas dashboard and relaunching the page from there.</p>
+<p>Sorry, something's gone wrong. Try returning to your Canvas assignment and relaunching the page from there.</p>
```
sadleb, line 1: When I originally proposed the solution of just removing the heading, I didn't fully consider accessibility. In other words, what the experience for someone using a screenreader would be like. It's generally bad to have no headings on a page. Given that, I think a better solution would be:
```
<h1>Sorry, something's gone wrong</h1>
<p>Try returning to your Canvas assignment and relaunching the page from there.</p>
```

joe-cena-git: Added header in latest commit

sadleb (approved):
lgtm.

nit: remove `T-1` from PR title?

PR #1049, opened 2022-04-13
T-2 fix spelling on the word permission in discord sign-up error page

joe-cena-git:
# Description
This PR fixes a misspelled word on the discord sign-up error page.

# Motivation and Context
Task: [T-2 As Platform Admin, I don't want to see an error Message with a misspelled word when accessing the Discord assignment with the wrong enrollment type](https://app.asana.com/0/1201916906514465/1202057432317451)

# Screenshots
TODO: test and submit screenshots


sadleb (approved):
Lgtm. What’s the deal with the T-2 prefix though? I don’t think that’s going to be meaningful when browsing PR titles looking for something. If the intention is to start giving all tasks a unique ID, we should talk more. Not sure if just naming each successive task with a new number is maintainable.

PR #1050, opened 2022-04-13
T-3 Add Atlanta to list of regions on Join The Braven Network page

joe-cena-git:
# Description
When signing up on the "Join The Braven Network" page, Atlanta was missing as a region/area. This PR adds Atlanta.

# Motivation and Context
Task: [T-3 As a prospective volunteer, I want to see all Braven sites, including Atlanta,  in the Braven Network dropdown](https://app.asana.com/0/1201916906514465/1202068192762824)

# Screenshots
TODO: test and upload screenshots before/after


app/views/champions/new.html.erb
```diff
@@ -73,7 +73,7 @@
     <fieldset>
       <legend class="radio-group-legend">All volunteer opportunities are currently virtual. However, Braven knows the power of local connections. If you would like to be informed of volunteer opportunities for a particular region, please indicate which one below. If you have no preference, select National.</legend>
       <div class="form-group">
-        <% areas = ['Newark, NJ', 'New York City, NY', 'San Francisco Bay Area, San Jose', 'Chicago', 'National'] %>
+        <% areas = ['Atlanta', 'Newark, NJ', 'New York City, NY', 'San Francisco Bay Area, San Jose', 'Chicago', 'National'] %>
```
sadleb, line 76: In `./app/models/champion.rb` we use this value to set `contact['BZ_Region__c']` (contact is the Salesforce Contact). We need to make sure the value `Atlanta` matches the region in Salesforce. Does it?

sadleb: Update: it does match the expect Salesforce value.

sadleb (approved):
lgtm

PR #1052, opened 2022-04-14
Include LC Capstone evals when grading capstone eval teamwork grade

elenabaurkot:
# Description
In this [Capstone Evaluations grade breakdown](https://github.com/bebraven/platform/pull/991) PR we added the ability to show who has and has not completed Capstone Evaluations and to calculate Capstone Evaluation teamwork grades. That PR only accounts for fellow Capstone Evaluations, not the LC Capstone Evaluations. This update incorporates LCs' evaluations into the grade too and displays LC names' for who has or has not completed their evaluations. 

# Motivation and Context
To make sure LC Capstone Evaluations are incorporated into the teamwork grade. 

Task: https://app.asana.com/0/1201131148207877/1202124220243067/f

# Screenshots
Includes the Leadership Coach evaluation ("test lc" is shown under "Users who have submitted Capstone Evaluations that are published")
Screen Shot 2022-04-14 at 3 24 45 PM


app/models/course.rb
```diff
@@ -87,6 +87,16 @@ def students
     sections.map { |s| s.students }.flatten.uniq
   end
 
+  def lcs
+    users = sections.map { |s| s.users }.flatten.uniq
+    lcs = users.select do |user|
+      candidate = HerokuConnect::Candidate.where(contact__c: user.salesforce_id, program: self.salesforce_program_id).first
```
elenabaurkot, line 93: Should there only be one matching Candidate for a given contact in a given program? I am just grabbing the first one found here because I was assuming there would only be one match, but wanted to double check this.

app/models/course.rb
```diff
@@ -87,6 +87,16 @@ def students
...
+      if candidate
+        candidate.is_lc?
```
elenabaurkot, line 95: I added the `if candidate` here because I didn't know if there will always be a candidate for a user (I think in production yes, but I know some of my test users in my dev env don't have candidates)

app/models/course.rb
rshipp: assuming candidates follow the same rules as participants, which I think is the case, there may be multiple candidates per contact, even in the same course / with the same role. there *shouldn't* be, but this does happen sometimes due to mistakes. i think we should try to handle that here, since we've had issues in the past (like when MIV participants messed things up) with arbitrarily selecting the "first" result for a participant.

app/models/course.rb
rshipp: seems like a good idea

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -33,7 +33,9 @@ def launch
   def score
     authorize :CapstoneEvaluationSubmission
 
+    # Merges ungraded CapstoneEvaluationSubmissions for Accelerator course and LC course
     new_capstone_eval_submissions = @course.capstone_evaluation_submissions.ungraded
+      .or(@lc_course.capstone_evaluation_submissions.ungraded)
```
rshipp, line 38: i like this syntax

app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -44,13 +46,17 @@ def score
   end
 
 private
-  def set_course
+  def set_courses
     @course = Course.find_by!(canvas_course_id: @lti_launch.course_id)
+
+    sf_program = HerokuConnect::Program.find(@course.salesforce_program_id)
+    @lc_course = Course.find_by!(canvas_course_id: sf_program.canvas_cloud_lc_playbook_course_id__c)
```
rshipp, line 53: i think we might want to break this up into `set_course` and `set_lc_course`, rather than setting two variables in one. when you list `before_action` callbacks in the top of the controller, they are called in order, so you should be able to safely use `@course` in `set_lc_course`.

app/controllers/capstone_evaluation_results_controller.rb
sadleb: Ditto. Did not know about this option.

app/models/course.rb
sadleb: I think you can do something similar to `students` where you use a scope on the section to grab folks with a `TaEnrollment` - then we can pudo y have to worry about the Salesforce stuff and multiple Candidates. Just rely on the sync to get LCs into the section properly (with a TaEnrollment). Actual Teaching Assistants and Staff don’t go into these sections, so I think it’s safe

app/services/grade_capstone_evaluations.rb
```diff
@@ -98,24 +99,27 @@ def user_has_grade?(current_user)
...
   end
 
-  def students_with_published_submissions
+  def users_with_published_submissions
     all_capstone_eval_submissions.filter_map { |s| s.user if !s.new }
   end
 
-  def remaining_students
-    @course.students.uniq - all_capstone_eval_submissions.map { |s| s.user }
+  def remaining_users
```
sadleb, line 106: We could probably create a `students_and_LCs` scope that returns this list. This is fine too if that turns out to be too much work.

sadleb (commented):
Some small comments. The most important one is that Candidate thing which I think we can get rid of completely

app/models/course.rb
sadleb: This concern goes away if we use the section to get LC submissions instead of looking them up from Salesforce data. See above.

app/models/course.rb
```diff
@@ -5,6 +5,12 @@ class Course < ApplicationRecord
   scope :unlaunched_courses, -> { where is_launched: false }
 
   has_many :sections
+  has_many :roles, :through => :sections
+  has_many :students, -> { where("roles.name = ?", RoleConstants::STUDENT_ENROLLMENT).distinct }, through: :roles, :source => :users
+  has_many :students_and_lcs, -> { where("roles.name = ?", RoleConstants::STUDENT_ENROLLMENT)
+    .or(where("roles.name = ?", RoleConstants::TA_ENROLLMENT))
+    .and(where("sections.name != 'Teaching Assistants'"))
+    .distinct }, through: :roles, :source => :users
```
elenabaurkot, line 13: I am filtering out the teaching assistants section since I only want to grab LC and Fellow users from the course, should I be filtering out the "Default Section" as well?

spec/services/grade_capstone_evaluations_spec.rb
```diff
@@ -244,37 +259,41 @@
...
 
-  describe '#remaining_students' do
-    let!(:graded_capstone_evaluation_submission) { create(:graded_capstone_evaluation_submission,
-      course_id: course.id,
-      user_id: fellow_with_already_graded_submission.id
-    ) }
+  describe '#remaining_users' do
+    let!(:fellow_with_no_submission) { create(:fellow_user, section: section) }
+    # Creates a TA user in a student section in the Accelerator, which makes them an LC
+    let!(:lc_with_no_submission) { create(:ta_user, section: section) }
```
elenabaurkot, line 283: Is this correct? (Any TAs in a student section in an Accelerator course is an LC because actual TAs would be in the "Teaching Assistants" section?)

app/models/course.rb
sadleb: There is a `section_type` column. I’d use that. To answer your question though, we’re only interested in section_type of `cohort` The other three types should be excluded. There is a constant in the `Section` model. So something like:

```
where("roles.name = ?", RoleConstants::STUDENT_ENROLLMENT)
    .or(where("roles.name = ?", RoleConstants::TA_ENROLLMENT))
    .and(where("sections.section_type == Section::Type::COHORT'"))
    .distinct
```

spec/services/grade_capstone_evaluations_spec.rb
sadleb: I’d make an `lc_user` factory. But yes, it’s the exact same role as a `ta_user`

app/models/course.rb
elenabaurkot: Should I include section_type of cohort_schedule too? I was going to use the scope from Section, which filters on both of them (`scope :cohort_or_cohort_schedule, -> { where(section_type: [Type::COHORT, Type::COHORT_SCHEDULE]) }`)
```
has_many :students_and_lcs, -> { where("roles.name = ?", RoleConstants::STUDENT_ENROLLMENT)
    .or(where("roles.name = ?", RoleConstants::TA_ENROLLMENT))
    .merge(Section.cohort_or_cohort_schedule)
    .distinct }, through: :roles, :source => :users
```

spec/services/grade_capstone_evaluations_spec.rb
elenabaurkot: I updated to make an lc_user factory, I just wanted to double check did you mean to use a cohort_section for the lc_users's section in this spec, or I should do that in the actual factory?

app/models/course.rb
sadleb: Yeah, that’s fine. It shouldn’t really happen that an LC is in a cohort schedule but not mapped to a cohort by this point in the semester, but I don’t think it hurts if we grab any that are in that situation

spec/services/grade_capstone_evaluations_spec.rb
sadleb: Nice! Re: `cohort_section` i meant that instead of `let(:section) { create :section … ` do `let(:section) { create :cohort_section … `

rshipp (approved):
lgtm

PR #1054, opened 2022-04-18
Sentry: set the User context so we can search for specific user exceptions

sadleb:
# Description
When an exception is sent to Sentry, add the user context to the information sent.

# Motivation and Context
Task: https://app.asana.com/0/1201131148207877/1202138912499633

This will make it easier to look up exceptions that apply to a particular
user when troubleshooting. E.g. in the Sentry search field you can now
do something like: `user.email:some@email` or `user.id:XX` to find exceptions
for a user.

# Screenshots
All the new tags added
image

Searching for exceptions by user email:
image

Searching for exceptions by user id:
image

The user associated with an exception
image


rshipp (approved):
seems good

PR #1057, opened 2022-04-21
Redis: speed up container restarts by truncating the Redis AOF file

sadleb:
# Description

See here for info about Append-only file persistence which we use
to make Sidekiq work across reboots:
https://redis.com/ebook/part-2-core-concepts/chapter-4-keeping-data-safe-and-ensuring-performance/4-1-persistence-options/4-1-2-append-only-file-persistence/

See here for some config options, which are supposed to rewrite/compact
the file when it gets too big:
https://redis.com/ebook/part-2-core-concepts/chapter-4-keeping-data-safe-and-ensuring-performance/4-1-persistence-options/4-1-3-rewritingcompacting-append-only-files/

However, there seems to be a bug with Docker restarts where the config
options don't work reliably b/c the base size of the file is set to what
it is on restart which is the uncompacted size. More info here:
https://github.com/sourcegraph/sourcegraph/issues/3300

This PR just manually send the `BGREWRITEAOF` command to redis on restart
telling it to manually run the rewrite/compaction.

Note: the bash syntax is a bit confusing, but it's just sending that
command to redis using a plain ol' tcp socket instead of needing to
install redis-cli in the Ruby container.

# Motivation and Context
Container restarts get slower and slower, pretty quickly for me. It's annoying to run
```
docker-compose down
docker volume rm platform_redis-persistent
```

and deal with the error I get requiring you to temporarily create and use a `docker-compose-redis.yml`.

Task: 

# Screenshots


elenabaurkot (commented):
lgtm 👍

PR #1058, opened 2022-04-21
Optimize SalesforceAPI calls

sadleb:
# Description
The POST to Salesforce to get the oauth token now accounts for a significant
portion of the time spent during a sync. This change just makes it so that the
token is only gotten once after a server restart. Since our server restarts once
a day, that's the oldest the token can ever be. I tested that it doesn't change
and remains valid for at least that period of time, so this should be safe.

This trace should basically flatline after this change:
https://ui.honeycomb.io/braven/datasets/platform.braven.org/result/HEcxMG3SQuB

# Motivation and Context
Task: https://app.asana.com/0/1201131148207877/1201661063922393

# Screenshots


rshipp (approved):
nice

elenabaurkot (commented):
Lgtm 👍

PR #1060, opened 2022-04-21
Fix publishing Modules to Canvas by setting content-type to empty

elenabaurkot:
# Description
Publishing modules to Canvas broke because the content-type of uploads to AWS S3 started setting `content-type` to binary/octet-stream by default, which were causing modules to download files instead of properly loading when opened. To prevent the content-type from defaulting to binary/octet stream, we now explicitly set it to an empty string.

# Motivation and Context
To enable Designers to publish new modules to Canvas.

Task: https://app.asana.com/0/1201131148207877/1202091130082326/f

# Screenshots
Previous content-type for AWS S3 uploads defaulting to binary/octet-stream:
Screen Shot 2022-04-21 at 2 56 42 PM

Content-type now set to empty for AWS S3 uploads:
Screen Shot 2022-04-21 at 2 58 45 PM


rshipp (approved):

rshipp:
this is a great solution, idk if i would have thought to try that

elenabaurkot:
> this is a great solution, idk if i would have thought to try that

Thanks @rshipp I tried a lot of other things before getting here lol

sadleb:
> > this is a great solution, idk if i would have thought to try that
> 
> Thanks @rshipp I tried a lot of other things before getting here lol

Ditto. I would not have thought of this. Solid solution!

PR #1062, opened 2022-04-27
Refactor Honeycomb alerts to use add_alert

rshipp:
# Description

Change `add_field('alert...` to `add_alert`, and add a relevant message as appropriate. I don't think any of these are product support alerts, but we can modify as needed to use the other function if they end up being.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202052488545511/f


elenabaurkot (commented):
This looks good to me 👍

PR #1063, opened 2022-04-28
Only include Capstone Evals for current course in Teamwork grade

elenabaurkot:
# Description
This is an update to the PR [Include LC Capstone evals when grading capstone eval teamwork grade](https://github.com/bebraven/platform/pull/1052). While QA'ing that task, it was realized that when grading the Capstone Evaluation submissions using the grading service to calculate Fellows' teamwork grade, if a Fellow had Capstone Evaluations submitted for them in another course, it would also factor those scores into their grade for the Course we were scoring. This update makes it so that only submissions for the course we are creating grades for (and submissions from the LC partner course) are included in a Fellow's teamwork grade. This update also addresses symbols that were appearing in the Canvas grade book for creating multiple submissions for an assignment. When we score Fellows' teamwork grade, we have to make a submission for them for the assignment so they can view their grade breakdown in the submission details of the grading tab. Previously, every time Fellows were regraded it made a new Canvas submission for the assignment, but now we only create one if they don't already have a submission. This makes it so we always see the Fellow's score from the grade book without having to click on the symbol marking that it was a resubmission.

# Motivation and Context
Update grades to ensure Fellows are receiving the correct Capstone Evaluation teamwork grade.

Task: https://app.asana.com/0/1201131148207877/1202124220243067/f


app/services/grade_capstone_evaluations.rb
```diff
@@ -18,6 +18,11 @@ def run
       .uniq
 
     grades = {}
+
+    # Returns all previous submissions for the assignment in the format:
+    # { canvas_user_id => submission }
+    previous_submissions = CanvasAPI.client.get_assignment_submissions(@course.canvas_course_id, @lti_launch.assignment_id, true)
```
rshipp, line 24: thanks for this comment, super helpful

sadleb (commented):
lgtm

PR #1064, opened 2022-04-28
Update sidekiq-scheduler

rshipp:
# Description

Upgraded sidekiq-scheduler, and verified both that the warning went away and that the scheduler/herokuconnect still worked.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1201999845627792/f


elenabaurkot (commented):
Lgtm 👍

PR #1065, opened 2022-05-02
Content Editor: Remove notice that makes it hard to use the page

rshipp:
# Description

Remove the green notice/flash/alert bar at the top of the page after saving in the content editor. 

I tried to find another way to do this that would still provide visual feedback that the page had saved, but couldn't come up with anything good. Some examples:

* Tried to make the alert push the page down instead of being on top (z-index) of everything... could probably get it eventually but even if I did, it would still refresh the page and lose work if you didn't click it immediately.
* Looked into inline dismissal, but the bootstrap alert JS requires jquery, and we don't want to enable that in our app and make it easier for people to accidentally start doing jquery stuff. Opens up a whole can of worms.
* Looked into rewriting the jquery inline dismissal in vanilla JS, but it's pretty complicated and probably not worth the effort.
* Thought about other ways to style the alert, but even if I could come up with a good fix, it's in the application layout, which means changing it here would impact every other page, and create a bunch of extra work.

In the end, I think the best solution in this case is just getting rid of this notice from the content editor so you don't accidentally save the page, do some work, click the X, and lose all your work. The downside is that it's harder to tell if the editor saved after you click Save, but we can just inform designers of this.

# Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200393036057897/f


elenabaurkot (commented):
Based on your description, this seems like a good solution to me 👍

sadleb:
We could do something janky, like pass a custom query param instead of `notice` and use that to put a homespun alert just for this page instead of in the application layout. Something like
```
<div class="alert">
  <span class="closebtn" onclick="this.parentElement.style.display='none';">&times;</span>
  Saved successfully!
</div>
```

I'm fine with just removing it though. I expect that if designers are worried it didn't save, they'll just refresh to make sure.

Actually, last idea... is there a way to detect if there have been changes since the editor was loaded? If so, we could default the Save button to be disabled and only enable it once they start making changes. That could serve as a visual cue b/n a saved vs unsaved state.

rshipp:
Do you think it's worth spending more time on this? I can come up with a better solution with more time, I just didn't think it was important enough to justify that.

sadleb:
> Do you think it's worth spending more time on this? I can come up with a better solution with more time, I just didn't think it was important enough to justify that.

I'd say no, let's not spend more time on this at the moment. Instead, let's do what you have and see how it works out in practice. We can prioritize introducing something to indicate save success in the future if designers actually need it after using it if it causes problems.

sadleb (approved):
lgtm

PR #1071, opened 2022-05-10
Capstone Evaluation grading skips failing users

elenabaurkot:
# Description
Previously, when the Capstone Evaluation grading service ran, if one user failed the entire grading service would fail. Now if a user fails it skips that user and sends an alert to Honeycomb. The Honeycomb alert triggers an alert in the #alerts-eng-highlander Slack channel to let us know the Capstone Evaluation grading service failed for one or more users and then continues on to grade the next user. 

# Motivation and Context
To make sure Capstone Evaluation grades don't fail for everyone if one user is failing. 

Task: https://app.asana.com/0/1201131148207877/1202237512873757/f

# Screenshots
Alert sent to Slack #alerts_eng-highlander channel for failing users
Screen Shot 2022-05-10 at 10 14 30 AM


app/services/grade_capstone_evaluations.rb
```diff
@@ -25,25 +25,33 @@ def run
...
-        # Create a submission for the Capstone Evaluations Teamwork assignment for the user if they don't already have one
-        # we are giving a grade, so they are able to see the grade breakdown from grades
-        unless previous_submissions[user.canvas_user_id]
-          CanvasAPI.client.create_lti_submission(
-            @course.canvas_course_id,
-            @lti_launch.assignment_id,
-            user.canvas_user_id,
-            launch_capstone_evaluation_results_url(protocol: 'https')
-          )
+      begin
```
sadleb, line 28: nit: you don't need a `begin` here. `rescue` can be used with any block, so having it below without this `begin` just says "rescue anything that raises inside the `users_to_grade.each do |user|` block"

sadleb (approved):
lgtm

app/services/grade_capstone_evaluations.rb
elenabaurkot: I edited this to get rid of the begin, one question about where I have `Sentry.set_user({})` (line 57) - should this be inside the loop right after line 46 to reset it before looping to the next user or is it OK where it is? @sadleb

sadleb: It's good where it is. `user.add_to_honeycomb_span('grade_capstone_evaluation')` sets the user context in Sentry for each loop. The `Sentry.set_user({})` at the end is just to unset the most recent user context since we're done processing each user.

elenabaurkot: ah okay got it thanks!

PR #1077, opened 2022-05-13
Add calling class/method info to RestClient Honeycomb instrumentation

sadleb:
# Description
We've had some noisy alerts for `RestClient::NotFound` that are expected
and handled in the code but we don't have a good way to tune our triggers
to ignore them since they go into an `error` and `error_detail` field that we
generally want alerts for. [Example here](https://bebraven.slack.com/archives/C01U1T16Y4A/p1652344291465939).

We've been using `restclient.request.url` and `restclient.response.body` when
we can in order to tune, but it's a little messy and sometimes not enough.
This adds the method being called in whatever API wrapper class
we're using (e.g. `CanvasAPI`) as well as the parent consumer class and
method to Honeycomb. See screenshot below

# Motivation and Context
Follow-up to the task below which we ended up solving by removing bad data:

Task: https://app.asana.com/0/1201131148207877/1202149718442972

I could also see us doing future tuning / optimization using these fields since it
easily let's us see the most expensive API methods and what is calling them.

# Screenshots
Sample of what these fields look like. Taken from running Module grading with a bad Canvas user
image

DiscordBot example
image


lib/restclient_instrumentation.rb
```diff
@@ -17,15 +17,37 @@ def execute &block
...
+          parent_calling_location = cl
+          break
+        else
+          # keep updating the caller to the deepest stackframe in that file
+          # so we get the actual method being called in that file, not the helper
+          # get/put/patch or retry_timeut stuff
+          calling_location = cl
+        end
+      end
+    end
```
sadleb, line 40: Below is an example of the `calling_location` for each stack-frame in this loop. These are [`Thread::Backtrace::Location` objects](https://ruby-doc.org/core-2.5.0/Thread/Backtrace/Location.html)
```
### index 2: caller_location: path = /app/vendor/bundle/ruby/3.0.0/gems/rest-client-2.1.0/lib/restclient.rb, label = 'get', base_label = 'get'
### index 3: caller_location: path = /app/lib/canvas_api.rb, label = 'block in get', base_label = 'get'
### index 4: caller_location: path = /app/lib/canvas_api.rb, label = 'retry_timeout', base_label = 'retry_timeout'
### index 5: caller_location: path = /app/lib/canvas_api.rb, label = 'get', base_label = 'get'
### index 6: caller_location: path = /app/lib/canvas_api.rb, label = 'get_latest_submission', base_label = 'get_latest_submission'
### index 7: caller_location: path = /app/app/services/grade_rise360_module_for_user.rb, label = 'canvas_submission', base_label = 'canvas_submission'
```

As you can see, we hit the parent calling class at index == 7 of the stacktrace for this example. I made the loop go to a max of 10 to be a little safer in case we add more logic to an API class or method that ends up nesting the original API method a bit deeper than this. I also defensively don't assume we found the `parent_calling_location` so it wouldn't raise an error if we didn't find it, those fields would just be empty.

Lastly, I tested this in the Rails console where the stacktrace is less than 10 deep just to make sure `caller_locations(2,10)` wouldn't fail in that scenario.

sadleb: Just tested it. Doesn't work properly but also doesn't break.

PR #1081, opened 2022-05-16
Add Generate Test Users Feature

elenabaurkot:
# Description
This update is meant to make it easier to generate test users in development and production. A new "Generate test users" button is added to the Platform Administration Dashboard that brings you to a form to add your test user information. Users have the ability to create Fellows, LCs, TAs, CPs, Staff or Faculty users. For Fellows, LCs and CPs users have the option of adding a cohort schedule (required for them to sync properly) and cohort section. There is also an option to select a TA assignment for Fellow users. Users have the ability to clone the test users they are creating and edit the fields or add empty users. Once the button is clicked to "Generate Test Users" it kicks off a job that kicks off the generate_test_users_job that kicks of the generate_test_users_service. When the service is done running it will send a success or failure email. The success email has the Canvas signup links (unless the service was unable to retrieve signup token from the Salesforce contact). The failure email contains any errors, failing users and any users that may have been successfully created. 

# Motivation and Context
The motivation is to make the test user creation process easier in production and development.

# Screenshots
Platform Dashboard View, 'Generate Test Users' option added at the bottom:
Screen Shot 2022-05-16 at 4 06 37 PM

Test User Creation Form:
Screen Shot 2022-05-16 at 4 09 00 PM

Test User Creation Form (view for Fellows, LCs and CPs also has cohort schedule and cohort section options, the cohort section option only shows once a cohort schedule has been selected- Fellows have the option to select a TA Assignment as well):
Screen Shot 2022-05-16 at 4 09 15 PM

Success Email:
Screen Shot 2022-05-16 at 4 15 25 PM

Failure Email:
Screen Shot 2022-05-16 at 4 17 49 PM

Failure Email with multiple types of failures:
Screen Shot 2022-05-16 at 4 17 12 PM


Gemfile.lock
```diff
@@ -131,7 +131,7 @@ GEM
       xpath (~> 3.2)
     childprocess (4.1.0)
     coderay (1.1.3)
-    concurrent-ruby (1.1.10)
+    concurrent-ruby (1.1.9)
```
sadleb, line 134: This stuff looks like a mistake. Maybe you didn’t rebase / rebuild off the `main` branch?

app/assets/stylesheets/test_users.scss
```diff
@@ -0,0 +1,4 @@
+label.required::after {
```
sadleb, line 1: I think we should target a specific `div` in the test users HTML. In dev, it’s possible that things work as expected, but in prod all the CSS is bundled together and put on all pages with the admin layout, meaning all required labels would get this applied. It’s prob fine since there likely aren’t any others, but I think it’s better to be safe in the PR and only target the specific divs we care about

Gemfile.lock
elenabaurkot: I was a little confused here, I didn't think I made these changes, but when I copy and pasted it back to the Gemlock without the changes there were still different changes so I wasn't sure what to go with. I rebased last week during one of our working sessions, but maybe I just need to rebase again? @sadleb

app/assets/stylesheets/test_users.scss
elenabaurkot: Ah okay that makes sense, updated to target a specific div here 👍

Gemfile.lock
elenabaurkot: I rebased and then copied and pasted in the Gemfile.lock from main and I think it's good now

app/assets/stylesheets/layouts/admin.scss
```diff
@@ -11,6 +11,7 @@
  * It is generally better to create a new file per style scope.
  *
  */
+ //*= require test_users
```
rshipp, line 14: instead of linking it into the layout, can you include this directly in the page with `stylesheet_link_tag`? that way the css specific to this page doesn't apply to every page using the admin layout

elenabaurkot: That makes sense, will update to that 👍

app/assets/stylesheets/test_users.scss
```diff
@@ -0,0 +1,6 @@
+div.test-user-form {
+  label.required::after {
+    content: ' *';
+    color: red;
```
sadleb, line 4: Can we use a brand color instead of red? It doesn't really matter since this is internal, but I'd rather always use a brand color so we don't start polluting the codebase with out of the box colors if people copy/paste.

The are in: `app/assets/stylesheets/shared/_variables.scss`. The standard `$bv-red` is probably good.

app/controllers/test_users_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+class TestUsersController < ApplicationController
+  before_action :get_program_select, only: [:index]
+  before_action :set_env, only: [:index, :post, :get_program_tas]
+
+  layout 'admin'
+
+  def index
+    authorize :TestUser
+    @role_options = ["Fellow", "Leadership Coach", "Teaching Assistant", "Coach Partner", "Staff", "Faculty"]
+      .map { |role| [role, role.split(" ").join] }
```
rshipp, line 13: took me a second to understand what this was doing, i think a more idiomatic way would be with gsub:
```suggestion
      .map { |role| [role, role.gsub(" ", "")] }
```

app/controllers/test_users_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+class TestUsersController < ApplicationController
+  before_action :get_program_select, only: [:index]
+  before_action :set_env, only: [:index, :post, :get_program_tas]
+
+  layout 'admin'
+
+  def index
+    authorize :TestUser
+    @role_options = ["Fellow", "Leadership Coach", "Teaching Assistant", "Coach Partner", "Staff", "Faculty"]
```
sadleb, line 12: nit: using the values from `SalesforceConstants::Role` would make it so that if we changed or added a role in the future, it would be more likely the dev would find this place in the code that would need to be changed as well.

app/controllers/test_users_controller.rb
sadleb, line 13: I don't get what this is doing, but if you use the constants as mentioned in the comment above you'll have to do a `to_s` b/c they are symbols, not strings.

app/controllers/test_users_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def post
+    authorize :TestUser
+    GenerateTestUsersJob.perform_later(current_user.email, params.to_unsafe_h, @env)
+    redirect_to generate_test_users_path, notice: 'The generation process was started. Watch out for an email'
+  end
+
+  def cohort_schedules
+    authorize :TestUser
+    cohort_schedules = HerokuConnect::Program.where(sfid: params[:id]).first.cohort_schedules
```
sadleb, line 24: I think `HerokuConnect::Program.find(sfid: params[:id])` is better than `HerokuConnect::Program.where(sfid: params[:id]).first`

app/controllers/test_users_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    render :json => cohort_schedules
+  end
+
+  def cohort_sections
+    authorize :TestUser
+    cohorts = HerokuConnect::CohortSchedule.where(sfid: params[:id]).first.cohorts
```
sadleb, line 30: Ditto on the above comment. If you expect one record, just use `find` instead of `where(...).first`

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    raise unless @sync_error.empty?
+  end
+
+  def generate_test_users()
+    Rails.logger.info('Started generating test users')
+
+    num_of_users = @params['email'].count
+    for user in 0...num_of_users
```
rshipp, line 48: could you rename `user` to something that makes it clearer what the variable represents? i can't tell from looking at this, is it like the index into the params array that got passed in from the frontend? i'd normally expect a variable called `user` to represent a `User` object.

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+    new_test_users = generate_test_users()
+
+    unless @failed_users.empty?
+      Rails.logger.error(@failed_users.inspect)
+      msg = "Failed to generate test users. The following users failed => #{@failed_users.inspect}"
+      Honeycomb.add_alert('generate_test_users.failed_users', msg)
+      raise
+    end
+
+    raise unless @sync_error.empty?
```
sadleb, line 41: Can we wrap this in a named Honeycomb span? E.g. We haven't been good about this in the past, but going forward I think the `run` method of all our services should have a dedicated HC span. See `sync_salesforce_participant.rb` for an example.

sadleb, line 41: This is internal test code, so it doesn't matter too much, but it will make it hard to troubleshoot if you just use `raise` instead of raising something that gives an indication of what and where the problem is. E.g. here is an equivalent for what I do in `sync_salesforce_program.rb`:
```
raise SyncParticipantsError, "Some participants failed to sync => #{@failed_participants.inspect}"
```

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+
+  def run
+    Honeycomb.add_field('generate_test_users.users.count', @params['email'].count)
+    new_test_users = generate_test_users()
+
+    unless @failed_users.empty?
+      Rails.logger.error(@failed_users.inspect)
+      msg = "Failed to generate test users. The following users failed => #{@failed_users.inspect}"
+      Honeycomb.add_alert('generate_test_users.failed_users', msg)
+      raise
```
sadleb, line 38: This is internal test code, so it doesn't matter too much, but it will make it hard to troubleshoot if you just use `raise` instead of raising something that gives an indication of what and where the problem is. E.g. here is an equivalent for what I do in `sync_salesforce_program.rb`:
```
raise SyncParticipantsError, "Some participants failed to sync => #{@failed_participants.inspect}"
```

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+    @params = params
+    @env = env
+    @user_contact_ids = []
+    @failed_users = []
+    @success_users = []
+    @programs_to_sync = []
+    @sync_error = ""
+  end
+
+  def run
```
sadleb, line 30: Can we wrap this in a named Honeycomb span? E.g. We haven't been good about this in the past, but going forward I think the `run` method of all our services should have a dedicated HC span. See `sync_salesforce_participant.rb` for an example.

app/services/generate_test_users.rb
sadleb, line 41: Ditto on raising something. Is @sync_error an actual error object? If so, does this work? `raise @sync_error if @sync_error.present?`

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+  def sync_error
+    return @sync_error
+  end
+
+private
+  def sync_programs
+    @programs_to_sync.each do |program_id|
+      SyncSalesforceProgramJob.new.perform(program_id)
+    rescue SyncSalesforceProgram::SyncParticipantsError => e
+      @sync_error = e.message
```
sadleb, line 137: If there are failures for multiple programs, you'll only get the latest one. That's probably fine. However, going back to the `raise if @sync_error` comment above, I think it might make more sense to store the actual error here. So `@sync_error = e` and then change `def sync_error` to `def sync_error_message` which returns the message.

Also, we should do `Sentry.capture_exception(e)` here to make it easier to troubleshoot.

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    # update @success_users with their signuptoken
+    @user_contact_ids.each do |u|
+      signup_token = SalesforceAPI.client.get_contact_signup_token(u['contact_id'])
```
sadleb, line 142: Since it's a read, this should be `HerokuConnect` but since you just set it, the value may not be available yet so I think it makes sense to use the`SalesforceAPI`. But can you add a comment about that?

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+    count = 0
+    while HerokuConnect::Contact.where(email: email).exists?
+      count = count + 1
+      email = original_email.downcase.gsub(/@/, "#{count}@")
+    end
+    count
+  end
+
+  def get_candidate_role_fields(role)
+    if role == "CoachPartner" || role == "Staff" || role == "Faculty"
```
sadleb, line 187: Ditto on using `SalesforceConstants::Role` if it's easy. For example, I think `CoachPartner` should actually be `Coach Partner`? If so, using the constants would have avoided that bug (assuming it's a bug)

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+      # Sets Coach_Partner_Role__c to the actual role, adds spaces before capital letters (ex: "CoachPartner" = "Coach Partner")
+      additional_fields = { 'Coach_Partner_Role__c' => role.gsub(/[A-Z]/, ' \0').strip }
+    else
+      additional_fields = { 'Coach_Partner_Role__c' => 'Test' }
+    end
+
+    if role == "TeachingAssistant" || role == "Staff" || role == "Faculty"
```
sadleb, line 194: Ditto on constants.

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
...
+    while HerokuConnect::Contact.where(email: email).exists?
+      count = count + 1
+      email = original_email.downcase.gsub(/@/, "#{count}@")
+    end
+    count
+  end
+
+  def get_candidate_role_fields(role)
+    if role == "CoachPartner" || role == "Staff" || role == "Faculty"
+      # Sets Coach_Partner_Role__c to the actual role, adds spaces before capital letters (ex: "CoachPartner" = "Coach Partner")
```
sadleb, line 188: I'm confused why this is set to `CoachPartner` in the first place. Nevertheless, I think `role.underscore.humanize` is some rails magic to do the same thing. It turns CamelCase into Camel Case.

app/controllers/test_users_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def get_program_tas
+    authorize :TestUser
+    tas = HerokuConnect::Program.find(params[:id]).ta_participants.filter_map {|p| p if p.is_teaching_assistant? }
+    render :json => tas
+  end
+
+private
+  def set_env
+    @env = Rails.env.production? ? :prod : :dev
```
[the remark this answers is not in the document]
elenabaurkot: I am moving this to the service because I realized it is actually just used there ( I think I was sending it to the job and then to the service so this definitely makes more sense). Is it still OK if I call it @env in there or should I name it something different? @rshipp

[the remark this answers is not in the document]
rshipp: i'd still try to call it something else, but i think my comment here https://github.com/bebraven/platform/pull/1081#discussion_r875128336 might remove the need for this variable entirely

app/assets/stylesheets/test_users.scss
elenabaurkot: Updating to user $bv-red here

app/services/generate_test_users.rb
```diff
@@ -0,0 +1,237 @@
+# frozen_string_literal: true
+
+class GenerateTestUsers
+  include Rails.application.routes.url_helpers
+
+  TEST_ACCOUNT = {
+    'dev': '0011100002DmVInAAN',
+    'prod': '0011J00001goGveQAE'
```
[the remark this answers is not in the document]
elenabaurkot: I realized I am also using env for the salesforce participant record type id for and the candidate record type id for TAs only (all the other roles have the same record type id in prod and dev) could I make environment variables for those as well? And if so are these names too long: `SALESFORCE_TA_PARTICIPANT_RECORD_TYPE_ID` and `SALESFORCE_TA_CANDIDATE_RECORD_TYPE_ID` @rshipp

app/services/generate_test_users.rb
elenabaurkot: I think I had a reason at one point, but honestly I have no clue what it was, so just going to switch this to "Coach Partner" instead and get rid of this

app/controllers/test_users_controller.rb
elenabaurkot: That does definitely seem better, but I think I am going to get rid of this all together based on [this comment ](https://github.com/bebraven/platform/pull/1081#discussion_r875148747 )

elenabaurkot: This was making the values in the dropdown into one work instead of two (while still displaying them as two words for the user, but I don't think I need this anymore since like you mentioned in another comment there was no need for me to use "CoachPartner" instead of "Coach Partner"

app/controllers/test_users_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def get_program_select
+    @program_options = SalesforceAPI.client.get_current_and_future_accelerator_programs
```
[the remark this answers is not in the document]
elenabaurkot: Keep forgetting that 🤦‍♀️  changing to use HerokuConnect

app/services/generate_test_users.rb
elenabaurkot: Would it make more sense to just use `i` you think? It represents the user params for the test user that is currently being created from all the params sent over - looping through to grab the name, program, cohort schedule ect. for each user (Example params = `{"authenticity_token"=>"vCBP_z5EGNKbNfiZEOR1-ZBCMgwIFCf6a4zqhucc-uSGky7-L1p7-nH_Hzhu-10b3cvu550b_mWMlcXRhRoTPA", "role"=>["Fellow", "Fellow"], "program_id"=>["a2Y11000001Hg7hEAC", "a2Y11000001Hg7hEAC"], "cohort_schedule"=>["a3311000001qTDzAAM", "a3311000001qTDzAAM"], "cohort_section"=>["", ""], "ta"=>["", ""], "first_name"=>["Elena", "Elena"], "tag"=>["CreateUser", "UserCreate"], "email"=>["elena.baurkot@bebraven.org", "elena.baurkot@bebraven.org"], "commit"=>"Generate Test User(s)", "controller"=>"test_users", "action"=>"post"}`) @rshipp

app/services/generate_test_users.rb
elenabaurkot: Yeah I think I actually updated to this for that reason, I will add a comment in for that

Gemfile.lock
sadleb: Probably related to needing to rebuild the container after rebasing

app/services/generate_test_users.rb
rshipp: yeah, i think `i` or `index` would be clearer

app/services/generate_test_users.rb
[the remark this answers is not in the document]
rshipp: hmm i think those names are fine, even if they are a little long. i don't know what a record type id is, it's interesting that the rest match in prod/dev but those two don't.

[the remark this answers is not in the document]
sadleb: The reason is bc anything that hasn’t changed since we created the sandbox is a direct copy from prod when we created it. Anything done later was done directly in the sandbox so the ids diverged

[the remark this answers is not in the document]
elenabaurkot: The record type id is the id associated with their role type in Salesforce for candidates and participants (so the id for Fellow, TA, ect.) @rshipp

app/services/generate_test_users.rb
elenabaurkot: Yes, updated to that 👍

app/services/generate_test_users.rb
elenabaurkot: Made these updates 👍

app/controllers/test_users_controller.rb
```diff
@@ -38,12 +43,8 @@ def get_program_tas
...
-  def set_env
-    @env = Rails.env.production? ? :prod : :dev
-  end
-
-  def get_program_select
-    @program_options = SalesforceAPI.client.get_current_and_future_accelerator_programs
-      .map { |program| [program['Name'], program['Id']] }
+  def set_program_options
+    @program_options = HerokuConnect::Program.current_and_future_program_ids
+      .map { |id| [HerokuConnect::Program.find(id).name, id] }
```
sadleb, line 48: Nit: you could create a `current_and_future_programs` scope and then do `@program_options = HerokuConnect::Program.current_and_future_program.pluck(:name, :id)` which would do one database call instead of one per program to map the name/id pairs

sadleb (approved):
Approved pending one last nit piece of feedback for performance (and for learning) purposes.

config/secrets.yml
```diff
@@ -30,6 +30,9 @@ development:
   salesforce_username: <%= ENV['SALESFORCE_PLATFORM_USERNAME'] %>
   salesforce_password: <%= ENV['SALESFORCE_PLATFORM_PASSWORD'] %>
   salesforce_security_token: <%= ENV['SALESFORCE_PLATFORM_SECURITY_TOKEN'] %>
+  salesforce_test_account: <%= ENV['SALESFORCE_TEST_ACCOUNT'] %>
+  salesforce_ta_participant_record_type_id: <%= ENV['SALESFORCE_TA_PARTICIPANT_RECORD_TYPE_ID'] %>
+  salesforce_ta_candidate_record_type_id: <%= ENV['SALESFORCE_TA_CANDIDATE_RECORD_TYPE_ID'] %>
```
sadleb, line 35: @elenabaurkot I think we need to set these in our dev envs and add the default values to `.env.example`.

sadleb: Done here: https://github.com/bebraven/platform/pull/1102

PR #1082, opened 2022-05-18
Remove fellow eval feature

rshipp:
# Description

Remove controllers and supporting code, but leave models/tables so we can maintain historical access to that data in case we ever need it.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202177242987753/f


app/controllers/lti_launch_controller.rb
```diff
@@ -99,8 +99,7 @@ def launch
          /#{completed_preaccelerator_survey_submissions_path}/,
          /#{completed_postaccelerator_survey_submissions_path}/,
          /#{capstone_evaluation_submission_path('')}\d+$/,
-         /#{launch_capstone_evaluation_results_path}/,
-         /#{fellow_evaluation_submission_path('')}\d+$/
```
sadleb, line 103: Not saying we should do this, but instead of deleting the path and controller we could just replace the controller logic to show a message about this feature having been removed and rip out all the guts? That way, we won't need to have a discussion again when an LC or staff member inevitably tries to open one of these in an old course?

sadleb (approved):
lgtm pending that one comment about potentially showing a message to the end-user instead of an error page.

elenabaurkot (commented):
Lgtm 👍

app/controllers/lti_launch_controller.rb
rshipp: whoops, I totally forgot to follow up on this. I don't personally think this is worth prioritizing, if that's what you're asking. are you proposing a new followup task?

PR #1084, opened 2022-05-23
When SalesforceAPI access_token becomes invalid, refresh it

sadleb:
# Description
Previously, through testing, we thought the SalesforceAPI `@access_token` remained valid for at least a day which is when we refresh it, but even so, if it ever does become invalid we need to refresh it then and not have everything broken until a server restart. This fixes that.

Task: https://app.asana.com/0/1201131148207877/1202318571379868

# Motivation and Context

Task: 

# Screenshots


rshipp (approved):

joe-cena-git (commented):
Looks good, reviewed this code and I can see how we are using the rest client with error handling.

PR #1085, opened 2022-05-24
Add due dates specs

rshipp:
# Description

Specs followup for #1027.

# Motivation and Context

Task: https://app.asana.com/0/1178906757852153/1201634733776799/f


elenabaurkot (commented):
This lgtm 👍

spec/controllers/canvas_assignment_overrides_controller_spec.rb
```diff
@@ -0,0 +1,99 @@
+require 'rails_helper'
+
+RSpec.describe CanvasAssignmentOverridesController, type: :controller do
+  render_views
+
+  context "with normal signin" do
+    let(:user) { create :admin_user }
+    let(:course) { create :course }
+    let(:section) { create :section, course: course }
```
sadleb, line 9: nit: can you do the following to make it clear the intention that this code runs only on cohort schedule sections?

```suggestion
    let(:section) { create :cohort_schedule_section, course: course }
```

spec/controllers/canvas_assignment_overrides_controller_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    let(:from_course) { create :course }
+    let(:from_section) { create :section, course: from_course }
```
sadleb, line 11: nit: can you do the following to make it clear the intention that this code runs only on cohort schedule sections?

```suggestion
    let(:from_section) { create :cohort_schedule_section, course: from_course }
```

spec/services/copy_canvas_assignment_overrides_spec.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe CopyCanvasAssignmentOverrides do
+
+  let(:course) { create(:course) }
+  let(:section) { create(:section, course: course) }
+  let(:from_section) { create(:section) }
```
sadleb, line 9: Ditto on the above comments about `cohort_schedule_section`

spec/services/copy_canvas_assignment_overrides_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+          override_hash['id'] = create(:canvas_assignment_override)['id']
+          override_hash['course_section_id'] = section.canvas_section_id
+          # Technically the assignment ID should match the assignment name, but
+          # for the purposes of these tests it doesn't matter.
+          override_hash['assignment_id'] = canvas_assignments.first['id']
+          override_hash
+        end
+      }
+
+      it "creates translated overrides" do
```
sadleb, line 63: Can you try adding a spec to ensure that the overrides sent don't have any assignment or section IDs from the original overrides? I'm thinking about the dup-shifted comment the fact that you're using `filter_map` to skip those where you don't make a change. I worry about a future dev / refactor accidentally sending some overrides to Canvas for the `from_section` or course.

rshipp: interesting, I can try doing this by just stubbing the canvas call

spec/controllers/canvas_assignment_overrides_controller_spec.rb
rshipp: it's the only intended use case for after 2022, based on your proposal that we only put due dates in cohort schedules moving forward. as far as i'm aware, we only have due dates in regular sections right now, so we won't be able to do that for a year

rshipp: talked about this in working session and realized the controller does limit by cohort schedule, and we should have cohort schedule due dates for even old courses since that's where they're set up initially before being copied to individual cohort sections. updated specs to match

spec/services/copy_canvas_assignment_overrides_spec.rb
rshipp: updated specs

PR #1086, opened 2022-05-24
Upgrade Rails from 6.1.4 -> 6.1.6

sadleb:
# Description
Prepping to [work on a Rails 7.0 upgrade](https://app.asana.com/0/1201131148207877/1200066294473271).

Now that the course is ending, it's the perfect time to upgrade all our gems/packages. Not sure how hard that's going to be, but a good first step is getting to the latest version of Rails 6.1.

This PR will require a container rebuild.

# Testing
* specs are passing
* sanity checked some stuff like Modules and Projects
* made sure the React submit button on projects works
* sanity checked that Honeycomb and trace propagation from Boomerang JS spans still works
* sanity checked publishing a new Rise360 module
* sanity checked Content Editor
* sanity checked that the recurring unique jobs sidekiq scheduler works
* sanity checked syncing something from SF to Canvas

# Notes
* When running specs, a ton of the following messages are output at the end. This was [Introduced here](https://github.com/getsentry/sentry-ruby/commit/38df003c8b9ccd40fe33d1887c914a9fad7543c6). I tried some stuff, like using the `enabled_environments` config and reducing the `background_worker_threads` config, but nothing worked. It's annoying but not worth spending more time on at the moment imo.
```
Killing session flusher
Shutting down background worker
Killing session flusher
Shutting down background worker
Killing session flusher
```

# Screenshots


Gemfile.lock
```diff
@@ -219,7 +218,7 @@ GEM
       moneta
       multi_json (>= 1.9.2)
       webrick
-    honeycomb-beeline (2.8.2)
+    honeycomb-beeline (2.10.0)
```
sadleb, line 221: TODO: check that stuff still works with this upgrade. I remember there being an issue with newer versions of `libhoney` and we pinned it at one point.

Gemfile.lock
```diff
@@ -344,26 +343,26 @@ GEM
...
+      rake (>= 12.2)
       thor (~> 1.0)
     rainbow (3.1.1)
     rake (13.0.6)
     rate_throttle_client (0.1.2)
     rb-fsevent (0.11.1)
     rb-inotify (0.10.1)
       ffi (~> 1.0)
-    react-rails (2.6.1)
+    react-rails (2.6.2)
```
sadleb, line 358: TODO: check that our React stuff (just the buttons I think?) still works. I ran into issues with react compatibility in the past.

Gemfile.lock
```diff
@@ -456,14 +455,12 @@ GEM
       connection_pool (>= 2.2.2)
       rack (~> 2.0)
       redis (>= 4.2.0)
-    sidekiq-scheduler (3.2.1)
-      e2mmap
-      redis (>= 3, < 5)
+    sidekiq-scheduler (4.0.0)
+      redis (>= 4.2.0)
```
sadleb, line 459: Look into this one too. I think we did an upgrade of one piece of this to get rid of a deprecation warning. Make sure this is all set.

Gemfile.lock
```diff
@@ -456,14 +455,12 @@ GEM
...
       rufus-scheduler (~> 3.2)
-      sidekiq (>= 3)
-      thwait
+      sidekiq (>= 4)
       tilt (>= 1.4.0)
-    sidekiq-unique-jobs (7.1.15)
+    sidekiq-unique-jobs (7.1.23)
```
sadleb, line 463: Ditto. Test this.

elenabaurkot (commented):
This looks good to me

spec/api_helper.rb
```diff
@@ -13,6 +13,12 @@ def result(evaluation)
       evaluation.notify(:after_json, json)
     end
   end
+
+  # See here for why this is here:
+  # https://bytemeta.vip/repo/thoughtbot/factory_bot/issues/1536
```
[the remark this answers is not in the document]
sadleb: fixed to use cached version

rshipp (approved):

PR #1087, opened 2022-05-24
Restrict accelerator surveys to only Fellows

rshipp:
# Description

Duplicate the behavior we use for waivers, with pre/post accelerator surveys, so LCs/TAs don't try to do them.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202323838358713/f

# Screenshots


elenabaurkot (commented):
Looks good to me 👍

sadleb (approved):
lgtm

PR #1088, opened 2022-05-24
Fix email duplicates in Test User Generation

elenabaurkot:
# Description
This is a fix for the [Add Generate Test Users Feature PR](https://github.com/bebraven/platform/pull/1081). When Test Users are generate it checks it checked if the email already exists in HerokuConnect for a user and if it does it increments the email. However, if someone generates two users with the same email while running the service the user doesn't already exist in the database so it tries to create them twice and fails the second time. This update checks if a user already exists in HerokuConnect or during this run of the service to generate test users and if so increments the email.  

# Motivation and Context


app/services/generate_test_users.rb
```diff
@@ -172,10 +172,14 @@ def get_email_and_last_name(user_index)
...
+    # The email can be used as is if it doesn't already exists in the database or in the users we
+    # previously generated while running this service now, otherwise we need to add an incrementor to the email
+    return if HerokuConnect::Contact.where(email: email).empty? && !previous_user_emails.include?(email)
 
     count = 0
-    while HerokuConnect::Contact.where(email: email).exists?
+    while HerokuConnect::Contact.where(email: email).exists? || previous_user_emails.include?(email)
       count = count + 1
       email = original_email.downcase.gsub(/@/, "#{count}@")
     end
```
sadleb, line 185: Would the following be cleaner?
```
# The email can be used as is if it doesn't already exist in the database or in the users we
# previously generated while running this service now, otherwise we need to add an incrementor to the email
count = nil
while HerokuConnect::Contact.where(email: email).exists? || previous_user_emails.include?(email)
  count ||= 0 # initialize it in the first loop
  count = count + 1
  email = original_email.downcase.gsub(/@/, "#{count}@")
end

count
```

app/services/generate_test_users.rb
```diff
@@ -172,10 +172,14 @@ def get_email_and_last_name(user_index)
   # Return the number to be added to the email and last_name for the new user
   def check_user_existence(original_email)
```
sadleb, line 173: I tried to think of a better name for this method, but couldn't come up with anything. Maybe `get_unique_suffix` ?

sadleb (approved):
Left non-merge blocking feedback around potentially more readable code.

PR #1090, opened 2022-05-26
Upgrade to Rails 7.0

sadleb:
# Description
Upgrade to Rails 7.0. Overview of major changes from 6.1 -> 7.0 here: https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-1-to-rails-7-0

Task: https://app.asana.com/0/1201131148207877/1200066294473271

Note that Rails 7.0 has made some drastic changes to its default approach to handling Javascript. However, it still works with the old ways too. That is a whole project unto itself, so I'm **not** making any changes to the JS pipeline and tools in this PR. Follow-up tasks:
* Upgrade Webpacker: https://app.asana.com/0/1201131148207877/1200066294473247
* Upgrade React: https://app.asana.com/0/1201131148207877/1200066294473253

# Deployment
* [x] Generate the following three ENV vars using [this guide](https://edgeguides.rubyonrails.org/active_record_encryption.html) and setting them in prod **BEFORE** deploy.
* `ENV['ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY']`
* `ENV['ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY']`
* `ENV['ACTIVE_RECORD_ENCRYPTION_DERIVATION_SALT']`

# Testing
* [x] specs are passing
* [x] sanity checked some stuff like Modules and Projects
* [x] made sure the React submit button on projects works
* [x] sanity checked that Honeycomb and trace propagation from Boomerang JS spans still works
* [x] sanity checked publishing a new Rise360 module
* [x] sanity checked Content Editor
* [x] sanity checked that the recurring unique jobs sidekiq scheduler works
* [x] sanity checked syncing something from SF to Canvas

# Screenshots

Example of new specs Honeycomb traces
image


config/secrets.yml
```diff
@@ -132,6 +141,9 @@ staging:
   discord_bot_token: <%= ENV['DISCORD_BOT_TOKEN'] %>
 
 production:
+  active_record_encryption_primary_key: <%= ENV['ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY'] %>
+  active_record_encryption_deterministic_key: <%= ENV['ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY'] %>
+  active_record_encryption_derivation_salt: <%= ENV['ACTIVE_RECORD_ENCRYPTION_DERIVATION_SALT'] %>
```
sadleb, line 146: TODO: generate some of these using [the guide](https://edgeguides.rubyonrails.org/active_record_encryption.html) and set prod specific ones.

Gemfile
```diff
@@ -31,17 +31,14 @@ group :development, :test do
...
   gem 'shoulda-matchers'
 end
 
 group :development do
   # Access an interactive console on exception pages or by calling 'console' anywhere in the code.
   gem 'listen', '>= 3.3.0'
   gem 'web-console', '>= 3.3.0'
-  # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
-  gem 'spring'
-  gem 'spring-watcher-listen', '~> 2.0.0'
```
sadleb, line 44: Spring and Bootsnap weren't playing nice together. Was getting this error:
`You have already activated digest 3.0.0, but your Gemfile requires digest 3.1.0.`

[See here](https://aboobacker.in/2021/06/28/bootsnap-and-spring-understanding-rails-boottime-optimisations.html) for any overview of what these gems are/do.

Rails 7 no longer ships with Spring enabled by default and the `spring-watcher-listen` gem doesn't seem to be
maintained anymore and doesn't support v3 or v4 of Spring.

So I just removed it and tried things out. Error goes away. No noticeable difference from some quick sanity testing of Guard reloading changes or rake tasks running.

app/controllers/cas_controller.rb
```diff
@@ -93,10 +93,10 @@ def login
           logger.info("User '#{tgt.username}' authenticated based on ticket granting cookie. Redirecting to service '#{@service}'.")
           # Devise tends to flash a "You need to sign in or sign up before continuing." alert.
           flash.delete(:alert)
-          redirect_to Utils.build_ticketed_url(@service, st) and return
+          redirect_to Utils.build_ticketed_url(@service, st), allow_other_host: true and return
```
sadleb, line 96: See the comment in `spec/rails_helper.rb` for what's going on here.

Gemfile
```diff
@@ -123,9 +120,6 @@ gem 'platform-api', require: false
 # Implementation of JSON Web Token (JWT) standard: https://github.com/jwt/ruby-jwt
 gem 'jwt'
 
-# Generates attr_accessors that encrypt and decrypt attributes transparently in the database
-gem 'attr_encrypted'
-
```
sadleb, line 128: See notes and links in `db/migrate/20220526120958_change_keypairs_to_active_record_encryption.rb` for more info

spec/spec_helper.rb
```diff
@@ -48,9 +48,32 @@
...
-  # If you turn logging up, this adds a log when each test starts to run
+  config.before(:suite) do
+    # libhoney introduced the excon gem for HTTP requests here:
+    # https://github.com/honeycombio/libhoney-rb/commit/820bc2fb6ac7860d01743db1f3726191981cde81
+    # This lead to tons of error messages logged (when using :debug level logging)
+    # with: "Libhoney::TransmissionClient: 💥 no stubs matched"
+    # and a huge hash of the request information. The below setting gets rid of that error
+    # by allowing actual network requests to go out when running tests.
+    # See: https://github.com/excon/excon#stubs
+    Excon.defaults[:allow_unstubbed_requests] = true
```
sadleb, line 59: Source code here that led to the error: https://github.com/excon/excon/blob/master/lib/excon/middlewares/mock.rb#L52

spec/spec_helper.rb
```diff
@@ -48,9 +48,39 @@
...
+     #
+     # Example span_name: "accelerator_survey_submissions_controller_spec.rb:58 - checks Canvas LTI Advantage API for a score"
+    span_name = "#{File.basename(x.metadata[:location])} - #{x.metadata[:description]}"
+    @root_honeycomb_span  = Honeycomb.start_span(name: span_name)
+    @root_honeycomb_span.add_field('rspec.example.name', x.metadata[:full_description])
+  end
+
+  config.after :example do |x|
+    @root_honeycomb_span.send
   end
```
sadleb, line 84: See the screenshot in the description for what this results in.

spec/spec_helper.rb
```diff
@@ -48,9 +48,39 @@
...
+     # tests easier. It also makes it easier to write, test, and see your Honeycomb fields.
+     #
+     # Example span_name: "accelerator_survey_submissions_controller_spec.rb:58 - checks Canvas LTI Advantage API for a score"
+    span_name = "#{File.basename(x.metadata[:location])} - #{x.metadata[:description]}"
+    @root_honeycomb_span  = Honeycomb.start_span(name: span_name)
+    @root_honeycomb_span.add_field('rspec.example.name', x.metadata[:full_description])
+  end
+
+  config.after :example do |x|
+    @root_honeycomb_span.send
```
sadleb, line 83: See the screenshot in the description for what this results in.

db/schema.rb
```diff
@@ -673,24 +671,24 @@
...
+    t.datetime "updated_at", null: false
     t.index ["course_custom_content_version_id"], name: "index_survey_submissions_on_course_survey_version_id"
     t.index ["user_id"], name: "index_survey_submissions_on_user_id"
   end
 
   create_table "ticket_granting_tickets", force: :cascade do |t|
     t.string "ticket", null: false
-    t.datetime "created_on", null: false
-    t.datetime "created_at", null: false
+    t.datetime "created_on", precision: nil, null: false
```
elenabaurkot, line 690: Does specifying `precision: nil` mean it won't have any decimals?

elenabaurkot (commented):
Lgtm 👍

db/schema.rb
sadleb: Looks like it just means "use the database default for that type" https://github.com/rails/rails/issues/44571#issuecomment-1054758649

app/controllers/cas_controller.rb
rshipp: can you explain this more? turning off this security feature just to get the tests to pass seems suboptimal

config/secrets.yml
rshipp: if you're making this assertion, we need to never copy production data into staging/local like we've done in the past

db/schema.rb
```diff
@@ -155,29 +154,29 @@
     t.bigint "canvas_course_id", null: false
     t.float "score"
     t.string "grade"
-    t.datetime "graded_at"
+    t.datetime "graded_at", precision: nil
```
rshipp, line 157: the fact that anything that previously had no precision now has `precision: nil`, while anything that had `precision: 6` now has no precision, makes me worry this is going to end up in a different state the next time somebody runs a migration or something

rshipp (changes requested):
adding the security bypass to all the redirects is not safe, we should try to find other options to get the tests to pass

app/controllers/cas_controller.rb
sadleb: Here is a little [article](https://blog.saeloun.com/2022/02/08/rails-7-raise-unsafe-redirect-error) with some links to the source code. We have to do it b/c we redirect from `platform.braven.org` to `braven.instructure.com`. It's not just for specs. It's just new behavior. The default used to be `true` for all redirects but now it's false, so I had to go through the redirects that went to Canvas and explicitly allow them

sadleb:
> adding the security bypass to all the redirects is not safe, we should try to find other options to get the tests to pass

It's not all redirects. That was the old behavior. See my response [here](https://github.com/bebraven/platform/pull/1090#discussion_r892521600).

db/schema.rb
sadleb: This whole thing is confusing to me, but reading through the change and specifically [this comment ](https://github.com/rails/rails/issues/44571#issuecomment-1056131789) makes me think it won't keep switching around in future migrations. It's a separate question whether we want to go back and be consistent or not. Outside the scope of this PR though.

app/controllers/cas_controller.rb
rshipp: can you verify which redirects *must* allow other domains? looking at this briefly, i can see this one https://github.com/bebraven/platform/pull/1090/files#diff-1cc94f8305c6c7787c0b817518c0b004af88947cf2b6fa6a8363c1f14a455890R117 at least does not and should not allow that behavior. we had to do a bunch of manual work on the URL to make sure it didn't point to other hosts, in fact.

app/controllers/lti_launch_controller.rb
```diff
@@ -24,7 +24,7 @@ def login
                                          .merge(:state => LtiLaunchController.generate_state, :nonce => SecureRandom.hex(10)))
 
     # Step 2 in the flow, do the handshake
-    redirect_to "#{LTI_AUTH_RESPONSE_URL}?#{@lti_launch.auth_params.to_query}"
+    redirect_to "#{LTI_AUTH_RESPONSE_URL}?#{@lti_launch.auth_params.to_query}", allow_other_host: true
```
sadleb, line 27: This is `https://canvas.instructure.com/api/lti/authorize_redirect`

app/controllers/cas_controller.rb
sadleb: Good call on that one. That was a mistake. I removed it and fixed the specs. The rest in this file are to the `service` param which is generally Canvas (but can be platform in some cases), so they need `allow_other_host`. Will double check the rest.

rshipp (approved):

sadleb:
Note that when I tried to deploy to heroku using `git push heroku main` it failed with this:
```
remote:        error Command "webpack" not found.
remote:        
remote: 
remote:  !
remote:  !     Precompiling assets failed.
remote:  !
remote:  !     Push rejected, failed to compile Ruby app.
remote: 
remote:  !     Push failed
```
[This article](https://discuss.rubyonrails.org/t/deploying-to-heroku-command-webpack-not-found/76409) helped me solve it by adding the nodejs buildpack:
image

PR #1091, opened 2022-05-31
Update participant lookups from SalesforceAPI to HerokuConnect

elenabaurkot:
# Description
Participant lookups using the SalesforceAPI only looks up participants in current and future programs. This creates errors with participant lookups when programs become former. This update changes over participant lookups from using the Salesforce API to use HerokuConnect instead so that Fellows can still use Canvas after their program becomes former and we can do participant lookups without having them receive errors. The only place that still does participant lookups using the SalesforceAPI is the Discord Bot because it doesn't have access to the HerokuConnect database. If someone tries to add a SalesforceAPI participant lookup to the code outside of the Discord Bot they will receive an error that says `All participant lookups should be made using HerokuConnect::Participant, except from the discord_bot because it doesn't have access to the HerokuConnect database.`

# Motivation and Context
To make sure Fellows can continue using Canvas after to their program ends, such as view Learning Lab Zoom links without receiving errors.

Task: https://app.asana.com/0/1201131148207877/1202020298489360/f


lib/salesforce_api.rb
```diff
@@ -404,6 +405,10 @@ def find_contact(id:)
   # TODO: rename to find_participants_by_program_id()
   # https://app.asana.com/0/1201131148207877/1201217979889312
   def find_participants_by(program_id:)
+    if defined?(Rails) && Rails.respond_to?(:logger)
+      raise ParticipantLookupError, "All participant lookups should be made using HerokuConnect::Participant,"\
+      "except from the discord_bot because it doesn't have access to the HerokuConnect database."
```
elenabaurkot, line 410: This error gets raise when the method is called anywhere outside of Discord Bot -> this also raises if I use the method in the console, wanted to double check if that seems correct?

rshipp: ah, that's interesting, i wondered what that check was for. seems reasonable to me

lib/salesforce_api.rb
```diff
@@ -404,6 +405,10 @@ def find_contact(id:)
   # TODO: rename to find_participants_by_program_id()
   # https://app.asana.com/0/1201131148207877/1201217979889312
   def find_participants_by(program_id:)
+    if defined?(Rails) && Rails.respond_to?(:logger) && !caller_locations.first.path.include?('salesforce_api_spec.rb')
```
elenabaurkot, line 408: Updated this to also check if SalesforceAPI participant lookups are being called from the SalesforceAPI spec and allow it to be called from there since it was causing tests to fail

app/models/heroku_connect/participant.rb
```diff
@@ -33,10 +33,37 @@ def self.default_columns
       :program__c,
       :cohort__c,
       :cohort_schedule__c,
-      :webinar_access_1__c, :webinar_access_2__c,
+      :webinar_access_1__c,
+      :webinar_access_2__c,
+      :webinar_access_3__c
```
sadleb, line 38: For future devs, adding this new field also requires doing the steps [here](https://github.com/bebraven/platform/wiki/Salesforce#steps-for-making-a-new-salesforce-field-available-in-ruby) to make it available. I double checked that was done.

app/models/heroku_connect/candidate.rb
```diff
@@ -69,7 +69,7 @@ def is_lc?
     role == SalesforceConstants::Role::LC_SUBSTITUTE
   end
 
-  # Checks if they're an actual TA.
+  # Checks if they're an actual TA, includes test user TAs
```
sadleb, line 72: Wait, if this includes test users what is the `is_test_user?` check above?

app/models/heroku_connect/participant.rb
```diff
@@ -33,10 +33,37 @@ def self.default_columns
...
+  alias_attribute :zoom_meeting_link_3, :webinar_access_3__c
+
+  scope :find_participant, ->(contact_id, program_id) {
+    where(program__c: program_id, contact__c: contact_id).first
+  }
+
+  # Strange thing to note: if you call 'HerokuConnect::Participant.with_discord_info.first' and
+  # print out the participant it won't show the joined info, but if you call
+  # 'HerokuConnect::Participant.with_discord_info.first.discord_user_id' you do in fact have
+  # access to the joined attributes
```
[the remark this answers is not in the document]
elenabaurkot: Ah that is super helpful, wish I had read the comment when I was looking at that before, going to add that in here too

app/models/heroku_connect/candidate.rb
elenabaurkot: This method does include test TAs and I was purposely excluding them on line 51 because of the comment that was on lines from 45-47. I'm thinking the purpose of this is if we aren't allowing TAs to sign up for Discord using the signup page we're trying to have test TAs mimic the same behavior, but I am not positive if this is what we want or not @sadleb @rshipp 
```
    # Show Staff/Faculty/Test TAs a different page, so they can't use this.
    # Note: We will likely adjust how this is determined later. See the Slack
    # discussions in this task for more info: https://app.asana.com/0/0/1201314155101899/f
```

elenabaurkot: Comment on line 47 (`# discussions in this task for more info: https://app.asana.com/0/0/1201314155101899/f`) points to these Slack conversations:
- https://bebraven.slack.com/archives/C01166ZLR1V/p1636993153158400
- https://bebraven.slack.com/archives/C011XKHHQ1J/p1636742054079800
- https://bebraven.slack.com/archives/C02H6RE7DB5/p1636741230071200

These give rationale for excluding test TAs. After reading through them, I'm not sure if I understood 100%, but I think what I took is that only actual TAs will have TA role on participant and a candidate role select of TA (fake TAs will have candidate role select of "Faculty" - a little confused if this definitely applied to Spelman too or not from some of the comments) @sadleb @rshipp

app/controllers/discord_signups_controller.rb
```diff
@@ -37,15 +37,18 @@ def launch
...
     participant.add_to_honeycomb_span()
 
     # Show Staff/Faculty/Test TAs a different page, so they can't use this.
     # Note: We will likely adjust how this is determined later. See the Slack
     # discussions in this task for more info: https://app.asana.com/0/0/1201314155101899/f
-    if participant.role == SalesforceAPI::TEACHING_ASSISTANT && participant.volunteer_role != 'Teaching Assistant'
+    #
+    # Checks if participant has a TA record type, but they aren't an actual TA (meaning they are staff or faculty)
+    # or if they have their role set to :Test (making them a test TA)
+    if participant.recordtypeid == Rails.application.secrets.salesforce_ta_participant_record_type_id && (!participant.is_teaching_assistant? || participant.is_test_user?)
```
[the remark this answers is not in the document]
elenabaurkot: If we do want to allow test TAs to use this page I am thinking of updating to check that the `participant.role_category` is a teaching assistant and if `.is_teaching_assistant?` is false (since that method counts test TAs as an actual TA) to check if they are staff or faculty. I would also remove the `.is_test_user?` method I added since we wouldn't need it anymore

```
    # Show Staff/Faculty a different page, so they can't use this.
    # Checks if participant has a TA record type, but they aren't an actual TA (meaning they are staff or faculty)
    if participant.role_category == SalesforceConstants::RoleCategory::TEACHING_ASSISTANT && !participant.is_teaching_assistant?
      return render :no_discord
    end
```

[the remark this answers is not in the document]
sadleb: Yeah, that sounds right

app/models/heroku_connect/candidate.rb
sadleb: Whew, that was a lot to read through and process. I think the intent was only to allow actual TAs access b/c Staff use a different process for getting roles (a command in the bot I think) and things can get messed up. We wanted to avoid accidentally giving university Faculty access b/c there is more to figure out and might differ per region. However, the "test" one wasn't really thought through in those threads imo. So I'd say, let's just allow Test TAs access and change the comment. It's not a breach of the privileges / permissions for actual university folks since it's just a test account. Plus is it makes the code cleaner and allows us to test this feature with a Test TA account.

rshipp: yeah test TAs should have access as normal TAs, as long as that's not a bunch of extra work. we legally must exclude faculty/guest "TA"s; they aren't allowed access to Discord at all.

sadleb (approved):
Lgtm!

PR #1092, opened 2022-05-31
Fix failing spec

rshipp:
# Description

Followup to #1085 to fix a failing spec, noticed by @elenabaurkot. 

Just have to manually remove these attributes before sending bc they aren't part of the canvas api.

# Motivation and Context

Task: none


elenabaurkot (commented):
Lgtm!

PR #1093, opened 2022-05-31
Fix failing spec

rshipp:
# Description

Better followup to the followup #1092.

These titles aren't hardcoded, I must have just been getting lucky.

# Motivation and Context

Task: None


elenabaurkot (approved):
Tried it and it works to fix the failing spec, lgtm!

PR #1096, opened 2022-06-08
Add reset password Heroku security banner alert after failed login

elenabaurkot:
# Description
We rotated all secrets stored in config variables on Heroku as a result of a Heroku security breach. Rotating some of these secrets invalidated all user passwords on Platform. When a user tries to log in with an invalid password, they will see a banner alert letting them know to reset their password because of a security incident. This banner will remain in place until around August 1, 2022. 

# Motivation and Context
As reported on [status.heroku.com](http://click.mail.salesforce.com/?qs=8f4374bab63f0724e3f5620f24b348340034a31de26f4e716df9c6a374d96044b528d911b5f0a91dafff78212f9b0435cae582ba4c06201c), on April 7, 2022, a threat actor obtained access to a Heroku database and downloaded stored customer GitHub integration OAuth tokens. On that same day, the threat actor downloaded data from another database that stores pipeline-level config vars for [Review Apps](http://click.mail.salesforce.com/?qs=8f4374bab63f0724d1f830548bd0b9967cca42f09d7bd31af80b624caa7daf252fe0d0f6e1c083b62d1f6e8d10081b48aa5c4c0ba4114f16) and [Heroku CI](http://click.mail.salesforce.com/?qs=8f4374bab63f07241158b6c91dd0b1e5e1bfc9c54c526efd7aabdc4d7ef8d3b310c9500aef69afae51f98ecb5870324e8f38b43a32f1cb1b). As a result we had to rotate all of the secrets we had stored in our config variables on Heroku. Rotating some of these config variables resulted in invalidating everyone's Platform passwords. This banner alerts users of the security breach and lets users know they need to reset their passwords.  

Task: https://app.asana.com/0/1201131148207877/1202375758620598/f

# Screenshots
Screen Shot 2022-06-08 at 2 41 51 PM


sadleb (approved):
Lgtm!

app/views/cas/login.html.erb
```diff
@@ -1,3 +1,9 @@
+<% if @show_banner %>
+  <div class="login-alert row mx-1 mt-2 p-2 alert" role="alert">
+    Braven reset your password on 6/9/2022. If you haven’t already, please click “Forgot your password?” below to create a new password and regain access to your account. This was a precautionary measure in response to a security incident with a 3rd party vendor. If you have already reset your password, you can ignore this message.
```
sadleb, line 3: Actually, small change request @elenabaurkot - can we put the text in a `<p role="alert">`? It’s more semantically correct that way and better for accessibility with the role too

sadleb (changes requested):
One small request

app/views/cas/login.html.erb
elenabaurkot: Updated to this 👍

sadleb (approved):

app/views/cas/login.html.erb
sadleb: If the change you made still looks the same, this is good to go! If not, I actually meant to just put the "text" in a `<p>` but the overall thing could be in a div so the bootstrap stuff would still apply there. E.g. this is what I was thinking
```
<div class="login-alert row mx-1 mt-2 p-2 alert">
  <p role="alert">TEXT HERE</p>
</div>
```

PR #1098, opened 2022-06-14
Prevent Fellows from submitting projects after grade_finalized_date

elenabaurkot:
# Description
To prevent Fellows from submitting projects during grade finalization, a `grades_finalized_date` field has been added to programs. If this date has passed, when Fellows launch projects they will be in read only mode so they can't enter anything into the inputs and there won't be a submit button or project save updates. A `grades_finalized_extension` date field was added to Participants to allow  Fellows who earned an "incomplete" in the course to submit projects after the grade finalization. If the `grades _finalization_date` for a program has passed, but a Fellow has a `grades_finalized_extension` date that hasn't passed, they can still edit and submit projects. If the `grades _finalization_date` for a program has passed and a Fellow does not have a `grades_finalization_extension` listed or they have one that has also passed, projects will be in read only mode and they won't be able to submit. 

# Motivation and Context
Fellows were submitting/re-submitting projects during the grade finalization period after the final due date had passed. This was resulting in a different overall grade in the Portal from what had been submitted to the University Partner.

Task: https://app.asana.com/0/0/1202365097128921/f

# Screenshots
Project before Program `grades_finalized_date` passes -> Fellows can type in input fields and Submit button showing:
Screen Shot 2022-06-14 at 9 44 37 AM

Project after Program `grades_finalized_date` passes (when the Participant doesn't have a `grades_finalized_extension` -> Fellows cannot type in input fields and Submit button is hidden:
Screen Shot 2022-06-14 at 9 43 52 AM


app/controllers/project_submissions_controller.rb
```diff
@@ -101,4 +102,17 @@ def set_has_previous_submission
...
 
+  def set_read_only
+    @read_only = false
+    program = HerokuConnect::Program.find_by(canvas_cloud_accelerator_course_id__c: @lti_launch.course_id)
+    # Return and launch project like normal if the program grades_finalized_date hasn't passed
+    return unless program.grades_finalized_date__c.past?
+
+    participant = HerokuConnect::Participant.find_participant(current_user.salesforce_id, program)
+    # Return and launch project like normal if the participant has a project submission extension
+    return if participant.grades_finalized_extension__c
```
rshipp, line 113: any extension, even if the extension has past?

rshipp:
can you double-check the project answers JS to make sure there's nothing in there that assumes "read only mode" is the same as "TA/gradebook mode"? i remember doing something along those lines, maybe just in the Honeycomb code?

app/controllers/project_submissions_controller.rb
elenabaurkot: Right now the extension is just a checkmark on the participant in Salesforce so maybe it would be better to make that a date and only allow them to launch it until that date passes @rshipp - I can check with Annie to see if we want the extension to only go until a certain date

elenabaurkot:
> can you double-check the project answers JS to make sure there's nothing in there that assumes "read only mode" is the same as "TA/gradebook mode"? i remember doing something along those lines, maybe just in the Honeycomb code?

I will take a look at that @rshipp

elenabaurkot:
> can you double-check the project answers JS to make sure there's nothing in there that assumes "read only mode" is the same as "TA/gradebook mode"? i remember doing something along those lines, maybe just in the Honeycomb code?

@rshipp I read through the `project_answers.js` and didn't see anything that stood out that would cause any of that to break - I also didn't see anything about the "TA/gradebook mode"

rshipp:
ok, thanks for checking!

app/controllers/project_submissions_controller.rb
elenabaurkot: `grades_finalized_extension__c` was updated to a date field

app/controllers/project_submissions_controller.rb
```diff
@@ -101,4 +102,17 @@ def set_has_previous_submission
     ).exists?
   end
 
+  def set_read_only
+    @read_only = false
+    program = HerokuConnect::Program.find_by(canvas_cloud_accelerator_course_id__c: @lti_launch.course_id)
+    # Return and launch project like normal if the program grades_finalized_date hasn't passed
+    return unless program&.grades_finalized_date__c&.past?
```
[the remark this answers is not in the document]
sadleb: We'll need to test out whether `past?` plays nice with the actual values that end up coming through from salesforce. If not, we might have to add a helper method to wrap the raw value in logic to compute if it's in the past, similar to what we did for the `due_in_past?` method on the `CanvasSubmission` model. I'm esp thinking about UTC and timezone bugs.

app/controllers/project_submissions_controller.rb
```diff
@@ -101,4 +102,24 @@ def set_has_previous_submission
...
+      msg = "No program found for Course with the canvas course id: #{@lti_launch.course_id}. " +
+      "Projects will automatically be launched in edit mode instead of read only mode if a program is " +
+      "missing, even if the 'Grades Finalized Date' has passed."
+      Honeycomb.add_support_alert('missing_program_error', msg)
+    end
+
+    # Return and launch project like normal if the program grades_finalized_date hasn't passed
+    return unless program&.grades_finalized_date__c&.past?
+
+    participant = HerokuConnect::Participant.find_participant(current_user.salesforce_id, program)
```
sadleb, line 118: Probably not worth doing now in this PR, but I am thinking about how common it may be to need to get a Participant from the LtiLaunch and whether it makes sense to create helper methods that all the different places can share. E.g. we could have helpers like `@lti_launch.participant` or a `program.participant_for(user)` that abstracts this stuff away (and where we can do central error handling / alerting over time as we run into issues.)

app/models/heroku_connect/program.rb
```diff
@@ -34,6 +34,7 @@ def self.default_columns
       :default_timezone__c,
       :program_end_date__c,
       :program_start_date__c,
+      :grades_finalized_date__c
```
sadleb, line 37: Ditto on making sure you map this in the prod app.

spec/controllers/project_submissions_controller_spec.rb
```diff
@@ -121,6 +121,13 @@
   end # 'GET #new'
 
   describe 'GET #edit' do
+    let!(:program) { build :heroku_connect_program,
+      canvas_cloud_accelerator_course_id__c: lti_launch.course_id,
+      grades_finalized_date__c: Date.today
```
sadleb, line 126: should we be using `utc` ?

sadleb (approved):
Overall, lgtm! I'm approving in advance of addressing the concerns around timezones.

app/controllers/project_submissions_controller.rb
rshipp: I already added `user.participant`, so i think that's good enough

spec/controllers/project_submissions_controller_spec.rb
rshipp: rails in heroku runs in utc, so it's a good idea to always use utc to avoid something working different locally than in prod

app/controllers/project_submissions_controller.rb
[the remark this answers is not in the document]
elenabaurkot: Updating to convert the time zone to the time zone of the program at the end of the day and checking if that time has passed

app/controllers/project_submissions_controller.rb
rshipp: i think its out of scope for this pr

spec/controllers/project_submissions_controller_spec.rb
elenabaurkot: Updated to use `Time.now.utc.to_date` to get the utc Date

PR #1102, opened 2022-06-16
Fix failing specs

sadleb:
# Description
The following PR hardcoded some IDs in the factories which is causing some specs that
require the associated models created to match up to fail. Reverting the hardcoded IDs
b/c I don't think they were needed in that PR anyway:

* https://github.com/bebraven/platform/pull/1091

# Motivation and Context
Failing specs looked like this:
image


elenabaurkot:
I feel like I ended up added these hardcoded values in because I was having some specs fail at one point because the `recordtypeids` were coming back with random values that didn't match the actual type of user they were, but if they are passing without them it sounds like it's probably better not to have them hardcoded here. Sorry for breaking your tests with the hardcoded values @rshipp and @sadleb!

PR #1106, opened 2022-06-22
Loosen the content security policy for Rise360 modules

sadleb:
# Description
Modules aren't loading in production (they work in dev). I'm not sure what changed, but we started getting errors like the following which resulted in a message to the user about their browser maybe being out of date:
```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' https: https://ajax.googleapis.com/". Either the 'unsafe-inline' keyword, a hash ('sha256-v68uDI5VvfFqPm0HqePyy/qY1WToOE6l3MiSxN5Casg='), or a nonce ('nonce-...') is required to enable inline execution.
```

My hunch is that previously the global content security policy (in `config/initializers/content_security_policy.rb`) wasn't being applied for the requests that are proxied through `lib/lti_rise360_proxy.rb` but now after the Rails7 upgrade they are.

# Motivation and Context
Task: https://app.asana.com/0/1201131148207877/1202476572511190

* We `iframe` in the Module content in `app/views/rise360_module_versions/show.html.erb`.
* The `src` of the `iframe` is set to a path like:
```
https://platformweb/rise360_proxy/lessons/sz9d8lx9240jnpasmvbfvia7vj6o/index.html?
  #   actor=%7B%22name%22%3A%22RISE360_USERNAME_REPLACE%22%2C%20%22mbox%22%3A%5B%22mailto%3ARISE360_PASSWORD_REPLACE%22%5D%7D&
  #   auth=LtiState%20SOMESTATE&endpoint=https%3A%2F%2Fplatformweb%2Fdata%2FxAPI
```
* When the browser tries to load the iframe and hits that path, it goes through the `lti_rise360_proxy` which fetches the data from an AWS S3 bucket.
* Since this is middleware, it never hits a controller so we can't use an example like what is found in `FormAssemblyController` to loosen the policy. E.g.
```
content_security_policy do |policy|
     global_script_src =  policy.script_src
     policy.script_src "#{Rails.application.secrets.form_assembly_url}:*", :unsafe_eval, :unsafe_inline, -> { global_script_src }
  end
```
* Instead, I dug through the [ContentSecurityPolicy middleware here](https://github.com/rails/rails/blob/3872bc0e54d32e8bf3a6299b0bfe173d94b072fc/actionpack/lib/action_dispatch/http/content_security_policy.rb#L174) to find some code I could use to get/set the policy.
* Note: I'm using `clone` b/c it appears that the global policy is a singleton of some sort and if I try to edit it directly, it impacts all endpoints and not just this one.

# Screenshots


elenabaurkot (commented):
This looks good to me 🤞

PR #1108, opened 2022-06-22
Performance tune Module grading for unregistered users

sadleb:
# Description
Don't grade Modules for unregistered Fellows since it's noisy in Honeycomb and leads to a bunch of unnecessary API calls.

# Motivation and Context
The motivation here was to tune out noisy alerts, [like this one](https://bebraven.slack.com/archives/C01U1T16Y4A/p1655628399554239), but I think they can still happen for other reasons, so I'm going to try and do that in the trigger using the new `restclient.parent_class` and related fields. _UPDATE_: [trigger was tuned here](https://ui.honeycomb.io/braven/datasets/platform.braven.org/triggers/JK46TVXuFHb) using a new `restclient.normal.error.rise360_module_grading` derived column

However, in the period of time where we've accepted a bunch of folks and they start syncing and getting Canvas accounts, but the program hasn't launched yet a failed `CanvasAPI` call will happen for each of them in the nightly grading. No
reason to start grading until they've created an account and registered.

Task: https://app.asana.com/0/1201131148207877/1202413649545453

# Screenshots


elenabaurkot (commented):
Lgtm!

rshipp (approved):

PR #1100, opened 2022-06-15
Bulk attendance CSV import/export

rshipp:
# Description

Messy initial PR to get something in front of regional consultants. See linked task for more information.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202276177659407/f

# Screenshots

PR #1109, opened 2022-06-24
Attendance import/export, second pass

rshipp:
# Description

Followup to #1100. Fixes some issues, refactors into services, and adds specs. 

When reviewing, you may first want to refer back to #1100, since we skipped review to get that out for the demo.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202276177659407/f


elenabaurkot (commented):
Lgtm 👍

app/controllers/attendance_event_submissions_controller.rb
```diff
@@ -167,63 +150,22 @@ def bulk_import_new
...
-      @unsaved_answers << AttendanceEventSubmissionAnswer.new(
-        attendance_event_submission: @attendance_event_submission,
-        for_user_id: row[CSVHeaders::PLATFORM_USER_ID.parameterize.underscore].to_i,
-        in_attendance: to_boolean(row[CSVHeaders::IN_ATTENDANCE.parameterize.underscore]),
-        late: to_boolean(row[CSVHeaders::LATE.parameterize.underscore]),
-        absence_reason: row[CSVHeaders::ABSENCE_REASON.parameterize.underscore]&.strip,
-      )
+    begin
+      @unprocessed_rows, @unsaved_answers = service.run
+    rescue CSV::MalformedCSVError
```
sadleb, line 162: Should we capture the exception in Sentry and add something to Honeycomb?

app/services/export_attendance_csv.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      # Sort by cohort, then last name, then loop over each.
+      # Note in order to sort_by with an array like this, there can't be any
+      # values inside the array that are unsortable (e.g. nil, true, false).
+      # That's what the `cohort || ''` part's doing, is changing nil to an
+      # empty string so it can be sorted appropriately.
+      @all_fellow_users.sort_by { |u| [u.cohort(@course)&.name || '', u.last_name] }.each do |user|
+        answer = @answers.find_by(for_user: user)
+
+        # Make sure these match the order defined in the CSV headers!
```
sadleb, line 26: Maybe add a comment where `ALL_HEADERS` is defined alerting devs to come here and update this if they change the order?

app/services/import_attendance_csv.rb
```diff
@@ -0,0 +1,82 @@
...
+                    encoding: 'bom|utf-8',
+                    converters: lambda { |f, _| f&.strip })
+      .map(&:to_h)
+
+    # Filter
+    unprocessed_rows = []
+    unsaved_answers = []
+    data.each do |row|
+      # Add to unprocessed if "Present?" is empty or if the user is not in @fellow_users.
+      user_id = row[AttendanceEventSubmissionsController::CSVHeaders::PLATFORM_USER_ID.parameterize.underscore]&.to_i
```
sadleb, line 30: I think you can do something like the following to shorten these:
```
class ImportAttendanceCsv
  class CSVHeaders < AttendanceEventSubmissionsController::CSVHeaders # just so we can use a shorter name
  ...
  user_id = row[CSVHeaders::PLATFORM_USER_ID.parameterize.underscore]&.to_i
```
or even define something like
```
def get_value_for_column(row, column_name)
  row[column_name.parameterize.underscore]
end
```
so you can do
```
user_id =  get_value_for_column(row, CSVHeaders::PLATFORM_USER_ID)
```

sadleb (approved):
More nit-picky feedback than anything else

app/controllers/attendance_event_submissions_controller.rb
rshipp: no, this is explicitly a user-facing error. they just uploaded the wrong file. also, it can't happen, since the upload is filtered to CSVs.

PR #1112, opened 2022-06-28
Use LC last name only in test cohorts, make tag field more clear

elenabaurkot:
# Description
This is an update to the PR Add Generate Test Users Feature to make cohort names shorter. While demoing this we ran into an error where when adding two LCs to a cohort it resulted in cohort names being too long because they have a max length of 80set by Salesforce (https://ideas.salesforce.com/s/idea/a0B8W00000GdfnEUAR/increase-name-field-length-from-80-to-255). When an LC user is created this update sets `Label_Last_Name_Only__c` to true on the cohort they are being added to, this way only the last name of the LC gets populated in the cohort name. This also makes the tag field more clear to let users know that it helps clarify the purpose of test users and how it will be formatted in the last name. 

# Motivation and Context
Make the purpose of the tag field more clear, prevent errors from cohort name being too long

# Screenshots
Screen Shot 2022-06-28 at 12 06 22 PM


sadleb (approved):
Lgtm, though I’m not familiar with the `small` html tag and whether it’s being used properly here.

rshipp (approved):

PR #1126, opened 2022-07-20
clean up noisy test logs

tomlovett:
# Description
Many tests in the codebase are expected to throw errors every time they are run; for the sake of clarity and runtime, those errors will not be logged when they are run as part of tests. (Note: This will silence unexpected errors in the logs as well; if you run into unexpected errors, removing `... unless ENV["RAILS_ENV"] == "test"` will return error-logging to your console. 

There were several tests that used `expect{...}.not_to raise_error(SpecificError)`. This syntax raises warnings in the console because this criteria will pass false-positive even if a different type of error is raised. So, changing it to less-specific `.not_to raise_error` is actually more robust (and less noisy).

# Motivation and Context
There are many unnecessary logs in our test suite that slow down and add clutter to test runs. There are also several warnings for test criteria to be more specific; updating that test criteria makes the tests both more robust and less noisy.

Task: 
https://app.asana.com/0/1201131148207877/1202621971606154


tomlovett (commented):
Context and explanatory notes for changes.

rshipp (approved):

elenabaurkot (approved):
This looks good to me 👍  Just to clarify, we should not expect not to raise specific errors, but we can expect to raise specific errors @tomlovett, is that correct?

tomlovett:
> This looks good to me 👍 Just to clarify, we should not expect not to raise specific errors, but we can expect to raise specific errors @tomlovett, is that correct?

Yup! `expect{ ... }.to raise_error(BlahError)` is totally good. `expect{ ... }.not_to raise_error(BlahError)` is the dangerous false positive, because the second statement will still pass with any other error. You can use `.not_to raise_error` instead, since that will catch all error types.

PR #1128, opened 2022-07-20
remove extraneous instances of `..._id` from models

tomlovett:
# Description
Replaces validation calls of `:user_id` with `:user`, since that is more idiomatic.

Tests were updated where needed to point at the model, rather than the `:..._id`

# Motivation and Context

Task: 
https://app.asana.com/0/1174274412967132/1199613829896323/f


elenabaurkot (commented):
This looks good to me - does this just validate an actual instance of the model exists instead of validating that theres an id for the instance @tomlovett?

app/models/discord_server_channel.rb
```diff
@@ -4,9 +4,8 @@ class DiscordServerChannel < ApplicationRecord
   # Note discord_server_id refers to local `DiscordServer.id`, not
   # the external Discord `server.id`.
   validates :name, :discord_server_id, :discord_channel_id, :position, presence: true
```
rshipp, line 6: this `discord_server_id` should probably be `discord_server` too. but still not `discord_channel_id`, which is a numerical string rather than a fkey bigint.

rshipp (commented):
the changes in this pr look good, but one comment on a missed id

tomlovett:
> This looks good to me - does this just validate an actual instance of the model exists instead of validating that theres an id for the instance @tomlovett?

Yup! I'm going off the ticket because I'm not an expert on this, but yes, it's performing the same check, except this updated syntax is more the way it's supposed to be done. - This isn't a major change/fix, just a way for me to get more tickets done. :)

app/models/discord_server_channel.rb
tomlovett: Discussed with @rshipp, and we'll leave it as is because changing this throws strange errors in `discord_server_channel_spec`

PR #1129, opened 2022-07-21
Change capstone eval scoring to run in background

rshipp:
# Description

Move logic to a background job, to avoid a timeout when computing a large number of scores. Also refactors some code to avoid referencing the service from the views.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202199801976847/f

# Screenshots

Screen Shot 2022-07-21 at 2 57 50 PM
Screen Shot 2022-07-21 at 2 58 02 PM


app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -29,6 +31,17 @@ class CapstoneEvaluationResultsController < ApplicationController
...
+
+    if current_user.is_enrolled_as_student?(@course)
+      service = GradeCapstoneEvaluations.new(
+        @course,
+        @lc_course,
+        @lti_launch
+      )
+      @user_has_grade = service.user_has_grade?(current_user)
+      @fellow_total_score = service.fellow_total_score(current_user).round(2)
+      @graded_results = service.grade_capstone_eval_questions(current_user, false)
```
elenabaurkot, line 43: This helps to see how to put the complicated logic in the service without having to expose the service to the view

spec/services/grade_capstone_evaluations_spec.rb
```diff
@@ -367,7 +369,7 @@
 
     subject { grade_capstone_evaluations_service.users_with_published_submissions() }
 
-    it 'returns an array of users (Fellows and LCs) with already graded submissions' do
+    xit 'returns an array of users (Fellows and LCs) with already graded submissions' do
```
elenabaurkot, line 372: This is my first time seeing `xit` instead of `it` - from my understanding it skips certain tests or marks them pending (I know we already have a handful that are pending in the current tests), just curious what is the purpose of marking certain tests pending? @rshipp

elenabaurkot (commented):
I know sometimes when we run something as a job we send a success or failure email at the end do we see any need for this @rshipp? If there's no reason to, then this looks good to me

spec/services/grade_capstone_evaluations_spec.rb
rshipp: for these, it's basically the same as commenting them out. i wanted them to not run, but I wasn't sure if we'd want to move/fix them later. how important do you think these tests are, should we try to find another way to test the same functionality?

rshipp:
i'm not sure if we need an email or not. is this likely to fail, do you think? I can bring this up at the next meeting and see what others think too.

spec/services/grade_capstone_evaluations_spec.rb
elenabaurkot: ah I see because these aren't in the service anymore so we don't actually need these specs, is that right? I don't think their vital, it's more to make sure the correct students are showing in the view properly it doesn't handle the actual grading part so it's probably fine

elenabaurkot:
I don't see it being super likely to fail, maybe if we start having issues with that where it does fail and we can see the grades aren't there, then we could add the email part in then if we need it, does that sound OK?

rshipp:
that's kinda what i was thinking. i'll ask next week just to double check though.

elenabaurkot:
Okay sounds good, approving for now 👍

elenabaurkot (approved):

rshipp:
@elenabaurkot Added the mailer stuff in the last commit, does that look ok?

elenabaurkot (approved):
Lgtm 👍

PR #1131, opened 2022-07-27
Add discord user info to Honeycomb before reset

rshipp:
# Description

When a user clicks the button to reset their discord account, do an extra API call and send their user information to Honeycomb (as long as their token isn't expired).

# Motivation and Context

Make it easier to tell which account should be kicked from Discord, by adding the old username#discriminator information to the Honeycomb trace.

Task: https://app.asana.com/0/1201131148207877/1200777784010605/f (related)


elenabaurkot (approved):
This looks good to me 👍

PR #1132, opened 2022-07-28
Delete assignments

elenabaurkot:
# Description
This update gives the ability to delete and publish latest for Projects, Surveys and Attendance Events that already have submissions created for the assignment (like we do for Modules). Before deleting or publishing latest, the user will see a warning page showing all the students who have submissions (for attendance events it will see the number of attendance submissions with answers). Projects and Attendance Events automatically have submissions created when they are opened so we only show the warning page if there are `ProjectSubmissionAnswers` or `AttendanceEventSubmissionAnswers` associated with the submissions (otherwise we don't care if they get deleted). When we publish latest, it does not delete scores in the Gradebook (it does delete RubricGrades for projects if there are any), but it seems unlikely that we would be deleting assignments that have already been graded- it could happen for Impact Surveys, but they get a score of 0/0 so it does not matter for this as much.

# Motivation and Context
Make it easier for designers to delete assignments that have already been opened or have been worked on. Also give them the ability to publish latest on assignments that have already been opened or worked on. 

Task: https://app.asana.com/0/1201131148207877/1202220042010363/f

# Screenshots
Before Unpublish pages for projects, surveys and attendance (note attendance looks slightly different than the other two because I thought it made more sense to show number of submissions instead of student names since they don't actually work on attendace):
Screen Shot 2022-07-28 at 1 39 33 PM

Screen Shot 2022-07-28 at 1 41 59 PM

Screen Shot 2022-07-28 at 1 43 37 PM

Before publish latest pages that show for Surveys and Projects (there is no publish latest for attendance events)
Screen Shot 2022-07-28 at 1 45 47 PM

Screen Shot 2022-07-28 at 1 49 14 PM


tomlovett:
At a quick glance, it's looking good so far! I really like how you abstracted a lot of the functions, it makes it very readable. 👍

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -53,9 +62,8 @@ def unpublish
   # Shows any messages related to unpublish that caller may have to act on before
   # allowing it.
   #
-  # TODO: instead of just saying that you can't unpublish, show a list of students whose data
-  # would get blown away and provide a button to unpublish anyway and purge that data.
-  # https://app.asana.com/0/1174274412967132/1200410628872768
+  # Instead of just saying that you can't unpublish, show a list of students whose data
+  # would get blown away and provides a button to unpublish anyway and purge that data.
   def before_unpublish
```
elenabaurkot, line 58: For the `unpublish` method in this file I was originally deleting the Rise360ModuleVersion on unpublish too, but realized when running specs it explicitly doesn't want me to do that. I am curious though when I deleted it, and checked the locally to make sure it wasn't there I was still able to publish it after with no problem- I think I am missing something here, do you know the reason for not deleting it @rshipp?

rshipp: not sure I understand. could it be making a new `Rise360ModuleVersion` when you publish, by copying the data from the `Rise360Module`?

elenabaurkot: That is kind of what I thought would happen, so I was confused why I shouldn't delete the `Rise360ModuleVersion`

app/controllers/concerns/publishable.rb
```diff
@@ -106,6 +106,15 @@ def publish
   end
 
   def publish_latest
+    if can_destroy_references?
+      authorize controller_path.classify.to_sym
```
rshipp, line 110: i think this authorize call needs to be moved to the top level? i'm not seeing another call in this method, which means if the `if can_destroy...` check returns false, this method will raise a `Pundit::AuthorizationNotPerformedError` instead of rendering the view as you would expect.

app/models/course_attendance_event.rb
```diff
@@ -29,4 +29,31 @@ class CourseAttendanceEvent < ApplicationRecord
...
+      return true if AttendanceEventSubmissionAnswer.where(attendance_event_submission_id: id).present?
+    end
+    false
+  end
+
+  # We only care about the number of submissions that have answers
+  # We don't care about ones created just from opening an attendance event
+  def submissions_with_answers_count
+    count = 0
+    attendance_submissions = AttendanceEventSubmission.where(course_attendance_event_id: self)
```
rshipp, line 47: is this different than calling `attendance_event_submissions` here? we do have an association defined at the top of this model.

app/models/course_attendance_event.rb
```diff
@@ -29,4 +29,31 @@ class CourseAttendanceEvent < ApplicationRecord
...
+    attendance_submissions.each do |submission|
+      count += 1 if AttendanceEventSubmissionAnswer.where(attendance_event_submission_id: submission.id).present?
+    end
+    count
+  end
+
+  # Get ids of AttendanceEventSubmissions for this CourseAttendanceEvent
+  # (with or without answers)
+  def attendance_submission_ids
+    AttendanceEventSubmission.where(course_attendance_event_id: self).pluck(:id)
```
rshipp, line 57: same question here as above, can we use the existing association?

app/models/course_project_version.rb
```diff
@@ -10,4 +10,39 @@ def new_submission_url
...
+
+  # Returns true if there are any ProjectSubmissionAnswers associated with this CourseProjectVersion
+  # Anytime a student (Fellow) opens a Project, a ProjectSubmission record is created.
+  # We only care about ProjectSubmission records if they have ProjectSubmissionAnswers
+  def has_student_data?
+    project_submission_ids = ProjectSubmission.where(course_custom_content_version_id: self).pluck(:id)
+    project_submission_ids.each do |id|
+      return true if ProjectSubmissionAnswer.where(project_submission_id: id).present?
+    end
+    false
```
rshipp, line 22: similar question, but since there's no existing association, can we add one?

app/models/course_project_version.rb
```diff
@@ -10,4 +10,39 @@ def new_submission_url
...
+  end
+
+  # The list of Users who are students that have submitted answers for this Project
+  # and therefore have associated data.
+  def students_with_data
+    project_submissions = ProjectSubmission.where(course_custom_content_version_id: self)
```
rshipp, line 28: see above comment

app/models/course_project_version.rb
```diff
@@ -10,4 +10,39 @@ def new_submission_url
...
+    submission_ids_with_answers = project_submissions.select do |submission|
+      ProjectSubmissionAnswer.where(project_submission_id: submission.id).present?
+    end
+    users_with_submissions = User.where(id: submission_ids_with_answers.pluck(:user_id))
+  end
+
+  # Return project_submission_ids of ProjectSubmissions that have answers
+  def get_projects_with_answers
+    project_submissions = ProjectSubmission.where(course_custom_content_version_id: self)
```
rshipp, line 37: see above comment

app/models/course_survey_version.rb
```diff
@@ -9,4 +9,21 @@ def new_submission_url
       protocol: 'https',
     )
   end
+
+  # Returns true if there is any student data that references this instance.
+  def has_student_data?
+    SurveySubmission.where(course_custom_content_version_id: self).present?
```
rshipp, line 15: another instance of one that'd need a new association defined

app/models/course_survey_version.rb
```diff
@@ -9,4 +9,21 @@ def new_submission_url
...
+  end
+
+  # The list of Users who are students that have submitted this Survey in Canvas
+  # and therefore have associated data.
+  def students_with_data
+    user_ids = SurveySubmission.where(course_custom_content_version_id: self).pluck(:user_id)
```
rshipp, line 21: see above

app/views/course_attendance_events/before_unpublish.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+<% content_for(:breadcrumbs) do %>
+  <li class="breadcrumb-item"><%= link_to 'Course Management', courses_path %></li>
+  <li class="breadcrumb-item"><%= link_to 'Edit Course', edit_course_path(@course_attendance_event.course) %></li>
+  <li class="breadcrumb-item active" aria-current="page">Unpublishing</li>
+<% end %>
+
+<div class="col-md-8">
+<strong><p>Number of attendance submissions for this Course Attendance Event: <%= @course_attendance_event.submissions_with_answers_count %></p></strong>
+<p>There are already attendance submissions for this Course Attendance Event. If you choose to continue, <strong><mark>THIS ATTENDANCE DATA BE BLOWN AWAY</mark></strong> and the Course Attendance Event will no longer show up in Canvas.</p>
```
rshipp, line 13: small typo
```suggestion
<p>There are already attendance submissions for this Course Attendance Event. If you choose to continue, <strong><mark>THIS ATTENDANCE DATA WILL BE BLOWN AWAY</mark></strong> and the Course Attendance Event will no longer show up in Canvas.</p>
```

app/controllers/concerns/publishable.rb
elenabaurkot: that makes sense, moved the authorize to the top 👍

app/models/course_attendance_event.rb
elenabaurkot: ah it is in fact the same thing, just updated to use the association instead

app/views/course_attendance_events/before_unpublish.html.erb
elenabaurkot: ah good catch, updating to this

app/controllers/discord_signups_controller.rb
```diff
@@ -212,10 +213,6 @@ def lti_launch_url
     launch_discord_signups_url(protocol: 'https')
   end
 
-  def can_publish_latest?
-    false
```
elenabaurkot, line 216: The Discord assignment doesn't actually have the ability to `publish_latest` so I took this out here, in publishable it sets `can_publish_latest` to true by default and overrides in controllers that are actually using `publish_latest`. Now looking at this again after having removed it, I'm wondering if it was added here for safety measure if there was somehow a way to `publish_latest` for the assignment, not sure if we need it or not

rshipp: it's weird that it's true by default. but if there's no way to publish latest for discord, it makes sense to have it set to false in the discord controller

rshipp (approved):
i don't know if the discord/publish latest comment breaks anything, but i'm approving the PR assuming it doesn't, and you can do whatever you think makes sense with that function

app/controllers/discord_signups_controller.rb
elenabaurkot: going to go with what you're saying here and leave this function in that sets it to false @rshipp

PR #1133, opened 2022-07-28
Fix capstone eval security issue

rshipp:
# Description

Fix a security issue with an improperly exposed `state` param in the `GROUP PROJECT: CAPSTONE CHALLENGE: TEAMWORK` assignment, maintain expected functionality for the iframe'd grade details for this assignment, and fix an unrelated typo in the exception handler (introduced in #1129). 

This PR adds a new, minimal action to the controller in order to get a unique URL that we can add to the `state` allowed-list in LtiLaunchController. That way we can remove the main `launch` action from that list, and prevent people who are just opening the assignment as normal from seeing a `state` param.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202676012728466/f


app/controllers/capstone_evaluation_results_controller.rb
```diff
@@ -44,6 +44,14 @@ def launch
     end
   end
 
+  # IMPORTANT NOTE!
+  # This action should only be used when inside a Canvas iframe, from create_lti_submission.
+  # It is allowed to expose the unsafe `state` param by LtiLaunchController.
```
elenabaurkot, line 49: When it's used inside of the Canvas ifram from `create_lti_submission` is it being exposed, or in this instance it doesn't get exposed? Also, how was it being exposed before @rshipp?

rshipp: When it's inside the iframe, there isn't an easy way to accidentally share the URL, because it's not visible in the browser URL bar. Anything in a Canvas iframe has to use the state param, because that's how it authenticates without 3rd-party cookies.

It was exposed before by LtiLaunchController; any URL in this list https://github.com/bebraven/platform/pull/1133/files#diff-1cc94f8305c6c7787c0b817518c0b004af88947cf2b6fa6a8363c1f14a455890L102-L103 gets the state param added to the URL. Where before we were using the same URL for launching the assignment and viewing it in the gradebook (the former needs to *not* have a state param, the latter needs to have it), we're now using `launch` to launch the assignment and `details` in the gradebook.

elenabaurkot: ah OK, that makes sense- this looks good to me 👍

elenabaurkot (approved):

PR #1136, opened 2022-08-02
Add more fields to Honeycomb automatically

rshipp:
# Description

I spent some time trying to figure out something useful for the original task, and eventually decided it didn't make sense as written. Instead, I added a default `add_to_honeycomb_span` method that can now be called on any model object (e.g. `course_project_version.add_to_honeycomb_span`) and called that method from ApplicationController, so any controller will now have autoinstrumentation for its own model. (E.g. visiting `/custom_contents/8/edit` will now add information about that CustomContent record to the Honeycomb span.)

If you're writing code that uses database records, you can now call `my_record.add_to_honeycomb_span` to add information about that record to the current span, without having to think about what fields might be important.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202052488545512/f

# Screenshots

Example of visiting https://platformweb/custom_contents/8/edit in my dev env:


elenabaurkot (approved):
This looks good to me 👍

tomlovett (approved):
Looks good to me!

sadleb:
Creeping. I freaking love this PR.

PR #1138, opened 2022-08-08
Upgrade CKEditor

rshipp:
# Description

Upgrade CKEditor5 packages from 26.x to the latest version as of this PR (35.x). Also upgrade related packages as-needed. 

Notes to inform designers:
* There's a new drag-and-drop support, but it may have some weird bugs. If you decide it's more painful than it is helpful, let us know and we can turn that feature off.
* If you notice any other bugs with the content editor, please let us know so we can take a look.

Some additional context in this upstream issue: https://github.com/ckeditor/ckeditor5/issues/12221

This PR requires a **container rebuild**, and may also require additional steps. I recommend planning for some additional troubleshooting time when you pull this down and rebuild, and let me know as soon as you run into any issues so I can help.

# Motivation and Context

Task: https://app.asana.com/0/1174274412967132/1200701153958461/f


app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -382,4 +384,39 @@ export default class ContentCommonEditing extends Plugin {
...
+            const selectedElement = viewDocument.selection.getSelectedElement();
+
+            // HACK: Fix the textArea/textInput onclick selection behavior.
+            const targetViewElement = domEvtData.target;
+            if ( !targetViewElement ) {
+                // Return early, because there's nothing to do.
+                return;
+            }
+            if ( ( targetViewElement.name === 'input' && targetViewElement.getAttribute( 'type' ) === 'text' ) ||
+                 ( targetViewElement.name === 'textarea' ) ) {
```
tomlovett, line 406: To abstract this function, you could do something like:

`if isTextInputOrTextArea(targetViewElement) {...}`

And then insert this function outside of the class, below the `import` statements:

```
const isTextInputOrTextArea = (elem) => (elem.name === 'textarea') || (elem.name === 'input' && elem.getAttribute('type') === 'text')
```

I'd also recommend putting simpler boolean checks first, for the sake of readability.

rshipp: is there a reason you're suggesting this `const x = ()` syntax rather than standard `x() {}` function syntax? i think the latter would let us move this down to the bottom of the file rather than the top, but idk about any other tradeoffs

tomlovett: They're identical in terms of functionality, I just prefer the `const` syntax in these cases because it'll be more concise.

I don't think there's any particular gain to declaring the function inside the class. In general, I prefer to have helper functions higher up in the file and the primary functions lower. This isn't a React component, but it's very good practice there, and we may as well aim to be consistent.

elenabaurkot (approved):
This lgtm

app/javascript/ckeditor/contentcommonediting.js
rshipp: ok. my preference is the opposite, but I'm fine with using this here. it seems to still work after this change.

app/javascript/ckeditor/contentcommonediting.js
```diff
@@ -402,8 +405,7 @@ export default class ContentCommonEditing extends Plugin {
                 // Return early, because there's nothing to do.
                 return;
             }
-            if ( ( targetViewElement.name === 'input' && targetViewElement.getAttribute( 'type' ) === 'text' ) ||
-                 ( targetViewElement.name === 'textarea' ) ) {
+            if ( isTextInputOrTextArea( targetViewElement ) ) {
                 // The target is a textInput or textArea.
```
tomlovett, line 409: I think you can take this line out now, too. LGTM

tomlovett (approved):

app/javascript/ckeditor/contentcommonediting.js
rshipp: which line, 409? github comments show up in weird places

tomlovett: Yeah, 409. By using a function that states its checking for textarea or textinput, we don't need a comment to explain that we're doing that.

sadleb:
Less code than I expected to make this happen. Nice.

PR #1142, opened 2022-08-11
Fix Honeycomb overages, try #1

rshipp:
# Description

Cache `accelerator_course` and `lc_playbook_course` lookups on HerokuConnect::Program objects, so they're only fetched once at the start of the program sync, instead of on every single participant sync.

It's hard to make local effectively match prod since this requires programs/participants, so this will be tested more fully in prod.

# Motivation and Context

# Screenshots

Before, with Course Loads:

After, with no Course Loads:


elenabaurkot (approved):
Lgtm 👍

tomlovett (approved):
Really nice find! LGTM

PR #1144, opened 2022-08-11
Fix Honeycomb overages, try #2

rshipp:
# Description

Followup to #1142.

While the last PR reduced spans significantly, we're still getting quite a lot from the same types of things.

Before and after previous PR:

You can see the how the purple line dropped after the deploy (dashed vertical line furthest to the right). However, we're still way over the daily event expectations, which will put us over the event cap by the end of the month.

This PR takes it a step further, and filters out ActiveRecord autoinstrumented spans where we're doing a select query (e.g. `Course Load`), and the result was found in the Ruby cache rather than querying postgres. We don't lose much information from this, since we add relevant IDs to spans, and the reduction in total spans should be significant.

See below for how drastic the number of cached load events vs all other events is:

# Motivation and Context


elenabaurkot (approved):
lgtm 👍

tomlovett (approved):
Nice! :+1: :+1:

sadleb:
24 million spans 👀

PR #1146, opened 2022-08-12
Fix a failing Sync spec

rshipp:
# Description

Followup to fix a spec I broke in #1142. See code comment for why this is a broken spec, and not broken model behavior.

# Motivation and Context


tomlovett (approved):
LGTM

PR #1147, opened 2022-08-12
CKEditor: make all dropdowns required

rshipp:
# Description

Mark all dropdowns as required in CKEditor components. When existing content is loaded in the editor, it will be automatically migrated, and can be re-published to take effect in our Impact Surveys and Projects.

This PR is based on #1138, so it should be merged after that.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202777829235221/f

# Screenshots


tomlovett:
To me it seems like the task is asking for the option to make a select required, rather than making all selects required by default. Can we double-check this with Jenna at stand-up?

elenabaurkot:
I think based on [this slack conversation](https://bebraven.slack.com/archives/C01166ZLR1V/p1660319849942959) it seems the decision was to make all selects required (not sure if that was finalized or not) @tomlovett @rshipp - if that is the case then this looks good to me 👍

elenabaurkot (approved):
Lgtm under the assumption we do want all selects to be required

rshipp:
closing, see task for more details

PR #1148, opened 2022-08-15
Rename user.contact_id to user.salesforce_contact_id

tomlovett:
# Description
I searched the codebase for every instance of `salesforce_id` and changed the ones referring to users to `salesforce_contact_id`. Other models have `salesforce_id`, so I had to make sure each particular instance was actually for users. In addition to manual checks, I ran `rspec` on the entire codebase and all tests passed.

# Motivation and Context
`User.salesforce_id` is slightly ambiguous in that it could refer to either its Contact or Participant records in Salesforce. This change makes it clear that we're referring to the Contact record.

Task: 
https://app.asana.com/0/1201131148207877/1202728460163281


elenabaurkot (approved):
This lgtm 👍

rshipp:
the changeset seems fine to me. could you check the Honeycomb triggers too https://ui.honeycomb.io/braven/triggers just to make sure there's nothing in there referencing `app.user.salesforce_id`?

tomlovett:
> the changeset seems fine to me. could you check the Honeycomb triggers too https://ui.honeycomb.io/braven/triggers just to make sure there's nothing in there referencing `app.user.salesforce_id`?

I looked through them and there wasn't anything for `app.user.salesforce_id`. (Also, I didn't fully understand what Jenna meant by this [comment](https://app.asana.com/0/0/1202728460163281/1202764730797350/f), so thank you for pointing it out.)

PR #1150, opened 2022-08-18
DRAFT: Create GitHub action for rspec and rubocop

tomlovett:
# Description
This will set up a CI task through GitHub actions that will run `rspec` and `rubocop` on each commit and pull request.
This is the first attempt at running this action, and may likely be broken on first attempt. I don't see any reason why this would negatively affect the repo, but the action may not run properly this time around. If so, it will be updated in another PR.

Having files in `.github/workflows` taps into the GitHub Actions capability. 
`on: push: [ main ]; pull_request: [ main ]` will trigger the `test` and `lint` tasks.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202807787148119


tomlovett:
Putting PR on hold because there is just too much funkiness and strange behavior. The url-parsing failure is particularly strange. Plus, I had to edit existing migrations just to get the app running on GitHub, and that's not good.

Leaving it around in case inspiration strikes, or to help future attempts.

PR #1151, opened 2022-08-23
Fix Zoom link sync for participants

elenabaurkot:
# Description
A small typo in the `sync_zoom_links_for_participant` service causes some participants Zoom links to be cleared out after getting set in Salesforce. On July 29, NIU had this occur for a handful of participants and on August 1 so did Spelman. For both of them the Zoom links were generated in the first sync and set in Salesforce and then in the following sync they were cleared out. The `SALESFORCE_SYNC_MAX_DURATION` is set to 15 minutes and if the sync is unable to finish in that time it exits early and finishes in the next sync. This issue with the clearing of the Zoom links only occurred the two times the sync exited early for these two programs. In the first sync, the `new_join_link_1` gets set to `nil` (line 42 in `sync_zoom_links_for_participant`), but then gets updated once the participant is registered (line 53). The second time the sync ran, the `new_join_link1` was set to an empty string because `clear_out_meeting_1_link` returned true when it should have returned false (line 46). `clear_out_meeting_1_link?` should have returned false because `zoom_meeting_1_configured?` was `true`, but instead it checked `zoom_meeting_2_configured?` which it wasn't since these participants don't have a zoom_meeting_2 set. It returned true that the `@participant_sync_info.zoom_meeting_id_1_changed?` (line 197) because the sync info didn't save in the last sync since it didn't complete so it was comparing the meeting id 1 to nil so it did change and therefor cleared out the meeting link. In the second sync it didn't run register_participant again (line 51) since the participant was already registered in the previous sync and `generate_zoom_meeting_1_link?` returned `false` since `zoom_meeting_1_configured?` was true (line 174). This is why the links got cleared out the second time around. Fixing this typo should prevent it from happening again. 

# Motivation and Context
Make sure participant Zoom links don't get cleared by accident from the sync. 

Task: https://app.asana.com/0/1201131148207877/1202801870467496/f


tomlovett (approved):
LGTM 👍 

(was a one-line fix, so it was easy to find, right?)

rshipp (approved):
wow..... that's wild. great find

PR #1152, opened 2022-08-24
Add Scout for memory monitoring in prod

rshipp:
# Description

Since we don't have any existing tools for memory monitoring / profiling, this PR adds Scout, which has a free Heroku addon that can help us see more details on what's using memory. I expect this additional gem to be temporary, and removed once we have a better idea what's happening.

# Motivation and Context

Task: https://app.asana.com/0/1201131148207877/1202850010343833/f


elenabaurkot (approved):
Lgtm!

PR #1153, opened 2022-08-25
add ruby and es6 linters to CI

tomlovett:
# Description
This will set up a CI task through GitHub actions that will run `eslint` and `rubocop` on each commit and pull request. It is not mandatory to pass these checks in order to merge code, but it is encouraged.

This will only run on files that have been changed on the branch in question. In addition to that check, ESLint only goes through files ending with `*.js` or `*.jsx`. RuboCop is set up to only check files ending in `*rb.` that are not in the `db/` folder. (It doesn't like our migrations.) We may have to add additional checks if it triggers too many false positives. 

We can modify/update this as we go

# Motivation and Context
This will help clean up and standardize our code.,

Task: https://app.asana.com/0/1201131148207877/1202807787148119

The changed files task gathers all the names of the changed files, then we send them through an each loop, and selectively run the linter on them. (Trying to write my own "grab changed files" script was torture and wildly ineffective.)

Screen Shot 2022-08-25 at 2 51 56 PM


tomlovett:
I've been looking for a way to remove that `continuous-integration/heroku` from the list - since it keeps the icon in the tab as "pending" - but haven't found the way, yet. I'll keep reading docs, but if you know how I'm all for it.

rshipp:
is there anywhere we can see what this looks like when it finds something to report on?

rshipp (approved):

PR #1154, opened 2022-08-26
run `yarn upgrade` to resolve security alerts

tomlovett:
# Description
We have multiple security alerts on our npm packages. `yarn upgrade` updates our npm to most recent stable versions, which will resolve some of our security alerts.

Task: N/A


rshipp:
this is good, but let's be sure there's a task and someone ready to prioritize QA after we merge it, so we don't get another instance of random JS breaking without us knowing about it. I'm also worried it will conflict with #1138, which we're still waiting on QA availability to merge.

PR #1155, opened 2022-09-06
Fix Honeycomb overages, try #3

rshipp:
# Description

Followup for #1142 and #1144. Remove all ActiveRecord `Load` events, not just the cached ones.

# Motivation and Context

# Screenshots

Bulk of the spans still coming from User Load autoinstumentation events.

Back over the daily event target.


tomlovett (approved):
LGTM

PR #1156, opened 2022-09-07
Set reminder to re-do Zoom calls after a rate limit error

tomlovett:
# Description
If we send too many requests to the Zoom API, set reminder to try again on the next sync after the rate limit has expired. (Zoom's rate limit expires on the next UTC day.)

(Additional formatting changes added per RuboCop.)

# Motivation and Context
There were Zoom meetings with outdated information for the participants. Tracing back what happened, updates to those meetings were being rejected because we had exceeded Zoom's rate limit of changes for the same email/meeting pair. 

Task: https://app.asana.com/0/1201131148207877/1202895413804903


app/services/sync_zoom_links_for_participant.rb
```diff
@@ -108,6 +109,30 @@ def unassign_zoom_links
...
+    # Gets midnight in the participant's program on the retry date
+    # We don't want to use midnight utc because it can fall during Learning Lab for some program
+    # Midnight in any US time zone is earlier than midnight utc so it will definitely be past midnight utc
+    # and we cam safely retry updating the participant's Zoom info
+    midnight_in_user_timezone = zoom_retry_date_utc.in_time_zone(
+      HerokuConnect::Program.find_by(sfid: @participant_sync_info.program).default_timezone__c
+    )
+
+    SalesforceAPI.client.update_participant(
+      @participant_sync_info.sfid, { 'Zoom_Rate_Limit_Until__c' => midnight_in_user_timezone }
```
elenabaurkot, line 132: This all looks good to me here as long as some manual testing has been done to make sure time zones are actually being set correctly here 👍

tomlovett: I did not manually check this, so I'm going back over it now. Good catch!

elenabaurkot: Time zones always trip me up (I tend to reverse them haha)

app/services/sync_zoom_links_for_participant.rb
```diff
@@ -108,6 +109,25 @@ def unassign_zoom_links
...
+  def reschedule_zoom_api_request(e)
+    Sentry.capture_exception(e)
+    Honeycomb.add_field('sync_zoom_links_for_participant.skip_reason', e.message)
+    Rails.logger.debug(e.message)
+
+    # Schedule re-attempt for midnight of the following day, local time
+    # Midnight UTC can fall during Learning Lab for some programs, so we do not use that
+    # Midnight UTC occurs before midnight in any US time zone so rate limiting will have been reset
+    user_timezone = HerokuConnect::Program.find_by(sfid: @participant_sync_info.program).default_timezone__c
+    midnight_after_next = DateTime.now.in_time_zone(user_timezone).end_of_day + 1.days
```
elenabaurkot, line 124: If the current date / time is September 8 at 7:13pm, I think the time we want to get is September 9 at the very start of the day 12:00am - I think this would end up getting us September 9 at 11:59pm, so the very end of the day. I'm thinking we could do something like `DateTime.now.in_time_zone(user_timezone).beginning_of_day + 1.day` (which I think would get us 12:00 am Sept 9) or `DateTime.now.in_time_zone(user_timezone).end_of_day + 1.minute` (which would get us 12:01 am on Sept 9). Otherwise I think we would be waiting over 24 hours before retrying the Zoom API request. I'm not sure if this would change what we want to name the variable or if that would still make sense.

elenabaurkot (changes requested):

app/services/sync_salesforce_participant.rb
```diff
@@ -198,13 +199,13 @@ def set_canvas_ta_caseload_enrollments
   # well as masquerade as other users. We use TA accounts for staff as well as real
   # Teaching Assistants in an "admin" capacity of sorts.
   def set_admin_permissions
-    return if !@participant_sync_info.has_canvas_staff_permissions?
```
elenabaurkot: The only reason I can think of to leave it this way is so that we don't have to go through the other logic if the person doesn't have staff permissions we can just exit the function. I'm not sure whether the time saving there makes much of a difference or not or if it just seems cleaner to write it the other way @tomlovett?

app/services/sync_salesforce_participant.rb
```diff
@@ -287,11 +287,11 @@ def sync_local_accelerator_enrollment(cohort_schedule_section, cohort_section, t
       raise ArgumentError, "For Participant ID: #{@participant_sync_info.sfid} the ta_section is nil" if ta_section.nil?
       @participant_sync_info.user.add_role @participant_sync_info.accelerator_course_role, ta_section
       return
-    elsif cohort_schedule_section.nil? && cohort_section.nil?
+    else
```
elenabaurkot, line 290: I'm not positive, but I think we may need to check the above conditions specifically because we only want to raise this ArgumentError in the instance where the cohort schedule and cohort section are both nil - otherwise I think there could be scenarios where all of the conditions above this don't evaluate to true and this will automatically raise even if they do have a cohort schedule / section

app/services/sync_salesforce_participant.rb
```diff
@@ -247,21 +248,20 @@ def find_or_create_section(course, section_name, salesforce_id, section_type)
...
-
-    synced_user = SyncSalesforceContact.new(
-      @participant_sync_info.contact_id,
-      @salesforce_program.time_zone
-    ).run
-
-    @participant_sync_info.user = synced_user
-    @participant_sync_info.user_id = synced_user.id
-    @participant_sync_info.canvas_user_id = synced_user.canvas_user_id
+    if @participant_sync_info.contact_changed?
```
elenabaurkot, line 255: I think I'm a little confused about this part, it might be more easily explained in a call - the way I am reading the original code is we want to return if `@participant_sync_info.contact_changed?` and not SyncSalesforceContact, but here it looks like the opposite. I don't know whether we want this change or not, but I think I am having a little trouble fully comprehending what the original code is doing here and maybe this is what we want

elenabaurkot (changes requested):

app/services/sync_salesforce_participant.rb
tomlovett: These were all local refactoring that I was messing around with, but didn't mean to commit. It's been taken out 😅

app/services/sync_zoom_links_for_participant.rb
elenabaurkot: Ah good call, did not think about that, I think originally I was getting the next day in utc and then converting time zones, but I think this update makes more sense 👍

app/services/sync_salesforce_participant.rb
elenabaurkot: Ah okay haha disregard

elenabaurkot (approved):
This all looks good to me!

app/services/sync_zoom_links_for_participant.rb
```diff
@@ -171,6 +194,11 @@ def set_zoom_link_infos
   # Note that a force_update doesn't actually update the link unless other info has changed,
   # like the password for the meeting for example.
   def generate_zoom_meeting_1_link?
+    if rerun_after_rate_limit?
+      # Do not remove the rate-limit flag here, because we still need it for zoom_meeting_2
```
sadleb, line 198: janky tech debt ftw

app/services/sync_zoom_links_for_participant.rb
```diff
@@ -171,6 +194,11 @@ def set_zoom_link_infos
...
+      # This is janky TECH DEBT; shoould fix when combining both links into the same flow
```
sadleb, line 199: Janky tech debt ftw

spec/models/participant_sync_info_spec.rb
```diff
@@ -27,14 +25,14 @@
...
-    let(:participant) { build :participant_sync_info_with_cohort }
+      let(:participant) { build :participant_sync_info_with_cohort }
       it 'returns true' do
         expect(participant.is_mapped_to_cohort?).to eq(true)
       end
     end
 
     context 'when not mapped to cohort' do
-    let(:participant) { build :participant_sync_info, cohort: nil }
+      let(:participant) { build :participant_sync_info, cohort: nil }
```
sadleb, line 35: All these whitespace changes are making me think. I was just raw dogging this stuff in Vim. But with a bunch of people working on the same files, a standard vscode (or some other IDE) setup feels important.

tomlovett: Yup Brian! We've started using RuboCop to standardize our formatting. It'll take a while before we clean up the entire codebase, but we'll get there little-by-little!

PR #1158, opened 2022-09-09
Set reminder to re-do Zoom calls after a rate limit error

tomlovett:
# Description
If we send too many requests to the Zoom API, set reminder to try again on the next sync after the rate limit has expired. (Zoom's rate limit expires on the next UTC day.)

(Additional formatting changes added per RuboCop.)

# Motivation and Context
There were Zoom meetings with outdated information for the participants. Tracing back what happened, updates to those meetings were being rejected because we had exceeded Zoom's rate limit of changes for the same email/meeting pair. 

Task: https://app.asana.com/0/1201131148207877/1202895413804903


elenabaurkot (approved):
This looks good to me!

PR #1159, opened 2022-09-21
Limit ZoomSync to three checks per day

tomlovett:
# Description
Zoom only allows us to update registrations 3 times a day; because of sync running 10 min, it gets hit multiple times, if there are changes on a participant record or contact record

Now we have a code that will flag ppl who hit the rate limit after 24 hours, so that fixes a next day Zoom link refresh

Remaining issue is still that same day changes/rate limiting result in incorrect Zoom info which could affect LLs happening that day


elenabaurkot (commented):
Looked at this together on call - everything looks good, only thing we decided is if there is to use the program default_time_zone for cohort schedule times

app/services/sync_salesforce_program.rb
```diff
@@ -178,9 +182,11 @@ def sync_participants
         end
         next unless unique_participant
 
+        # Note that we run this before the enrollment stuff b/c we want to have the Zoom
+        # links in Salesforce even if the Participant Canvas sync fails.
         zoom_sync_success = with_participant_error_handling(participant) do
-          # Note that we run this before the enrollment stuff b/c we want to have the Zoom
-          # links in Salesforce even if the Participant Canvas sync fails.
+          break unless zoom_sync_cohort_schedule?(participant.cohort_schedule_id)
```
elenabaurkot, line 188: I'm looking into the sync and I am thinking we might end up having some issues here, the `sync_participants` method only runs if there are participants with changes, if there aren't changes the sync returns early ([see here](https://github.com/bebraven/platform/blob/b8f839b54c4a1790e244ef71466ffc0ead3f3367/app/services/sync_salesforce_program.rb#L89-L94)). It decides if there are changes in the ParticipantSyncInfo, changed? method, which just compares the current sync info to the previous sync info ([see here](https://github.com/bebraven/platform/blob/main/app/models/participant_sync_info.rb#L305-L316)). When the participant does have changes, it will run `sync_participants` and skip over syncing Zoom links unless it is one of the schedule times for a zoom sync. I think the problem is, if it skips over syncing Zoom links, it will still sync the rest of the participant and then when it is really time to sync zoom links it might not happen because if the participant doesn't have any changes then it will never get to this point in the code until the next time the participant has changes. The changed? method checking for ParticipantSyncInfo changes checks for zoom changes, which again just compares last sync info to this sync info - I'm wondering if we can update the `zoom_info_changed` method to check the ParticipantSyncInfo against the participant's ZoomLinkInfo instead so it actually picks up if there are new Zoom changes even if the participant has been properly synced otherwise  (I don't think that ZoomLinkInfo would have gotten updated correctly if we haven't synced Zoom links- not 100% have to look into that logic a little more around ZoomLinkInfo updates). We don't need to run the zoom sync at all if none of the participants had any Zoom changes based on that, but we would still probably need to update the last zoom sync time. I might be missing something here, but I think we should look into this in the working session.

elenabaurkot (changes requested):

app/models/participant_sync_info.rb
```diff
@@ -291,25 +291,24 @@ def initialize(last_sync_info, new_sync_info)
     # If you did, when they did get a Cohort Schedule then other parts of the
     # sync may not run b/c we think it was already synced.
     def should_sync?
-      if requires_cohort_schedule? &&
-         cohort_schedule_id.blank? &&
-         user_id.blank?
+      !(requires_cohort_schedule? && cohort_schedule_id.blank? && user_id.blank?)
+    end
```
tomlovett, line 295: Behaves the same as behavior, just cleaned up the syntax.

app/services/sync_salesforce_program.rb
```diff
@@ -81,10 +85,11 @@ def run
       end
 
       @participants = ParticipantSyncInfo.diffs_for_program(@salesforce_program)
+      @participants = participants_with_syncable_changes # filter out Zoom-only changes if not a Zoom sync time
       @count = @participants.count
       Honeycomb.add_field('sync_salesforce_program.participants.count', @count)
 
-      if @force_canvas_update == false && @force_zoom_update == false && @participants.none? { |p| p.changed? }
+      if @count == 0 && !@force_canvas_update && !@force_zoom_update
```
tomlovett, line 92: Changed order of the check to make it more readable.

PR #1162, opened 2022-09-30
Updated code to account for Participant status of 'Withdrew'

elenabaurkot:
# Description
Updated the code to account for the new Participant status of "Withdrew". The "Withdrew" status just means the participants dropped after the course add/drop date, which means they receive a "W" grade on their transcript, whereas "Dropped" participants dropped before the add/drop date and don't receive a grade. This status was added for the data team to make sure their data for course completers is accurate. For Platform, we treat the "Withdrew" status exactly the same as the "Dropped" status for a participant. If a Participant has a status of "Withdrew" or "Dropped", we want to make sure they are unenrolled/lose Canvas access and are kicked from their Discord server.

# Motivation and Context
The new "Withdrew" status was added to Salesforce and the sync was not accounting for Participants with this status, leaving them enrolled in their course and causing some errors. 

Task: https://app.asana.com/0/1201131148207877/1203048160470112/f


app/models/participant_sync_info.rb
```diff
@@ -184,6 +184,14 @@ def is_dropped?
     status == HerokuConnect::Participant::Status::DROPPED
   end
 
+  def is_withdrew?
+    status == HerokuConnect::Participant::Status::WITHDREW
+  end
+
+  def left_course?
+    self.is_dropped? || self.is_withdrew?
```
elenabaurkot, line 192: decided against adding `|| self.is_failed?` to this method because for my next task I have to make sure failed participants don't get kicked from Canvas / removed from the course

app/services/sync_salesforce_participant.rb
```diff
@@ -281,7 +281,7 @@ def sync_local_accelerator_enrollment(cohort_schedule_section, cohort_section, t
 
     # If they fail the course, treat that the same as dropping and remove their access.
     # We can revisit this in the future if we decide to do something else.
-    return if @participant_sync_info.is_dropped? || @participant_sync_info.is_failed?
+    return if @participant_sync_info.left_course? || @participant_sync_info.is_failed?
```
elenabaurkot, line 284: see comment [here](https://github.com/bebraven/platform/pull/1162/files#r984937920) for leaving `is_failed?` out of `left_course` method

elenabaurkot:
Before merging this I may need a reminder on Monday of how to run rubocop on the files I edited @tomlovett

tomlovett (approved):
Great stuff! 👍  👍

PR #1163, opened 2022-10-03
Make sure Failed Participants maintain Canvas access

elenabaurkot:
# Description
Fellows who failed the Accelerator were being removed from Canvas, we want them to maintain access to the course if they completed it, even if they failed. 

# Motivation and Context
We want Fellows who fail the Accelerator course to maintain access to course content

Task: https://app.asana.com/0/1201131148207877/1202942869924566/f


elenabaurkot:
I think I'll probably need to merge this after the withdrew status PR because I'll need to rebase on main after that is merged so I'm holding off on running the `rubocop` on this until after rebasing because i think a lot of the files I touching in here already had `rubocop` ran on it in that PR

tomlovett:
Given how massive the changes were in the previous PR, it might make sense to wait until we're both inbetween tasks and then do the "full codebase rubocop."

Don't feel obligated to run the rubocop lint on this PR. We can get it alter.

elenabaurkot:
> 



> Given how massive the changes were in the previous PR, it might make sense to wait until we're both inbetween tasks and then do the "full codebase rubocop."
> 
> Don't feel obligated to run the rubocop lint on this PR. We can get it later.

That sounds good 👍 👍

PR #1165, opened 2022-10-07
Don't send Zoom cancellation emails to Failed Participants

elenabaurkot:
# Description
This is an add-on to the [PR Make sure Failed Participants maintain Canvas Access](https://github.com/bebraven/platform/pull/1163). That PR made sure Fellows who failed the accelerator didn't get kicked out of Canvas. This makes sure when participants fail they don't have their Zoom registration cancelled and receive an email about it. Failed Participants are still course-completers so  they keep Canvas access and don't get removed from Zoom (even though the course is over - they are treated the same as Enrolled participants). 

# Motivation and Context
It feels weird for people to receive Zoom cancellation emails when they weren't dropped from the course. 

Task: https://app.asana.com/0/1201131148207877/1202942869924566/f


tomlovett (approved):
LGTM nice work!

PR #1167, opened 2022-10-13
Allow Highlander to handle requests to Lowlander subdomain

tomlovett:
# Motivation and Context
`join.bebraven.org` will sunset Nov 10. We still need to support users who go those routes. These (commented-out) changes will allow Highlander to accept requests from the Lowlander subdomain, and re-route them to Highlander/Canvas.

I tried testing this locally, but unfortunately the local instance of the app can handle `https://join.platformweb.com` on its own; so I could not verify that this solution actually properly handles traffic.

NOTE: You may have to add `127.0.0.1 join.platformweb` to the `/etc/hosts` file on your workstation.

Task:  https://app.asana.com/0/1201131148207877/1202865857905024


config/routes.rb
```diff
@@ -296,22 +302,18 @@
...
 
   # Catch-all route for missing ones to return an unauthorized response rather than a 404
-  get '*other', to: 'home#missing_route'
-  post '*other', to: 'home#missing_route'
-  put '*other', to: 'home#missing_route'
-  patch '*other', to: 'home#missing_route'
-  delete '*other', to: 'home#missing_route'
-
+  match '*other', to: 'home#missing_route', via: :all
 end
```
tomlovett, line 319: Simplify syntax

tomlovett:
Screen Shot 2022-11-02 at 2 47 00 PM
Screen Shot 2022-11-02 at 2 47 00 PM (2)

config/routes.rb
```diff
@@ -6,6 +6,10 @@
 Rails.application.routes.default_url_options[:host] = Rails.application.secrets.application_host
 
 Rails.application.routes.draw do
+  # Re-routes all requests to Old Portal on to Highlander
+  constraints subdomain: 'join' do
+    match '*any', to: redirect("https://#{ENV['APPLICATION_HOST']}/cas/login?from_join_portal=1"), via: :all
```
elenabaurkot, line 11: Just double-checking my understanding here - anything with subdomain join will redirect to Platform with the parameter `from_join_portal=1`

elenabaurkot (approved):
One question to check my understanding in there, but as long as I'm understanding correctly this looks good to me 👍

config/routes.rb
tomlovett: Yup! `constraints subdomain` catches requests to `join.(host)` and `match *any` will catch any method of request.

PR #1173, opened 2022-10-26
change Braven icons to grey on development

tomlovett:
# Description
This changes the Braven icons and the favicon (browser tab icon) to be grey on `platformweb`. 

# Motivation and Context
This will make it will be easier for developers to tell the difference between dev and prod environments. This is a little safer, and it will make navigation easier when devs may have both prod and dev tabs open.

Task: https://app.asana.com/0/1201131148207877/1202700658085900

# Screenshots
Screen Shot 2022-10-26 at 2 02 06 PM


tomlovett:
Don't mind the poor formatting. RuboCop does not check ERB files, so I'm adding a task to add an ERB linter.

PR #1179, opened 2022-10-27
Launch course for Portal PAFs without SF program (code will be removed after launching)

elenabaurkot:
# Description
We are launching a Canvas course for all PAFs that took the Accelerator course using Portal (what we used before Platform). We are shutting down Portal and these users need to maintain access to course materials so we are bringing them all into a new course in Canvas. This course will be launched without a Salesforce program ID because all of these users are already tied to their original SF program so we don't need to make a new participant with a new program in SF, instead we will create a Canvas user for them and add them to course using an SIS Import for users and enrollments. After adding them, we will update their Salesforce contact with their Platform user ID and Canvas user ID. Any users who used the Portal and already have a Platform/Canvas user will just have an enrollment added to the new course. This code is for one time use and will be removed after the course has been launched. 

# Motivation and Context
So we can launch a course without a SF program ID so that Portal PAFs can maintain access to course materials after Portal is shut down. 

Task: https://app.asana.com/0/1201131148207877/1202865857905018/f

# Screenshots
The fields squared in red will be omitted when launching the Portal PAF course, which will be called "Pre-Fall 2021 Braven Accelerator - PAFs" because it won't be linked to a SF program ID and we won't need and LC playbook course. 
Screen Shot 2022-10-27 at 12 39 11 PM


app/controllers/courses_controller.rb
```diff
@@ -43,12 +43,23 @@ def edit
...
     @salesforce_program = HerokuConnect::Program.find(salesforce_program_id)
     validate_program_for_create(@salesforce_program, @source_course)
 
-    CloneCourseJob.perform_later(current_user.email, @source_course, create_params[:name].strip, @salesforce_program)
+    name == 'TEST PAF course'
+    if (params["fellow_course_name"] == "Pre-Fall 2021 Braven Accelerator - PAFs" ||
+        params["fellow_course_name"] == 'TEST PAF course') &&
+        @ready_to_launch_paf_portal
+      CloneCoursePafPortalJob.perform_later(current_user.email, @source_course, create_params[:name].strip)
+    elsif
```
tomlovett, line 58: This `elsif` should require another boolean statement, right?

Maybe do the name check on one line, then an if-else inside that, based on `@ready_to_launch_paf_portal`?

app/models/course.rb
```diff
@@ -64,13 +64,19 @@ def is_template?
   end
 
   def is_accelerator_course?
+    return true if is_paf_portal_course?
     id == program.accelerator_course&.id
```
tomlovett, line 68: this can be combined into one boolean check

app/models/course.rb
```diff
@@ -64,13 +64,19 @@ def is_template?
...
   end
 
   def is_lc_playbook_course?
+    return false if is_paf_portal_course?
     id == program.lc_playbook_course&.id
```
tomlovett, line 73: this too can be combined into one boolean check

app/controllers/courses_controller.rb
```diff
@@ -105,26 +116,38 @@ def launch_new
...
-                                                                                                                                                     :fellow_course_name,
-                                                                                                                                                     :lc_source_course_id,
-                                                                                                                                                     :lc_course_name
-                                                                                                                                                   ])
-      raise ActionController::BadRequest.new("Can't use the same source course for Fellow and LC course") if fellow_source_course_id == lc_source_course_id
-    rescue ActionController::ParameterMissing, ActionController::BadRequest => e
-      redirect_to launch_courses_path, alert: "Error: #{e.message}" and return
-    end
+    ready_to_launch_paf_portal = true
```
tomlovett, line 119: Delete this, right?

elenabaurkot: ah good call, removing all the @ready_to_launch since we are ready now

app/controllers/courses_controller.rb
elenabaurkot: Not exactly sure what happened with my elsif here, removing the @ready_to_launch stuff now so taking out the elsif too 👍

app/models/course.rb
elenabaurkot: I put the first check on line 67 to check if it `is_paf_portal_course?` because if it is, I don't want it to get to line 68 where it checks the [program method](https://github.com/bebraven/platform/blob/bd99dcce6dd20a983a6ffb54d82242f95785ec85/app/models/course.rb#L82-84) will fail since we don't have a linked SF program. If I made it as an || statement and put the top statement first, will it skip checking the second condition?
```
def is_accelerator_course?
     is_paf_portal_course? || id == program.accelerator_course&.id
end
```

tomlovett: Yup! That will return on the first check.

app/models/course.rb
elenabaurkot: Updated both of these to the single liner boolean check 👍

tomlovett (approved):
LGTM :+1: Nice work!

PR #1184, opened 2022-11-04
Sync Zoom maximum three times per day

tomlovett:
Continuation of https://github.com/bebraven/platform/pull/1164

Zoom only allows us to update registrations three times in a single day. However, Data Ops may make multiple changes to Participant records in one day at the start of a semester. These changes will only check and send Zoom requests three times per day. This will prevent rate-limit errors (unless multiple Zoom sync's are forced.)


app/services/sync_salesforce_program.rb
```diff
@@ -232,34 +233,22 @@ def with_participant_error_handling(participant, &block)
   end
 
   def raise_if_duplicate_participant!(participant)
```
tomlovett, line 235: Simplify syntax; no functionality changes

app/services/sync_zoom_links_for_participant.rb
```diff
@@ -265,39 +241,44 @@ def zoom_meeting_2_configured?
 
   def create_or_update_link_infos(registrant1, registrant2)
     upsert_data = []
-    upsert_data << @new_zoom_link_info_1.attributes.except('id', 'created_at',
-                                                           'updated_at') if registrant1 && registrant1['error'].blank?
-    upsert_data << @new_zoom_link_info_2.attributes.except('id', 'created_at',
-                                                           'updated_at') if registrant2 && registrant2['error'].blank?
-
-    if upsert_data.present?
```
tomlovett, line 273: Syntax simplification. Old code put all this code inside the `if` statement, whereas PR code pre-emptively exits-out if the `if` statement fails

db/migrate/20221007170320_add_pending_zoom_updates_to_participant_sync_info.rb
```diff
@@ -0,0 +1,5 @@
+class AddPendingZoomUpdatesToParticipantSyncInfo < ActiveRecord::Migration[7.0]
```
tomlovett, line 1: Migration originally generated October 7. RIP

app/services/sync_zoom_links_for_participant.rb
```diff
@@ -201,55 +178,29 @@ def set_zoom_link_infos
...
-    return true if @force_update
-    return true if @existing_zoom_link_info_1.blank?
-    if @existing_zoom_link_info_1.registrant_details_match?(@new_zoom_link_info_1)
-      return false
-    else
-      return true
-    end
+
+    @force_update || @existing_zoom_link_info_1.blank? ||
+      !@existing_zoom_link_info_1.registrant_details_match?(@new_zoom_link_info_1)
```
elenabaurkot, line 184: my only concern with combining the logic that was here before into the shorthand is the readability- it's easy for me to pretty quickly make sense of what is going on with the longer code, if I was trying to figure out what was going on here reading it for the first time it would definitely take me a minute or two to think through what's going on. It does look nicer though, so if you think this way makes more sense they will definitely both get the job done.

app/services/sync_zoom_links_for_participant.rb
```diff
@@ -201,55 +178,29 @@ def set_zoom_link_infos
...
-      return true
-    end
-  end
 
-  def rerun_after_rate_limit?
-    # When true: Participant exceeded Zoom's rate-limting, but 24 hours have passed since then, so try again
-    !@participant_sync_info.zoom_rate_limit_until.nil? &&
-      @participant_sync_info.zoom_rate_limit_until.past?
+    @force_update || @existing_zoom_link_info_2.blank? ||
+      !@existing_zoom_link_info_2.registrant_details_match?(@new_zoom_link_info_2)
```
elenabaurkot, line 191: same thing here as mentioned above

app/services/sync_salesforce_program.rb
```diff
@@ -154,59 +156,59 @@ def run
...
         zoom_sync_success = with_participant_error_handling(participant) do
-          # Turn off this behavior until everything is fixed
-          # break unless zoom_sync_cohort_schedule?(participant.cohort_schedule_id)
+          # use `next` because `break` will cause zoom_sync_success to be `nil`
+          next unless should_zoom_sync_participant?(participant) || @force_zoom_update
 
           SyncZoomLinksForParticipant.new(participant, @force_zoom_update).run
+
+          # And error in hitting Zoom API will leave pending_zoom_updates as true, so that we will try again next sync
+          participant.new_sync_info.pending_zoom_updates = false
```
elenabaurkot, line 193: When the `SyncZoomLinksForParticipant` runs above this, if they hit the rate limit I saw it raises in there for hitting it. If it raises in there, will it ever get to this point of setting pending_zoom_updates to false? I'm thinking we would want to leave it set to true so that it syncs whenever it stops getting rate limited, so if it wouldn't make it to this line then I think we're good.

elenabaurkot (approved):
Looks good, one comment is just syntax stuff and go with whatever your preference is, the other is just checking how hitting the zoom rate limit could affect this, but if it doesn't then this LGTM!

app/services/sync_salesforce_program.rb
tomlovett: Yup! That's the plan. The raise _should_ exit out of the parent block, leaving `pending_zoom_updates` as `true`.

app/services/sync_zoom_links_for_participant.rb
tomlovett: Yeah, I think this section is due for refactoring after this PR. I can definitely make a note to come back and clean this up and make it more readable.

PR #1185, opened 2022-11-07
Remove lowlander/highlander redirect.

peterdeweese:
# Description

Removing redirect that needs to be done differently.

# Motivation and Context

We assumed routing from a subdomain, but we'll need to attack this differently.

Task: 

# Screenshots


tomlovett (approved):
LGTM

PR #1186, opened 2022-11-08
set error message on failed login attempt by lowlander email

tomlovett:
# Description
We want to display a special error message when our Lowlander users are re-rerouted to Canvas. This new logic checks whether a failed login attempt came from someone who was imported as a Lowlander PAF.

Line 152 in `cas_controller.rb` splits user flow based on whether the users successfully passed login or not. We set the value for `failed_lowlander_login` there. Logic splits at 215 into different scenarios for why a user failed login: they still have to register, they are a valid user but put in the wrong password, or we don't have a user for that email.

I changed line 14 of `login.html.rb` so it would not simultaneously show the "Incorrect login/password" message if it's going to show the Lowlander warning message, because I feel like having both would be confusing.

Task: https://app.asana.com/0/1201131148207877/1202865857905024


elenabaurkot (approved):
Talked on call about adding one comment for paf_portal_course

PR #1188, opened 2022-11-14
Update schema with production dump.

peterdeweese:
# Description

This includes the schema dump from production. The codebase also contains each table that was dropped or renamed and related foreign keys and constraints.

# Motivation and Context

Task: 

# Screenshots


db/schema.rb
```diff
@@ -10,8 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_09_01_152706) do
+ActiveRecord::Schema[7.0].define(version: 2022_10_07_170320) do
   # These are extensions that must be enabled in order to support this database
+  enable_extension "hstore"
+  enable_extension "pg_stat_statements"
```
peterdeweese, line 16: I think we should keep hstore and not keep pg_stat_statements in the codebase copy. Even if we aren't using it at the moment, we'll match production which will be important if we ever use it in the future or want to remove it.

db/schema.rb
```diff
@@ -10,8 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_09_01_152706) do
+ActiveRecord::Schema[7.0].define(version: 2022_10_07_170320) do
```
peterdeweese, line 13: We can keep the codebase version of the schema version.

db/schema.rb
```diff
@@ -603,6 +450,7 @@
     t.string "cohort_schedule_id", limit: 18
     t.json "ta_caseload_enrollments"
     t.datetime "zoom_rate_limit_until"
+    t.boolean "pending_zoom_updates", default: false
```
peterdeweese, line 453: This is in production but not in the codebase! I do not see this column in migrations, but I recommend we pull it back to the schema if it is in production, so it reflects the truth and reality of production. If we don't want it, a migration can remove it. If we do, we can reflect this in a test.

tomlovett: This is probably an artifact from merging in Zoom sync, then reverting it. The PR that we will merge in this week includes a migration to add this column.

But I don't know how that affects whether or not to keep it in this PR. If it's in prod, keep it in the PR?

db/schema.rb
```diff
@@ -10,8 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_09_01_152706) do
+ActiveRecord::Schema[7.0].define(version: 2022_10_07_170320) do
   # These are extensions that must be enabled in order to support this database
+  enable_extension "hstore"
```
peterdeweese, line 15: This is really in production and is not in our dev dbs. I don't think we use it. Regardless of the difference, this both documents it, and allows migrations to add/remove it in the future to be consistent.

peterdeweese: I removed a stats extension that is probably used for Heroku or APM monitoring. The line should be configured per environment, not universally.

db/schema.rb
peterdeweese: That migration would have succeeded in dev and failed in prod, so it sounds like we should have it in the schema and change the PR to not add it.

tomlovett (approved):
Nice work! LGTM

PR #1189, opened 2022-11-15
Remove fancy-but-incorrect syntax in user.lowlander?

tomlovett:
# Description
On the initial commit, I changed the syntax of `User.lowlander?` from `courses.any? { |c| c.is_paf_portal_course? }` to `courses.any? & :is_paf_portal_course?` on the advice of Rubocop. However, the latter syntax returns false positives. (This false positive caused a non-Lowlander to see the Lowlander login message.)

Revert to the original syntax, which can effectively discriminate between Lowlanders and non-Lowlanders.


elenabaurkot (approved):
Lgtm!

PR #1190, opened 2022-11-16
add program name and sfid to Salesforce sync email

tomlovett:
# Description
Add a program's name and Salesforce ID to the subject of the emails sent out to users after a manual Salesforce sync has completed.

# Motivation and Context
When internal users run a manual Salesforce sync, they will receive an email saying the sync was either successful or it failed. Product Support has requested we include more program information in these emails, so they can better differentiate between results for specific programs.

Task: https://app.asana.com/0/1201131148207877/1203270264288131


peterdeweese (approved):
LGTM

PR #1192, opened 2022-11-17
deleted package-lock.json

ThotsaphonB:
# Description

Deleted "package-lock.json" so we could remain on Yarn

# Motivation and Context

1. Should not use package-lock.json as that is part of npm. Better to use Yarn

Task:
1. Ran `git rm package-lock.json` per https://dashboard.heroku.com/apps/braven-platform-highlander/activity/builds/e07456f8-894a-4902-8a18-d3bffdb624d3

# Screenshots


tomlovett (approved):
LGTM (assuming tests pass)

PR #1193, opened 2022-11-17
Fix issue with cutting and pasting industry selectors.

peterdeweese:
# Description

Renames industry selector names using uuids. Note updated comments and several lint fixes applied. I was going to refactor the clipboard filter to allow it to be modified further, which is why I touched that file. I decided not to pursue it because the pasting cases we find in production are of just the text field, and that will work fine now.

# Motivation and Context

When copying and pasting the text field from the industry selector to make more text fields, all fields end up with one value.

Task:


app/javascript/ckeditor/clipboardattributeediting.js
```diff
@@ -13,17 +13,18 @@ export default class ClipboardAttributeEditing extends Plugin {
   }
 
   init() {
-    this._getNewId = this.editor.plugins.get('UniqueId').getNewId;
-    this._getNewName = this.editor.plugins.get('UniqueId').getNewName;
+    this.getNewId = this.editor.plugins.get('UniqueId').getNewId;
+    this.getNewName = this.editor.plugins.get('UniqueId').getNewName;
```
peterdeweese, line 17: Cleaning up a few reasonable lint issues.

app/javascript/ckeditor/insertindustryselectorcommand.js
```diff
@@ -37,15 +33,15 @@ function createIndustrySelectorContainer(writer) {
   const uniqueId = new UniqueId();
   const container = writer.createElement('fieldset');
 
-  const selector_prompt = 'Which two of these industries best fit your interests and goals:';
-  const textarea_prompt = "If your industry of interest doesn't appear above, please enter it here:";
+  const selectorPrompt = 'Which two of these industries best fit your interests and goals:';
+  const textareaPrompt = "If your industry of interest doesn't appear above, please enter it here:";
```
peterdeweese, line 37: js should not use snake-case in general.

app/javascript/ckeditor/insertindustryselectorcommand.js
```diff
@@ -37,15 +33,15 @@ function createIndustrySelectorContainer(writer) {
...
 
-  writer.append(createParagraph(writer, selector_prompt), container);
-  writer.append(createSelector(writer, SELECTOR_NAMES[0], uniqueId.getNewId()), container);
+  writer.append(createParagraph(writer, selectorPrompt), container);
+  writer.append(createSelector(writer, uniqueId.getNewName(), uniqueId.getNewId()), container);
```
peterdeweese, line 40: This is the meat. The names now include uuids and a prefix that the clipboard paster will update.

elenabaurkot (approved):
This lgtm!

PR #1194, opened 2022-11-17
Allow Fellows to submit Post Accelerator survey multiple times temporarily

elenabaurkot:
# Description
Some Fellows completed their post-accelerator survey early, which is meant to be completed at the end of the course. We usually only allow one submission, but in order to get more accurate response, we are allowing Fellows to submit multiple times for the Fall 2022 semester. We will remove this code at the end of the semester. 

# Motivation and Context
Make sure Post-Accelerator Survey data is accurate.

Task: https://app.asana.com/0/1201131148207877/1203375190760129/f


tomlovett (changes requested):
Great work! 

Can you just add the "after Fall 2022 semester" to `accelerator_survey_submissions_controller` and `controllers/concerns/submittable`? Otherwise looks good to me!

tomlovett (approved):
LGTM!

PR #1196, opened 2022-11-18
Add helper functions to models for readability

tomlovett:
For a recent task, we added program name and SF id to an email header to help Product Support differentiate what program just finished sync'ing. I added a helper function to the emailer to display a Program's name and SF id, when something similar would help in our error logging.

So, this PR overwrites the `#to_s` method for our HerokuConnect models, so that we can succinctly display readable information in our logs.

# Description
Overwrite `#to_s` method in HerokuConnect base model. Use `email` specifically on Contact records, since that it is more appropriate for that model.

Implement this method in places where end-users would benefit from increased readability, like mailers.

General cleanup and refactoring.


tomlovett:
I've done so many commits to fix linting; at this point I'm over it.

elenabaurkot (approved):
This all LGTM - only comment is on the user model and course model the `canvas_href` method that was added doesn't seem to be used anywhere - is this just for if we need it in the future? Seems like it could be useful, just didn't actually see it used

peterdeweese (approved):

tomlovett:
> This all LGTM - only comment is on the user model and course model the `canvas_href` method that was added doesn't seem to be used anywhere - is this just for if we need it in the future? Seems like it could be useful, just didn't actually see it used

Yeah, and I noticed there's an existing function that does the same thing. I'll cut this one out in the next tech-debt/refactor PR I'm going to work on.

PR #1204, opened 2022-11-22
Reduce eslint errors considerably.

peterdeweese:
# Description

Resolves many eslint errors in ckeditor, not all but everything automatically fixable and some others.

# Motivation and Context

To learn our ckeditor code better and make it so we can use `eslint -fix` consistently.

Task: 
https://app.asana.com/0/0/1203315183781647/f


.eslintrc.js
```diff
@@ -19,6 +19,17 @@ module.exports = {
   ],
   rules: {
     'react/prop-types': 'off',
+    "max-len": [
+      "error",
+      {
+        code: 120,
```
peterdeweese, line 25: I think 120 is reasonable unless it doesn't fit with someone's workflow. 80 is how we did it on green screens way back. 😆

tomlovett: We have the backend set to 120 as well, so that sounds good to me!

tomlovett (approved):
LGTM!

PR #1205, opened 2022-11-22
Canvas access

elenabaurkot:
# Description
In Periscope and Salesforce some users were showing they did not have Canvas access even though they did. On their SF contact 'Canvas - Has Highlander Access' was not checked and 'Canvas User ID - Highlander' was not set even though these Contacts did actually have access to Canvas (their canvas user id was set on the User in our database). The users who had this issue were users that the sync exited early while in the process of syncing. The sync has 15 minutes to run and if it takes longer than 15 minutes it exits early (the sync only seems to have exited early the very first sync of the semester for programs because it has the most work to do during that sync since it's the first time syncing all program Participants so it takes the longest). 

In the `SyncSalesforceContact` service, when syncing a contact, it checks if the `signup_token_sent_at` on a user is blank, if it is it calls the `setup_signup_token` method. This method sets the `signup_token_sent_at` on the user and then makes a Saleforce API call to update the contact with their Signup Token and the Canvas user ID. The problem we think was happening is that when the sync exited early in the middle of syncing a contact, it set the `signup_token_sent_at` on the user, but didn't complete the `update_contact` call to Salesforce to update the Canvas user ID. When the subsequent sync ran and it got to the contact it had exited on, the user already had the `signup_token_sent_at` field set so it didn't call the `setup_signup_token` method again and the SF contact never got the Canvas User ID set (this is needed to show that someone has Canvas access). We updated the code to run the `setup_signup_token` method if it has a blank `signup_token_sent_at` field  or if the `HerokuConnect::Contact` `canvas_cloud_user_id__c` field is blank. This way if the `signup_token_sent_at` field got set, but the Canvas User Id wasn't updated on Salesforce because the sync exited early, we still make sure it gets set in the next sync when it gets to this user. 

# Motivation and Context
To ensure that Salesforce and Periscope are accurately showing who has access to Canvas and that Salesforce contacts have the Canvas user ID field set if it is set on the user.

Task: https://app.asana.com/0/1201131148207877/1203016603273752/f


tomlovett (approved):
LGTM!

PR #1206, opened 2022-11-28
Add tests and clearer email subjects to Mailers

tomlovett:
# Description
Add tests to the codebase so that all of our Mailers have at least basic test coverage.

# Motivation and Context
Product Support relies on these emails to do their job successfully. The more information we can include for them - whether tasks succeed or fail - the easier their lives will be.

It also important to have even simple tests like these for our Mailer jobs, as they will alert us when changes somewhere else in the codebase are having unintended downstream consequences.

Task: https://app.asana.com/0/1201131148207877/1203411749175431


app/services/generate_test_users.rb
```diff
@@ -216,14 +217,10 @@ def get_candidate_role_fields(role)
     additional_fields
   end
 
-  # For Candidate and Participant record type ID:
-  # LC and CP have the same record type (both get record type LC --> CP gets different Coach_Partner_Role__c)
-  # TA, Staff and Faculty have the same record type (both get record type TA --> Staff & Faculty gets different Coach_Partner_Role__c)
```
tomlovett, line 221: Comment's line is too long, and doesn't add much value, so deleting it.

app/services/generate_test_users.rb
```diff
@@ -216,14 +217,10 @@ def get_candidate_role_fields(role)
...
   def get_candidate_record_type_id(role)
     case role
-    when SalesforceConstants::Role::FELLOW
-      return "0121J000001De1qQAC"
-    when SalesforceConstants::Role::LEADERSHIP_COACH,
```
tomlovett, line 226: Remove redundancy.

app/services/generate_test_users.rb
```diff
@@ -235,9 +232,8 @@ def get_candidate_record_type_id(role)
 
   def get_participant_record_type_id(role)
     case role
-    when SalesforceConstants::Role::FELLOW
-      return "0121J000001De1vQAC"
-    when SalesforceConstants::Role::LEADERSHIP_COACH,
```
tomlovett, line 240: Remove redundancy.

app/services/grade_capstone_evaluations.rb
```diff
@@ -103,15 +99,9 @@ def grade_capstone_eval_questions(user, all_submissions)
...
   def submissions_have_been_graded?
-    # If any submissions have been graded for this course, set submissions_have_been_graded to true
-    all_capstone_eval_submissions.each do |submission|
-      unless submission.new?
-        return true
-      end
-    end
-
-    false
+    all_capstone_eval_submissions.any? { |submission| !submission.new? }
```
elenabaurkot, line 104: I like this syntax a lot, much cleaner 👍

app/services/generate_test_users.rb
elenabaurkot: I originally added this in because it's one of the more confusing aspects of how roles are assessed in Platform, but not sure if it's actually needed. LC and CP both get a role of leadership coach role, but a CP needs to have the Coach_Partner_role__c (candidate role select field in SF set) so that Platform knows they aren't actually a LC, but they are a CP -- same thing for TA,Staff and Faculty -- they are all given TA role, but have to have the Coach_Partner_role__c field set to their correct role or they don't get interpreted as a TA and have incorrect permissions. I was thinking the explanation might be useful context here for anyone digging through this code editing things, but it might need to be more clear or maybe it makes more sense to add that explanation to somewhere in the wiki instead. Do you think I should put this in the wiki as opposed to adding the info here @tomlovett ?

app/services/generate_test_users.rb
elenabaurkot: I think these are returning two different record types - when Fellow it ends in qQAC for the Fellow record type and when LC or CP it ends in rQAC because they  get the LC record type

app/services/generate_test_users.rb
elenabaurkot: Same thing as above here the Fellow is getting the Fellow role ending with "vQAC" and LC and CP get the Leadership Coach role ending with "wQAC"

peterdeweese (approved):

app/services/generate_test_users.rb
tomlovett: Good catch! Thanks!

app/services/generate_test_users.rb
```diff
@@ -218,34 +211,34 @@ def get_candidate_role_fields(role)
...
-      return "0121J000001De1qQAC"
+    	SalesforceConstants::CandidateRecordTypeId::FELLOW
     when SalesforceConstants::Role::LEADERSHIP_COACH,
         SalesforceConstants::Role::COACH_PARTNER
-      return "0121J000001De1rQAC"
+			SalesforceConstants::CandidateRecordTypeId::LEADERSHIP_COACH
     when SalesforceConstants::Role::TEACHING_ASSISTANT,
         SalesforceConstants::Role::STAFF,
         SalesforceConstants::Role::FACULTY
-      return Rails.application.secrets.salesforce_ta_candidate_record_type_id
```
tomlovett, line 235: I don't think this or the participant version need to be kept in secrets.

peterdeweese (approved):
I like the improvements!

elenabaurkot (approved):
LGTM!

app/services/generate_test_users.rb
tomlovett: Great point, will do!

PR #1207, opened 2022-11-29
fix frozen string literal error

tomlovett:
# Motivation and Context
This syntax tries to modify a frozen string literal, which was throwing an error when ProdSupport launched a test course: https://bebraven.slack.com/archives/C02H6RE7DB5/p1669732971867349

# Description
Change the syntax to avoid this error.

https://bebraven.slack.com/files/U03Q5B3VD8T/F04DLV7GQQ0/screen_shot_2022-11-29_at_9.41.28_am.png


elenabaurkot (approved):
LGTM!

PR #1208, opened 2022-11-29
Remove 'Test' from candidate role select in Generate test users tool

elenabaurkot:
# Description
When creating test users with the 'Generate Test Users' tool, we add the candidate role select of "Test" for Fellows, LCs and TAs (Coach Partners, Faculty and Staff need their actual candidate role select set so that they are interpreted as their actual role since Coach Partners get a role of Leadership Coach and Staff and Faculty get a role of Teaching Assistant). By adding 'Test' as the candidate role select for LCs, it was giving their Candidate a prefix of "TBD:" because of a Salesforce formula, by removing 'Test' from their candidate role select, they will get their normal name for their candidate. 

# Motivation and Context
To make sure candidate names are formatted properly for LC test users


tomlovett (approved):
LGTM!

app/services/generate_test_users.rb
```diff
@@ -19,7 +19,7 @@ def run
       Honeycomb.add_field('generate_test_users.users.count', @params['email'].count)
       new_test_users = generate_test_users()
 
-      if @failed_users.present?
+      unless @failed_users.empty?
```
peterdeweese, line 22: Why is the double negative preferred here? Although I think we initialize @failed_users as an empty array, note that `#blank?` will work on nil.

peterdeweese (approved):

PR #1213, opened 2022-12-02
Remove unused secrets

elenabaurkot:
# Description
The record type IDs for TA participants and TA candidates were previously different in development and production so we had them in config and .env variables. After the Salesforce refresh, they are not the same in development and production Salesforce. We refactored the code in [this PR](https://github.com/bebraven/platform/pull/1206) to use Salesforce constants so we no longer need the secrets `SALESFORCE_TA_PARTICIPANT_RECORD_TYPE_ID` and `SALESFORCE_TA_CANDIDATE_RECORD_TYPE_ID`. The variables should also be removed from developers .env file. 

# Motivation and Context
Remove secrets that are no longer used.


tomlovett (approved):
LGTM! Sorry you had to pick up my slack!

peterdeweese (approved):
Looks good!

elenabaurkot:
> LGTM! Sorry you had to pick up my slack!

No need to be sorry!! bad variables I added haha @tomlovett

PR #1214, opened 2022-12-02
Fix rubocop -a offenses for rise360 and lrs xapi.

peterdeweese:
# Description
Ran safe rubocop offense safe corrections.

# Motivation and Context
To learn about rise 360 code.
To achieve a cleaner lint before making changes so I can use safe corrections as I make changes.

Task: 
https://app.asana.com/0/1201131148207877/1203414606491884/f
https://app.asana.com/0/1201131148207877/1203459577927950/f


app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -7,9 +7,9 @@ class CourseRise360ModuleVersionsController < ApplicationController
   # Adds the #publish, #publish_latest, #unpublish actions
   include Publishable
 
-  prepend_before_action :force_destroy_references, only: [:publish_latest, :unpublish]
+  prepend_before_action :force_destroy_references, only: %i[publish_latest unpublish]
   prepend_before_action :set_model_instance,
-                        only: [:before_publish_latest, :publish_latest, :unpublish, :before_unpublish]
+                        only: %i[before_publish_latest publish_latest unpublish before_unpublish]
```
peterdeweese, line 12: I don't care much either way about these % array creators, but autocorrect makes it easy. I guess it tends to save some characters and enforce the type of each item in the array.

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -41,14 +41,16 @@ def points_possible
...
-      protocol: 'https',
+      protocol: 'https'
     )
   end
 
   def versionable_instance
-    params[:rise360_module_id] ?
-      Rise360Module.find(params[:rise360_module_id]) : # publish
-      @course_rise360_module_version.rise360_module_version.rise360_module # publish_latest
+    if params[:rise360_module_id]
```
peterdeweese, line 49: I think this was caught because it was a multi-line ternary operation.

app/controllers/course_rise360_module_versions_controller.rb
```diff
@@ -81,15 +83,15 @@ def destroy_references
       canvas_assignment_id = @course_rise360_module_version.canvas_assignment_id
 
       # Delete interactions for deleted modules, so they don't break grading.
-      Rise360ModuleInteraction.where(canvas_assignment_id: canvas_assignment_id).destroy_all
+      Rise360ModuleInteraction.where(canvas_assignment_id:).destroy_all
```
peterdeweese, line 86: I like this newer notation like ES6 for the conventional case of passing a local variable of the same name as the named parameter.

app/models/rise360_module_version.rb
```diff
@@ -26,8 +26,8 @@ def launch_url
   end
 
   def self.find_by_lti_launch_url(url)
-    rise360_module_version_id = url[/.*\/rise360_module_versions\/(\d+)/, 1]
-    return Rise360ModuleVersion.find_by(id: rise360_module_version_id)
+    rise360_module_version_id = url[%r{.*/rise360_module_versions/(\d+)}, 1]
+    Rise360ModuleVersion.find_by(id: rise360_module_version_id)
```
peterdeweese, line 30: Implicit return.

app/services/grade_rise360_module_for_user.rb
```diff
@@ -73,17 +73,19 @@ def run
     Honeycomb.add_field('grade_rise360_module_for_user.skipped_computing_grade', !run_computation)
     return nil unless run_computation
 
-    Rails.logger.debug("Computing grade for: user_id = #{@user.id}, " \
-                       "canvas_course_id = #{@course_rise360_module_version.course.canvas_course_id}, " \
-                       "canvas_assignment_id = #{@course_rise360_module_version.canvas_assignment_id}")
+    Rails.logger.debug do
```
peterdeweese, line 76: I don't know which autocorrect this is but splitting a line for a single argument might be more clear this way. Interesting.

app/services/grade_rise360_module_for_user.rb
```diff
@@ -276,8 +278,8 @@ def total_quiz_questions
   def grader_full_name
     return @grader_full_name if @grader_full_name
 
-    if !grade_is_manually_overridden?
-      raise RuntimeError.new('grade_module_for_user#grader_full_name was called for a submission that was not manually graded')
+    unless grade_is_manually_overridden?
```
peterdeweese, line 281: unless instead of `!`. Thats when to use it!

lib/lrs_xapi_mock.rb
```diff
@@ -28,7 +28,7 @@ class LrsXapiMockError < StandardError; end
 
   # Returns a hash with keys {:code, :body}, OR nil.
   def self.handle_request!(request, endpoint, user)
-    Honeycomb.start_span(name: 'lrs_xapi_mock.handle_request!') do |span|
+    Honeycomb.start_span(name: 'lrs_xapi_mock.handle_request!') do |_span|
       Honeycomb.add_field('lrs_xapi_mock.path', "/#{endpoint}")
```
peterdeweese, line 32: Hmm. I think it would be typical to add fields to span here, maybe `span.add_field(...`. This refactor won't break anything but something to consider when we work on these.

app/controllers/course_rise360_module_versions_controller.rb
tomlovett: So much better! Thanks for sharing this with us.

spec/controllers/rise360_module_grades_controller_spec.rb
```diff
@@ -5,51 +5,51 @@
...
   let(:user_with_grade) { create :fellow_user, canvas_user_id: '987432' }
   let(:user_viewing) { user_with_grade }
-  let(:rise360_module_grade) {
-    create :rise360_module_grade, user: user_with_grade, course_rise360_module_version: course_rise360_module_version
-  }
-  let(:canvas_api_user_id) { 9954321 }
+  let(:rise360_module_grade) do
+    create :rise360_module_grade, user: user_with_grade, course_rise360_module_version:
+  end
+  let(:canvas_api_user_id) { 9_954_321 }
```
tomlovett, line 18: This is kinda strange, but if there tests still pass, it's all good.

spec/controllers/rise360_module_versions_controller_spec.rb
```diff
@@ -33,36 +33,36 @@
...
         allow(LtiScore).to receive(:new_module_submission).and_return(lti_score_request)
 
-        get :show, params: { :id => rise360_module_version.id, :state => state }
+        get :show, params: { id: rise360_module_version.id, state: }
 
         grade = Rise360ModuleGrade.find_by(
-          user: user,
-          course_rise360_module_version: course_rise360_module_version
+          user:,
+          course_rise360_module_version:
```
tomlovett, line 65: Nitpick: Squish to one line?

spec/controllers/rise360_modules_controller_spec.rb
```diff
@@ -5,65 +5,65 @@
...
-    context "with invalid params" do
-      it "raises an error when zipfile param is missing" do
-        expect {
+  describe 'POST #create' do
+    context 'with invalid params' do
+      it 'raises an error when zipfile param is missing' do
+        expect do
           post :create
-        }.to raise_error ActionController::ParameterMissing
+        end.to raise_error ActionController::ParameterMissing
```
tomlovett, line 43: Nitpick: This can be one line, too.

app/services/grade_rise360_modules.rb
```diff
@@ -110,7 +110,7 @@ def grade_assignment(canvas_assignment_id, user_ids)
       on_time_credit_model_ids = []
 
       # Skip assignments with zero interactions; they probably are future ones
-      unless Rise360ModuleInteraction.where(canvas_assignment_id:).exists?
+      unless Rise360ModuleInteraction.exists?(canvas_assignment_id:)
```
tomlovett, line 113: Nice! This is really cool syntax.

tomlovett (approved):
Some nitpicks here and there, but LGTM! Nice work!

app/services/grade_rise360_module_for_user.rb
```diff
@@ -236,7 +238,7 @@ def grade_is_manually_overridden?
 
   # After calling run(), this will hold the breakdown of the grade into it's components and total.
   def computed_grade_breakdown
-    raise RuntimeError.new('grade_module_for_user#run was never called') if @computed_grade_breakdown.nil?
```
elenabaurkot, line 239: Curious about the changing of `raise RuntimeError` to just `raise`, don't know too much about this - is this just not actually a runtime error @peterdeweese?

app/services/grade_rise360_modules.rb
```diff
@@ -28,17 +27,17 @@ def run
...
-        Rails.logger.info("Exit early: no current/future accelerator programs with a Canvas course ID set.")
+        Rails.logger.info('Exit early: no current/future accelerator programs with a Canvas course ID set.')
         return
       end
 
       # Eliminate courses with no module interactions, and exit early if that
       # leaves us with an empty list.
       courses = Course.where(canvas_course_id: canvas_course_ids)
-                      .filter { |c| Rise360ModuleInteraction.where(canvas_course_id: c.canvas_course_id).exists? }
+                      .filter { |c| Rise360ModuleInteraction.exists?(canvas_course_id: c.canvas_course_id) }
```
elenabaurkot, line 37: didn't know you could use `.exists?` like this 👍

lib/lti_rise360_proxy.rb
```diff
@@ -33,14 +33,14 @@ def perform_request(env)
...
-      path = $1 # from regex match above
+      path = Regexp.last_match(1) # from regex match above
       uri = URI(Rise360Util.presigned_url(path))
 
       env['HTTP_HOST'] = uri.host           # e.g. some-bucket.s3.amazonaws.com
       env['PATH_INFO'] = uri.path           # The path matched in the regex above
       env['QUERY_STRING'] = uri.query       # The AWS query params signing it
 
-      loosen_content_security_policy(request) if path =~ /index\.html/
+      loosen_content_security_policy(request) if /index\.html/.match?(path)
```
elenabaurkot, line 43: strange that rubocop didn't change this regex into %r format but did for all the others

elenabaurkot (approved):
LGTM 👍

lib/rise360_util.rb
```diff
@@ -88,8 +88,8 @@ def self.update_metadata!(rise360_module, bucket = nil)
       quiz_breakdown: tincan_data.xpath(QUIZ_QUESTION_XPATH)
         .map { |q| q.attributes['id'].value.split('/')[-2] }
         .group_by(&:itself)
-        .map { |k, v| v.count },
-      activity_id: tincan_data.xpath(COURSE_XPATH).first['id'],
+        .map { |_k, v| v.count },
```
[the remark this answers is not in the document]
peterdeweese: @elenabaurkot, "_" is accepted as the leading in any variable name, but is conventional in this case to mark it as an unused variable since it may still need to be specified to use variables past it in the list. Also, this particular map could also be rewritten as `.values.map(:&count)`.

lib/lti_rise360_proxy.rb
peterdeweese: Interesting. I wonder if it is because this one is essentially matching a constant, and not using the match results? It seems it could check equality, instead.

app/services/grade_rise360_module_for_user.rb
peterdeweese: `raise` raises a RuntimeError by default!

elenabaurkot: ah good to know!!

lib/rise360_util.rb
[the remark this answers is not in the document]
elenabaurkot: good to know - thanks for the explanation @peterdeweese

PR #1215, opened 2022-12-02
Test launch no lc

elenabaurkot:
# Description
We are launching a staff course for onboarding so we only need the one Accelerator course template for this we don't need an LC course. This is non-standard and a one time thing so this will be reverted once the staff course is launched. 

# Motivation and Context
Launch a staff course for onboarding without an LC course

Task: https://app.asana.com/0/1201131148207877/1203373600201856/f


app/controllers/courses_controller.rb
```diff
@@ -107,22 +107,35 @@ def launch_create
 
     # Validate form.
     begin
-      salesforce_program_id,
-      notification_email,
-      fellow_source_course_id,
-      fellow_course_name,
-      lc_source_course_id,
-      lc_course_name = params.require([
```
tomlovett, line 115: We'll probably go over this in the call, and this isn't super necessary, but you can also use `params.permit` to grab all of the variables, but not fail if one is missing.

https://api.rubyonrails.org/v7.0.4/classes/ActionController/Parameters.html#method-i-permit

tomlovett: Ahh yeah, I didn't realize we were pulling out and using all of these variables within this function. That would explain why it's complaining about the variables not being set.

This looks good!

PR #1217, opened 2022-12-06
Refactor sync with rubocop.

peterdeweese:
# Description
Reduces rubocop warnings considerably through many safe, minor refactors and cleans up test output.

# Motivation and Context

To learn more about sync.
To improve code consistency and understandability,  lint and test output so one can jump in and make changes.
To encourage team code convention.

Task: 


.rubocop.yml
```diff
@@ -32,6 +32,9 @@ Rails/I18nLocaleTexts:
 RSpec/ExpectInHook:
   Enabled: false
 
+RSpec/LetSetup:
+  Enabled: false
+
```
peterdeweese, line 37: This is for `let!` macros in tests and makes it, so you have to refer to them directly to set them up outside a before block. It would prevent unused setup, but to fix the warnings while leaving this I'd blindly stick them all into before blocks. I think we can trust devs to set the state that is needed.

Gemfile
```diff
@@ -115,8 +115,8 @@ gem 'devise_cas_authenticatable', '~> 1'
 
 gem 'react-rails'
 
-gem 'sentry-ruby'
 gem 'sentry-rails'
+gem 'sentry-ruby'
 gem 'sentry-sidekiq'
```
peterdeweese, line 121: each list in this file is alphabetized by rubocop.

app/jobs/sidekiq_jobs/sync_salesforce_changes_job.rb
```diff
@@ -10,22 +10,22 @@
...
-      validate_config()
+      validate_config
 
       # Kick off all program jobs and then exit. They run in parallel and each is
       # self contained. A second run of this job before one of them has finished will
       # simply queue up the program job which will skip and log unless
       # SALESFORCE_SYNC_MAX_DURATION has expired.
       HerokuConnect::Program.current_and_future_program_ids.each do |program_id|
-        Rails.logger.debug("Kicking off SyncSalesforceProgramJob for #{program_id}")
+        Rails.logger.debug { "Kicking off SyncSalesforceProgramJob for #{program_id}" }
```
peterdeweese, line 25: Sending a block performs better because it is only evaluated if it is needed based on the environment's logging level. It can reduce a surprising amount of memory churn.

app/jobs/sidekiq_jobs/sync_salesforce_program_job.rb
```diff
@@ -129,8 +128,8 @@ def perform(program_id, email = nil, force_canvas_update = false, force_zoom_upd
     if e.is_a?(SyncSalesforceProgram::MissingCourseModelsError) && Rails.env.development?
       # This is expected in dev since each dev uses a different Sandbox Salesforce Program
       # for their local Courses. No need to create noise.
-      Rails.logger.debug("Skipping sync for Program Id: #{program_id} in dev. Missing local Course models.")
-      return
+      Rails.logger.debug { "Skipping sync for Program Id: #{program_id} in dev. Missing local Course models." }
+      nil
```
peterdeweese, line 132: This nil might not matter, but I decided to keep it equivalent on this pass.

tomlovett (approved):
LGTM!

elenabaurkot (commented):
A couple questions for understanding, but looks good to me

spec/services/sync_salesforce_program_spec.rb
```diff
@@ -114,24 +114,26 @@
...
     end
 
     context 'when previous Canvas SisImport had errors' do
-      let(:last_sis_import_status) { SisImportStatus.new(create :canvas_sis_import_failed) }
+      let(:last_sis_import_status) { SisImportStatus.new(create(:canvas_sis_import_failed)) }
+
       it 'turns diffing mode off' do
         program.accelerator_course.update!(last_canvas_sis_import_id: last_sis_import_status.sis_import_id)
-        expect(canvas_client).to receive(:get_sis_import_status).and_return(last_sis_import_status)
+        allow(canvas_client).to receive(:get_sis_import_status).and_return(last_sis_import_status)
```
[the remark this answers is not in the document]
peterdeweese: So this is the RSpec/StubbedMock cop, which is not very well documented. I think the idea is to encourage single expectations as if it returns something you'd need to expect it elsewhere. It also separates mocking from expecting. Since these expectations seem useful I might undo these changes and list them as things to consider later.

.rubocop.yml
peterdeweese: I disabled the cop so we don't need to worry about it. It is supposed to make you have to be more explicit such as calling the let in a before block, but that is redundant. I think it should be normal to set the before state with a let!– and have to know what you are doing. Adding it to the before block is no more nor less route. The opinions in certain RSpec cops are a bit more preference than most.

app/services/sync_canvas_grades.rb
```diff
@@ -4,9 +4,6 @@
 require 'salesforce_api'
 
 class SyncCanvasGrades
-  def initialize
```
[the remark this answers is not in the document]
peterdeweese: Yes. The empty block is doing the same thing as having no initialize in this particular case. The default class it extends from is the same.

app/services/sync_salesforce_contact.rb
```diff
@@ -136,7 +136,7 @@ def sync_canvas_email
       canvas_login_email = canvas_login['unique_id']
       Honeycomb.add_field('canvas.login.unique_id', canvas_login_email)
 
-      canvas_email_matches_platform = (canvas_login_email.downcase == @user.email.downcase)
+      canvas_email_matches_platform = canvas_login_email.casecmp(@user.email).zero?
```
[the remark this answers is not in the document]
peterdeweese: So casecmp is more efficient but has different output. It returns `0` if the case-insensitive compare matches!

PR #1224, opened 2022-12-09
Prevent sync alert for missing LC course for staff onboarding program

elenabaurkot:
# Description
We recently launched a staff onboarding program specifically for staff to onboard, which only needed an Canvas Accelerator course and no LC Playbook. We temporarily added code in the PR [Launch Staff onboarding course with no LC Playbook](https://github.com/bebraven/platform/pull/1215) that let us launch the Staff onboarding course without an LC Playbook. The code was reverted after launching since we do always want an LC Playbook when launching our standard courses, but since this was a one-time non standard course we didn't want to keep the code in there. 

The Salesforce sync that runs every five minutes to sync changes for current and future courses, checks that each program being synced has an Accelerator course and an LC Playbook course linked to the program. After launching this course we were receiving alerts that the staff onboarding program was missing the LC Playbook course. This update prevents alerting for a missing LC Playbook course only for the staff onboarding program since it shouldn't have one. 

# Motivation and Context
We don't need an LC Playbook for the staff onboarding course, so we don't need an alert the LC Playbook course is missing. 

Task: https://app.asana.com/0/1201131148207877/1203511615467171/f


app/services/sync_salesforce_program.rb
```diff
@@ -78,7 +78,11 @@ def run
         return self
       end
 
-      if @salesforce_program.accelerator_course.blank? || @salesforce_program.lc_playbook_course.blank?
+      # Don't alert on missing LC playbook for the staff onboarding course (sfid = "a2Y5c000001UOsxEAG")
+      # It purposely does not have an LC playbook because it's not needed
+      staff_onboarding_program_sfid = 'a2Y5c000001UOsxEAG'
```
ThotsaphonB, line 83: Move staff_onboarding_program_sfid to the Salesforce Constants file

app/services/sync_salesforce_program.rb
```diff
@@ -381,7 +385,14 @@ def translate_error_to_user_message(e, participant)
...
 
     case e
-    when DuplicateContactError, DuplicateParticipantError, MissingContactError, UserSetupError, ZoomAPI::ZoomMeetingEndedError, ZoomAPI::RegistrationNotEnabledForZoomMeetingError, ZoomAPI::ZoomMeetingDoesNotExistError, ZoomAPI::TooManyRequestsError, ZoomAPI::BadZoomRegistrantFieldError
+    when DuplicateContactError,
+      DuplicateParticipantError,
+      MissingContactError,
+      UserSetupError,
+      ZoomAPI::ZoomMeetingEndedError,
+      ZoomAPI::RegistrationNotEnabledForZoomMeetingError,
+      ZoomAPI::ZoomMeetingDoesNotExistError, ZoomAPI::TooManyRequestsError,
```
ThotsaphonB, line 394: Separate into 2 lines

ThotsaphonB (changes requested):
2 issues

tomlovett (approved):
LGTM after Ben's notes.

Thanks for writing tests to handle this new behavior!

PR #1225, opened 2022-12-09
Partially upgrade gems for sidekiq and redis.

peterdeweese:
# Description

Updates redis, sidekiq, and related gems to stop issues caused by upgraded OpenSSL that might have started when we moved to the heroku-22 stack or upon other gem changes. I expect that the fix included in redis 4.7.1 (we are upgrading to 4.8.0) resolves our issues with OpenSSL 3.0: https://github.com/redis/redis-rb/blob/master/CHANGELOG.md

This does not update to the latest but bumps the versions of several related gems to the latest minor or patch version.

Using connection_pool 2.3.0 (was 2.2.5)
Using fugit 1.8.0 (was 1.5.3)
Using rack 2.2.4 (was 2.2.3.1)
Using redis 4.8.0 (was 4.6.0)
Using rufus-scheduler 3.8.2 (was 3.8.1)
Using sidekiq 6.5.8 (was 6.5.1)
Using sidekiq-scheduler 4.0.3 (was 4.0.1)
Using tilt 2.0.11 (was 2.0.10)
Using tzinfo 2.0.5 (was 2.0.4)

# Motivation and Context

We get dozens of errors a day preventing background tasks from being run.

Task: 
https://app.asana.com/0/1201131148207877/1203515968983549/f


elenabaurkot (approved):
Lgtm!

PR #1226, opened 2022-12-09
Configure rubocop to run -a cleanly on spec and db.

peterdeweese:
# Description

Improves rubocop config.

# Motivation and Context
Makes it so `rubocop -a` can be run safely on all specs. Part of learning and cleaning up sync, improving continuity, etc.

Task: 


.rubocop.yml
```diff
@@ -18,7 +18,10 @@ require:
 AllCops:
   NewCops: enable
   Exclude:
-    - 'db/**/*'
```
peterdeweese, line 21: This now includes the db dir except for schema.rb and old migrations before 2023.

.rubocop.yml
```diff
@@ -65,6 +68,22 @@ Metrics/MethodLength:
 Metrics/PerceivedComplexity:
   Max: 16
 
+# These use deprecated capybara DSL
```
peterdeweese, line 71: When autocorrected, these cops broke these tests. The tests need to be modernized by hand at some point.

tomlovett (approved):
LGTM! I like the idea of linting all new migrations without worrying about the old ones.

PR #1232, opened 2022-12-13
Improve rubocop config to allow -A on all spec files.

peterdeweese:
# Description

Makes it so `rubocop -A sync` will work with all tests running afterward.

# Motivation and Context

So we can automatically improve our tests without fear for many rubocop offenses.

Task: 


ThotsaphonB (approved):
LGTM

PR #1235, opened 2022-12-15
Add missing fields to cohort and cohort schedule.

peterdeweese:
# Description

Adds use of 2 heroku connect columns from salesforce that were being ignored.
Cleans related rubocop warnings.

# Motivation and Context

Making it so we can use some Salesforce fields needed to complete this task. We'll test data using them in production while solving.
Removing lint so we can see what we break.

Task: 


elenabaurkot (approved):
Lgtm!

PR #1236, opened 2022-12-15
Refactor sync zoom links into new service

tomlovett:
# Motivation and Context
We originally were going to modify behavior on the SyncZoomLinksForParticipant module. That turned out to be unnecessary; however, the existing code was quite janky and the tests were borderline unreadable.

In the interest of making this module easier for the next developer to understand, I have refactored the service and written clear and comprehensive tests.

# Description
The old service ran both Zoom meetings at the same time. I found no easy way to run both at the same time, so instead the service handles one at a time and is called twice.

The service itself should be self-explanatory now (even if it is a little dense). The test suite - also a little dense - verifies expected behavior for all expected scenarios.

Task: https://app.asana.com/0/1201131148207877/1203468570749032


app/models/participant_sync_info.rb
```diff
@@ -306,6 +306,14 @@ def initialize(last_sync_info, new_sync_info)
       @new_sync_info.pending_zoom_updates = @last_sync_info&.pending_zoom_updates || zoom_info_changed?
     end
 
+    def last_sync_info
+      @last_sync_info
+    end
+
+    def participant
+      HerokuConnect::Participant.find_by(sfid:)
+    end
```
tomlovett, line 315: Maybe this should be handled by a relationship? Either way, it's a very useful connection.

app/services/register_participant_for_zoom.rb
```diff
@@ -0,0 +1,121 @@
...
+    add_honeycomb_field('registration_record', @registration_record.inspect)
+
+    add_honeycomb_field('cancel_registration?', false)
+    add_honeycomb_field('register_participant_for_meeting?', false)
+    add_honeycomb_field('zoom_link_info_record_created?', false)
+  end
+
+  def add_honeycomb_field(field_name, value)
+    Honeycomb.add_field("register_participant_for_zoom.#{field_name}", value)
+  end
```
tomlovett, line 120: Highly recommend helper functions like these whenever possible. Makes the rest of the code significantly more readable.

app/services/sync_salesforce_program.rb
```diff
@@ -189,11 +189,12 @@ def sync_participants
         # links in Salesforce even if the Participant Canvas sync fails.
         zoom_sync_success = with_participant_error_handling(participant) do
           # use `next` because `break` will cause zoom_sync_success to be `nil`
-          next unless should_zoom_sync_participant?(participant) || @force_zoom_update
```
tomlovett, line 192: Moved `@force_zoom_update` check inside `should_zoom_sync_participant?`. More readable (and logical) to keep it all one place.

spec/services/sync_salesforce_program_spec.rb
```diff
@@ -410,132 +414,134 @@
...
-    context "with a new CohortSchedule that has not yet sync'd Zoom links" do
-      let(:cohort_schedule) do
-        build :heroku_connect_cohort_schedule,
-              program:,
-              time__c: '7pm EST',
-              latest_zoom_sync__c: nil
-      end
-      let(:last_sync_info1) { create :participant_sync_info_fellow, program:, cohort_schedule: }
-      let(:mock_now) { DateTime.parse('2:05am EST') }
+    describe '#set_should_sync_zoom_links_for_cohort_schedule' do
```
tomlovett, line 417: Added a new `describe` scope since these tests are quite large but isolated to that one function.,

app/models/participant_sync_info.rb
peterdeweese: Optional. It can be, even across databases. Of course, a chain of AR queries going across this will always lazy load and can't join or have mixed where clauses. The main advantage is the lazy load will automatically cache it.

app/models/participant_sync_info.rb
```diff
@@ -306,6 +306,14 @@ def initialize(last_sync_info, new_sync_info)
       @new_sync_info.pending_zoom_updates = @last_sync_info&.pending_zoom_updates || zoom_info_changed?
     end
 
+    def last_sync_info
+      @last_sync_info
```
peterdeweese, line 310: `attr_reader :last_sync_info`

app/services/register_participant_for_zoom.rb
peterdeweese: This is an improvement! I'll bet we can go further after we start consolidating more of these and see patterns.

peterdeweese (approved):
A few comments on optional changes. I see a lot of functionalities moved/reorganized, and the tests as well provide some confidence, but I couldn't map every detail in my mind. I think you should closely analyze and monitor when released.

peterdeweese (approved):
Same comment as before regarding having difficulty mapping every change in my head. The feature flag mitigates this considerably. We can release it, choose a time to flip the flag, babysit, and back off if needed. Then we need follow-up to make it permanent if successful, like your feature flag process write-up mentions.

elenabaurkot (approved):
This looks good to me, code looks a lot cleaner! My only real concern was solved in your most recent commit- one minor comment, but not necessary just a suggestion.

app/services/register_participant_for_zoom.rb
```diff
@@ -0,0 +1,132 @@
...
+    add_honeycomb_field('cancel_registration?', true)
+
+    begin
+      ZoomAPI.client.cancel_registrants(@registration_record.meeting_id, [@participant_sync_info.email])
+    rescue StandardError => e
+      # Error in a cancellation is not a big deal. Log it and carry on.
+      add_honeycomb_field('cancellation_request_error', e)
+    end
+
+    upsert_join_url_to_salesforce('') # sending an empty string will clear the field on Salesforce
```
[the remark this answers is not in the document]
tomlovett: Very good point! Salesforce logic automatically clears join links when the meeting ID changes, so we can just code the cut altogether.

That `unless` statement would fix the issue if something bad happened on Salesforce, but for the sake of clarity I'm going to leave it out. If the link was not cleared out, it'll mean the Salesforce side had an issue.

PR #1237, opened 2022-12-15
Add section__c to HerokuConnect::Candidate.

peterdeweese:
# Description

Add section__c to HerokuConnect::Candidate. I added this mapping to staging and verified it is in production.
Cleans rubocop warnings from participant and candidate so I can tell if my changes make any messes. Some other files are changed to support.

# Motivation and Context

So we can use this data for discord roles.

Task: 

# Screenshots


app/models/heroku_connect/candidate.rb
```diff
@@ -1,89 +1,93 @@
...
+      [
+        :id, :sfid, :createddate, :isdeleted,
+        :name,
+        :status__c,
+        :recordtypeid,
+        :contact__c,
+        :participant__c,
+        :program__c,
+        :coach_partner_role__c, # Currently called "Candidate Role Select" in the UI
+        :section__c
```
peterdeweese, line 31: This is the only change other than rubocop resolutions.

PR #1238, opened 2022-12-15
Add permission for users to generate Zoom links

elenabaurkot:
# Description
Adding a role `CanGenerateZoomLinks` to allow users other than admins to access and use generate zoom links tool if they have this role on their user.

# Motivation and Context
To allow certain people to generate Zoom links even if they are not admins.


tomlovett (approved):
LGTM! Nice work!

PR #1240, opened 2022-12-16
Make NoCohortScheduleError message more clear

elenabaurkot:
# Description
If we are receiving a lot of NoCohortScheduleError messages, it's possible that regions have not received all learning lab days for participants and a good first step would be to check in with someone from the regional team and make sure they aren't still waiting to receive learning lab days from university partners. If they are not waiting, it's possible that the Salesforce roster upload happened before the sections were uploaded so cohort schedules weren't set properly. Check in with the Salesforce team to make sure sections were uploaded before rosters and everything looks good. This upates the NoCohortScheduleError error message and should hopefully make it more clear what's going on when we start receiving these errors.

# Motivation and Context
We started receiving a lot of NoCohortScheduleError messages for some participants and not others as a result of roster uploads happening before sections were uploaded to Salesforce so cohort schedules didn't get set properly. Participants will not go through the Salesforce sync if their cohort schedule isn't set unless their Platform user ID on their contact is set, so we were only receiving the errors for people who had their user ID set. We want to make it clear what's going on when we receive these errors.

Task: https://app.asana.com/0/1201131148207877/1203536358928603/f


peterdeweese (approved):
Much more clear!

PR #1243, opened 2022-12-19
Create FeatureFlag tool

tomlovett:
# Motivation and Context
We are beginning to use feature flags as a regular tool in the codebase. This helper abstracts the logic to get the value of a feature flag, and does so in a way that can handle bad values for the flag.

# Description
Calling `FeatureFlag.new.enabled?('FLAG_[NEW_FEATURE])` will call up that flag's value in from config. The flag itself - this will be covered in the documentation - should be a number between 0 and 100, representing what percent of users should pass through the new feature. `0` will send all users through old functionality, `100` will send all users through new functionality, and any number between will send that percentage of users through the new feature.

The helper function can handle flags that are not set, and ones that cannot be converted to a number.


lib/feature_flag.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class FeatureFlag
+  def enabled?(feature_flag_name)
+    @flag_value = ENV.fetch(feature_flag_name, nil)
+
+    return false if @flag_value.nil? || @flag_value == '0' || bad_flag?
+    return true if @flag_value == '100'
+
+    # Random.rand returns a random float between 0.0 and 1.0
+    # Use `Random.rand` instead of the simpler `rand` because it is easier to mock in tests
```
peterdeweese, line 11: Makes sense. 👍

peterdeweese (approved):
Great idea!

lib/feature_flag.rb
```diff
@@ -0,0 +1,23 @@
...
+    Random.rand <= Float(@flag_value) / 100
```
elenabaurkot, line 12: This is cool - so from what I understand if you set the flag value to 80, if the random number is less than than it returns true (and since 80 is pretty high most people would pass through using the new flag feature since the random number will usually be lower? If you set it to 100 everyone will pass through using the new flag feature?

elenabaurkot (approved):
One clarifying question, but lgtm! Wiki documentation is super helpful too 👍

lib/feature_flag.rb
tomlovett: Yup! That's it!

For what it's worth, I had it totally backwards at first. Changing the tests to 80 actually showed me that I had it wrong.

PR #1245, opened 2022-12-20
Clarify a few aspects of lrs xapi.

peterdeweese:
# Motivation and Context

While analyzing this code closely to try to support the below task and seeing confusing honeycomb warnings that people were concerned were bugs, I saw easy ways to nudge the code toward clarity while the understanding of the code was still present. 

# Code Description

Adds an error description when lrs state data is missing instead of presenting a confusing JSON parse error in honeycomb.
Combines two identical logical branches and moves comments to their own context, making an outer method more clear.
Extracts lzw javascript compilation to its own method to clarify the bug fix method a little.

The next step would be extracting it to its own library or replacing with more native lzw compression. There is a library that uses unix compress libraries, or it might be found with UNIX or ruby tar libraries available in Heroku and locally. Javascript compilation is resource-heavy and slow for such a simple case. There has to be a better way.

Task:
https://app.asana.com/0/1201131148207877/1203414606491884/f


lib/lrs_xapi_mock.rb
```diff
@@ -211,18 +199,23 @@ def self.handle_request!(request, endpoint, user)
...
 
   # Fix the weird broken suspend_data bug.
-  private_class_method def self.fix_broken_suspend_data(data_json, quiz_breakdown)
+  # There's a weird bug where sometimes your quiz can think it's on the last
+  # (results) page, but somehow doesn't have all the fields it's supposed to
+  # compute once you reach that page. We don't know how to reproduce this, but
+  # it makes it so people can't continue past the quiz, so we need to fix it.
+  # If the data wasn't broken, this will just leave it as-is.
+  # Optional parameters included to simplify testing.
+  private_class_method def self.fix_broken_suspend_data(data_json, lti_launch: nil, quiz_breakdown: nil)
```
peterdeweese, line 208: Changed the signature to allow for more work to be done inside this method, but to allow simple testing. See argument error to make sure that we get one or the other optional parameter.

lib/lrs_xapi_mock.rb
```diff
@@ -211,18 +199,23 @@ def self.handle_request!(request, endpoint, user)
...
+
+    if lti_launch.present?
+      module_version = CourseRise360ModuleVersion
+                       .find_by(canvas_assignment_id: lti_launch.assignment_id)
+                       .rise360_module_version
+      Honeycomb.add_field('rise360_module_version.id', module_version.id.to_s)
+      quiz_breakdown = module_version.quiz_breakdown
+    end
     Honeycomb.start_span(name: 'xapi.fix_broken_suspend_data') do
-      # First, decompress the data. It's compressed using lzwCompress.js, which
```
peterdeweese, line 216: This was extracted to the method below. This method is too complicated and this makes it one step easier to understand.

spec/lib/lrs_xapi_mock_spec.rb
```diff
@@ -385,7 +385,7 @@
   end
 
   describe 'self.fix_broken_suspend_data' do
-    subject { described_class.send(:fix_broken_suspend_data, data, quiz_breakdown) }
+    subject { described_class.send(:fix_broken_suspend_data, data, quiz_breakdown:) }
```
peterdeweese, line 388: The optional parameters are now named parameters.

lib/lrs_xapi_mock.rb
```diff
@@ -325,4 +318,14 @@ def self.handle_request!(request, endpoint, user)
     # If JSON.parse fails, just return the original data.
     data_json
   end
+
+  private_class_method def self.lzw_compressor
+    # Data is compressed using lzwCompress.js, which doesn't seem to have a Ruby equivalent, so unfortunately that
+    # means we have to run JavaScript here.
+    @lzw_compressor ||= Honeycomb.start_span(name: 'xapi.fix_broken_suspend_data.compile') do
```
tomlovett, line 325: Is there another way we could log this to Honeycomb, since we have a limited number of spans to use per month? (We can put an infinite number of fields on spans, though)

tomlovett (approved):
LGTM, but asking to find another logging route other than creating new spans

peterdeweese:
@tomlovett , understood. I'll try something else.

elenabaurkot (approved):
lgtm!

lib/lrs_xapi_mock.rb
peterdeweese: OK. I just got rid of the span completely, as we should still get error reporting. I think it might not even be hitting honeycomb when there is no error, so it might not have done any harm, but it clutters the code with the only gain being essentially an error category that we don't need.

PR #1246, opened 2022-12-20
Implement Faker gem in test factories

tomlovett:
# Motivation and Context
Our factories used a lot of `sequence` generators. These are not great for string values, and they are even jankier when it comes to generating Salesforce ID's. 

Faker is a gem that can generate semi-random values for our factories. These values are both more standardized and more random.

Implementing this gem and replacing subpar field generators will increase the robustness of our tests and set an example for future tests.

There is a risk in breaking our existing tests, but our CI should alert us to that before we get to merging in.

# Code Description
First, I added Faker to the gemfile and ran `bundle install`.

Then, I went through each factory in the codebase and replaced generators with Faker wherever possible. Most important changes were changing the `sfid` generators to use random numbers, rather than the previous generator functions.


Gemfile
```diff
@@ -151,6 +151,9 @@ gem 'rubyzip'
 
 gem 'rails_same_site_cookie'
 
+# Generates test data
+gem 'faker'
```
peterdeweese, line 155: Maybe it should be in the group where factorybot is, if we only use faker for test and development.

Gemfile.lock
```diff
@@ -178,6 +178,8 @@ GEM
     factory_bot_rails (6.2.0)
       factory_bot (~> 6.2.0)
       railties (>= 5.0.0)
+    faker (3.1.0)
```
peterdeweese, line 181: I have a PR updating Rails. Conflicts possible.

spec/controllers/capstone_evaluation_results_controller_spec.rb
```diff
@@ -129,27 +129,27 @@
...
-              )
-            }
-          end
-
-          it 'shows page telling them grade has not been calculated' do
-            subject
-            expect(response.body).to include('grade has not yet been calculated')
-          end
-        end
+        # context 'when a user doesn\'t have a score for their teamwork grade' do
```
peterdeweese, line 132: This test might be made pending instead of commenting out.

spec/factories/access_tokens.rb
```diff
@@ -1,9 +1,7 @@
 FactoryBot.define do
   factory :access_token do
-    sequence(:name) { |i| "Access Token #{i}" }
-    sequence(:key)  { |i|
-      "N2M2ODZiYzgtMTQ5MC00YTZhLThlOTQtYjhhZjJjNTZjODc0.4bgJEIOOI6n2ubk9dlvIYkcMjVln-7u0OPgWgIj7osAGBfxAs67AlWEy2zAb3mM2SBW0lM-U6Dz-4zbhQK-TKt4RR4Tqeqt7dfHpPzrT-mV-1kypNOdtNgM3FHOC70-v-3zqxW088b6Uf36Ugt6hsfR2XaHyCeKH0Bt4iEKQFKdaMUuB6o5cqDY6myY-5HAPsWk4k6qaxxPvG#{i[0]}"
-    }
+    name { Faker::Dessert.variety }
+    key { Faker::Alphanumeric.alphanumeric(number: 255) }
```
peterdeweese, line 4: Much better!

spec/factories/courses.rb
```diff
@@ -1,16 +1,16 @@
 FactoryBot.define do
   factory :course do
-    sequence(:name) { |i| "Course #{i}" }
+    sequence(:name) { |i| "#{Faker::Dessert.variety} ##{i}" } # Janky workaround since we seem to persist Course names
```
peterdeweese, line 3: 👍

spec/spec_helper.rb
```diff
@@ -66,6 +68,17 @@
...
   end
 
+  config.before(:suite) do
+    DatabaseCleaner.strategy = :transaction
+    DatabaseCleaner.clean_with(:truncation)
+  end
+
+  config.around(:each) do |example|
+    DatabaseCleaner.cleaning do
+      example.run
```
peterdeweese, line 78: 🥇

peterdeweese (approved):
Looks good in general! Please consider the gem group comment.

PR #1249, opened 2022-12-21
Add program name/info to failure email for SyncSalesforceJob

tomlovett:
# Motivation and Context
Product Support alerted us that the failure email for SyncSalesforceJob did not include the name of the program in the email. This makes that email less useful.

These changes add the program field to the call to the failure email.

# Code Description
I added safety-wrapping to the function calls in the event that variables are not set before the function is called. 

I added a separate `rescue` statement for "missing program" since `sync_service` will still be nil by the time that flow is hit.

I changed the calls to `error_detail` in SyncSalesforceProgram service because these threw errors when I tested locally. We must have forgotten to update the `error_detail` during rubocop/refactoring.

Task: https://app.asana.com/0/1201131148207877/1203411749175431

(Screenshot on Asana task)


app/services/sync_salesforce_program.rb
```diff
@@ -245,7 +245,7 @@ def with_participant_error_handling(participant)
     Sentry.capture_exception(e)
     error_detail = translate_error_to_user_message(e, participant)
     Honeycomb.add_field('error', e.class.name)
-    Honeycomb.add_field('error_detail', error_detail)
+    Honeycomb.add_field('error_detail', e)
```
peterdeweese, line 248: So we don't need to also log the results of `translate_error_to_user_message`?

peterdeweese (approved):
Looks like it'll work! I did have one question about honeycomb logging that I don't have a real qualm with.

elenabaurkot (approved):
lgtm!

app/services/sync_salesforce_program.rb
tomlovett: Ahh! That's what that line is supposed to be! Good catch! Re-implementing.

PR #1251, opened 2022-12-21
Add more Honeycomb fields to debug incorrect timezones

elenabaurkot:
# Motivation and Context
When we launched our most recent courses, all timezones were set to eastern in Canvas when they should be set to the Salesforce Program's timezone. I tested launching a course in my development environment using a template course that had an eastern time zone and a Salesforce program with a pacific time zone and the Canvas course was correctly set to have a pacific time zone. I tried doing the same thing in production and the Canvas course was incorrectly set to have an eastern time zone instead of Pacific. The purpose of this PR is to add more debugging information to figure out why this is happening. This is an addition to the PR [Add Honeycomb fields for launching course to debug incorrect time zones](https://github.com/bebraven/platform/pull/1248).

# Code Description
Add Honeycomb spans when cloning courses to see why the time zone is being incorrectly set.


ThotsaphonB (approved):
Also LGTM

app/services/initialize_new_course.rb
```diff
@@ -18,6 +18,7 @@ def initialize(new_course, salesforce_program)
 
   def run
     Honeycomb.start_span(name: 'initialize_new_course.run') do
+      Honeycomb.add_field('initialize_new_course.new_course', @new_course)
```
peterdeweese, line 21: I think this is what Tom was talking about, that we should add fields and avoid spans.

peterdeweese (approved):
LGTM

tomlovett:
LGTM!

PR #1252, opened 2022-12-22
Remove extraneous call to `e`

tomlovett:
# Motivation and Context
This call to `e` threw an error during a SF sync that had failed participants. We usually have those error objects, just not in this location.

# Code Description
Cut out that call to `e`, also add another field for `failed_participants`, since that will help in bug-solving.


ThotsaphonB (approved):
LGTM

elenabaurkot (approved):
Lgtm!

PR #1253, opened 2022-12-22
Use nil-friendly check in register_participant_for_zoom

tomlovett:
# Motivation and Context
A sync on a test program failed with this error: `NoMethodError: undefined method `empty?' for nil:NilClass return if @registration_record.nil? && @current_sf_meeting_id.empty?`

`.empty?` is not a nil-friendly call, but `.blank?` is, so use that instead. 

I don't want to use `.nil?` because the value should be a string. I believe it only returned nil because of test user funkiness.

# Code Description
Replace `@current_sf_meeting_id.empty?` with nil-friendly `@current_sf_meeting_id.blank?`


peterdeweese (approved):
It looks like it'll work

PR #1258, opened 2022-12-23
Update timezone on Canvas course after course launch

elenabaurkot:
# Motivation and Context
When we launch a new course, previously it would create the course using the default timezone of the Salesforce program and then copy the content of the course template over and the timezone would remain set to the correct Salesforce program timezone. When we launch courses now it still creates the course with the correct time zone, but when it copies the content over (using the CanvasAPI copy_course call which hits the Canvas `content_migrations` endpoint) it seems to copy the timezone set in course template that it is copying the content over from and results in setting the incorrect timezone on the Canvas course (they all get set to eastern like the course template now). A couple strange things to note: 1) it doesn't appear that any code affecting course launch has changed since we launched Fall courses 2) When I launch courses from my development environment the resulting course still has the correct time zone even after the copy_course call. As of now, the workaround is to update the course time zone after it creates and I am planning on opening a Canvas ticket to see if we can gain anymore insight into why this started happening (I checked the Canvas API change logs and can't see anything that would have caused this there). 

# Code Description
Updated the code to call `update_course` after it finishes launching and updates the timezone to the correct timezone. 


app/services/clone_course.rb
```diff
@@ -79,7 +83,7 @@ def start_canvas_course_clone!(source_course, destination_course)
     destination_course.update!(canvas_course_id: canvas_course_id)
     course_copy = CanvasAPI.client.copy_course(source_course.canvas_course_id, destination_course.canvas_course_id)
 
-    course_info = CanvasAPI.client.get_course(source_course.canvas_course_id)
+    course_info = CanvasAPI.client.get_course(destination_course.canvas_course_id)
```
peterdeweese, line 86: 😆

app/services/clone_course.rb
```diff
@@ -47,6 +47,10 @@ def wait_and_initialize
       add_to_honeycomb_span()
 
       after_canvas_launch_completes! do
```
peterdeweese, line 49: 👍

peterdeweese (approved):
It seems this should work around it!

PR #1259, opened 2023-01-04
Add section name to discord roles.

peterdeweese:
# Motivation and Context

NLU has multiple locations on a certain day, so this separates the discord roles for each section.

# Code Description

Looks up a new flag from program, use_section_codes_for_discord_roles__c. If this is set and a section is available, it is postfixed to the discord role name.

Task:


lib/discord_bot.rb
```diff
@@ -924,12 +924,24 @@ def self.compute_cohort_schedule_role(participant)
...
+    role = if participant.role == SalesforceAPI::FELLOW
+             'Fellow'
+           elsif SalesforceAPI.is_lc?(participant)
+             'LC'
+           else
+             return
+           end
+
+    # TODO: Prevent duplicate calls by passing in when available.
+    program_info = SalesforceAPI.client.get_program_info participant.program_id
```
peterdeweese, line 936: I made a task to resolve this TODO. We may have extra SF API calls.

spec/lib/discord_bot_spec.rb
```diff
@@ -1050,74 +1049,79 @@
...
-                                                   cohort_schedule_day: cohort_schedule_day))
-      }
+    context 'with Fellow and use section code flag' do
+      let(:factory) { :salesforce_participant_fellow }
+      let(:use_section_codes_for_discord_roles__c) { true }
 
-      it 'returns role name' do
-        expect(subject).to eq("LC: #{cohort_schedule_day} LL")
-      end
+      it { is_expected.to eq('Fellow: Monday LL CH') }
```
peterdeweese, line 1081: Here is the expected result. Many of the other changes here are selected rubocop improvements.

spec/lib/discord_bot_spec.rb
```diff
@@ -112,11 +112,11 @@
...
 
       context 'with message from admin' do
-        let(:roles) {
+        let(:roles) do
           [
-            instance_double(Discordrb::Role, name: ADMIN_ROLES.first),
+            instance_double(Discordrb::Role, name: ADMIN_ROLES.first)
           ]
-        }
+        end
```
tomlovett, line 119: NITPICK: There's a few instances like this that could _probably_ fit on one line with curly-brackets.

tomlovett (approved):
Have a few optional nitpicks about formatting, but LGTM!

spec/lib/discord_bot_spec.rb
```diff
@@ -1050,74 +1049,79 @@
...
-      let(:participant) {
-        SalesforceAPI.participant_to_struct(create(:salesforce_participant_fellow,
-                                                   cohort_schedule_day: cohort_schedule_day))
-      }
+      let(:factory) { :salesforce_participant_fellow }
 
-      it 'returns role name' do
-        expect(subject).to eq("Fellow: #{cohort_schedule_day} LL")
-      end
+      it { is_expected.to eq('Fellow: Monday LL') }
```
elenabaurkot, line 1074: @peterdeweese I feel like you explained this in the working session once, but I'm forgetting how this works - what is the reason that you don't need to call the subject here?

lib/discord_bot.rb
```diff
@@ -924,12 +924,24 @@ def self.compute_cohort_schedule_role(participant)
...
+    suffix = ''
+    if program_info['use_section_codes_for_discord_roles__c']
+      # TODO: update to use the virtual Participant#section__c in apex code in salesforce instead of using this.
+      candidate_info = SalesforceAPI.client.find_candidate_info(participant_id: participant.id)
+      section = candidate_info&.dig(:section__c)
+      suffix = " #{section}" if section.present?
     end
+    "#{role}: #{ll_day} LL#{suffix}"
```
elenabaurkot, line 944: I don't think this was made super clear by the task, but I just talked to Product Support to clarify this and for programs that do have the special section codes, they want to have both the regular role and this special role - so returning something like `["#{role}: #{ll_day} LL",  "#{role}: #{ll_day} LL#{suffix}"]` when they have the special section and if they don't just returning the normal `"#{role}: #{ll_day} LL"` and adding any returned roles to the `role_names` array on line 821. I think this is because they want to be able to message individual sections, but at times also message learning lab days as a whole even if there are a handful of sections for that day. I don't think these roles get added automatically to Discord so we will have to actually add whatever new roles that come from these sections to the Discord server that's using this.

elenabaurkot (commented):
One small change after clarifying the roles with product support, but otherwise looks good!

PR #1260, opened 2023-01-04
Gracefully handle re-registration error from Zoom

tomlovett:
# Motivation and Context
While testing out the new Zoom sync, I got a 400 response from Zoom that did not include much in the way of details
See `error_detail` field on this [span](https://ui.honeycomb.io/braven/datasets/platform.braven.org/result/oubgYMrUvip/trace/FrXQgLNLddZ?fields[]=c_name&fields[]=c_sql.active_record.sql&span=8392eacaa6d0bd4e)

It turns out we were getting an error from Zoom that we could not re-register a participant for a Zoom meeting if we had previously cancelled them: https://devforum.zoom.us/t/error-3072-when-trying-to-re-add-user-to-webinar/4207

This code allows us to handle that error gracefully, and perhaps update functionality of the `RegisterParticipantForZoom` service if we so choose.

# Code Description
Adds a new `case` statement for the Zoom error for re-registering a participant.

Also cleans up `ZoomAPI` client's specs, and adds a test case that returns the `3072` code.

Task: https://app.asana.com/0/1201131148207877/1203642977148289


peterdeweese (approved):
LGTM!

PR #1261, opened 2023-01-04
Remove Zoom re-registration functionality while preserving test refactoring

tomlovett:
# Motivation and Context
Functionality was added in #1260 to handle the possible need to re-register participants for the same meeting. Discussions with ProdSupport made it clear that that behavior will never happen; so we can remove that code from the codebase.

#1260 also included some test refactoring that should be kept, so it was better to manually remove that functionality than revert the PR.

This PR also includes a little more refactoring of that test.

# Code Description
Remove the error-handling for code `3072`

The test scenario on line 47 was deleted. In addition, the `stub(...)` calls were moved up to the parent context for the sake of readability.


elenabaurkot (approved):
Lgtm!

PR #1263, opened 2023-01-05
Add discord role.

peterdeweese:
# Motivation and Context

NLU has multiple locations on a certain day, so this separates the discord roles for each section.
Instead of replacing the old role as the previous PR implemented, this adds it.

# Code Description

Separates methods for creating each role type and adds both if they are created by name on discord.

Task:


Gemfile
```diff
@@ -34,6 +34,15 @@ gem 'jbuilder'
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', '>= 1.4.2', require: false
 
+group :production do
```
peterdeweese, line 37: Stops us from loading and initializing judoscale in our development and testing environments.

lib/discord_bot.rb
```diff
@@ -963,34 +981,30 @@ def self.compute_cohort_role(participant)
...
   def self.get_role(server, role_name)
-    Honeycomb.start_span(name: 'bot.get_role') do |_span|
-      Honeycomb.add_field('role.name', role_name)
-      LOGGER.debug "Fetching role '#{role_name}'"
+    Honeycomb.add_field('role.get_role.name', role_name)
+    LOGGER.debug "Fetching role '#{role_name}'"
 
-      server.roles.find { |r| r.name == role_name }
-    end
+    server.roles.find { |r| r.name == role_name }
```
elenabaurkot, line 987: just checking my understanding here -- if it tries to find a role that doesn't exist it will just return nil so it won't actually add any roles to members until we manually add the role to the Discord server? If that sounds right then I think this all looks good to me

elenabaurkot (approved):
One double check for understanding and as long as I'm understanding correctly, this lgtm!

Gemfile.lock
```diff
@@ -298,8 +298,6 @@ GEM
     nio4r (2.5.8)
     nokogiri (1.13.10-aarch64-linux)
       racc (~> 1.4)
-    nokogiri (1.13.10-x86_64-linux)
-      racc (~> 1.4)
```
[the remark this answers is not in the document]
peterdeweese: I'll revert that. Thanks!

lib/discord_bot.rb
peterdeweese: @elenabaurkot, correct!

PR #1264, opened 2023-01-05
Pass uncaught Zoom client error message

tomlovett:
# Motivation and Context
ZoomAPI's error-handling defaults to `raise` if the specific code is not accounted for. This means the actual information from the error is not passed up to the parent context. This unnecessarily complicates debugging.

# Code Description
Change the error type to `RestClient::BadRequest` and include the error body.


elenabaurkot (approved):
Good thinking, this looks good to me 👍

peterdeweese (approved):

PR #1265, opened 2023-01-05
Allow Zoom host error to fail silently

tomlovett:
# Motivation and Context
We were getting loud error messages for trying to register the Host of a Zoom meeting. This was handled gracefully by the old Zoom service, but the new one did not account for it.

We want this error to pass silently (and without being treated as an error). We also want the error message to be uploaded to Salesforce.

# Code Description
Add error-handling for this error to the new Zoom service.

The error was also renamed to be more clear and to remove duplicate reference to Zoom. (The error is attached to `ZoomAPI` class; no need to include "Zoom" in its name.)

Task: https://app.asana.com/0/1201131148207877/1203651181813327


app/services/register_participant_for_zoom.rb
```diff
@@ -101,15 +109,6 @@ def upsert_join_url_to_salesforce(join_url)
...
 
-  def save_registrant_id_to_zoom_registration_record(registrant_id)
-    if @registration_record.registrant_id? # is existing record
-      @registration_record.update!(registrant_id:)
-    else
-      @registration_record.registrant_id = registrant_id
-      @registration_record.save!
-    end
-  end
-
```
tomlovett, line 112: Simplify save logic

peterdeweese (approved):
LGTM!

elenabaurkot (approved):
Lgtm!

PR #1268, opened 2023-01-09
Remove two uses of activesupport in discord_bot.

peterdeweese:
# Motivation and Context

Methods from active support cause discord_bot to fail because it is not run in a rails environment.

# Code Description

Replaced use of Array#excluding and Rails#env#test?

Task:


elenabaurkot (approved):
Lgtm!

PR #1272, opened 2023-01-11
Update Scout gem and reduce messages sent to sentry that are not working.

peterdeweese:
# Motivation and Context

Scout needs to be updated, and we are sending more to sentry than we are using.

# Code Description

Updates Scout. Disables extra sentry data that our plan doesn't use.

Task:
https://app.asana.com/0/1201131148207877/1203643906856834/f


tomlovett (approved):
LGTM!

PR #1274, opened 2023-01-12
Fix migration issue in production.

peterdeweese:
# Motivation and Context

Migrations have been failing whenever we push!

# Code Description

Comments out line that likely targeted a local environment that was out of sync with production.

Task:


tomlovett (approved):
LGTM

PR #1278, opened 2023-01-12
Reduce jobs and sentry in non-server processes.

peterdeweese:
# Motivation and Context

We can't run tasks without docker. This will allow us to run some tasks and reduce unneeded resource use while doing so. Also, a job is created in production by all tasks instead of just on server start.

# Code Description

It adds a check to see if we are running in server mode before adding server start jobs. It also doesn't start sentry if in development mode and not running in the server.

Task:


config/application.rb
```diff
@@ -10,8 +10,11 @@
 Bundler.require(*Rails.groups)
 
 module Platform
-  class Application < Rails::Application
+  def self.running_as_server?
```
peterdeweese, line 13: Checks to see if we are in server mode.

config/application.rb
```diff
@@ -21,11 +24,14 @@ class Application < Rails::Application
 
     # Enable ActiveRecord encryption: https://edgeguides.rubyonrails.org/active_record_encryption.html
     config.active_record.encryption.primary_key = Rails.application.secrets.active_record_encryption_primary_key
-    config.active_record.encryption.deterministic_key = Rails.application.secrets.active_record_encryption_deterministic_key
-    config.active_record.encryption.key_derivation_salt = Rails.application.secrets.active_record_encryption_derivation_salt
+    config.active_record.encryption.deterministic_key = Rails.application
```
peterdeweese, line 27: There are a lot of rubocop corrections. I'll point out other significant changes.

config/initializers/delayed_jobs.rb
```diff
@@ -1,12 +1,12 @@
 # Add delayed_jobs that should be run on app startup in this file
 Rails.configuration.after_initialize do
+  # These are annoying when running specs or rake tasks. We should test the jobs directly anyway and not as part of
+  # app startup.
+  next if Rails.env.test? || !Platform.running_as_server?
```
peterdeweese, line 5: Here it is. We don't want to start any delayed jobs just because rails is starting in a task like db:migrate or rails console.

config/initializers/sentry.rb
```diff
@@ -1,9 +1,12 @@
 require 'filter_logging'
 
 # https://docs.sentry.io/platforms/ruby/guides/rails/configuration/options/
+
+# Do not init in development unless running as server.
+return if Rails.env.development? && !Platform.running_as_server?
+
```
peterdeweese, line 7: Here is the other significant piece. We don't want sentry in development for tasks at all. In fact, we might not want it in development at all.

ThotsaphonB (approved):
LGTM!

